PR #7, opened 2016-01-20
Update README with more detailed instructions

mispy:
I got it all working, but it required a bit of fiddling. Hopefully this will make it easier for others :)


zdenekhynek:
Thanks so much for all the pull requests!

PR #18, opened 2017-04-15
Added Country code column for exported csv files

aaldaber:
Just added the country code column, which already was in the result set of database query, to the list of columns to be written to csv.


mispy:
First commits, yay! 🎉 You can go and mark the related issues on Basecamp as completed :)

PR #25, opened 2017-06-19
Adding the wdi_importer tool, and changes to the country_name tool

aaldaber:


importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+                    '%s' \
+                    '%s' \
+                    '%s' \
+                    '%s' \
+                    '</p>' \
+                  '</div>'
+
+
+wdi_zip_file_url = 'http://databank.worldbank.org/data/download/WDI_excel.zip'
+wdi_downloads_save_location = os.path.dirname(os.path.realpath(__file__)) + '/wdi_downloads/'
```
mispy: Note that the grapher repo gets overwritten on deployment, so these directories will be deleted unless they're symlinked in by deploy.sh.

I'd suggest we keep any persistent output data in the one location (to avoid symlinks all over the place)-- probably BASE_DIR + '/data' or similar

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+
+        column_number = 0
+
+    insert_string = 'INSERT into data_values (value, year, fk_ent_id, fk_var_id) VALUES (%s, %s, %s, %s)'  # this is used for constructing the query for mass inserting to the data_values table
+    data_values_tuple_list = []
+
+    for category in wdi_categories_list:
+        newdataset = Dataset(name='World Development Indicators - ' + category,
+                             description='This is a dataset imported by the automated fetcher',
+                             namespace='wdi', fk_dst_cat_id=DatasetCategory.objects.get(name=wdi_category_name_in_db),
```
mispy: Small optimization: don't need to fetch DatasetCategory for every loop here

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+                                                        global_cat[indicator_code]['concept'] + '<br>' if global_cat[indicator_code]['concept'] else '',
+                                                        '<i style="text-decoration: underline;">Related source links: </i><br>' +
+                                                        global_cat[indicator_code]['sourcelinks'] + '<br>' if global_cat[indicator_code]['sourcelinks'] else '',
+                                                        '<i style="text-decoration: underline;">Other web links: </i><br>' + global_cat[indicator_code][
+                                                            'weblinks'] + '<br>' if global_cat[indicator_code]['weblinks'] else ''
+                                                        ),
+                                                       datasetId=newdataset.pk)
+                                    newsource.save()
+                                    backup_sql_file.write("INSERT INTO %s (name, description, datasetId, created_at, updated_at) "
```
mispy: We should never use raw string formatting to construct SQL statements. While obviously we don't expect an injection attack from the World Bank, it is still possible that their data will be unintentionally formatted in a way that breaks the SQL. 

Since we're just generating strings here we can't use parameterized queries... I'd suggest writing a simple string formatting function that just runs any params through pymysql.escape_string first

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+                                    with connection.cursor() as c:
+                                        c.executemany(insert_string, data_values_tuple_list)
+                                    backup_sql_file.write('INSERT INTO %s (value, year, fk_ent_id, fk_var_id) VALUES ' % DataValue._meta.db_table + ','.join(['("%s", %s, %s, %s)' % x for x in data_values_tuple_list]) + ';\n')
+                                    logger.info("Dumping data values...")
+                                    data_values_tuple_list = []
+
+
+            column_number = 0
+            if row_number % 10 == 0:
+                time.sleep(0.001)  # this is done in order to not keep the CPU busy all the time, the delay after each 10th row is 1 millisecond
```
mispy: Good thinking!

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+    row_number = 0
+    for row in country_ws.rows:
+        row_number += 1
+        for cell in row:
+            if row_number > 1:
+                column_number += 1
+                if column_number == 1:
+                    country_code = cell.value
+                if column_number == 3:
+                    country_name = unidecode.unidecode(cell.value)
```
mispy: Is the use of unidecode necessary? We should be able to handle unicode strings throughout the grapher-- if we can't then that's a bug that should be fixed.

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+import_history = ImportHistory.objects.filter(import_type='wdi')
+
+
+# if wdi imports were never performed
+if not import_history:
+    logger.info("This is the very first WDI data import.")
+    backup_sql_file = open(os.path.join(wdi_backups_location, '%s.sql' %
+                           timezone.now().strftime("%Y-%m-%d %H:%M:%S")), 'w', encoding='utf-8')
+
+    wb = load_workbook(excel_filename, read_only=True)
```
mispy: Good to specify read-only!

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+        row_number += 1
+        indicator = None
+        for cell in row:
+            if row_number > 1:
+                column_number += 1
+                if column_number == 1:
+                    indicator = cell.value
+                    global_cat[cell.value] = {}
+                if column_number == 2:
+                    global_cat[indicator]['category'] = cell.value.split(':')[0]
```
mispy: We can probably make this code faster and less verbose by assigning global_cat[indicator] to a variable on column_number == 1 instead of doing a lookup for every cell

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+                        global_cat[indicator]['weblinks'] = cell.value
+                    else:
+                        global_cat[indicator]['weblinks'] = ''
+                global_cat[indicator]['saved'] = False
+
+        column_number = 0
+
+    category_vars = {}  # categories and their corresponding variables
+
+    for key, value in global_cat.items():
```
mispy: I was about to say "use iteritems" but nope, Python 3 got rid of that 🎉

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+        if category_vars.get(value['category'], 0):
```
mispy: I think you can just do "if value['category'] in category_vars" here

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+
+    existing_categories = DatasetCategory.objects.values('name')
+    existing_categories_list = []
+
+    for each in existing_categories:
+        existing_categories_list.append(each['name'])
+
+    if wdi_category_name_in_db not in existing_categories_list:
+        the_category = DatasetCategory(name=wdi_category_name_in_db)
+        the_category.save()
```
mispy: Are we using transactions here? What happens if the importer fails in the middle?

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+    with open(filename, "rb") as f:
+        while True:
+            buffer = f.read(blocksize)
+            if not buffer:
+                break
+            m.update(buffer)
+    return m.hexdigest()
+
+
+source_template = '<table>' \
```
mispy: Sorry that this was necessary... I should've found a better way to store source information by now

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+    existing_categories = DatasetCategory.objects.values('name')
+    existing_categories_list = []
+
+    for each in existing_categories:
+        existing_categories_list.append(each['name'])
+
+    if wdi_category_name_in_db not in existing_categories_list:
+        the_category = DatasetCategory(name=wdi_category_name_in_db)
+        the_category.save()
+        backup_sql_file.write("INSERT INTO %s (name, created_at, updated_at) VALUES ('%s', '%s', '%s');\n" %
```
mispy: I'm not sure I understand the purpose of having a separate backup file. If we're writing out all of the statements in full raw SQL anyway, why not just generate this sql file and then run that for the actual import?

importer/wdi_importer.py
```diff
@@ -1,63 +1,922 @@
...
+                               the_category.updated_at.strftime('%Y-%m-%d %H:%M:%S')))
+        logger.info("Inserting a category %s." % wdi_category_name_in_db.encode('utf8'))
+
+    else:
+        the_category = DatasetCategory.objects.get(name=wdi_category_name_in_db)
+
+    existing_subcategories = DatasetSubcategory.objects.filter(fk_dst_cat_id=the_category.pk).values('name')
+    existing_subcategories_list = []
+
+    for each in existing_subcategories:
```
mispy: When constructing data used solely for existence testing, it's generally better to use set() or dict() instead-- O(1) instead of O(n)

PR #26, opened 2017-07-08
Related to dataset_categories

aaldaber:
Add a 'fetcher_autocreated' field to the dataset_categories table, don't show categories/subcategories of bulk imported data on the manual data upload screen, minor fixes to grapher_admin/views.py


mispy:
Looks sensible! Should fix the tests though (you probably just need to run data_export.py again)

PR #27, opened 2017-07-16
Add version tracking system, add qog_fetcher and unwpp_fetcher, add user data editing

aaldaber:


grapher_admin/views.py
```diff
@@ -1434,3 +1547,359 @@ def register_by_invite(request: HttpRequest, code: str):
...
+        if len(each.split('->')) > 1:  # finding the status R (file renamed)
+            filename = each.split('->')[1].strip()
+        else:
+            filename = each[each.find(' ') + 1:].strip()
+        if filename.startswith('"') and filename.endswith('"'):
+            changed_files_list.append(filename[1:-1])
+        else:
+            changed_files_list.append(filename)
+
+    if dataset_filename in changed_files_list or metadata_filename in changed_files_list:
```
mispy: If we extract the commit info directly from the repo rather than storing it in the database, we should be able to remove this section. We can then just try a `git commit` and if nothing has changed it won't create one.

grapher_admin/views.py
```diff
@@ -1434,3 +1547,359 @@ def register_by_invite(request: HttpRequest, code: str):
...
+
+    metadata_file_path = os.path.join(current_dataset_folder, metadata_filename)
+    dataset_file_path = os.path.join(current_dataset_folder, dataset_filename)
+    metadata_file_path_escaped = shlex.quote(os.path.join(current_dataset_folder, metadata_filename))
+    dataset_file_path_escaped = shlex.quote(os.path.join(current_dataset_folder, dataset_filename))
+
+    delete_previous = False
+    if old_dataset_name is not None:
+        if os.path.isfile(old_dataset_file_path):
+            subprocess.check_output('git -C %s rm %s' %
```
mispy: I think this can be simplified a bit by using `git mv`. I'd suggest batching the repo-modifying git operations into a single subprocess call at the end, to avoid leaving the repo in an intermediary state if errors occur. (you already mostly do this!)

grapher_admin/views.py
```diff
@@ -1434,3 +1547,359 @@ def register_by_invite(request: HttpRequest, code: str):
...
+            delete_previous = True
+            commit_message = 'Updating the dataset: %s'
+        else:
+            commit_message = 'Creating the dataset: %s'
+    else:
+        commit_message = 'Creating the dataset: %s'
+
+    # now saving the dataset to file
+
+    with open(dataset_file_path, 'w', newline='', encoding='utf8') as f:
```
mispy: It would be wise to write the dataset to a /tmp/{hash} location first and then atomically `mv` into the repo after the write is completed, in the same command sequence as the commit. This would reduce the risk of data corruption due to interrupted write operations.

grapher_admin/views.py
```diff
@@ -503,6 +509,9 @@ def store_import_data(request: HttpRequest):
                     with connection.cursor() as cursor:
                         cursor.execute("DELETE FROM sources WHERE sources.id NOT IN (SELECT variables.sourceId FROM variables)")
 
+                csv_write_thread = Thread(target=write_dataset_csv, args=(dataset.pk, dataset_old_name, request.user.get_full_name(), request.user.email))
```
mispy: I'd recommend not doing this in a thread. If the thread fails for some reason, the SQL database will be updated but the corresponding commit will be lost, creating inconsistency between the export archive and the grapher output.

Instead, it should be done synchronously within the same transaction block as the rest of the import, so that if it errors out, the SQL changes are also rolled back. It's fine if the import process takes a little while.

grapher_admin/migrations/0005_datasetcommithistory.py
```diff
@@ -0,0 +1,29 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('grapher_admin', '0004_datasetcategory_fetcher_autocreated'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DatasetCommitHistory',
+            fields=[
```
aaldaber: Yep, this can be done using git log. Although it might be slower to query for commit info from git itself instead of looking up from the db, it is a better design to have commit info extracted from git rather than having duplicate info in two places.

grapher_admin/views.py
aaldaber: Yes, I absolutely agree on this one.

mispy:
Thanks! Will start testing and deployment of this now.

PR #28, opened 2017-07-27
Query optimizations for slow grapher_admin views (chart listing, chart editing, data import)

aaldaber:


mispy:
Neat that Django does iterators like that!

Issue #51, opened 2017-12-05
Slope charts should allow highlighting and searching entities like scatterplots


mispy:
We want to take the little legend from the scatterplot chart type, e.g. https://ourworldindata.org/grapher/urbanization-vs-gdp

And put it on the "slope chart" chart type, e.g. https://ourworldindata.org/grapher/access-to-a-financial-account-or-services

Performance is an important consideration here; we want to try to keep the number of SVG elements and updates to them to a minimum.

---

The relevant files to look at:

https://github.com/owid/owid-grapher/blob/master/js/charts/ScatterPlot.tsx
https://github.com/owid/owid-grapher/blob/master/js/charts/ScatterColorLegend.tsx
https://github.com/owid/owid-grapher/blob/master/js/charts/SlopeChart.tsx
https://github.com/owid/owid-grapher/blob/master/js/charts/ControlsFooter.tsx

Issue #50, opened 2017-12-05
Slope charts should have region legends like scatterplots


mispy:
This one should be reasonably straightforward.

We want to take the little legend from the scatterplot chart type, e.g. https://ourworldindata.org/grapher/urbanization-vs-gdp

And put it on the "slope chart" chart type, e.g. https://ourworldindata.org/grapher/access-to-a-financial-account-or-services

---

The relevant files to look at:

https://github.com/owid/owid-grapher/blob/master/js/charts/ScatterPlot.tsx
https://github.com/owid/owid-grapher/blob/master/js/charts/ScatterColorLegend.tsx
https://github.com/owid/owid-grapher/blob/master/js/charts/SlopeChart.tsx

sherinkurian:
hi @mispy - I was hoping to contribute to this repo since I really enjoy OurWorldInData. 

to test the charts, this page seemed ideal http://localhost:8000/grapher/admin/testall?type=scatter. but the local urls are not working. 

1. 404 on embedCharts.js. it seems like this file is dynamically created and not sure how to generate that. 

<img width="1388" alt="screenshot 2018-05-26 19 09 22" src="https://user-images.githubusercontent.com/4923428/40581876-7c28b98e-6118-11e8-8a19-4fbbe5f87427.png">

2. There is a link to open the local chart in the html source ^  but when I try to access it from the browser, it throws a 404. 

```
<figure data-grapher-src="http://localhost:8000/grapher/top-MRTs-on-individual-income?tab=chart"></figure>
```

thank you!

mispy:
Hi Sherin, thanks for taking a look! You are on the right track, cool that you found the testing url.

In order to test the "public" chart urls for the initial database charts you'll need to run a static build: `node dist/src/bakeCharts.js`. I've updated the README to reflect this.

sherinkurian:
Hi Jaiden, thank you so much. Managed to get the embedCharts.js but got stuck on getting individual charts to load. I can take a look again to see what's up.

Meanwhile, I ended up using the node server instead at http://localhost:3030/admin/charts which worked without any changes. It seemed like the project is moving towards node? 

There is a PR for your review to set legend on slope charts :)

PR #62, opened 2018-05-28
add features to slope charts

sherinkurian:
On slope charts:

- [x] Differentiate between hover and focus states
- [x] Add legend
- [x] Add search
- [x] Add foreground and background groups

As requested in:
https://github.com/owid/owid-grapher/issues/50
https://github.com/owid/owid-grapher/issues/51

@mispy

### DIFF
--- a/js/charts/SlopeChart.tsx
+++ b/js/charts/SlopeChart.tsx
@@ -46,7 +46,6 @@ export default class SlopeChart extends React.Component<{ bounds: Bounds, chart:
     }
 
     @action.bound onLegendMouseOver(color: string) {
-        this.hoverKey = undefined
         this.hoverColor = color
     }
 
@@ -68,7 +67,7 @@ export default class SlopeChart extends React.Component<{ bounds: Bounds, chart:
     }
 
     // All currently hovered group keys, combining the legend and the main UI
-    @computed get hoverKeys(): string[] {
+    @computed.struct get hoverKeys(): string[] {
         const { hoverColor, hoverKey, transform } = this
 
         const hoverKeys = hoverColor === undefined ? [] : uniq(transform.data.filter(g => g.color === hoverColor).map(g => g.key))
--- a/js/charts/SlopeChart.tsx
+++ b/js/charts/SlopeChart.tsx
@@ -1,15 +1,105 @@
 import * as React from 'react'
-import { computed } from 'mobx'
+import { intersection, without, uniq, getRelativeMouse } from './Util'
+import { observable, computed, action } from 'mobx'
 import { observer } from 'mobx-react'
 import Bounds from './Bounds'
 import ChartConfig from './ChartConfig'
-import LabelledSlopes from './LabelledSlopes'
+import LabelledSlopes, { SlopeProps } from './LabelledSlopes'
 import NoData from './NoData'
+import ScatterColorLegend, { ScatterColorLegendView } from './ScatterColorLegend'
 
 @observer
 export default class SlopeChart extends React.Component<{ bounds: Bounds, chart: ChartConfig }> {
+    // currently hovered individual series key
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -19,3 +19,4 @@ dist
 public
 grapher_admin/static/build
 .netlify
+*.swp
```
sherinkurian: for vim

js/charts/SlopeChart.tsx
```diff
@@ -1,25 +1,119 @@
...
 
         const { bounds, chart } = this.props
         const { yAxis } = chart
         const { data } = this.transform
+        const { legend, focusKeys, hoverKeys, focusColors, activeColors, sidebarWidth } = this
 
-        return <LabelledSlopes bounds={bounds} yDomain={yAxis.domain} yTickFormat={this.transform.yTickFormat} yScaleType={yAxis.scaleType} yScaleTypeOptions={yAxis.scaleTypeOptions} onScaleTypeChange={(scaleType) => { chart.yAxis.scaleType = scaleType }} data={data} fontSize={chart.baseFontSize} />
+        return <g>
+            <LabelledSlopes bounds={bounds} yDomain={yAxis.domain} yTickFormat={this.transform.yTickFormat} yScaleType={yAxis.scaleType} yScaleTypeOptions={yAxis.scaleTypeOptions} onScaleTypeChange={(scaleType) => { chart.yAxis.scaleType = scaleType }} data={data} fontSize={chart.baseFontSize} focusKeys={hoverKeys} onMouseOver={this.onSlopeMouseOver} onMouseLeave={this.onSlopeMouseLeave}  />
+            <ScatterColorLegendView legend={legend} x={bounds.right - sidebarWidth} y={bounds.top} onMouseOver={this.onLegendMouseOver} onMouseLeave={this.onLegendMouseLeave} focusColors={focusColors} activeColors={activeColors} />
```
sherinkurian: may be rename `ScatterColorLegendView` to something generic?

mispy:
Hi Sherin! Thanks so much for your help with this!

The group hover highlighting works perfectly. However, the bounds calculation is a bit off; the legend can overlap with the chart like this:

screen shot 2018-06-06 at 4 51 46 pm

To fix this I think you would need to correct the bounds for the slope chart based on the bounds of the legend (in the way the scatterplot also does)

sherinkurian:
Hi Mispy, thanks for the review :) - the overlap is corrected.

minor thing - I felt the on-mouse-over experience for slope lines were a bit jumpy. but that could be because of too many data points.

mispy:
Looking good!

It's true that the mouseover can be a bit jumpy. We might not want to reduce the area that much though-- it can be tricky to get the mouse exactly on very thin lines. Might be good to have an upper bound on distance, so that it activates the closest line but only if it is within a certain range.

There are a couple of compatibility things I need to work out with existing charts before we can merge this. I'll see if I can write a better export script so I can get you a more extensive copy of our database for testing with. :)

mispy:
Hi Sherin, for this one I think the only blocker is compatibility with existing slope charts. I've ported the testing pages to the node admin on master now, and in conjunction with the database exports, you should hopefully be able to test how your code renders the charts in comparison to the live version by going here: http://localhost:3030/admin/test/embeds?type=SlopeChart

sherinkurian:
Woah, this is great. I should be able to test this once the static build completes and all the urls start loading. Thank you for improving the development experience 💯

sherinkurian:
Hi Jaiden, the features should be backwards compatible now. Thank you for the test pages. They were super helpful.

mispy:
All works perfectly! Merged and deployed 🎉https://ourworldindata.org/grapher/child-mortality-by-country

sherinkurian:
yay, that makes me so happy :)

Issue #63, opened 2018-06-06
Port the country name standardizer to JavaScript


mispy:
Aibek wrote this useful tool for standardizing country names between different formats, a common problem in working with economics datasets: https://github.com/owid/owid-grapher/tree/master/country_name_tool

This is currently server-side code, but we'd like to do most of it client-side. It should be possible to preload the necessary entity data and process the CSV entirely in the browser.

I think the only server-side step should be the remembering of choices in the case when manual resolution is needed.

Once ported to client-side JS, we could then think about moving the standardizer out of the admin and making it a standalone public tool.

sherinkurian:
hi Jaiden, would it be possible to share data from `country_name_tool_countrydata` and `country_name_tool_countryname` - to test client-side performance. it's not a blocker though.

OurWorldInData-User:
Dear Sherin,

Your version of the Country Tool is fantastic! 
https://owid.cloud/admin/standardize

I am sure many researchers will be very, very happy to be able to use this tool as it is one of the most tedious recurring tasks to do deal with these country name standardizations when you combine many datasets.

I tried it out with a dataset and it works very nicely!

One thing that will be needed – especially if we want to make it public – is a bit of explanation of how to use it.
Could there be a little text on the same page – maybe simply above the tool – where you explain the steps one has to take to standardize the country names?
– The file has to be a csv
– The country has to be saved in a row called "country"
– The output file will appear there to download
etc.

And more generally it is really, really cool that you are becoming part of the new team! Great to be working with you and if you have any questions, concerns, ideas just let me know.

Best,
Max

HannahRitchie:
Hey Sherin,

Just tried out the new soon-to-be-public version of the Country Tool. It is really great! Super-fast and very straightforward to use. It it really cool that you are helping us out in this way - I think the people who use our stuff will be really grateful (and we are too, of course).

I have only a few small suggestions in addition to Max above:
- for it to be more intuitive/obvious for people to use I think next to the 'Potential candidates' title I would put 'Potential candidates (select below)'. This makes it obvious that they have to choose one for themselves.
- similarly in the next column I would change 'Custom Name' to 'Or enter a Custom Name'. This makes it obvious to people that they have to do one or the other (some may think they have to do both, or get confused there).

Thanks again for all of your help!
Hannah

JoeHasell:
Hi Sherin,

Really well done – this is going to be an incredibly useful tool for a lot of people! As Hannah says, it runs impressively fast.

[...]

PR #68, opened 2018-06-25
move country standardizer tool to js

sherinkurian:
hi Jaiden,

As requested in https://github.com/owid/owid-grapher/issues/63

screenshot 2018-06-25 07 15 40

Note:

1. Use fuzzysort to determine string similarity
2. Custom selections from dropdown are saved back only when selected input format is the "Non standard country name" and output format is the OWID name (same as the current tool)
3. The country-update tool and instruction pages need to be moved over as well


mispy:
Woah, this looks great! Pretty much exactly what we were after, in terms of the design and architecture.

Here's a mysqldump of the country name tables: [owid-grapher-country-name-tool.sql.zip](https://github.com/owid/owid-grapher/files/2136740/owid-grapher-country-name-tool.sql.zip)

We can port the update and instructions separately yep

By the way, if you were interested in either of these jobs we'd be very happy if you applied: https://ourworldindata.org/why-you-should-come-do-programming-at-oxford But otherwise your assistance is much appreciated :)

sherinkurian:
Thank you @mispy :)

I believe this PR seems to be in a good place. Looking forward to any feedback.

mispy:
Thanks Sherin, looks really good! Could you test it with this file to verify the behavior against the old version? 
[rugged_data.csv.zip](https://github.com/owid/owid-grapher/files/2178794/rugged_data.csv.zip) Then we should be able to merge :)

You might've seen already, but I also uploaded some more complete database exports so you can check out how the code works for all of our charts https://github.com/owid/owid-grapher#database-setup

sherinkurian:
Thank you for the input file! It caught some wrong assumptions in my code (not converting to ASCII and being case sensitive while matching). PR is ready to merge if there are no concerns with below:

1. One small difference between the old and new tool is in the output CSV column headers. I am using the more verbose readable name (eg: `Our World In Data Name`) in the header instead of `owid_name`. Calling it out in case there is a good reason for it to be this way.

Eg: output with current/python tool
```
Country,owid_name,Year,Terrain Ruggedness Index 100m (Nunn and Puga 2012)
Aruba,Aruba,1996,0.462
Afghanistan,Afghanistan,1996,2.518
```

Eg: output with new tool
```
Country,Our World In Data Name,Year,Terrain Ruggedness Index 100m (Nunn and Puga 2012)
Aruba,Aruba,1996,0.462
Afghanistan,Afghanistan,1996,2.518
```

2. Final thing to note is that this tool is accessible from a different link (shown below) to be safe but happy to replace the currently used link with this one.

screenshot 2018-07-11 08 41 39



Thanks a ton for providing more data for charts. Looking forward to using them for tests :)

mispy:
🎉 Thanks Sherin!! We'll test it out for a bit and then replace the old Python one and look into creating a public version based on this :)

PR #69, opened 2018-06-25
improve sample env

sherinkurian:
cc @mispy


mispy:
Thanks! I also removed the need to configure log file location. Eventually we should remove Django and make it a fully node project but that will take a while :)

PR #77, opened 2018-07-12
fix issue when chosen input format does not have an entry 

sherinkurian:
Certain OWID country names (like Rest of the World, World) do not have a corresponding entry for formats like ISO alpha 2 code. This fix will prevent the tool from breaking when those input formats are chosen.

cc @mispy


sherinkurian:
Closing this. Bundling all of the improvements in here https://github.com/owid/owid-grapher/pull/78

Issue #64, opened 2018-06-08
Slope chart data transform should be implemented like other charts


mispy:
The slope chart transform code currently uses this Observations thing I was experimenting with. I ended up deciding not to use that and process data in a simpler way; it should work directly from DimensionWithData.

PR #86, opened 2018-08-04
Slope transform cleanup

sherinkurian:
As requested in https://github.com/owid/owid-grapher/issues/64


js/charts/SlopeChartTransform.ts
```diff
@@ -71,17 +70,36 @@ export default class SlopeChartTransform implements IChartTransform {
...
+    // dimension data saves size a level deeper. eg: { Afghanistan => { 1990: 1, 2015: 10 }}
+    // this returns that data in the form { Afghanistan => 1 }
+    @computed get sizeByEntity(): Map<string, any> {
+        const { sizeDim } = this
+        const sizeByEntity = new Map<string, any>()
+
+        if (sizeDim !== undefined) {
+            sizeDim.valueByEntityAndYear.forEach((yearToSizeMap, entity) => {
+                const values = Array.from(yearToSizeMap.values())
+                sizeByEntity.set(entity, values[0]) // hack: default to the value associated with the first year
```
sherinkurian: defaulting to first year to be consistent with original implementation

js/charts/SlopeChartTransform.ts
```diff
@@ -124,37 +142,35 @@ export default class SlopeChartTransform implements IChartTransform {
     @computed get data(): SlopeChartSeries[] {
         if (!this.yDimension) return []
 
-        const { variableData, sizeDim, yDimension, xDomain, colorDim, colorScale, chart } = this
-        let data = variableData
+        const { yDimension, xDomain, colorByEntity, sizeByEntity, chart } = this
         const entityKey = this.chart.vardata.entityMetaByKey
 
-        // Make sure we're using time bounds that actually contain data
```
sherinkurian: note that this logic of calculating the longest range of years is not ported over. it is not needed looking at the charts on the test page. let me know if i am missing something here

js/charts/SlopeChartTransform.ts
```diff
@@ -124,37 +142,35 @@ export default class SlopeChartTransform implements IChartTransform {
...
+        const entities = yDimension.entitiesUniq
+        let data: SlopeChartSeries[] = entities.map(entity => {
+            const slopeValues: SlopeChartValue[] = []
+            const yValues = yDimension.valueByEntityAndYear.get(entity)
+            if (yValues !== undefined) {
+                yValues.forEach((value, year) => {
+                    if (year === minYear || year === maxYear) {
+                        slopeValues.push({
+                            x: year,
+                            y: typeof value === "string" ? parseInt(value) : value
```
sherinkurian: dirty way to get around ts complaints

PR #103, opened 2018-11-01
Add & deduplicate foreign key constraints

danielgavrilov:
Removes a duplicate foreign key constraint from `datasets.createdByUserId` and adds foreign key constraints to `dataset_files.datasetId` and `sources.datasetId`.


mispy:
Yayy first pull request! 🎉

PR #104, opened 2018-11-02
Create a promise-based exec()

### LINKED ISSUE #100: Node scripts with calls to shell commands should fail fast
Currently we use shell.exec from shelljs: https://github.com/shelljs/shelljs

However, shell.exec doesn't throw an exception on a non-zero exit code, and so things can fail in weird ways. We should replace these execs with stricter ones and (importantly) make sure there's no code that implicitly relies on the existing behavior to work.

Is there a better node shell library, maybe one that uses Promises so we can use async/await too?

danielgavrilov:
Fixes #100.

Creates promise-based `exec()` and `tryExec()` utility functions and removes `shelljs` as a dependency, which [does not yet support async commands](https://github.com/shelljs/shelljs/issues/2).

Things I'm unsure about: 

- I have created `tryExec()` anticipating that in some cases we don't want to throw an error if a command returns a non-zero exit code, but didn't end up using it anywhere. 
- I wrote unit tests but didn't test the other code I changed—it seems most of it deals with deploys and I don't think that would be set up on my machine?
- [Overengineering urge] we should maybe redirect stdout & stderr to the server process to easily see logs?

Also, do mention _any_ comments you might have!

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -134,7 +134,7 @@ module.exports = {
 
   // The glob patterns Jest uses to detect test files
   testMatch: [
-    "<rootDir>/test/*.js",
+    "<rootDir>/test/**/*.js",
 //    "**/?(*.)+(spec|test).js?(x)"
   ],
 
--- a/package.json
+++ b/package.json
@@ -115,7 +115,6 @@
     "save-svg-as-png": "^1.2.0",
     "sharp": "^0.18.4",
     "shell-quote": "^1.6.1",
-    "shelljs": "^0.7.8",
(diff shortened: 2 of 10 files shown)


mispy:
All looks good! 🎉 Makes sense to disambiguate the different exec functions that were hanging around. And awesome that you wrote a test! I'll double check the deploy code.

The logging is pretty ad hoc at the moment. It does some redirection to logs in e.g. https://github.com/owid/owid-grapher/blob/master/src/admin/api.ts#L66. Definitely room for improvement there!

PR #106, opened 2018-11-05
Ensure a git repo exists before attempting to deploy

### LINKED ISSUE #105: ChartBaker should ensure git repository exists at repoDir
Otherwise the commands it runs might commit to the code repository, which is no good

danielgavrilov:
Fixes #105. 

`git init` is supposed to be safe to run in an existing repo. Pretty small change, but since it deals with deploys my confidence levels are very low.


mispy:
It seems you can indeed init another repo from within an existing one! That's neat

PR #107, opened 2018-11-08
Attach Slack error reporting before catch-all middleware

### LINKED ISSUE #101: Fix Slack error reporting
The admin server is supposed to hit a webhook when it encounters an unhandled error to post it to Slack, using an express middleware, but it doesn't seem to be doing so. Probably the other error handling in `app.tsx` that sends stack traces down to the client is interfering with it, it would be good if they played nice together.

danielgavrilov:
Should fix #101, but I haven't tested.

@mispy I was actually wondering:

1. Is there any way to test this?
2. Why are some imports (including this one & cookie parser) still using `require()`?

### DIFF
--- a/src/admin/app.tsx
+++ b/src/admin/app.tsx
@@ -1,6 +1,7 @@
 import * as express from 'express'
 require('express-async-errors')
 const cookieParser = require('cookie-parser')
+const errorToSlack = require('express-error-slack')
 import "reflect-metadata"
 
 import AdminSPA from './AdminSPA'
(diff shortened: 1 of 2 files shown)


mispy:
Thanks Daniel! Good questions.

1. Possibly, but since it relies on integration with an external service and isn't public-facing it's probably not worth the effort.
2. The `require()` is for libraries which don't have published TypeScript definitions, so they come in as `any` types. If you feel like it, a useful upstream OSS contribution would be to write definitions for them and submit them as pull requests to `DefinitelyTyped`.

PR #109, opened 2018-11-13
Show bar chart if variable only contains a single year

### LINKED ISSUE #70: Admin variable previews should pick chart based on kind of data
At the moment it's always maps which doesn't work for non-geographic variables. Should pick like:

Mappable entities => map
Single-year non-mappable data => discrete bar chart
Multi-year non-mappable data => line chart

danielgavrilov:
Should fix #70. 

@mispy Let me know if there's anything you'd do differently, mainly to understand any nuances I might've missed. 

E.g. does the `else` branch need to switch back to a line chart?


mispy:
Looks good! 👍 Thanks for catching the missing dispose. It would make sense to have the else branch switch back (to be explicit about it), but it doesn't need to because it's the default chart type.

PR #111, opened 2018-11-20
Default to variable name if there is no dimension displayName

danielgavrilov:
When a chart is in single entity mode and the dimension has no `displayName`, the labels are empty strings e.g. https://ourworldindata.org/grapher/long-term-price-index-in-food-commodities-1850-2015:

long-term-price-index-in-food-commodities-1850-2015

This PR makes the label default to `variable.name` if there is no `displayName`, which works well for this chart, but not sure if it would work well in other cases?

### DIFF
--- a/js/charts/ChartData.ts
+++ b/js/charts/ChartData.ts
@@ -311,7 +311,7 @@ export default class ChartData {
                 if (isSingleVariable) {
                     label = entity
                 } else if (isSingleEntity) {
-                    label = `${dim.displayName}`
+                    label = `${dim.displayName || variable.name}`
                 }
 
                 keyData.set(key, {


danielgavrilov:
Ah it turns out this is not a new idea: 

https://github.com/owid/owid-grapher/blob/6840bb920514f120259576c29eab9b29dc49dab3/js/charts/DimensionWithData.ts#L19-L21

The problem is that the display name in the chart config for each dimension is set to `""`, overriding anything else. It seems the Display name is correctly defaulted to the variable name when creating a brand new chart, so this was either changed by a human or caused by a change in implementation.

Either way, I think the fix for this (and any other chart that might be affected) would be to set the Display name in the UI?

mispy:
Yep, looks like you identified the problem! It's an issue with the chart configuration (and possibly the editor code that made the chart configuration?), not with the rendering code.

The grapher makes a clear distinction between the meaning of `""` and `undefined` in most places. This is because there are cases where we do want to be able to overwrite a non-blank default with a blank string (I think Max specifically requested it at some point).

danielgavrilov:
Yes, it all makes sense (and really well architected!), I ended up just updating the chart. 🙂

Issue #108, opened 2018-11-10
Basic version history for chart configuration


mispy:
Talking to Esteban, there a couple of cases where published charts have been broken in some way by a mistake in editing and it was tricky to find out what the original settings were. It'd be great to have some way of seeing previous published versions of a chart configuration and rolling back. 

This doesn't have to be fancy, just fulfil the function of being able to undo. Storing old jsons in a table somewhere with a date on them and a little interface in the admin editor to select them is all that is needed.

sherinkurian:
Hi guys, I should have asked a day sooner but hope no one is working on this. Can I claim this :)

mispy:
Thanks Sherin! This is live now. The tweaks I made were mainly to make it like Wordpress revision history (since our authors are likely familiar with how that works) https://github.com/owid/owid-grapher/commit/012cfb435e94c1e61bcd1e4c0bb025e73d2574a3

HannahRitchie:
Thank you very much Sherin for your work on this! It is incredibly useful for us! You're a star :-)

sherinkurian:
@HannahRitchie! So kind of you. Jaiden tells me the team is going to be in California next year. I would love to meet up with you if it works out :)

P.S: @mispy I wouldn't call your changes "tweaks". LOL - Just teasing.. happy to learn.

PR #115, opened 2018-11-23
Sherin/mig audit

sherinkurian:
Track configuration changes to charts as requested in https://github.com/owid/owid-grapher/issues/108

This adds a simple `chart_logs` table to track history of configs. I went back and forth on whether to save the current config being added or the config being replaced. I chose the latter since the History view will always show what was replaced (instead of current + history) and who replaced/invalidated it.

screenshot 2018-11-23 11 50 25

Also, added an index. We may want to add pagination to do this at some point but for now it will only show the the last 50 config updates.

PS: HI Jaiden!! It is thanksgiving holidays out here so I am glad to have found time to relax with owid :)

### DIFF
--- a/src/model/ChartLog.ts
+++ b/src/model/ChartLog.ts
@@ -1,0 +1,36 @@
+import {Entity, PrimaryGeneratedColumn, Column, BaseEntity, ManyToOne, JoinColumn} from "typeorm"
+
+import * as db from '../db'
+import { Chart } from './Chart'
+import User from './User'
+
+@Entity("chart_logs")
+export class ChartLog extends BaseEntity {
+    @PrimaryGeneratedColumn() id!: number
+    @Column() chartId!: number
+    @Column({ type: 'json' }) config: any
+    @Column() userId!: number
+
+    @Column() createdAt!: Date
+    @Column() updatedAt!: Date
+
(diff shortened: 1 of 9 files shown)


src/model/ChartLog.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export function createChartLog(chartId: number|undefined, userId: number, config: any) {
+    if (chartId === undefined) return
+
+    const log = new ChartLog()
+    log.chartId = chartId
+    log.userId = userId
+    log.config = config
+    // TODO: the orm needs to support this but it does not :(
```
sherinkurian: in other words, i was not able to figure this out and gave up 😬

mispy:
Fantastic, thanks Sherin! This will be a very nice little feature to have for reliability and transparency purposes. I might tweak it a little before I merge it.

Little mobx thing: don't use `@action.bound` together with `async`, since an action only applies to code executed synchronously within it. To use actions together with async code you call `runInAction` from within the async function.

I don't know what's up with the timestamps either, I'll have to work that out at some point.

By the way: the OWID team is going to be in California for a few months early next year, so if you feel like it we'll hopefully get a chance to meet up with you! 😄

PR #118, opened 2018-11-27
Namespace filter for test pages

danielgavrilov:
~~**Not ready to merge**~~ Now ready to be reviewed.

Allows specifying a `namespace` query parameter in the test pages, e.g. http://localhost:3030/admin/test/embeds?namespace=wdi

...so that we can show charts from live and test grapher side-by-side:

screenshot 2018-11-27 at 22 27 57

(Something funny happening with the unit here! We should probably discuss separately.)

The purpose is to make it easier to go through the chart differences and flag up things that don't look right. 

Since the import will run on the test grapher first, there will be no (easy) way to fix anything at this point, except flag things. Do you think it's a good idea to show the chart IDs and ask the reviewer to put them in e.g. a spreadsheet? (Or is there some easier way?) We would then run the import on the live grapher and fix all the flagged charts before baking them. 🍪

### DIFF
--- a/src/admin/testPages.tsx
+++ b/src/admin/testPages.tsx
@@ -17,12 +17,17 @@ const IS_LIVE = NODE_BASE_URL === "https://owid.cloud"
 
 const testPages = Router()
 
+interface ChartItem {
+    id: number,
+    slug: string
+}
+
 interface EmbedTestPageProps {
     prevPageUrl?: string,
     nextPageUrl?: string,
     currentPage?: number,
     totalPages?: number,
-    slugs: string[]
+    charts: ChartItem[]
 }
 
 function EmbedTestPage(props: EmbedTestPageProps) {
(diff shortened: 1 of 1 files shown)


src/admin/testPages.tsx
```diff
@@ -13,13 +13,25 @@ import * as querystring from 'querystring'
 import * as _ from 'lodash'
 import * as url from 'url'
 
+const IS_LIVE = NODE_BASE_URL === "https://owid.cloud"
```
danielgavrilov: @mispy Not sure if this is a good way to check whether this is the live environment—let me know if there's a better one!

mispy: That's the way I did it for the little indicator in the navbar, yep. There's probably something more systematic but it works for now.

mispy:
Useful filter to have!

Anything that can be fixed by adding exceptions to the import script (i.e. actual issues with the data) probably should be for reproducibility. 

The other type of issue we might encounter is one where the data is logically imported correctly, but there's some change in context between the WDI versions such that the charts no longer make sense. Those would need to be manually corrected.

I thiiink we can probably wait to see how much that's actually a problem before deciding how to deal with it (if it's hundreds of charts vs. just one or two)

PR #119, opened 2018-11-28
Increase exec() maxBuffer to 10MB

danielgavrilov:
The exec() call was reaching the maxBuffer limit on stdout in the chart baker (with a fresh setup). I think the default limit was 200KB.


danielgavrilov:
This is preventing deploying the static charts from the test server so I will merge it—sorry!

PR #120, opened 2018-11-28
Create a data export script for a namespace

danielgavrilov:
This is basically how I reproduce the charts on a test server for a specific namespace. The dump will contain all variables that are used in charts that have **at least one** variable from the specified namespace. 

Thought it might be useful to have it in the repo.


mispy:
Agreed!

PR #122, opened 2018-11-29
Bring back shelljs

danielgavrilov:
When I removed it, I wondered why shelljs was creating files for stderr, stdout and parameters... I now figured out that it's the most robust way to handle lots of data without hitting the maxBuffer limit.

`netlifyctl` in particular is pretty noisy and easily reaches the maxBuffer limit, which also means all deploys will fail with the existing code, which I think is not good?

Anyway, the return value is now changed to be the exit code as a string (but we don't use it anywhere anyway).

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
     "save-svg-as-png": "^1.2.0",
     "sharp": "^0.18.4",
     "shell-quote": "^1.6.1",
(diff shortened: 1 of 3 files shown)


mispy:
Nice thing about having it all go through the one utility function is that we can swap the underlying implementation out easily 👍

PR #130, opened 2018-12-19
Chart tags

mispy:
Make charts taggable and introduce fast UI for mass tagging charts and datasets


danielgavrilov:
Hmm entering "Uncategorized" while the chart is already Uncategorized will remove the tag. Is this the actual intent? 

I don't quite understand how Uncategorized is supposed to work—can't we just imply that a chart without categories is Uncategorized?

danielgavrilov:
Ah I can see this is already implemented on the database level to treat anything without a tag as "Uncategorized", but the UI still shows a tag (except in this view https://owid.mispy.me/admin/tags/375). Is the reason for showing the "Uncategorized" tag so that people can find the /tags/375 page? Seems like a good reason! 😄

js/admin/Forms.tsx
```diff
@@ -487,4 +488,93 @@ export class Timeago extends React.Component<{ time: Date }> {
...
+@observer
+export class EditableTags extends React.Component<{ tags: Tag[], suggestions: Tag[], onSave: (tags: Tag[]) => void, disabled?: boolean }> {
+    @observable isEditing: boolean = false
+    base: React.RefObject<HTMLDivElement> = React.createRef()
+
+    @observable tags: Tag[] = _.clone(this.props.tags)
+
+    @action.bound onAddTag(tag: Tag) {
+        this.tags.push(tag)
+        this.tags = _.uniq(this.tags).filter(t => t.name !== 'Uncategorized')
```
danielgavrilov: ...just testing out this new GitHub feature 😄

```suggestion
        this.tags = _.uniq(this.tags).filter(t => t.name !== 'Uncategorized')
        this.ensureUncategorized()
```

mispy:
That suggest changes github thing is super nifty!!

Yep, the main purpose of Uncategorized being a special tag is so people can find all the uncategorized stuff (and then categorize it all 😄)

js/admin/Forms.tsx
```diff
@@ -487,4 +487,94 @@ export class Timeago extends React.Component<{ time: Date }> {
...
```
danielgavrilov:
```suggestion
        this.tags = _.uniqBy(this.tags, t => t.id).filter(t => t.name !== 'Uncategorized')

```
The uniq otherwise doesn't do what it should do 🙂

danielgavrilov:
This looks great for mass-tagging! One possible issue is that the ChartEditPage doesn't have a tags UI, but I guess we _probably_ don't want to add tags chart-by-chart, and if there is a need we'll find out about it. 

Another not important but personally annoying issue is that I can't figure out why the "Uncategorized" tag is not shown here: https://owid.mispy.me/admin/tags/375 

For some reason, when loading that page, the `suggestions` are an empty array when `ensureUncategorized()` runs, and only then they get populated. But when opening the Datasets or Charts page, the suggestions are populated before `ensureUncategorized()` runs. Why is that happening?! 😱

mispy:
Thanks for having a look! Yep, I also thought it ought to have tags on the chart editor page as well (and autopopulate them from the tags of the datasets being used by the chart) but I figure I've already spent too much time writing internal admin UI code for this... time to shipp

The suggestions sometimes not being populated was a race condition between the loading and the rendering I think, sometimes EditableTags was mounted before it had suggestions, and since it only ran `ensureUncategorized` in `componentDidMount` it missed it. The right way to do it is to run it in `componentDidUpdate` as well 🙂

PR #158, opened 2019-02-01
Less annoying dropdown

danielgavrilov:
Will now close if clicked and has already been open for 500ms. Will also not open unless mouse has been over it for more than 100ms.


mispy:
ooh, you're using the mobx things! Very good 😄

PR #161, opened 2019-02-06
Change chart logo

danielgavrilov:
@mispy We discussed this with Max—interested to hear your thoughts!

We switched it to use the `Lato` font as we do now for all body text and we changed the colours to "Oxford Blue" and "New Red", which we use in the current header. It seems you intentionally made the colours less opaque on charts so happy to hear if you have any comments.


mispy:
I like it, is more subtle now!

PR #164, opened 2019-02-22
Link tracking for header, footer & homepage

danielgavrilov:
It makes navigation slower but gives us more data about what people click on. For every link, there is up to a 350ms delay before navigating away from a page, in order to have time to send an event to Amplitude.

Google Analytics can tell us which pages were visited next, but since we have duplicate links in the header, footer and across the homepage, without this we don't know which link they clicked on.


danielgavrilov:
One problem with that for people with ad blockers (like me) the Amplitude SDK loads fine, but requests to the API are blocked. In that case, the `Amplitude.logEvent()` function never calls the passed callback, so every click takes the maximum wait time (350ms). Feels super slow!

PR #171, opened 2019-05-27
Fix suppoerters link in footer

madhums:
Supporters link in the footer seems to be broken. This PR fixes it.


danielgavrilov:
Thanks a lot for the PR @madhums!

PR #273, opened 2019-11-28
isExplorable chart field

danielgavrilov:
- Adds an `EXPLORER` flag in settings (need to set `EXPLORER=true` in your local `.env`)
- An `isExplorer` column on `charts` table
- An "Explorable chart" checkbox in the Grapher admin
- Logic to ensure map & chart tab are force enabled for explorable charts
- Logic to ensure only `LineChart` and `DiscreteBar` can be made explorable (arbitrarily chosen, can change)

Notion: https://www.notion.so/owid/Boolean-is_explorable-on-Chart-310eafedc0924ad78afb616a718c8200

### DIFF
--- a/admin/client/EditorBasicTab.tsx
+++ b/admin/client/EditorBasicTab.tsx
@@ -12,6 +12,7 @@ import { Toggle, SelectField, EditableList, FieldsRow, Section } from "./Forms"
 import { ChartEditor } from "./ChartEditor"
 import { VariableSelector } from "./VariableSelector"
 import { DimensionCard } from "./DimensionCard"
+import { canBeExplorable } from "utils/charts"
 
 @observer
 class DimensionSlotView extends React.Component<{
@@ -199,16 +200,30 @@ export class EditorBasicTab extends React.Component<{ editor: ChartEditor }> {
                         options={ChartTypeDefs.map(def => def.key)}
                         optionLabels={ChartTypeDefs.map(def => def.label)}
                     />
+                    {editor.features.explorer && (
+                        <FieldsRow>
+                            <Toggle
+                                label="Explorable chart"
+                                value={chart.props.isExplorable}
+                                onValue={value =>
+                                    (chart.props.isExplorable = value)
+                                }
+                                disabled={!canBeExplorable(chart.props)}
+                            />
+                        </FieldsRow>
+                    )}
                     <FieldsRow>
                         <Toggle
                             label="Chart tab"
                             value={chart.props.hasChartTab}
                             onValue={value => (chart.props.hasChartTab = value)}
+                            disabled={chart.props.isExplorable}
                         />
                         <Toggle
                             label="Map tab"
                             value={chart.props.hasMapTab}
                             onValue={value => (chart.props.hasMapTab = value)}
+                            disabled={chart.props.isExplorable}
                         />
                     </FieldsRow>
                 </Section>
--- a/charts/ChartConfig.tsx
+++ b/charts/ChartConfig.tsx
@@ -25,6 +25,7 @@ import { IChartTransform } from "./IChartTransform"
 import { ChartDimension } from "./ChartDimension"
 import { TooltipProps } from "./Tooltip"
 import { LogoOption } from "./Logos"
+import { canBeExplorable } from "utils/charts"
 
 declare const App: any
 declare const window: any
@@ -106,6 +107,7 @@ export class DimensionSlot {
 // under the same rendering conditions it ought to remain visually identical
 export class ChartConfigProps {
     @observable.ref type: ChartTypeType = "LineChart"
+    @observable.ref isExplorable: boolean = false
 
     @observable.ref id?: number = undefined
     @observable.ref createdAt?: Date = undefined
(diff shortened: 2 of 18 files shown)


admin/client/EditorBasicTab.tsx
```diff
@@ -199,16 +200,30 @@ export class EditorBasicTab extends React.Component<{ editor: ChartEditor }> {
...
                     />
+                    {editor.features.explorer && (
+                        <FieldsRow>
+                            <Toggle
+                                label="Explorable chart"
+                                value={chart.props.isExplorable}
+                                onValue={value =>
+                                    (chart.props.isExplorable = value)
+                                }
+                                disabled={!canBeExplorable(chart.props)}
```
jasoncrawford: Why is this a utility function, instead of a method on the chart object?

charts/ChartConfig.tsx
```diff
@@ -270,6 +272,17 @@ export class ChartConfig {
                 this.props.dimensions = this.validDimensions
             }
         })
+
+        autorun(() => {
+            if (this.props.isExplorable) {
+                if (canBeExplorable(this.props)) {
+                    this.props.hasChartTab = true
+                    this.props.hasMapTab = true
```
jasoncrawford: I thought a component was not supposed to update its own props?

test/charts/ChartConfig.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    it("explorer not available for scatter plots", () => {
+        const props = new ChartConfigProps()
+        props.type = "ScatterPlot"
+        props.hasChartTab = true
+        props.isExplorable = true
+        const config = createConfig(props)
+        expect(config.isExplorable).toBe(false)
+    })
+})
```
jasoncrawford: yay tests! 🙌

utils/charts.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// A centralized predicate to test whether a chart can be explorable.
+// Used for validation on both server & client.
+export function canBeExplorable(config: ChartConfigProps) {
+    return EXPLORABLE_CHART_TYPES.includes(config.type)
+}
+
+export function isExplorable(config: ChartConfigProps): boolean {
+    return config.isExplorable && canBeExplorable(config)
+}
```
jasoncrawford: Again not sure why these are in a separate utils file instead of just being methods on ChartConfigProps or something?

admin/client/EditorBasicTab.tsx
danielgavrilov: So my thought process was:
- `ChartConfigProps` doesn't have any methods and the convention is to treat it as a dumb object that has no behaviour, only fields.
- The constructor on `ChartConfig` initializes a whole bunch of stuff, including data loading, so having it there is not an option if we want to check whether a chart is explorable on the back-end.
- Checking for `isExplorable` on the back-end is important in case we change the logic, e.g. we exclude bar charts. Then, having this check on the back-end will make sure a previously created bar chart doesn't get published as an explorable chart. The alternative is that we migrate the config when we make a change to the logic, which is more error-prone.

Do you think it's better to turn `ChartConfigProps` into an object with behaviour?

jasoncrawford: What's the motivation for making it a dumb object with no behavior? I generally feel it's better to put behavior on the object rather than to have a random separate file with utility functions that apply to that one object.

Needing the logic on the back-end makes sense to me, would be nice if there were a more elegant way to do it, but this seems fine

charts/ChartConfig.tsx
danielgavrilov: It's true that a component shouldn't _directly_ modify its props, and this codebase breaks that principle in many cases. It's basically built on breaking that principle – the chart config is often changed directly within components. See e.g. how the settings menu works: 

https://github.com/owid/owid-grapher/blob/b3b70fbb8eb7fad2d02935d416effe562a88e133/charts/Controls.tsx#L218-L220

Whether it's a good idea is debatable – it makes arbitrary config changes easy to make at the cost of making components harder to reason about. I would prefer the conventional React approach where mutating functions are explicitly passed, but that _would_ add a lot of boilerplate code.

Anyway, I was following the convention for applying "constraints" to config props; see the other autoruns above my additions. We can of course change that, how do you think we should ideally apply constraints to the config?

jasoncrawford: OK. If this is already an established pattern let's just keep it for now. We can talk separately about whether we want to change to a different pattern, but it doesn't nee to hold up this PR.

admin/client/EditorBasicTab.tsx
danielgavrilov: The `class ChartConfigProps` should really be an `interface` since we are **always** treating it as such and it's a useful no-behaviour thing to have. We can then have an object with behaviour that accepts `ChartConfigProps` and is used to set & get props. And then finally the `ChartConfig` that has all the initialization behaviour – fetching the data, passing it to all the objects, etc.

I would be in favour of re-architecting it that way but I think it's maybe not a trivial amount of work.

danielgavrilov: Maybe it's not too involved, we just need to move all the types to an `interface` and have the class `implements` that interface. Then it's a question of whether mobx `@observable` would have an identical behaviour as it does now applied to individual properties. Still, it's some work to verify and test that...

jasoncrawford: Cool. Up to you if you want to do that work now or leave it for another PR, I'm good with merging this

danielgavrilov: I think it's better to leave as a refactoring task, wouldn't like to dive into that now.

PR #275, opened 2019-12-02
Explore view

jasoncrawford:
Add an ExploreView to the ExplorePage. Currently contains a ChartView (with a hardcoded chart).

Some minor details not handled yet, noted those in TODOs.

Notion: https://www.notion.so/owid/DataExplorerView-component-f27979015b9c48a68a5b2074f0636155


jasoncrawford:
Added tests, prettified. @danielgavrilov please review, thanks!

charts/ExploreView.tsx
```diff
@@ -0,0 +1,46 @@
...
+    dimensions: [{ display: {}, property: "y", variableId: 104402 }],
+    selectedData: [{ index: 0, entityId: 355 }]
+}
+
+// This component was modeled after ChartView.
+//
+// TODO that ChartView handles but this doesn't:
+// * re-render on window resize event (throttled)
+// * FullStory event logging on bootstrap
+// * error logging via Analytics.logEvent on componentDidCatch
```
danielgavrilov: Is the intent eventually that this will replace `ChartView` or would it always be a wrapper?

jest.config.js
```diff
@@ -1,10 +1,33 @@
...
+            displayName: "server",
+            testEnvironment: "node",
+            testMatch: ["<rootDir>/test/**/?(*.)+(spec|test).[jt]s?(x)"]
+        },
+        {
+            ...common,
+            displayName: "client",
+            testEnvironment: "jsdom",
+            setupFilesAfterEnv: ["<rootDir>/test/enzymeSetup.ts"],
+            testMatch: ["**/__tests__/**/*.[jt]s?(x)"]
```
danielgavrilov: I can see this works fine but I don't understand the regex: it's a `j` or `t` followed by an optional `s` and a required `x`? Is it not supposed to be `[jt]sx?` instead?

charts/ExploreView.tsx
jasoncrawford: ExploreView will always contain a ChartView, I think

jest.config.js
jasoncrawford: Yeah it's a weird regex. I was confused too. It actually does mean what you think it ought to be saying. Instead of `x?` they use `?(x)`. See https://github.com/micromatch/micromatch#extended-globbing

PR #276, opened 2019-12-03
Add CI checks: prettier & typecheck

danielgavrilov:
- Moves prettier check into `yarn testcheck`, which is basically the command that runs _all_ pre-deploy checks
- Runs `yarn testcheck` on Travis CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
         "typecheck": "tsc --noEmit",
         "build": "rm -rf dist && ENV=production webpack -p --progress",
         "test": "jest",
-        "testcheck": "npm-run-all --parallel typecheck test",
+        "testcheck": "yarn test && yarn typecheck && yarn prettify:check",
         "deploy": "./bin/deploy.sh",
         "serve": "yarn tsn admin/server/server.tsx",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
         "typecheck": "tsc --noEmit",
         "build": "rm -rf dist && ENV=production webpack -p --progress",
         "test": "jest",
-        "testcheck": "npm-run-all --parallel typecheck test",
+        "testcheck": "yarn test && yarn typecheck && yarn prettify:check",
```
jasoncrawford: Why not run in parallel? Seems that helps speed up build/deploy?

danielgavrilov: It garbles the output since they all write to stdout. Some of them also have the fancy loading indicators that completely overwrite messages...

If we only care about a final pass/fail then we can do that, `npm-run-all` will have a non-zero exit code if a single or more processes have a non-zero exit code.

But if we care about seeing what the errors are, then it's better to do it this way.

danielgavrilov:
👇 Feel free to hit the big green button!

PR #278, opened 2019-12-04
Add .nvmrc specifying Node version 12

jasoncrawford:
Just specifying "12" matches our travis.yml file, which according to @danielgavrilov is the one other place this is specified.

Unclear if a .nvmrc file is the best way to do this. I thought the canonical place would be in package.json, but apparently nvm doesn't read that by default, and I read somewhere that it's only “advisory”. I didn't spend a long time researching this.

Also unclear if we should be specifying a minor or even patch level? That might be overly prescriptive, but it could also be good to make sure all developers are using the exact same version of Node, and that it's what we deploy with too. The latest LTS Node is 12.13.1, we could specify that (and maybe match it in travis.yml). Thoughts?


danielgavrilov:
`.nvmrc` sounds like the right place and yes I also think it should be pinned down to the patch level.

Currently the server is also running `v12.13.1` so let's go with that?

jasoncrawford:
> Currently the server is also running `v12.13.1` so let's go with that?

@danielgavrilov Cool. Should I update travis.yml also?

danielgavrilov:
@jasoncrawford Yeah it's probably a good idea?

PR #279, opened 2019-12-04
Chart types in Explorer

jasoncrawford:
(Now ready for review) This adds a set of chart type buttons to the ExploreView; clicking a button changes the chart type.

Also adds some testing support:
* xhr-mock library for mocking HTTP calls from React components
* VSCode launchers for Jest

Ignore the individual commits, which are pretty messy, and just review the final diff.


charts/ExploreView.tsx
```diff
@@ -27,19 +49,95 @@ const DUMMY_JSON_CONFIG = {
...
+        // to be recreated all the time. They aren't pure value objects and have behaviors on
+        // instantiation that include fetching data over the network. Instead, we rely on their
+        // observable properties, and on this autorun block to connect them to the Explore controls.
+        // -@jasoncrawford 2019-12-04
+        autorun(() => {
+            this.chart.tab = this.tab
+            this.chart.props.type = this.configChartType
+            this.chart.props.hasMapTab = this.isMap
+            this.chart.props.hasChartTab = !this.isMap
+        })
```
danielgavrilov: It feels a bit icky having these synced up like this. This ensures changes to ExploreView get applied to the ChartView, but not the other way around... but I think it's fine for what we are doing. 

One potential issue is that the tab will get switched whenever any of the properties change, which is again probably fine for now, but just something to watch out for, since the user can switch to the source, data & download tabs.

On the nitpicky side, `autorun` returns a disposer that should run in `componentWillUnmount` like here: https://github.com/owid/owid-grapher/blob/b3b70fbb8eb7fad2d02935d416effe562a88e133/admin/client/VariableSelector.tsx#L382-L384

We always store it in an attribute: https://github.com/owid/owid-grapher/blob/b3b70fbb8eb7fad2d02935d416effe562a88e133/admin/client/VariableSelector.tsx#L364

jasoncrawford: > This ensures changes to ExploreView get applied to the ChartView, but not the other way around

What do you mean? What's an example?

> the tab will get switched whenever any of the properties change

Not sure what this refers to either?

> autorun returns a disposer that should run in componentWillUnmount

Great feedback, thanks for catching that! Will fix

jasoncrawford:
@danielgavrilov that patch looks good to me, thanks! I will merge it into my branch

charts/ExploreView.tsx
danielgavrilov: > > This ensures changes to ExploreView get applied to the ChartView, but not the other way around
> 
> What do you mean? What's an example?

I meant `this.configChartType` is not guaranteed to be `== this.chart.props.type`, and same for `this.tab`. They shouldn't be used to check the current state basically, they are only pushed to the chart config when any property used within the `autorun` changes. The user could've selected the sources tab for example – `config.props.tab` would be `"sources"` while the value of `this.tab` would be `"chart"`.

One example I didn't think of earlier and that complicates things a bit is that when a user selects a Line Chart and collapses the timeline to a single year, the chart becomes a `DiscreteBar`. This is not reflected in `chart.props.type` right now, but the information needs to reach the explorer UI somehow so that the right type is highlighted.

> > the tab will get switched whenever any of the properties change
> 
> Not sure what this refers to either?

`autorun` runs whenever any of the properties used within it change. So a change to any of `this.tab`, `this.isMap` or `this.chartConfigType` will force a re-sync for all properties, and as I mentioned above, they may not necessarily be in sync.

---

As I said, this is all fine for now! Just expressing what I'm thinking of when I look at the code.

jasoncrawford: Hmm, good point, I'd forgotten about the sources/data tabs. I experimented with this. The result of this is just that if you click Sources or Data, and then you change the chart type, it removes the overlay and pops you back to the chart/map. I think that's fine behavior.

jasoncrawford:
@danielgavrilov I think this is ready. I finished the tests and styled the buttons.

Note the icons are placeholder, Font Awesome doesn't have all the icons we need for this… someone needs to sit down and craft them, or perhaps grab a set from The Noun Project or something

Also I didn't factor out a separate component for the chart types. We can do that later if ExploreView gets too big, right now it's only 170 lines

Take a look at the latest?

charts/ExploreView.tsx
```diff
@@ -1,9 +1,21 @@
...
+import { observer } from "mobx-react"
+import { extend } from "lodash"
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
+import { IconDefinition } from "@fortawesome/fontawesome-svg-core"
+import {
+    faChartLine,
+    faChartBar,
+    faChartArea,
+    faMap
+} from "@fortawesome/free-solid-svg-icons"
```
danielgavrilov: This is the case where webpack doesn't do a good job treeshaking – you can see that the build has now grown from 1.8MB to 2.4MB. 

You can import the icons like this to avoid the problem:
```js
import { faChartLine } from "@fortawesome/free-solid-svg-icons/faChartLine"
import { faChartBar } from "@fortawesome/free-solid-svg-icons/faChartBar"
import { faChartArea } from "@fortawesome/free-solid-svg-icons/faChartArea"
import { faMap } from "@fortawesome/free-solid-svg-icons/faMap"
```

danielgavrilov: Fixed this! (Hope you don't mind me commiting, I only wanted to merge this into my branch quicker so I can build on it without conflicts)

jasoncrawford: I see. Cool, thanks for the tip

PR #280, opened 2019-12-06
Add basic tests for ChartPage

jasoncrawford:
This is the result of an aborted attempt to write more extensive tests for ChartPage. I was going to test the URL binding functionality at this level. However, I realized that to do this requires a full browser-to-server integration test.

We should really set up that kind of test at some point, but I don't want to do all that now just to test some simple URL binding. So I'm punting on this for now.

But by that point I had already written these simple shallow-render tests, and I figured they were still useful, so I'm adding them.


danielgavrilov:
Thanks Jason, makes sense!

PR #281, opened 2019-12-06
URL binding refactor

jasoncrawford:
Ready for review.

For motivation/plan, see the notes on https://www.notion.so/owid/Chart-type-in-URL-params-4e7cd507297c48afabb6c0e66762e248. This is part 1, the refactor.

* Replace `URLBinder` with a `ChartUrl` object and a static utility, `UrlBinding`
* Move all manipulation of `window` and `document` into a method `bindToWindow` that is only ever invoked from top-level JavaScript
* Also move to the top level: grabbing the global window query string to configure the chart with
* Remove `isSinglePage` from ChartConfig
* Rename and refactor some methods for clarity, mostly in the URL utils

### DIFF
--- a/charts/ChartView.tsx
+++ b/charts/ChartView.tsx
@@ -85,7 +85,7 @@ export class ChartView extends React.Component<ChartViewProps> {
                 chart.props.hideConnectedScatterLines,
             compareEndPointsOnly_bool: chart.props.compareEndPointsOnly,
             entityType_str: chart.entityType,
-            isSinglePage_bool: chart.isSinglePage,
+            isEmbed_bool: chart.isEmbed,
             hasChartTab_bool: chart.hasChartTab,
             hasMapTab_bool: chart.hasMapTab,
             tab_str: chart.tab,
--- a/charts/ChartView.tsx
+++ b/charts/ChartView.tsx
@@ -1,6 +1,6 @@
 import * as React from "react"
 import * as ReactDOM from "react-dom"
-import { observable, computed, action } from "mobx"
+import { observable, computed, action, autorun } from "mobx"
 import { observer } from "mobx-react"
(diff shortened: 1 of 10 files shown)


charts/ChartView.tsx
```diff
@@ -69,6 +70,7 @@ export class ChartView extends React.Component<ChartViewProps> {
 
         render()
         window.addEventListener("resize", throttle(render))
+        urlBinding.bindUrlToWindow(chart.url)
```
danielgavrilov: There should be an `isSinglePage` check here?

jasoncrawford: Good catch… but this is in the bootstrap call, wouldn't `isSinglePage` always be true in that case?

danielgavrilov: `bootstrap` is also used for the (iframeless) embeds: https://github.com/owid/owid-grapher/blob/d0b59de2fb930aec763c7aa24bee07db9845fa20/site/client/Grapher.ts#L81

jasoncrawford: I moved this to a different function that is only called from the top level, and I got rid of isSinglePage completely

charts/ChartView.tsx
```diff
@@ -417,8 +418,16 @@ export class ChartView extends React.Component<ChartViewProps> {
...
 
-    componentDidMount() {
+    // Binds chart properties to global window title and URL. This should only
+    // ever be invoked from top-level JavaScript.
+    bindToWindow() {
         window.chartView = this
+        window.chart = this.chart
+        urlBinding.bindUrlToWindow(this.chart.url)
+        autorun(() => (document.title = this.chart.data.currentTitle))
+    }
```
danielgavrilov: Great to have these in one place!

PR #282, opened 2019-12-09
Indicator switching

danielgavrilov:
**EDIT 2019-12-12**: Now ready to review for merge.

This implements:

- An `/expore/indicators.json` file that contains all explorable charts
- Forcing a few select chart IDs to be explorable (see `utils/charts.ts:12` for reasoning behind this)
- A dropdown to search & select a single indicator
- An `ExploreViewContext` that passes a store of indicators containing fetching & caching logic
- A global store containing logic for loading indicators

Left to do:
- [x] Refactor store _(added 2019-12-10)_
- [x] Tests for indicator switching

One problem is that the chart disappears in between switching indicators; I haven't yet looked into this, but it might be an issue for another PR.

~~No tests yet for the indicator switching logic, wanted to hear @jasoncrawford's feedback on how this is implemented – specifically the store that is passed through a context and the async handling.~~ Tests now added.

Some thoughts/opinions behind this:
1. We want to decouple the fetching logic from the rest of the code in order to make it easier to test & maintain.
2. ~~Having `get()` and `search()` be async means we can more easily swap the current static file with an API in the future.~~ We only want `search()` to be async, `get()` should be synchronous to avoid asynchronous code.
3. We should eventually move variable data to the store too.


charts/IndicatorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+@observer
+export class IndicatorDropdown extends React.Component<IndicatorDropdownProps> {
+    static contextType = ExplorerViewContext
+    context!: ExplorerViewContextType
+
+    static defaultProps = {
+        placeholder: "Type to search..."
+    }
+
+    @observable.ref currentIndicator?: Indicator | null
```
jasoncrawford: TypeScript question: since this is marked optional with `?`, isn't the `| null` redundant?

charts/IndicatorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+            this.props.selectedId &&
+            this.props.selectedId !==
+                (this.currentIndicator && this.currentIndicator.id)
+        ) {
+            const fetchedId = this.props.selectedId
+            const indicator = await this.context.indicatorStore.get(
+                this.props.selectedId
+            )
+            // this.props.selectedId could've changed during the await, need to
+            // make a second check.
```
jasoncrawford: I'm not totally sure what's causing this “could have changed during the wait” stuff, but this is the second place it's shown up so far. Seems awkward and error-prone. Why is this needed? Can we re-architect or re-design somehow to eliminate this?

charts/IndicatorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+        } else if (!this.props.selectedId) {
+            this.currentIndicator = null
+        }
+    }
+
+    @bind onChange(indicator: ValueType<Indicator>) {
+        // The onChange method can return an array of values (when multiple
+        // items can be selected) or a single value. Since we are certain that
+        // we are not using the multi-option select we can force the type to be
+        // a single value.
```
jasoncrawford: Nice comment

charts/IndicatorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+        this.props.onChangeId((indicator as Indicator).id)
+    }
+
+    @bind async loadOptions(input: string): Promise<Indicator[]> {
+        return await this.context.indicatorStore.search({
+            query: input
```
jasoncrawford: Super-minor but when it makes sense, I like to name my variables the same as their keys to take advantage of the more succinct object syntax. If you name the param `query` then this could just be `.search({query})`

charts/IndicatorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+    render() {
+        return (
+            <AsyncSelect
+                className="indicator-dropdown"
+                onChange={this.onChange}
+                placeholder={this.props.placeholder}
+                defaultOptions={true}
+                loadOptions={this.loadOptions as any}
+                getOptionValue={this.getValue as any}
+                getOptionLabel={this.getLabel as any}
```
jasoncrawford: Why are the `as any`s needed here?

charts/IndicatorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+            />
+        )
+    }
+}
+
+export class SingleValue extends React.Component<SingleValueProps<Indicator>> {
+    static contextType = ExplorerViewContext
+    context!: ExplorerViewContextType
+
+    @observable.ref indicator?: Indicator | null
```
jasoncrawford: same comment here about `?` vs. `| null`

charts/client/chart.scss
```diff
@@ -1,7 +1,7 @@
 $sans-serif-font-stack: Lato, "Helvetica Neue", Arial, sans-serif;
 $serif-font-stack: "Playfair Display", Georgia, "Times New Roman", serif;
 
-$controls-color: #3F9EFF;
+$controls-color: #3f9eff;
```
jasoncrawford: Looks like a bunch of unrelated style changes in this file owing to Prettier? Not a big deal but personally I would be inclined to do a separate PR for that. It's a bit of a pain but it keeps unrelated changes apart.

site/server/devServer.tsx
```diff
@@ -98,6 +99,12 @@ devServer.get("/explore", async (req, res) => {
     res.send(await renderExplorePage())
 })
 
+devServer.get("/explore/indicators.json", async (req, res) => {
+    // TODO remove this doing-undoing madness, it's only here to send correct
+    // mimetype so that the browser extension for syntax highlighting kicks in.
+    res.json(JSON.parse(await renderExplorableIndicatorsJson()))
```
jasoncrawford: I think you can just do `res.type('json').send(…)`

Alternately: `res.format({json: () => …})`

charts/IndicatorStore.ts
```diff
@@ -0,0 +1,59 @@
+import { fetchJSON } from "./Util"
+import { Indicator } from "./Indicator"
+import { BAKED_BASE_URL } from "settings"
+
+export class IndicatorStore {
```
jasoncrawford: Have we thought about using an existing solution for this, like Redux or Relay or something?

charts/IndicatorStore.ts
```diff
@@ -0,0 +1,59 @@
...
+    private async fetchAll() {
+        try {
+            const json = await fetchJSON(
+                `${BAKED_BASE_URL}/explore/indicators.json`
+            )
+            const indicators: Indicator[] = json.indicators
+            return indicators
+        } catch (error) {
+            console.error("Failed to fetch indicators", error)
+            return []
```
jasoncrawford: If we fail to fetch, do we want to return an empty array, or trigger some sort of error?

charts/IndicatorStore.ts
```diff
@@ -0,0 +1,59 @@
...
+                resolve()
+            })
+        }
+        return this.fetchPromise
+    }
+
+    private populate(indicators: Indicator[]) {
+        for (const indicator of indicators) {
+            this.indicatorsById[indicator.id] = indicator
+        }
```
jasoncrawford: Isn't there a util for this? `byId` or `byKey` or something

charts/IndicatorStore.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    async search(props: { query: string }): Promise<Indicator[]> {
+        await this.loadAllIdempotent()
+        const { query } = props
+        const indicators = Object.values(this.indicatorsById)
+        if (!query) {
+            // If there is no search query, return full list
+            return indicators
+        } else {
```
jasoncrawford: Again minor, and related to my previous style comment: there's no need for else after return. (I have even worked in codebases where this was disallowed by the style checker…)

charts/IndicatorStore.ts
```diff
@@ -0,0 +1,59 @@
...
+            return indicators.filter(indicator => {
+                const titleLower =
+                    (indicator.title && indicator.title.toLowerCase()) || ""
+                const queryLower = query.toLowerCase()
```
jasoncrawford: Pretty minor, but you could move this outside the inner function for a tiny perf increase

charts/Util.ts
```diff
@@ -423,3 +423,30 @@ export function urlToSlug(url: string): string {
...
+                }
+            }
+        })
+        req.open("GET", url)
+        req.send()
+    })
+    // window.fetch() implementation below. Decided to use XMLHttpRequest for
+    // the time being, since mocking window.fetch() seemed not to allow
+    // specifying an endpoint, rather you just intercept function calls, which
+    // feels more verbose.
```
jasoncrawford: Yeah, this is weird, the Jest mock stuff and even the packages explicitly for mocking fetch don't seem to work the way xhr-mock works. I agree with you that specifying url patterns is better. We could easily roll our own stuff for this using Jest `mock`, and I think using `fetch` would be more elegant than XHRs, but this is fine for now

charts/ExploreView.tsx
```diff
@@ -77,34 +72,66 @@ export class ExploreView extends React.Component<ExploreProps> {
     // This is different from the chart's concept of chart type because it includes "WorldMap" as
     // an option, and doesn't include certain chart types we don't support right now, such as
     // scatter plots
-    @observable chartType: ExplorerChartType = ChartType.LineChart
+    @observable.ref chartType: ExplorerChartType = ChartType.LineChart
```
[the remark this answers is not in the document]
danielgavrilov: > Do we just use it by default unless there's a reason not to? 

Yeah, it doesn't really matter here. But I guess it's better if reference comparison is the default (less mobx overhead), forcing you to make a conscious decision when you want nested properties to be observed.

charts/ExploreView.tsx
```diff
@@ -77,34 +72,66 @@ export class ExploreView extends React.Component<ExploreProps> {
...
-        // instantiation that include fetching data over the network. Instead, we rely on their
-        // observable properties, and on this autorun block to connect them to the Explore controls.
-        // -@jasoncrawford 2019-12-04
-        this.dispose = autorun(() => {
-            this.chart.props.type = this.configChartType
-            this.chart.props.hasMapTab = this.isMap
-            this.chart.props.hasChartTab = !this.isMap
-            this.chart.tab = this.tab
-        })
+    componentDidMount() {
```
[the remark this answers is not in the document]
danielgavrilov: It's not a very clear-cut good call, `componentDidMount()` runs after `render()` so components get unpopulated config. But the React convention is to have light constructors... so anyway, happy to do it either way

charts/ExploreView.tsx
```diff
@@ -77,34 +72,66 @@ export class ExploreView extends React.Component<ExploreProps> {
...
+        this.disposers = [
+            // We need these updates in an autorun because the chart config objects aren't really meant
+            // to be recreated all the time. They aren't pure value objects and have behaviors on
+            // instantiation that include fetching data over the network. Instead, we rely on their
+            // observable properties, and on this autorun block to connect them to the Explore controls.
+            // -@jasoncrawford 2019-12-04
+            // Turned the autorun into a reaction in order to make the dependencies (type & isMap) more
+            // explicit -@danielgavrilov 2019-12-09
```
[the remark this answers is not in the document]
danielgavrilov: Yeah sorry for doing this in this PR, but I was looking at it, and it's really only about excluding `ExplorerView.tab` from the dependent properties... 

`ExploreView.tab` is a scary prop for me because its meaning is "what the tab property of chart config should be set to when the chart type is switched", it's not "the current tab" which I expect. 

I pushed a change to remove the tab property which I think is the better approach here possibly. If you disagree I am happy to return it as it was since it's out of scope for this PR (sorry for starting this whole thing)

charts/ExploreView.tsx
```diff
@@ -156,12 +183,33 @@ export class ExploreView extends React.Component<ExploreProps> {
...
 
-    render() {
+    renderIndicatorSwitching() {
         return (
-            <div>
-                {this.renderChartTypes()}
-                <ChartView chart={this.chart} bounds={this.bounds} />
+            <div className="indicator-bar">
+                <IndicatorDropdown
+                    placeholder="Select X axis variable"
```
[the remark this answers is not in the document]
danielgavrilov: :+1: changed!

charts/IndicatorDropdown.tsx
danielgavrilov: Without `| null` you can't later do:
```js
this.currentIndicator = null
```
That will be a type error. You can still set it to `undefined` though... I chose `null` because `undefined` is set while the indicator is still loading.

charts/IndicatorDropdown.tsx
danielgavrilov: It's very fiddly, I agree it should be handled differently.

It's to handle this case:

```
switches to indicator 1 ---> waits impatiently ---> switches to indicator 2 ---> waits ---> indicator 2 shown ---> indicator 1 shown
                          \                      \                                      /                      /
                           \                      -- indicator 2 loading ------ loaded -                      /
                            \                                                                                /
                              -- indicator 1 loading ----------------------------------------------- loaded--
                                                  X at this point, indicator 1 should be cancelled
                                                  but promises are not cancellable
```

Indicator 1 gets delayed but its handler is still called last, so the current indicator is Indicator 1 despite the user choosing Indicator 2.

danielgavrilov: I will work on this, I agree it's a problem

charts/IndicatorDropdown.tsx
danielgavrilov: :+1: I agree, changed

charts/IndicatorDropdown.tsx
danielgavrilov: The true answer is because I was lazy to figure out the right types. Now fixed

But generally, I think this way of rendering labels is temporary – we should use our own components to render them so that we can change layout/style, e.g. showing sourceDesc in a new line, with a smaller font, etc.

charts/IndicatorDropdown.tsx
danielgavrilov: Same answer 😄

charts/IndicatorStore.ts
danielgavrilov: Yeah implemented the equivalent with lodash now

charts/client/chart.scss
danielgavrilov: :+1: I agree

site/server/devServer.tsx
danielgavrilov: Ah, thanks!

charts/IndicatorStore.ts
danielgavrilov: :+1: I just like the symmetry, but really don't mind it either way. Removed the `else` now

charts/IndicatorStore.ts
danielgavrilov: We should definitely let the user know, but before that we should also retry the request a few times, since there isn't much the user can do to fix it.

We don't have anything in the current codebase to show error dialogs or retry attempts, and this PR is a bit big as it is – maybe we should tackle that in another task?

charts/IndicatorStore.ts
danielgavrilov: :+1: Will answer in main thread

charts/ExploreView.tsx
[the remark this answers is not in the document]
jasoncrawford: Oh, hmm. Yeah, I was not thinking about the difference between componentDidMount and componentWillMount. Except it appears that componentWillMount is now deprecated, for reasons I don't understand.

According to the React docs, if you do this (update state in componentDidMount), then you end up rendering a second time, but both renders happen before the actual DOM render, so there should be no flashing on screen at least.

So yeah, agree it's not a clear-cut call… maybe it's actually better in the constructor?

charts/ExploreView.tsx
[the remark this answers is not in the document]
jasoncrawford: Not sure why `tab` is scary? If it's just confusing, we could rename it to something more clear?

Not sure what distinction you're making between "what the tab property of chart config should be set to when the chart type is switched" and "the current tab"?

Anyway `tab` is completely a function of `isMap`, which is itself a function of `chartType`, so … I still don't understand the goal of this?

I'll review your new code though.

charts/IndicatorDropdown.tsx
jasoncrawford: I see, the `?` only allows undefined, not null. Didn't realize that

charts/IndicatorDropdown.tsx
jasoncrawford: OK. We can discuss, I probably have ideas

charts/IndicatorStore.ts
jasoncrawford: Yeah sure

charts/ExploreView.tsx
```diff
@@ -95,18 +101,12 @@ export class ExploreView extends React.Component<ExploreProps> {
...
-                    this.chart.tab = this.tab
-                },
-                { fireImmediately: true }
-            ),
+            autorun(() => {
+                this.chart.props.type = this.configChartType
+                this.chart.props.hasMapTab = this.isMap
+                this.chart.props.hasChartTab = !this.isMap
+                this.chart.tab = this.isMap ? "map" : "chart"
+            }),
```
jasoncrawford: OK I see, this is good

charts/IndicatorDropdown.tsx
jasoncrawford: So I think a more elegant solution, that would be more in line with the reactive style that React/mobx are meant for, would be to have `currentIndicator` not be an observable attribute, but a computed property that looks up the indicator in the store by id. That should be a *synchronous* method that returns either the indicator, if it's loaded, or an empty object or other placeholder value, if not. The currently selected ID, itself, should be an observable property, so that when it changes, it triggers `currentIndicator` to recompute.

Then there just needs to be some mechanism whereby the indicator store also retriggers the computation when it has loaded the data. That should be straightforward using mobx if we do this right.

Does that make sense? Basically, be more strict about the rendering being a pure computation derived from first principles.

jasoncrawford: Curious to hear your thinking on rolling our own store vs. using an existing one, too, since I think that's relevant here… using a good one would make some of these basic architecture/design choices for us and maybe smooth the way if we find a good impedance match with what we're doing.

charts/ExploreView.tsx
```diff
@@ -77,34 +72,66 @@ export class ExploreView extends React.Component<ExploreProps> {
...
-    @observable chartType: ExplorerChartType = ChartType.LineChart
+    @observable.ref chartType: ExplorerChartType = ChartType.LineChart
+
+    // TODO: Inidialize indicatorId from URL query paramter
+    @observable.ref indicatorId?: number = 677
 
     chart: ChartConfig
 
-    dispose!: IReactionDisposer
+    disposers!: IReactionDisposer[]
```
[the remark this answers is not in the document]
danielgavrilov: Yeah that's a good idea

charts/ExploreView.tsx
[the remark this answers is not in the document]
danielgavrilov: Now moved!

charts/ExploreView.tsx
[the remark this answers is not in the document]
danielgavrilov: Let's keep it in mind for another PR 😀

charts/IndicatorDropdown.tsx
danielgavrilov: Agree with exactly your proposal in the above comment re synchronous method!

Re switching to Redux: it comes with "declarative actions" and they are kind of the immutable solution to the problem mobx tries to solve with mutation. The two approaches might not work very well together I think, things might not render when they should, or re-render when they don't need to...

jasoncrawford: Ah, interesting. Is there another model store that would be a better fit?

jasoncrawford: Mobx has some recommendations here, we should both review this: https://mobx.js.org/best/store.html

jasoncrawford: And there's a bunch of relevant things here, including a whole category of model libraries: https://github.com/mobxjs/awesome-mobx

danielgavrilov: For mobx probably this: https://github.com/mobxjs/mobx-state-tree
Docs: https://mobx-state-tree.js.org/intro/philosophy

Haven't investigated much yet

danielgavrilov: The store issue ended up being quite fiddly. In order to be able to return “placeholder” values while the real indicator is loading, we could: 

1. Create our own getter function (`get()`) that creates and returns the placeholder value and at the same time schedules a fetch to populate the store with the real value.
2. Create a `fetch()` function that populates the store with a placeholder value and schedules a fetch request, but then retrieve the placeholder value with property access, e.g. `indicators[indicatorId]`.

The first option is problematic because [mobx tracks property access, not values](https://mobx.js.org/best/react.html#mobx-tracks-property-access-not-values), so observers that call `get()` but do nothing else will not automatically subscribe to value changes. I chose this option and that's why there's a `noop()` call in ExploreView.

The second option seems bad because we always have to remember to call load before accessing, but then we don't have the problems with property access.

jasoncrawford: One way I have seen this handled is like this:

* The first time you call store.get(id), an empty model object is created with that id, put in the store, and returned to you
* At the same time, the store makes a fetch call for the data
* When the data comes back, the model object is updated. So if the view is reacting to its properties, it re-renders

That solves the problem with option 1 and it's what I'd be inclined to do. There could be other ways to solve this, but I like this pattern in general.

charts/Store.ts
```diff
@@ -0,0 +1,115 @@
...
+                    // Mark all other indicator as failed
+                    missingIds.forEach(id => {
+                        const storeEntry = this.get(id)
+                        storeEntry.isLoading = false
+                        storeEntry.lastRetrieved = undefined
+                        storeEntry.error = new Error(
+                            `No indicator with id ${id}`
+                        )
+                    })
+                })
```
jasoncrawford: I don't really understand why all this is needed. When are we ever going to have missing indicators?

charts/Store.ts
```diff
@@ -0,0 +1,115 @@
...
+            return titleLower.indexOf(queryLower) > -1
+        })
+    }
+}
+
+export class RootStore {
+    public indicators: IndicatorStore
+
+    constructor() {
+        console.log("abc")
```
jasoncrawford: log message left in

charts/IndicatorDropdown.tsx
```diff
@@ -69,9 +35,12 @@ export class IndicatorDropdown extends React.Component<IndicatorDropdownProps> {
     }
 
     @bind async loadOptions(query: string): Promise<Indicator[]> {
-        return await this.context.indicatorStore.search({
+        const entries = await this.context.store.indicators.search({
             query
         })
+        return entries
+            .filter(entry => entry.entity) // remove entities that haven't loaded
```
jasoncrawford: I'm confused by this too. How are we going to get only partial loading? We're loading a single static JSON file that's a list of indicators. Why do we have to filter it for missing things?

charts/Store.ts
danielgavrilov: I think when we delete a chart or we remove its "explorable" status. The user would arrive on the page, a placeholder indicator will be created, but the indicator won't exist in `indicators.json`.

charts/IndicatorDropdown.tsx
danielgavrilov: Yeah not really necessary since `search()` returns loaded entities only. 

I added the filter because Typescript complained that `entry.entity` can technically be `undefined`.

danielgavrilov: Removed it now since `search()` probably won't ever return entities that aren't loaded.

charts/Store.ts
jasoncrawford: Ah, is the concern about stale links basically? With old IDs? That makes sense

danielgavrilov:
Ok I think this should be good to go. It adds 100kB to the bundle due to the fancy <select>, but maybe that’s alright.

I’ve commented out the test for `time` URL parameter: https://github.com/owid/owid-grapher/pull/282/commits/e9c9c6ad700da393aa6145a6daea3d8f08ab7cff

I think we should address it separately, the HTMLTimeline logic is quite tangled.

Noted down future to-dos on Notion: https://www.notion.so/owid/Indicator-switching-todos-91cc0252516141bb97b3b0e0802493eb

- Persist `selectedData` somewhere on ExploreModel
- Persist `time` chart state between switching
- Avoid disappearing chart when switching

charts/ExploreView.tsx
```diff
@@ -26,14 +19,24 @@ import { Indicator } from "./Indicator"
...
+        ]
+    }
 }
 
-function noop(value: any): void {
-    return
-}
+import * as urlBinding from "charts/UrlBinding"
+import { ExploreUrl } from "./ExploreUrl"
+import { ExploreModel, ExplorerChartType } from "./ExploreModel"
```
jasoncrawford: All imports should be at the top, no?

charts/__tests__/ExploreView.test.tsx
```diff
@@ -31,30 +50,59 @@ async function updateViewWhenReady(exploreView: ReactWrapper) {
...
+        // problem separately.
+        //
+        // -@danielgavrilov 2019-12-12
+        //
+        // it("applies the time params to the chart", async () => {
+        //     const view = await renderWithQueryStr("time=1960..2005")
+        //     const style: any = view.find(".slider .interval").prop("style")
+        //     expect(parseFloat(style.left)).toBeGreaterThan(0)
+        //     expect(parseFloat(style.right)).toBeGreaterThan(0)
+        // })
```
jasoncrawford: Instead of commenting out, a slightly better thing to do is `it.skip`. This reminds you on every run that a test was skipped

danielgavrilov: Ah nice! Changed now

charts/ExploreView.tsx
danielgavrilov: Ah, didn't notice when merging. Now fixed, thanks!

jasoncrawford:
Looks great, ship it? 🚢 :shipit:

PR #290, opened 2019-12-13
Explore model/view refactor

jasoncrawford:
Mostly move a bunch of model concerns out of the ExploreView, into the ExploreModel. Added tests for ExploreModel.

Note that the ExploreModel now takes a RootStore as a required param. I tend to think this is the way models should work.

If every view has a model, and every model has a store, then we don't need to create context providers to pass the store around. We can just pass the model down through the props (or if we don't want to do that, put the model itself in a context).

Also cleaned up the way the view was doing the chart type buttons.

### DIFF
--- a/charts/ExploreView.tsx
+++ b/charts/ExploreView.tsx
@@ -133,6 +133,11 @@ export class ExploreView extends React.Component<ExploreProps> {
         return this.props.bounds
     }
 
+    get childContext() {
+        return {
+            store: this.props.store
+        }
+    }
     renderChartTypeButton(button: ChartTypeButton) {
         const isSelected = button.type === this.model.chartType
         return (
@@ -172,17 +177,6 @@ export class ExploreView extends React.Component<ExploreProps> {
         )
     }
 
-    get childContext() {
-        return {
-            store: this.props.store
-        }
-    }
-
-    bindToWindow() {
-        urlBinding.bindUrlToWindow(this.model.url)
-        autorun(() => (document.title = this.chart.data.currentTitle))
-    }
-
     render() {
         return (
             <ExplorerViewContext.Provider value={this.childContext}>
@@ -194,5 +188,10 @@ export class ExploreView extends React.Component<ExploreProps> {
                 </div>
             </ExplorerViewContext.Provider>
         )
+    }
+
+    bindToWindow() {
+        urlBinding.bindUrlToWindow(this.model.url)
+        autorun(() => (document.title = this.chart.data.currentTitle))
     }
 }
--- a/charts/ExploreView.tsx
+++ b/charts/ExploreView.tsx
@@ -101,21 +101,14 @@ export class ExploreView extends React.Component<ExploreProps> {
     }
 
     model: ExploreModel
-    chart: ChartConfig
-    url: ExploreUrl
 
     disposers: IReactionDisposer[]
 
     constructor(props: ExploreProps) {
         super(props)
 
-        const chartProps = new ChartConfigProps()
-        this.chart = new ChartConfig(chartProps)
-
         this.model = this.props.model
-
-        this.url = new ExploreUrl(this.model, this.chart.url)
-        this.url.populateFromQueryStr(this.props.queryStr)
+        this.model.populateFromQueryStr(this.props.queryStr)
 
         this.disposers = [
             // We need these updates in an autorun because the chart config objects aren't really meant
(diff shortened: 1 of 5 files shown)


danielgavrilov:
> If every view has a model, and every model has a store, then we don't need to create context providers to pass the store around. We can just pass the model down through the props (or if we don't want to do that, put the model itself in a context).

But how would the store reach the constructor of a model in this case? It seems like there needs to be some way of instantiating a model somewhere down the component tree, not just at the top-level as it is currently with ExploreModel?

I’m thinking of, e.g. creating thumbnail views of related indicators. We can’t anticipate the models that will be required by all the components in the tree, so there needs to be something like a "factory" passed down that can create instances of models.

charts/ExploreView.tsx
```diff
@@ -88,115 +60,58 @@ export class ExploreView extends React.Component<ExploreProps> {
...
-                } else {
-                    const indicator = this.indicatorEntry.entity
-                    if (indicator) {
-                        this.chart.update(chartConfigFromIndicator(indicator))
-                    }
-                }
-            })
-        ]
+    componentWillUnmount() {
+        this.model.dispose()
```
danielgavrilov: This creates a bit of an asymmetry – the model is created outside of the view, but is disposed when the view is removed. It feels like in that case the model should be created by the view itself? That would make the relationship more explicit (one is disposed when the other is disposed).

charts/__tests__/ExploreView.test.tsx
```diff
@@ -24,18 +24,16 @@ const indicatorsJson = fs.readFileSync("test/fixtures/indicators.json")
 const indicatorsUrl = /\/explore\/indicators\.json/
 const indicator: Indicator = JSON.parse(indicatorsJson.toString()).indicators[0]
 
-function getStore() {
-    return new RootStore()
-}
+const store = new RootStore()
```
danielgavrilov: This means the store state changes get shared from one test to another. It feels like we'd want to avoid that?

jasoncrawford:
> But how would the store reach the constructor of a model in this case? It seems like there needs to be some way of instantiating a model somewhere down the component tree, not just at the top-level as it is currently with ExploreModel?

So there are different ways to do this.

1. The path we're starting to head down has an explicit Store, and you pass it to the model. The cleanest evolution of this path is that you never actually call `new Model()`, instead, the store has methods to give you models. So you might call `store.newModel()`, and the store creates a model and injects itself. In this world you always get models from the store.

2. A variation on this is that models (and collections, which are a kind of model) give you other models. Again you never instantiate a model directly, but you might, for instance, ask a collection to give you a model by id, or (in a more complex relational data model, which we don't really have right now) you might ask a model to give you a child or parent model.

2b. I guess another variation is to say that the collections themselves are the stores. Whenever you create a new collection, you're creating an independent store. If you don't want this, be careful to only create a single collection of each type, and put it in some global place.

3. Alternately, you can say that you don't care about dependency-injecting the store, and instead you're going to optimize for having the models be simple and obvious to work with. In this world you have a global/singleton store, you make models with `new Model()`, and the model just references the global store whenever it needs it.

I'm most familiar with (3) and it works well. But I'd be willing to try a (1) and/or (2) (they can sort of be combined but I think (2) is a bit more natural).

charts/ExploreView.tsx
jasoncrawford: Good point. Creation and disposal are not entirely symmetric… except that creation sets up the autorun. We could put that in a separate “setup” call that the view invokes, for symmetry; or we could do as you suggest and create the model in the view; or we could have the top-level that creates the model also dispose… or we could not bother at all because this is a top-level thing anyway and we're not creating and destroying them all the time.

charts/__tests__/ExploreView.test.tsx
jasoncrawford: Hmm, good point. I'll take a look, maybe better to create a new store each time

jasoncrawford: Done

charts/ExploreView.tsx
jasoncrawford: After having played with this a bit and tried it different ways, I'm inclined to either just leave it as-is despite the asymmetry, or to remove the componentWillUnmount/dispose call completely and maybe make a note in the bootstrap call that there's no need to dispose of the model because it lasts the lifetime of the window.

I'm leaving it as is for now but let me know if you have more input or strong preferences here

jasoncrawford:
Fixed merge conflicts, addressed or responded to comments. @danielgavrilov take another look?

charts/ExploreView.tsx
danielgavrilov: :+1: Yeah I agree it doesn't matter for now, can stay as it is.

danielgavrilov:
> I'm most familiar with (3) and it works well. But I'd be willing to try a (1) and/or (2) (they can sort of be combined but I think (2) is a bit more natural).

Great to have those written up like that – I'm most familiar with (3) too, but I don't have strong feelings about which of those approaches would be best in combination with MobX.

PR #292, opened 2019-12-16
Filter out blank rows in data table

jasoncrawford:
Don't display blank rows in the data table.

* Refactored data manipulation logic into a DataTableTransform, to separate it from the view logic in the DataTable component
* Also switched from formatValueLong to Short; this is better for table display

I realized an issue while testing this: the data table isn't using any “tolerance” on the year, the way the map does. Added this to backlog task: https://www.notion.so/owid/Data-table-current-year-tolerance-0bd50618b2ed4f259caf3856253052cb


charts/__tests__/DataTable.test.tsx
```diff
@@ -6,23 +6,22 @@ import { ChartConfig, ChartConfigProps } from "../ChartConfig"
...
 import * as fixtures from "test/fixtures"
 
-const chartFixture = fixtures.readChart(677)
-const variableFixture = fixtures.readVariable(104402)
-
-describe(DataTable, () => {
+function setupChart(id: number, varId: number) {
     const props = new ChartConfigProps()
-    let chart: ChartConfig
+    extend(props, fixtures.readChart(id))
```
danielgavrilov: I made it possible to pass initializing props to the `ChartConfigProps` constructor, it might be more convenient to use that (you also get typechecking with it)

jasoncrawford: Nice, much cleaner

PR #294, opened 2019-12-18
Fix TextWrap bounds calculations for HTML

danielgavrilov:
I knew [this one](https://github.com/owid/owid-grapher/commit/5ef074a48380ff38a31af4d1630251cbefc0b635) was coming to bite me at some point.

I recently noticed that the license for the charts was aligned to the left and it used to be aligned to the right:

Screenshot_2019-12-18_at_14_18_37

This happens because **the text bounds calculation was done including the HTML content, so the text width calculation treated the HTML tags as renderable text.**

**This PR strips HTML before calculating the width.**

Maybe a useful note while on this topic:

We use `pixelWidth()` to calculate the rendered space used by text – it's just a library that contains the rough dimension of letters and sums them up. This is not bulletproof and the charts would break if the used font is not available on the client, or if the text rendering of the OS is very different from others (like some Linux distros).

### DIFF
--- a/charts/Bounds.ts
+++ b/charts/Bounds.ts
@@ -78,6 +78,8 @@ export class Bounds {
             fontFamily?: string
         } = {}
     ): Bounds {
+        // Collapse contiguous spaces into one
+        str = str.replace(/ +/g, " ")
         const key = `${str}-${fontSize}`
         let bounds = this.textBoundsCache[key]
         if (bounds) {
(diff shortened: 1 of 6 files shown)


danielgavrilov:
Yeah we should definitely switch to browser rendering, the challenge is that it needs to work in node environments as well (and be pretty cheap to do so that deploys are still reasonably quick). 

I've added this as a task on Notion for the future: https://www.notion.so/owid/Find-text-boundaries-in-TextWrap-using-browser-rendering-681fe80980bc441eba6ce2cd2e39613a

PR #299, opened 2020-01-06
Don't override entities on indicator switch

jasoncrawford:


jasoncrawford:
@danielgavrilov this is the code we were just looking at together

jasoncrawford:
Actually hold this one

PR #300, opened 2020-01-06
Add some margin below indicator dropdown

jasoncrawford:


charts/client/chart.scss
```diff
@@ -1012,6 +1012,7 @@ html.iframe .ChartPage {
 .ExplorePage .indicator-dropdown {
     position: relative;
     z-index: $zindex-IndicatorDropdown;
+    margin: 20px auto;
```
danielgavrilov: 😄Thanks for adding!

PR #302, opened 2020-01-08
Fix disappearing chart while indicator is loading

danielgavrilov:
It used to be the case that the chart view disappears (element is fully removed from the DOM tree) while the data is loading, e.g. when switching an indicator.

Changes:
- The ChartView is always shown, even while loading the data
- If available, the header and notes/sources of a chart are shown (indicator data)
- The chart plot is replaced with a loading indicator while the data is loading. This is the same loading indicator that appears

### DIFF
--- a/charts/Bounds.ts
+++ b/charts/Bounds.ts
@@ -66,7 +66,7 @@ export class Bounds {
     }
 
     static forText(
-        str: string,
+        str: string = "",
         {
             x = 0,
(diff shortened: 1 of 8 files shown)


danielgavrilov:
This affects the chart loading state on the live site, so I will do some manual checks to see if anything seems broken (and if we can add any more automated tests)

Issue #256, opened 2019-09-11
Dual-point interactive timeline on slope charts


danielgavrilov:
Currently, the start & end years on slope charts are defined by authors and they cannot be changed: https://ourworldindata.org/grapher/change-in-female-height-slope

It should be possible to change these with the timeline we already have below many charts, like line charts: https://ourworldindata.org/grapher/annual-co2-emissions-per-country?tab=chart&year=2016&time=1979..2017

This is one step in the direction of making _line chart → slope chart_ switching possible for users.

PR #307, opened 2020-01-26
show timeline for slope charts

sherin:
Hello 👋

It has been a while :) I had some free time and decided to check back in on one of my favorite github projects.

This is a very simple PR to get interactive timeline for slope charts as requested here: https://github.com/owid/owid-grapher/issues/256

It works for the most part. Except playing a video starts with a view of no data which can definitely be improved. Let me know and I can make that improvement in this PR. 

Note that all current slope charts will start showing timeline control by default.

Example screenshot:

Screenshot 2020-01-25 16 23 53

### DIFF
--- a/charts/Controls.tsx
+++ b/charts/Controls.tsx
@@ -338,7 +338,7 @@ class TimelineControl extends React.Component<{ chart: ChartConfig }> {
         this.props.chart.map.targetYear = targetStartYear
     }
 
-    @action.bound onScatterTargetChange({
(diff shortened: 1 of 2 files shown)


danielgavrilov:
Hi Sherin! Thank you very much for this!

I will merge and remove the play button – I think there isn't any reasonable way to "play" this chart. :)

sherin:
> I will merge and remove the play button – I think there isn't any reasonable way to "play" this chart. :)

Good idea!

Issue #170, opened 2019-04-04
NumberField doesn't allow typing decimal points


danielgavrilov:
Because it always converts to `Number` and then renders that number, decimal points get removed immediately.

Ideally there should be an internal string attribute on the component, which:
- If `undefined`, the value of the prop is used (which is a `Number`)
- If user interacts, the contents of the text box are persisted to the string attribute

danielgavrilov:
Fixed in #311

PR #311, opened 2020-02-04
Allow user to type decimal values

markledwich2:
This is a fix for   #170 
### Before
Users have prevented form typing in decimals in appropriate fields (a workaround is to paste them)
 
decimal_input_before

### After
`NumberField` now allows users to type in negative/decimal numbers that may not yet be parsable.

There are two new properties ( `allowDecimal`, `allowNegative`  ) which define what a user is allowed to type into the control.

decimal_input_after

### TODO
- [x] Update use of `NumberField` with appropreate  `allowDecimal`, `allowNegative`
- ~~Upgrade/Migration from old state~~


admin/client/Forms.tsx
```diff
@@ -157,13 +157,40 @@ export interface NumberFieldProps {
...
+ * Like NumberField, but also capture the users text input
+ */
+export class NumberInputField extends React.Component<
+    NumberFieldProps<NumericInputValue>
+> {
+    render() {
+        const { props } = this
+
+        const textFieldProps = extend({}, props, {
+            value: props.value && props.value ? props.value.input : undefined,
```
jasoncrawford: `props.value && props.value`? Is that a mistake?

admin/client/Forms.tsx
```diff
@@ -157,13 +157,40 @@ export interface NumberFieldProps {
...
+export class NumberField extends React.Component<
+    NumberFieldProps<number | undefined>
+> {
     render() {
         const { props } = this
 
         const textFieldProps = extend({}, props, {
             value:
-                props.value !== undefined ? props.value.toString() : undefined,
+                props.value && props.value ? props.value.toString() : undefined,
```
jasoncrawford: This line does `props.value && props.value` again, I must be missing something, what's going on with this?

jasoncrawford: In any case, whatever logic is on here might make sense as a method on the NumericInputValue class? Then it could be given a meaningful method name, and wouldn't have to be repeated between the two field components

admin/client/Forms.tsx
markledwich2: Yeah, that's a mistake. I believe I was intending to check props.value.input but there is no need. Thanks - will fix.

admin/client/Forms.tsx
markledwich2: The logic here is just to prevent calling a method on undefined. 
I'll update both places to use optional chaining e.g. `props.value?.toString()`. I didn't realize they had added that to typescript.

danielgavrilov:
Hi @markledwich2! Thank you very much for this, and sorry for the slow response – I was away last week and catching up now.

As you noted, this approach requires migrating the chart config object for each field that we choose to manage through NumberInputField.

I haven’t thought about this deeply, so do let me know if you disagree, but I don’t think it’s important to persist the input field state to the database. 

If a valid input is entered, then whether we use the input state or the numeric value doesn’t matter because we get the same outcome. And if the input is invalid, it’s better not to persist it? Otherwise the next user who edits the chart has a potential error they didn’t cause.

I think it’s probably best to keep the input state to the local component state which would avoid migrations. I’m not exactly sure on the best way to achieve that, but here are 2 options:

One way could be that the numeric value (from the global chart config) is shown by default in the input field. If the user changes the field, then we store a string locally on the component state and we persist the parsed numeric value to the chart config object. A potential problem with this is that the component might not be unmounted when a user switches to a different chart and stale state will be shown. Not sure if we'd encounter that case in the Grapher as it is currently implemented.

Another way that might be more robust is that only while the input field is in focus, the input state is stored and shown from the local component state. On blur, the field switches to showing the persisted chart config value and discards the local component state.

Sorry for the long text – I thought it might be helpful to dump my thinking here. Let me know if you disagree or have any questions! And thanks again!

markledwich2:
I think you are right - avoiding persisting this in the chart config would be better overall. I'll take a look at this next week and try what you suggested.

markledwich2:
@danielgavrilov  I have changed this to use the approach you suggested.

`NumberField` now allows users to type in negative/decimal numbers that may not yet be parsable. 

- There are two new properties ( `allowDecimal`, `allowNegative`) that define what a user is allowed to type into the control. I have set these appropriately for the different form fields. The control now provides better restrictions on what input is allowed. 
- The user input state is stored as state on the component.  I tested switching between editing charts and there doesn't seem to be any stale state. So I haven't implemented clearing that on blur.

danielgavrilov:
Thanks @markledwich2!

I just realised we also have a two-way binding in some cases:



In this case even the onBlur approach I mentioned will by some definition fail, though I feel it's acceptable that the value doesn't change when the field is focussed. 

This might be overengineering it, but a robust approach might be to check after every keystroke if `configValue.toString() === localInputState`, and if true, then set `localInputState = undefined` and fall back to rendering the `configValue` in the input field (the idea is that whenever `localInputState` is `undefined` we fall back to the config value). Then again, on blur always set `localInputState = undefined` to fall back to the config value.

Sorry that this turned out to be a fiddly issue!

markledwich2:
Oh, well spotted. I will take a look at this next week.

markledwich2:
I implemented it as you described.

Behavior differences:
- The two-way binding now works.
- When losing focus on a partially completed field (e.g. "-0.0" or "1".), it will update to the parsed value. I think this is good, given that is how it will be saved/appear next time you open the chart.

danielgavrilov:
Thanks a lot @markledwich2! I will just add `allowDecimal` and `allowNegative` in a few more places and deploy this.

PR #323, opened 2020-03-06
Add analytics to Charts and additional site events

breck7:
From the Notion issue, this adds tracking for the following:

- [ ]  Record current chart state (more context on this one?)
- [x]  Record user-initiated tab changes
- [x]  Record download SVG/PNG
- [x]  Record download data

I also added tracking to Feedback opens and Timeline plays. Once these are live I'll make a dashboard for chart events in Amplitude.

The existing code is simple enough and I just stuck with the patterns I found, though I guess there is room for marginal improvements. Most events currently have the type OWID_SITE_CLICK and then "note" is used for further bucketing, but some events use different types. At some point if someone is spending a lot of time in Amplitude it might be helpful to organize the analytics code a little better and standardize these event names. For now Amplitude has great querying capabilities so even with a few naming conventions it is easy to sort through the events.

### DIFF
--- a/site/client/owid.entry.ts
+++ b/site/client/owid.entry.ts
@@ -57,6 +57,7 @@ new SmoothScroll('a[href*="#"][data-smooth-scroll]', {
 
 const search = document.querySelector("form#search-nav") as HTMLFormElement
 if (search) {
+    // Todo: Not seeing this event in Amplitude. Can this be removed?
     const input = search.querySelector("input[type=search]") as HTMLInputElement
     search.addEventListener("submit", ev => {
         ev.preventDefault()
--- a/site/client/owid.entry.ts
+++ b/site/client/owid.entry.ts
@@ -55,51 +55,21 @@ new SmoothScroll('a[href*="#"][data-smooth-scroll]', {
     popstate: false
 })
 
-const search = document.querySelector("form#search-nav") as HTMLFormElement
-if (search) {
-    // Todo: Not seeing this event in Amplitude. Can this be removed?
-    const input = search.querySelector("input[type=search]") as HTMLInputElement
-    search.addEventListener("submit", ev => {
-        ev.preventDefault()
-        Analytics.logEvent("OWID_SITE_SEARCH", { query: input.value })
-            .then(() => search.submit())
-            .catch(() => search.submit())
-    })
-}
-
-const trackedLinkExists: boolean = !!document.querySelector(
-    "[data-track-click]"
-)
-
-function createFunctionWithTimeout(callback: () => void, timeout: number = 50) {
-    let called = false
-    function fn() {
-        if (!called) {
-            called = true
-            callback()
-        }
+document.addEventListener("click", async ev => {
+    const targetElement = ev.target as HTMLElement
+    const trackedElement = getParent(
+        targetElement,
+        (el: HTMLElement) => el.getAttribute("data-track-click") !== null
+    )
+    if (trackedElement) {
+        // Note that browsers will cancel all pending requests once a user
+        // navigates away from a page. An earlier implementation had a
+        // timeout to send the event before navigating, but it broke
+        // CMD+CLICK for opening a new tab.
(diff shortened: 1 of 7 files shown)


danielgavrilov:
> Record current chart state (more context on this one?)

So, I definitely wrote this, and I'm not sure exactly what I meant at the time, but this relates to tracking users interacting with the charts using the chart config as a proxy.

The chart controls available to the user are dependent on the chart state. In order to know the click rate of e.g. the _Add country_ button or the timeline, we need to know how many people saw those controls, because there are chart configurations where those controls are not available.

Then, instead of adding tracking logic to each control, we could potentially just track the changes to the chart config.

Some potential issues with this:
- A changing chart state is not always directly caused by an interaction. E.g. playing the timeline will continuously change the year, but the user is not dragging the timeline
- There might be config changes that happen as side effects because of other actions. E.g. clicking a country on a map will take you to a line chart with that country. The user didn't switch tabs, add/remove countries, but it's impossible to tell if we just look at config changes

Anyway, I think this should be done separately (not in this PR) and we need to think through the potential issues a bit more. Do you agree? It might turn out to be a bad idea after all.

site/client/owid.entry.ts
```diff
@@ -57,6 +57,7 @@ new SmoothScroll('a[href*="#"][data-smooth-scroll]', {
 
 const search = document.querySelector("form#search-nav") as HTMLFormElement
 if (search) {
+    // Todo: Not seeing this event in Amplitude. Can this be removed?
```
danielgavrilov: Ah yes, it can be removed, or we can fix the selector to match the current markup. I think it should be `form.HeaderSearch` to target the header search box.

breck7:
> Anyway, I think this should be done separately (not in this PR) and we need to think through the potential issues a bit more. Do you agree? It might turn out to be a bad idea after all.

I agree that we could do this in a separate PR. That's a good point that these changes won't give us good click rates. That could be really nice to have. For now I was thinking let's start with tracking absolute feature usage. Then we can say Feature X is used N times per day. And then the next iteration we could have Feature X is used N times per day and 10% of the time it appears on screen.

> Then, instead of adding tracking logic to each control, we could potentially just track the changes to the chart config.

I like the idea of tracking the chart config in the sense that it would be useful to know what components have been visible to the user so we could get usage rates and not just usage. I think tracking control usage too would be conceptually simple and helpful.

I like the idea of not having to add tracking code to each control. 

For control usage, maybe we could do it once globally by hooking onto the methods decorated with "@action.bound"? It looks like there are ~273 methods with that decorator (sitewide, not just on charts). We could either have another "@track" decorator that we put on a select number of those or track all of them and have an opt-out decorator "@donttrack" so we don't log every mouseover, etc.

Or another idea: what if we added a higher level set of "command" methods? We could either cleanup the "@action.bound" methods or have those methods call the new ones. There are also ~24 anonymous onClick functions that we might promote and name. So we would have "playTimelineCommand", "switchToMapTabCommand", "saveAsPngCommand", "addCountryCommand", "showCountryLineChartFromMapClickCommand", "downloadCsvCommand", "switchChartTypeCommand", etc. Then we do the same thing as above and add a decorator to track those. The benefit of this approach is we could use the method names as the event name, so it might be more informative in stats than things like "@action.bound onClick()". We might also make it easily create new ways for users to interact with the site, by separating the high level commands from a lower level method of triggering those commands (ie, you could write high level integration tests, or add keyboard shortcuts, or add a "Spotlight" type of UI component, etc).

This might be way overkill. It is more naming work. Or maybe there are simple well known ways to do it in the React/Mobx community.

site/client/owid.entry.ts
breck7: I will remove.

I noticed we have a little bit of search tracking now in Amplitude via OWID_PAGE_LOAD with an href filter. I started a chart and dashboard for that. That excludes instant searches though.

We could move this tracking event to `HeaderSearch.tsx`. I think it's probably best for now to stick to Algolia for tracking search queries, instead of duplicating that info into Amplitude.

PR #332, opened 2020-03-13
Always show footer timeline and additional controls on their own rows like on mobile

breck7:
Working on the issue: "Display timeline slider on its own row for date-based charts"

After working with the code we currently have 8 different code paths:
(>700px?)*(inlineControls?)*(timelineControls?)

If we added (isDayBased?) we would have more to support.

Instead what if we just always present the chart as on the mobile device, with timelines and extra controls on their own lines?

I have another branch too where I try to cleanly implement all the existing paths with the addition of the new paths but thought I'd put this one for review first, since it removes a decent amount of complexity.

### DIFF
--- a/charts/Controls.tsx
+++ b/charts/Controls.tsx
@@ -497,10 +497,6 @@ export class Controls {
         return this.props.chart.tab === "map"
     }
 
-    @computed get widthIsGreaterThan700(): boolean {
-        return this.props.width > 700
-    }
-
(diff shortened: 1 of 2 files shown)


danielgavrilov:
@breck7 Awesome! I am in favour of this approach where we put the timeline in its own row for every chart. I don't think we should do it differently depending on whether it's days or not.

PR #337, opened 2020-03-14
Allow inline controls in chart tabs row

danielgavrilov:
I think it's good to have this back actually. I forgot about this case:

Screenshot 2020-03-14 at 12 15 42

(current on left, this PR on right)


breck7:
Fine with me. Maybe down the line we can get code coverage in to ensure we've got automated tests of all the cases but for now I think this is fine.

PR #338, opened 2020-03-16
ScatterPlot: Use date formatting when labels are days instead of years

danielgavrilov:
Notion issue: https://www.notion.so/owid/Show-axis-value-instead-of-year-on-ScatterPlot-points-d8069816d451406ab267b2c116ebc679

### To-do

- [x]  Format dates correctly on scatter plots
- [ ]  ~~Allow authors to change whether points are labeled with time, X, or Y value~~


breck7:
I checked out and played with this branch. LGTM!

I think showing the dates on hover isn't so bad and a big improvement over current situation.

danielgavrilov:
Thanks @breck7! I will merge as you suggest, we can leave the next item for another PR.

Issue #316, opened 2020-02-12
Support a log scale on DiscreteBar


danielgavrilov:
Selecting a single year on the timeline of a line chart turns it into a bar chart. But if a log scale is used on the line chart, then the bar chart is broken: https://ourworldindata.org/grapher/historys-hockey-stick-gross-domestic-product-per-capita-using-the-ratio-scale-1990?year=730&time=2016&country=BWA+ZAF+USA

Ideally, DiscreteBar should also support a log scale.

breck7:
Pushing a branch with a fix. Adding a screenshot of current behavior:

![Screen Shot 2020-03-17 at 2 33 48 PM](https://user-images.githubusercontent.com/74692/76913888-6d677c80-685c-11ea-935e-59f763675a44.png)

breck7:
And the same issue on the Corona charts:

![Screen Shot 2020-03-17 at 2 35 52 PM](https://user-images.githubusercontent.com/74692/76913952-a0aa0b80-685c-11ea-8dbb-91ef39aa73f6.png)

PR #340, opened 2020-03-18
Add support for Log Scale to Discrete Bar. #316

breck7:


breck7:
Before and after:
Screen Shot 2020-03-17 at 2 33 48 PM

Screen Shot 2020-03-17 at 2 33 53 PM

PR #341, opened 2020-03-18
Add tick marks to line charts and stacked bar charts.

breck7:
Before and afters:

Screen Shot 2020-03-17 at 4 21 12 PM
Screen Shot 2020-03-17 at 4 17 13 PM
Screen Shot 2020-03-17 at 4 16 11 PM

### DIFF
--- a/charts/StackedBarChart.tsx
+++ b/charts/StackedBarChart.tsx
@@ -8,6 +8,7 @@ import { includes, guid, uniq, makeSafeForCSS } from "./Util"
 import { ChartConfig } from "./ChartConfig"
 import { Bounds } from "./Bounds"
 import { AxisBox, AxisGridLines } from "./AxisBox"
+import { AxisTickMarks } from "./AxisTickMarks"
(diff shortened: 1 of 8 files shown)


charts/StackedBarChart.tsx
```diff
@@ -418,6 +419,18 @@ export class StackedBarChart extends React.Component<{
         const { stackedData, xValues } = this.transform
         const { innerBounds } = axisBox
 
+        const textColor = "#666"
+
+        const tickMarks = true ? (
```
danielgavrilov: This conditional is always true? Is it accidentally here or intentionally?

breck7: Accident! Thanks.

Making a minor note to myself to look into adding the no-unnecessary-condition rule to our tslint "https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md"

PR #344, opened 2020-03-20
format "year" properly given context (either year or day, at time of …

shaahmad:
…commit)

### DIFF
--- a/charts/DataTab.tsx
+++ b/charts/DataTab.tsx
@@ -45,7 +45,7 @@ export class DataTab extends React.Component<{
                 const row = [
                     entity,
                     vardata.entityMetaByKey[entity].code || "",
-                    year
+                    chart.formatYearFunction(year)
                 ]
 
                 let rowHasSomeValue = false


charts/DataTab.tsx
```diff
@@ -45,7 +45,7 @@ export class DataTab extends React.Component<{
                 const row = [
```
danielgavrilov: TypeScript incorrectly infers the type of `row` now that the `formatYearFunction` is used.

The previous inference for `row` was `(string | number)[]`, because `year` was a `number`. The new inference is `string[]`. In the code lower below, when the rest of the columns (the variable values) are added, the `string[]` inference doesn't work.

One way to fix this is to set the type explicitly:

```suggestion
                const row: (string | number)[] = [
```

shaahmad:
> Great that you used the `formatYearFunction`! One more thing would be to change the CSV column header from `"Year"` to `"Date"`. I think you will need to check the `yearIsDay` flag explicitly to achieve that, and you need to iterate through the variables in the same way `formatYearFunction` does.
> 
> Maybe it makes sense to somehow expose the `yearIsDay` field through the `ChartConfig` object so that the same logic could be used within `formatYearFunction` too.
> 
> Let me know if you want to chat about this.

charts/DataTab.tsx
shaahmad: Ah ok, I thought we had agreed not to modify references to `year`, but now I see you may have been referring just to actual code references, not the string used for the row title. I will address this.

Wrt to the TypeScript reference, good to know! I haven't encountered this type of TypeScript error before. I'll look into it to find best practices.

PR #347, opened 2020-03-23
Use dimension-specific formatYear where possible

danielgavrilov:
In some cases, we may use different time formatting on different dimensions, e.g. a scatter plot where X axis has a years variable, and Y axis has a days variable: https://owid.cloud/admin/charts/4041/edit (Slack discussion: https://owid.slack.com/archives/CUQSW9STU/p1584912362004000?thread_ts=1584907551.000900&cid=CUQSW9STU)

total-confirmed-cases-of-covid-19-per-million-people-vs-gdp-per-capita

From an implementation point of view, this breaks the timespan logic (and probably more) and it isn't a good idea to combine variables with different time-bases. But authors are already creating these charts, and I imagine there will be more as we find interesting correlations between COVID-19 and various health/economics metrics.

Anyway, I thought we should try to use dimension-specific year formatting to address the problem as much as possible.

This PR fixes a problem here:

Screenshot 2020-03-23 at 11 05 10

And here: 

Screenshot 2020-03-23 at 11 08 16

### DIFF
--- a/charts/SourcesTab.tsx
+++ b/charts/SourcesTab.tsx
@@ -39,9 +39,9 @@ export class SourcesTab extends React.Component<{
         const maxYear = max(variable.years)
         let timespan = ""
         if (minYear !== undefined && maxYear !== undefined)
-            timespan = `${this.props.chart.formatYearFunction(
+            timespan = `${dimension.formatYear(
                 minYear
-            )} – ${this.props.chart.formatYearFunction(maxYear)}`
+            )} – ${dimension.formatYear(maxYear)}`
 
         return (
             <div key={source.id} className="datasource-wrapper">
--- a/charts/DimensionWithData.ts
+++ b/charts/DimensionWithData.ts
@@ -6,10 +6,13 @@ import {
     some,
     isString,
     sortBy,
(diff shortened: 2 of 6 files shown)


breck7:
The GDP vs Cases is definitely an interesting chart.

This looks like a good step in the right direction. formatYearFunction almost certainly shouldn't be on chart and should be closer to the dimension/variable class.

Perhaps in the future we might benefit from a pattern like this:

- TableClass/DatasetClass
- VariableClass/ColumnClass
- DatumClass/RowClass/ObservationClass

The format functions would likely belong on VariableClass.

charts/SourcesTab.tsx
```diff
@@ -39,9 +39,9 @@ export class SourcesTab extends React.Component<{
         const maxYear = max(variable.years)
         let timespan = ""
         if (minYear !== undefined && maxYear !== undefined)
-            timespan = `${this.props.chart.formatYearFunction(
+            timespan = `${dimension.formatYear(
```
breck7: Nice. Definitely makes more sense to pull from dimension instead of chart.

PR #350, opened 2020-03-24
separate dayIsYearVar getter function, format years in csv download

shaahmad:
I created a separate "Date" vs "Year" row header for the two different cases, but it may be ok to just use "Date" all the time. Lmk, I can change that.

### DIFF
--- a/charts/DataTab.tsx
+++ b/charts/DataTab.tsx
@@ -34,7 +34,9 @@ export class DataTab extends React.Component<{
 
         const rows: string[] = []
 
-        const titleRow = chart.yearIsDayVar ? ["Entity", "Code", "Date"] : ["Entity", "Code", "Year"]
+        const titleRow = chart.yearIsDayVar
+            ? ["Entity", "Code", "Date"]
+            : ["Entity", "Code", "Year"]
         dimensions.forEach(dim => {
             titleRow.push(csvEscape(dim.fullNameWithUnit))
         })
--- a/charts/DataTab.tsx
+++ b/charts/DataTab.tsx
@@ -34,9 +34,12 @@ export class DataTab extends React.Component<{
 
         const rows: string[] = []
 
-        const titleRow = chart.yearIsDayVar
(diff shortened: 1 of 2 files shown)


charts/DataTab.tsx
```diff
@@ -34,18 +34,20 @@ export class DataTab extends React.Component<{
 
         const rows: string[] = []
 
-        const titleRow = ["Entity", "Code", "Year"]
+        const titleRow = chart.yearIsDayVar
+            ? ["Entity", "Code", "Date"]
+            : ["Entity", "Code", "Year"]
```
breck7: nit: avoid repetition and only change the 3rd word:

    ["Entity", "Code", chart.yearIsDayVar ? "Date" : "Year"]

jasoncrawford:
@shaahmad I think `Date` vs. `Year` is good for the header

shaahmad:
I need to still address charts with two different time units on the x vs y axis (COVID cases vs GDP). But this fix will address issues with charts with just one time unit (e.g. total COVID cases)

breck7:
Nice!

> I need to still address charts with two different time units on the x vs y axis (COVID cases vs GDP). But this fix will address issues with charts with just one time unit (e.g. total COVID cases)

Great. Seems like this will help in the interim. LGTM.

Long term csvBlob looks like the type of thing that would benefit from unit tests.

Though long term perhaps we should have some sort of "DataFrame/DataTable" class that all charts use, and then toCsv methods on those.

breck7:
In the interim while we don't have unit tests could you share a before and after showing current CSV behavior with the new behavior?

shaahmad:
Before:
Screen Shot 2020-03-24 at 9 55 40 PM
After:
Screen Shot 2020-03-24 at 9 55 50 PM

danielgavrilov:
Looks great @shaahmad! I will merge & deploy this, we can address the issue with the different time units separately.

Congrats on the first PR! 🙌

PR #354, opened 2020-03-26
Add ability to dump a JSON file of all chart feature usage to better find all charts that might be affected by a change.

breck7:


breck7:
@danielgavrilov I took the JSON dump and tried a few things (Google Data Studio, Ohayo/Vega, DataTables). Maybe someday if we have Grapher working on general tabular data could analyze Grapher data in Grapher :).

I like your approach too. I haven't used the JSON functions in MySQL much before. I should practice those and perhaps that will be easier.

breck7:
I should add some discoverability to this in the admin console.

PR #389, opened 2020-04-04
Add Entity Annotations Feature to Line Charts

breck7:
Before & After:
Screen Shot 2020-04-03 at 5 44 48 PM

Screen Shot 2020-04-03 at 5 44 13 PM

This hopefully improves styling of those charts and will let us use standard entity names and get the Map view working. It requires researchers to paste in an entity-annotation map at the variable level. The preferred way would be to have an annotation column and I tried to implement that but was not able to come up with a good approach yet. If we get our data warehouse model refactored hopefully that will be easy to move, if we want to keep this annotations feature around.

I also tried to clean up a few things as I went like adding more privates and renaming the vague "DataKey" to "EntityDimensionKey". Not that the latter rolls off the tongue but I think it's slightly better to have wordy but more informative names. These are all nits and would be happy to separate those out into a separate commit.

Finally, I added the owidDataset property to ChartConfig so you can just load a full chart from 1 JSON file to ease debugging.

### DIFF
--- a/admin/client/EditorDataTab.tsx
+++ b/admin/client/EditorDataTab.tsx
@@ -3,7 +3,7 @@ import { clone, map } from "charts/Util"
 import { computed, action, observable } from "mobx"
 import { observer } from "mobx-react"
 import { ChartConfig } from "charts/ChartConfig"
-import { DataKey } from "charts/DataKey"
+import { EntityDimensionKey } from "charts/EntityDimensionKey"
 import {
     EditableList,
     EditableListItem,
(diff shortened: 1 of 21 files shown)


danielgavrilov:
I just added some "usability" improvements, but still need to test this a bit more.

But I agree with the approach @breck7! As well as increased use of `private`s and longer, more descriptive variable names.

danielgavrilov:
Just added some more cosmetic changes 🙂

Everything works well – deploying this!

PR #392, opened 2020-04-07
Add sandbox to folder baking

breck7:
This allows you to bake a current chart in the sandbox along with your local development version of Grapher charts to a self-contained folder in order to share interactive development prototypes with the rest of the team for feedback.

How to use:

1. Go to the admin for any chart on prod and copy/paste the debug version into sandbox/chart.js (paste it with `const sandboxChart =`—see chart.sample.js for a sample).
2. Run "ts-node bakeSandbox.ts /some/destionation/folder

### DIFF
--- a/public/sandbox/bakeSandbox.ts
+++ b/public/sandbox/bakeSandbox.ts
@@ -1,10 +1,12 @@
 #!/usr/bin/env ts-node
 
 import * as fs from "fs-extra"
-import * as http from "http"
 
 const bakeSandbox = async (outputPath: string) => {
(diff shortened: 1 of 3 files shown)


danielgavrilov:
This is definitely helpful!

But why not use `yarn build` and just copy over the contents of the dist folder? Is that equivalent?

Otherwise, you need the webpack server running for this command, and it also spits out the live-reloading code and quite a heavy bundle (commons.js is 12.6MB).

breck7:
Yes copying the reqs from the 'dist/' folder is a simpler approach. Thanks! Will revise.

PR #393, opened 2020-04-07
Take annotation into account when computing legend width

danielgavrilov:
### Before:

Screenshot 2020-04-07 at 20 37 01

### After:

Screenshot 2020-04-07 at 20 37 35

---

There is a bit too much padding? This could go out as it is, but that wide gap is kind of a bit much. 

I thought this would be a 2min fix and I've spend much longer, so I'm leaving this here to merge or for anyone to extend and fix that padding.

I think the problem is that when legend items get offset, they have lines on the left, and this padding is reserved for those lines. But the padding is reserved in `HeightedLegend`, before we even know whether we need those lines. So when the lines are unnecessary, we end up with unused space.

### DIFF
--- a/charts/HeightedLegend.tsx
+++ b/charts/HeightedLegend.tsx
@@ -120,7 +120,12 @@ export class HeightedLegend {
                 item,
                 textWrap,
                 annotationTextWrap,
-                width: leftPadding + textWrap.width,
+                width:
+                    leftPadding +
+                    Math.max(
+                        textWrap.width,
+                        annotationTextWrap ? annotationTextWrap.width : 0
+                    ),
                 height:
                     textWrap.height +
                     (annotationTextWrap ? annotationTextWrap.height : 0)


danielgavrilov:
In the case when the lines are added, it looks fine:

Screenshot_2020-04-07_at_20_49_21

breck7:
Here is an approach that first computes the max legend title width and then uses that to set the annotation width. Solves the padding problem but at the cost of splitting annotations too aggressively. I'm going to merge yours now and just leave the code for this other approach here.

Screen Shot 2020-04-07 at 11 20 56 AM

```
const { fontSize, leftPadding, maxWidth } = this
        const maxTextWidth = maxWidth - leftPadding

        const textWraps = this.props.items.map(
            item =>
                new TextWrap({
                    text: item.label,
                    maxWidth: maxTextWidth,
                    fontSize
                })
        )

        const maxAnnotationWidth = Math.max(
            ...textWraps.map(wrap => wrap.width)
        )

        return this.props.items.map((item, index) => {
            const annotationTextWrap = item.annotation
                ? new TextWrap({
                      text: item.annotation,
                      maxWidth: maxAnnotationWidth,
                      fontSize: fontSize * 0.9
                  })
                : undefined
            const textWrap = textWraps[index]
            return {
                item,
                textWrap,
                annotationTextWrap,
                width: leftPadding + textWrap.width,
                height:
                    textWrap.height +
                    (annotationTextWrap ? annotationTextWrap.height : 0)
            }
        })
```

breck7:
I'll fix for the case without lines.

breck7:
I added a maximum annotation width of 150 in a91fc8e482211b2c0810e9643189715e3e001283 on top of this which seems to strike a nice balance for the case when we have long annotations.

I think that should be good enough for this one.

Issue #257, opened 2019-09-11
Show static chart while interactive chart is loading


danielgavrilov:
**NOTE**: testing and implementing this requires setting up Wordpress locally. We unfortunately don't yet have instructions and/or public resources to make this possible.

---

We currently serve server-generated static SVG charts on initial page load, so that users on slow connections can see the charts before all JavaScript loads and executes.

But, when the scripts load, they remove the static chart and replace it with a loading indicator while the data for the chart is fetched.

The phases roughly look like this:

![chart-loading-phases](https://user-images.githubusercontent.com/1308115/64704325-0a102280-d4a6-11e9-9af3-3bfeb8c36f38.png)
(Static chart on left, interactive on right)

Ideally, we should remove the middle phase where a loading indicator is shown without the chart. The static chart should be shown up until all dependencies for rendering the chart are ready and the interactive chart is successfully rendered, and then immediately swapped with the static chart.

Note that this is the behaviour we want for embedded charts – the thumbnails for those are rendered with the URL params considered. But for standalone charts the page thumbnail does not consider URL params (since we are a static site), so showing the thumbnail may be misleading... we will have to test it out.

Maybe the ideal solution is that the thumbnail should only be kept if it matches the query params, otherwise a loading indicator can be shown.

danielgavrilov:
The phases have now changed (see https://github.com/owid/owid-grapher/pull/302) – we now load and show the chart title, subtitle & notes as soon as the chart metadata is loaded (it's always loaded first). Before it used to be the case that you see a loading indicator until both metadata & data are fully loaded. 

It would still be great to keep the preview (as long as the query params match) and seamlessly transition from preview to interactive chart without a loading state.

andrewramacher:
Hey I was hoping to help with this project so I wrote the code for this feature and created a branch, but I am unable to push due to a permissions error. When running `git push --set-upstream origin replace-loading-chart-with-static` I get the error `ERROR: Permission to owid/owid-grapher.git denied to andrewramacher.` Is there another way I can go about uploading the fix?

shaahmad:
Hey Andrew,

Thanks for reaching out. I will give you a call to handle this.

Thanks,
Shahid

breck7:
I think the new behavior is good enough. Closing.

PR #409, opened 2020-04-14
Replace loading chart with static

andrewramacher:
The goal of this change is to replace the loading chart with a static image of the chart when loading a specific interactive chart as requested in https://github.com/owid/owid-grapher/issues/257. Before, there would be two loading animations upon opening the page. One while the text, overview, and title of the chart loads, and another after this while the interactive chart loads. The changed code displays a static image of the chart instead of the "loading" animation during the second load, so the user only sees one loading wheel animation. 

During second load, before:
Screen Shot 2020-04-14 at 2 36 26 PM

During second load, after:
Screen Shot 2020-04-14 at 2 36 39 PM
In this image the chart is not interactive

I tested this by running it locally and with yarn test.

### DIFF
--- a/charts/ChartTab.tsx
+++ b/charts/ChartTab.tsx
@@ -12,7 +12,7 @@ import { DiscreteBarChart } from "./DiscreteBarChart"
 import { StackedBarChart } from "./StackedBarChart"
 import { ChartLayout, ChartLayoutView } from "./ChartLayout"
 import { TimeScatter } from "./TimeScatter"
-import { LoadingChart } from "./LoadingChart"
+import { BAKED_GRAPHER_URL } from "settings"
 
 @observer
 export class ChartTab extends React.Component<{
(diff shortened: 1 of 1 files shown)


danielgavrilov:
Hi @andrewramacher,

Thanks for the PR. This solution which works for standalone charts (`/grapher/` pages), but we also need this to take query parameters into account and support embedded charts.

For example, if you visit the [COVID testing page and scroll down to a country](https://ourworldindata.org/covid-testing#argentina), as you scroll you can see the same chart embedded but showing different countries. When the embed images are created, the [URL query parameters are taken into account](https://github.com/owid/owid-grapher/blob/812faa1a314bda9641e4400ecf6acc971d5857d9/site/server/bakeChartsToImages.tsx#L38-L75) and you will see the image URLs contain an md5 hash of the params: https://ourworldindata.org/exports/full-list-cumulative-total-tests-per-thousand-3469a55cef624c7091294a7831cb417e_v93_850x600.svg (to see these images it's easiest to disable JavaScript on the page).

I think the ideal solution is still as mentioned on the [issue](https://github.com/owid/owid-grapher/issues/257#issue-492267898):

> Maybe the ideal solution is that the thumbnail should only be kept if it matches the query params, otherwise a loading indicator can be shown.

This is a bit involved to implement – we need to pass the image URL of the thumbnail and ideally render the identical HTML at the loading stage. We can also achieve this (for embeds only) by changing the `MultiEmbedder` to wait for all resources (including variable data) to fully load before mounting the React chart. I'm not sure which is better.

Either way, this will likely be difficult to implement, mostly because we don’t yet have a Wordpress setup released publicly that you can use to test this. While we do intend to release one at some point, it's unlikely to be in the next few months unfortunately. Very sorry about that! I have now left a note on the issue regarding this.

danielgavrilov:
I will close this, but please reopen if you make changes to support the case for embeds. Thanks!

PR #410, opened 2020-04-14
Heterogenous time variable data download fix

shaahmad:
**Summary:** This PR fixes data downloads for charts which have a mix of day-based variables (COVID tests, etc.) and year-based variables (GDP-per-capita, etc.). Charts were formerly displaying buggy output interpreting days as years or vice versa. The desired behavior in data downloads for these charts is to display all day-based data for every entity and every day, and then to display only the latest data point for any year-based variables along with the year of that data point.

**Before:**
Screen Shot 2020-04-14 at 6 21 24 PM

**After:**
Screen Shot 2020-04-14 at 6 24 32 PM


charts/DataTab.tsx
```diff
@@ -1,11 +1,12 @@
-import { flatten, uniq, sortBy, extend, csvEscape } from "./Util"
+import { flatten, uniq, sortBy, extend, csvEscape, last, first } from "./Util"
 import { Bounds } from "./Bounds"
 import * as React from "react"
 import { computed, action } from "mobx"
 import { observer } from "mobx-react"
 import { ChartConfig } from "./ChartConfig"
 import { faDownload } from "@fortawesome/free-solid-svg-icons/faDownload"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
+import { DimensionWithData } from "./DimensionWithData"
```
breck7, line 9: Note: you might want to rebase onto master. I pushed a change yesterday where I started moving the data classes and interfaces to a subfolder. `DimensionWithData` is now temporarily `ChartDimensionWithOwidVariable` to be more clear specifically what it is referring to. I'm trying to reduce and standardize terminology across the codebase hoping that might make somethings easier (and hopefully I'm not creating merge conflicts to just shuffle bits around).

charts/ChartConfig.tsx
```diff
@@ -237,6 +248,13 @@ export class ChartConfig {
             : formatYear
     }
 
+    // returns entities sorted
+    @computed get uniqueEntitiesAcrossDimensions() {
```
breck7, line 252: nit: maybe just call the function `sortedUniqueEntitiesAcrossDimensions` and ditch the line comment. I prefer long concrete symbol names if it means you can drop a comment.

charts/DataTab.tsx
```diff
@@ -20,49 +21,59 @@ export class DataTab extends React.Component<{
     // Here's where the actual CSV is made
     @computed get csvBlob() {
         const { chart } = this.props
-        const { vardata } = chart
+        const yearIsDayVar = chart.yearIsDayVar
+        const dayIndexedCSV = yearIsDayVar ? true : false
 
         const dimensions = chart.data.filledDimensions.filter(
             d => d.property !== "color"
```
breck7, line 28: This is unrelated to your commit but it surprises me that we'd filter this one out. My understanding is that most of the time color dimension is used it is used for  continents. Not sure what the harm is in including that field in the outputted CSV. As a user I'd expect all data used to make a chart would be available in the csv.

charts/DataTab.tsx
```diff
@@ -20,49 +21,59 @@ export class DataTab extends React.Component<{
...
-        ) as string[]
-        const yearsUniq = sortBy(
-            uniq(flatten(dimensions.map(d => d.yearsUniq)))
-        ) as number[]
+        const uniqueEntitiesAcrossDimensions =
+            chart.uniqueEntitiesAcrossDimensions
+
+        // only get days if chart has a day-indexed variable, else get years across dimensions
+        const indexingYears = sortBy(
+            dayIndexedCSV
```
breck7, line 35: Have you thought about splitting this into some smaller methods? A general csvBlob() method, that does the check is dayIndexedCSV, and then 2 methods one that generates the CSV for the Year style charts and the other for the day style charts? Might then be easier to unit test those 2. Or might not be necessary.

charts/DimensionWithData.ts
```diff
@@ -189,6 +194,14 @@ export class DimensionWithData {
         return this.variable.entities
     }
 
+    /*  returns [year of latest value, latest value] for a given entity and variable
```
breck7, line 197: nit: I think you can remove this comment any rely on TypeScript's inference

breck7:
Before and after's look good!

Tests I think would help a lot for this one.

Long term what do you think about moving these methods closer to something like "OwidVariableSet.ts"? I'm wondering if in theory it would make more sense to generate the CSV off all the variables alone, and not take into account the chart or dimensions at all. That would probably be some cleanup to do down the line though separately.

charts/DataTab.tsx
shaahmad: Got it, will fix.
Do you think it is necessary to specify "Owid" variable?

charts/ChartConfig.tsx
shaahmad: Makes sense, will do

charts/DimensionWithData.ts
shaahmad: I did want to specify more than the type here, though. I wanted to specify what the values in the tuple represented. I could rename the function, but I'm afraid for clarity it would have to be yearAndValueOfLatestValueforEntity(). Alternatively, I could create local variables in the function called latestValue, latestYear, and return a tuple of those. What do you think is the best style?

charts/DataTab.tsx
breck7: > Do you think it is necessary to specify "Owid" variable?

Maybe? Not sure. Sometimes when in doubt about these things I pick the more obscure word that will be easier to do a global rename (easier to rename all OwidVariables to Variables than vice versa). What I like about OwidVariable is that it maps to something very specific in our domain (maps to a database table, etc) whereas the general term "Variable" is heavily overloaded.

charts/DimensionWithData.ts
breck7: Ah, I get it. I'm fine with `yearAndValueOfLatestValueforEntity`.

Otherwise I'd vote for local variables over comments so the code is self-documenting. But it's a nit.

Tangential: I wonder if we should have type aliases "declare type owidValue = number` and `declare type owidYear = number`, which would indicate that they are ultimately coming from the DB. That might also be a helpful step in identifying all the places we need to update the code for generalizing year.

shaahmad: > Ah, I get it. I'm fine with yearAndValueOfLatestValueforEntity.

Sounds good, I agree.

> Tangential

I like that idea -- maybe as oYear or something short if we do it across the codebase :)

charts/DataTab.tsx
shaahmad: @vanwoerden @jasoncrawford Thoughts? Easy fix.

charts/DataTab.tsx
shaahmad: That makes sense.

charts/DataTab.tsx
shaahmad: So I thought hard about this one as I agree the method has at least three different points where it checks if the current chart is day-based or year-based, suggesting there might be two different branches of logic that could be separated into two different methods. I decided against it for now, for the following reasons:
1. The amount of shared logic exceeds the amount of unshared logic for the two kinds of charts. `csvBlob()` has a particular logic for iterating through entities and years to generate the final data download, and I think separating the two kinds of charts might require me to reproduce that shared logic across two different functions. The checks for day-based vs year-based charts are small modifications on formatting, not on how the CSV is basically constructed. I'm not sure I can improve readability here significantly without created reproduced logic.
2. I could create a `csvConfig` object with `titleRow` and other parameters that differ between day-based and year-based charts, but I'm not sure I could simplify `csvBlob()` that much except for getting rid of the lines populating `titleRow` and `indexingYears`.

If you have the time and desire, and would be interested in suggesting an alternative, I'm very open to learning!

shaahmad:
> Great to see the tests.
> 
> For the text fixtures, do you think you could filter these variable sets, reduced to be closer to the minimum size necessary to recreate the potential issue? Maybe shrink them so there is only about 30 rows in the generated csv?
> 
> DataTab.test.tsx.snap and the variableset files add about 3MB+ which is enough that you might notice it on every test run from here on out.

Good catch. I'll do that.

shaahmad:
> > Great to see the tests.
> > For the text fixtures, do you think you could filter these variable sets, reduced to be closer to the minimum size necessary to recreate the potential issue? Maybe shrink them so there is only about 30 rows in the generated csv?
> > DataTab.test.tsx.snap and the variableset files add about 3MB+ which is enough that you might notice it on every test run from here on out.
> 
> Good catch. I'll do that.

Didn't quite make it to 30 lines, but I was able to cut the test footprint by ~95%, which is roughly in line with our other tests.

Although I could split it into separate unit tests that assess different behaviors and reduce filesizes further, I wanted to capture all the implicit rules we have for building data downloads e.g. filtering color (currently), entity followed by code, etc., filtering empty rows, formatting days vs years, fixing specific columns, or any other behaviors I haven't formally picked up on. I thought it might have a good idea to have such a "baseline" test for DataTabs as none yet exist.

I could further reduce the footprint by removing entities that don't add additional behavior checks, but thought the few KBs wouldn't be worth the developer time at just this moment. I can swing back around at a later point to clean them up further.

breck7:
> Didn't quite make it to 30 lines, but I was able to cut the test footprint by ~95%, which is roughly in line with our other tests.

Great much better than a few megs!

PR #420, opened 2020-04-24
Floating country select

danielgavrilov:
Implements a sticky country select component that changes the countries on all charts:

### Wide mode (desktop)

Screenshot 2020-04-24 at 11 21 47

### Narrow mode (mobile)

Screenshot 2020-04-24 at 11 22 03


danielgavrilov:
I will merge this right away because it's been super delayed and the authors are waiting for it.

Most of the code is in `GlobalEntityControl.tsx` and I don't think that component is important because it's so isolated, and we might drop it altogether – it's just an experiment.

But `GlobalEntitySelection.ts` is maybe something worthwhile reviewing because we might build on it...

PR #433, opened 2020-05-05
COVID dashboard with feature flag

danielgavrilov:
@breck7 I based this on top of the `covid-chart-builder` branch, if some commits are missing feel free to push to it, or create another PR.

I think now that we are both working on this it would help to keep everything on master, but if there are parts that touch the live code we're not sure about, happy to keep basing off `covid-chart-builder`.

There is a `COVID_DASHBOARD` environmental variable that is enabled on the `staging` and `hans` servers, but not on `live`.


breck7:
Great, I think this is fine. Most of this stuff is built on top of/orthoganal to the rest.

PR #441, opened 2020-05-12
Consistent country colors on COVID Data Explorer

danielgavrilov:
Notion issue: https://www.notion.so/owid/Ensure-country-colors-don-t-change-as-you-select-countries-a0a033f788aa415d844a657fc2b839ce

- Assigns a color to a country on select. Uses that color for any chart, until the country is unselected.
- Adds a small, colored rectangle in the country list.

Screenshot 2020-05-12 at 16 31 24

### DIFF
--- a/charts/Util.ts
+++ b/charts/Util.ts
@@ -81,6 +81,7 @@ export {
     every,
     min,
     max,
+    minBy,
     maxBy,
     compact,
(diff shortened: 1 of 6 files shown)


breck7:
Nice! I like the color legend.

PR #446, opened 2020-05-20
Change delimiter in country query parameter to support Facebook URLs

breck7:
This addresses https://github.com/owid/owid-grapher/issues/397

It maintains backwards compatibility with the current URLs (called "v1"), while from here on out using a new schema ("v2"). v2 uses "|" as the delimiter, and concatenates the entities and then encodes the whole string. v2 always contain a "|", so if there is only 1 entity I append an extra "|" just to differentiate between v1 and v2.

The downside is the URLs are not as pretty, so very open to suggestions. Another potential tweak is to only use v2 style URLs for the Facebook share button (but then we'd have the problem if someone just copies/pastes the url into FB). Or we could use a different delimiter that doesn't have the Facebook problems (~ or _ perhaps).

### DIFF
--- a/charts/ChartData.ts
+++ b/charts/ChartData.ts
@@ -378,11 +378,7 @@ export class ChartData {
     }
 
     @computed get selectedEntityCodes(): string[] {
-        return uniq(
-            this.selectedKeys
-                .map(k => this.lookupKey(k).shortCode)
-                .map(encodeURIComponent)
-        )
+        return uniq(this.selectedKeys.map(k => this.lookupKey(k).shortCode))
     }
 
     @computed get selectedKeys(): EntityDimensionKey[] {
--- a/charts/ChartUrl.ts
+++ b/charts/ChartUrl.ts
@@ -49,15 +49,22 @@ export interface ChartQueryParams {
     endpointsOnly?: string
 }
 
+// Todo: ensure entityCodeOrName never contain the delimiter "|"
+declare type entityCodeOrName = string
+
 export class EntityUrlBuilder {
     private static v1Delimiter = "+"
     private static v2Delimiter = "|"
 
-    static entitiesToQueryParams(countries: string[]) {
-        return encodeURIComponent(countries.join(this.v2Delimiter))
+    static entitiesToQueryParams(entities: entityCodeOrName[]) {
+        // Always include a | in a v2 link
+        if (entities.length === 1)
+            return encodeURIComponent(entities[0] + this.v2Delimiter)
+
+        return encodeURIComponent(entities.join(this.v2Delimiter))
     }
 
-    static queryParamToCountries(queryParam: string) {
+    static queryParamToEntities(queryParam: string) {
         // First preserve handling of the old v1 country=USA+FRA style links. If a link does not
         // include a | and includes a + we assume it's a v1 link. Unfortunately link sharing
         // with v1 links did not work on Facebook because FB would replace %20 with "+".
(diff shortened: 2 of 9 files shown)


breck7:
This is live for testing on Playfair:

https://playfair-owid.netlify.app/coronavirus-data-explorer?casesMetric=true&totalFreq=true&smoothing=0&country=USA%7CGBR%7CCAN%7CBRA%7CAUS%7CIND%7CESP%7CDEU%7CFRA

https://playfair-owid.netlify.app/grapher/share-of-violent-deaths-prehistoric-archeological-sites?country=Bogebakken%20(Denmark)%3B%204300%20-%203800%20BCE%7CBritish%20Columbia%20(30%20sites)%3B%203500%20BCE%20-%201674%20CE%7CBrittany%3B%206000%20BCE

jasoncrawford:
Maybe `,` for delimiter? Thanks for tackling this bug @breck7

MarcelGerber:
> Maybe `,` for delimiter?

I think the issue with that is that `,` can likely be part of some entity name, and then it wouldn't be unique any more. `|` is (more) unlikely to be included in an entity name.

breck7:
@danielgavrilov exactly, "+" are encoded to "%2B". 

@jasoncrawford as @MarcelGerber says "," appears frequently in names (398). "_" is also used (13). "." is used (124). "-" is used (695).

However, does not look like "~" is used at all and that does not need escaping so the URLs will be more readable than pipe. Will try that now.

MarcelGerber:
Here's a list of all the characters that don't need url escaping ("unreserved characters"): https://en.wikipedia.org/wiki/Percent-encoding#Types_of_URI_characters

There are not many special characters available there, and with other delimiters the URLs get less human-readable (not the main goal, but not entirely irrelevant nonetheless).
So `~` is probably a good choice; I can't imagine a case right now where it is used in an entity. I don't know if we would need an escape hatch for that scenario, or if simply trusting or requiring all entity keys to not contain `~` is enough.

breck7:
Okay I tested "~" with Facebook and everything works. I think that's the way to go. v2 links will always contain a "~", and no existing links contain a "~", so we can easily distinguish between the two and not break existing links. Going forward all links will use "~" as the entity delimiter.

danielgavrilov:
Actually, could `,` possibly work, because 
```js
encodeURIComponent(",") === "%2C"
```

breck7:
@danielgavrilov "," appears 398 times in entity names, so we'd have to do a double escape (to distinguish between a "%2C" as a delimiter versus a part of the character name).

It also loses the nice readability aspect of using a character that doesn't require encoding "JPN~USA~GB" verus "JPN%2CUSA%2CGB".

Another approach suggested by @jasoncrawford is to ensure all entities have a slug friendly URL name—we could just generate a CODE for them?

MarcelGerber:
> Actually, could `,` possibly work, because

To make it clear, the issue here is that both "part-of-entity" commas and delimiters would get encoded to `%2C`, since `encodeURIComponent` would have to be applied to both of them.
That way we would have less human-readable URLs _and_ ambiguous parameters.

danielgavrilov:
Ah right :) I missed that wikipedia list. I am happy with `~`!

MarcelGerber:
Hm, I just noticed that both Github and Twitter don't include a trailing `~` in an URL.

image

https://ourworldindata.org/coronavirus-data-explorer?zoomToSelection=true&deathsMetric=true&dailyFreq=true&perCapita=true&smoothing=7&country=AGO~


Maybe to stop this we can _prepend_ the `~` to the countries?

jasoncrawford:
Uh-oh. Do they include internal `~`s? Or does a `~` break the URL? If the
latter then this may be a showstopper.

MarcelGerber:
No, internal ones are still included, at least for these two services (they're the only ones I tried).
So it's just an issue if there's only one country, and should be easily fixable.

breck7:
Oh nice catch @MarcelGerber! Whether to put the `~` as a prefix or suffix was arbitrary and I didn't have a reason to prefer one to the other. Now I do! Will push that now. Thanks.

PR #447, opened 2020-05-21
Sparkbars match chart color scheme

shaahmad:
**Before:**
image

**After:**
image

**Before:**
image

**After:**
image


MarcelGerber:
LGTM!

But through http://localhost:3030/admin/test/embeds?type=ChoroplethMap, I've found one example where it doesn't work at all:
image
(http://localhost:3099/grapher/pupil-teacher-ratio-for-primary-education-by-country)

I don't know what's the solution here - maybe a way to calculate the contrast between the white background and the color?

danielgavrilov:
Code looks good to me!

But we need to do something clever for certain color schemes. Multi-hue color scales might be a bit misleading, e.g. red-blue, because colors used in the legend have a "quantitative meaning" (e.g. a country is blue, but its bars are red).

Would be good to hear what @vanwoerden thinks!

shaahmad:
@danielgavrilov @MarcelGerber Multicolor-scale, inverted, and custom maps should be all addressed now.

charts/ColorSchemes.ts
```diff
@@ -4,44 +4,60 @@ import { clone, lastOfNonEmptyArray } from "./Util"
...
-    Paired: "Paired colors",
-    Pastel1: "Pastel 1 colors",
-    Pastel2: "Pastel 2 colors",
-    Set1: "Set 1 colors",
-    Set2: "Set 2 colors",
-    Set3: "Set 3 colors",
-    PuBu: "Purple-Blue shades",
-    "hsv-RdBu": "HSV Red-Blue",
-    "hsv-CyMg": "HSV Cyan-Magenta"
+type colorSchemeProps = { longSchemeName: string; singleColorScale: boolean }
```
MarcelGerber, line 7: I would go for `displayName` instead of `longSchemeName`.
Also types are usually uppercase (`ColorSchemeProps`).

charts/ColorSchemes.ts
```diff
@@ -65,21 +81,32 @@ function interpolateArray(scaleArr: string[]) {
...
-    constructor(name: string, colorSets: Color[][], isDistinct?: boolean) {
+    constructor(
+        name: string,
+        colorSets: Color[][],
+        singleColorScale?: boolean,
+        isDistinct?: boolean
+    ) {
         this.name = name
         this.colorSets = []
+        this.singleColorScale = singleColorScale
```
MarcelGerber, line 109: Maybe you can go with `this.singleColorScale = !!singleColorScale` to get rid of `undefined`, like in the line below?

shaahmad: done

shaahmad:
Will go ahead and merge this if no one has any further comments?

charts/MapData.ts
```diff
@@ -426,6 +426,10 @@ export class MapData extends ChartTransform {
         return colorScheme !== undefined ? colorScheme : this.defaultColorScheme
     }
 
+    @computed get singleColorScale(): boolean {
+        return this.colorScheme?.singleColorScale
```
danielgavrilov, line 430: Can `colorScheme` be undefined? If it can, we should change its type!

shaahmad: Ah it cannot -- must've been a fragment of an earlier revision I had. Fixed and will push.

PR #450, opened 2020-05-23
Add tracking to determine how often broken entity names are happening

breck7:
Before I deploy the `~` fix, I thought it might make sense to see how often we are seeing broken country links coming in. Then hopefully this will go down to zero when we switch to `~`.


danielgavrilov:
Great idea! One small issue is that the floating country select also calls that method to set countries. Seems like the easiest way to get around that is to pass a flag to `setSelectedEntitiesByCode`, though not the cleanest?

PR #456, opened 2020-05-28
Use shorter date format on x tick marks for all but last tick

breck7:
Before:
Screen Shot 2020-05-27 at 4 40 30 PM

After:
Screen Shot 2020-05-27 at 4 40 37 PM

I'm not crazy about passing the ChartConfig all the way down to HorizontalAxis, but I also didn't like passing 2 format functions or 2 format strings down, particularly because I'm not sure if we would ever use that option. If this becomes a pattern where we often want to decorate the last tick mark differently, perhaps we could implement something like that.


danielgavrilov:
Closing this in favour of #457

PR #457, opened 2020-05-28
Use shorter date format on x tick marks for all but last tick (alternative approach)

danielgavrilov:
@breck7 An alternative approach to #456 that doesn't pass `chart` down.

Was curious if it can be done, I might've missed some details!

### DIFF
--- a/charts/ChartConfig.tsx
+++ b/charts/ChartConfig.tsx
@@ -383,9 +383,7 @@ export class ChartConfig {
             ? (day: number, options?: TickFormattingOptions) =>
                   formatDay(
                       day,
-                      options && options.isFirstOrLastTick === false
-                          ? { format: "MMM D" }
-                          : undefined
+                      options?.isFirstOrLastTick ? {} : { format: "MMM D" }
                   )
             : formatYear
     }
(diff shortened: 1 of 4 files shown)


danielgavrilov:
Both of our approaches will break in 7 months though. And they may become broken if we include data from 2019.

I think the solution to that might be to generate better ticks that fall onto exact crossover points, e.g. first day of the month, and then use [multi-format time formatting as D3 does](https://github.com/d3/d3-time-format#d3-time-format). D3's multi-format basically relies on `roundYear(date) === date`.

breck7:
I like it!

Pushed 2 adjustments.

1) Regarding the multi years problem, I think it's a slight improvement to show the year on the first tick as well. We do already have some 2019-2020 charts. So I changed `isLastTick` to `isFirstOrLastTick`. (Sidenote: I second the idea of improving the general tick strategy for dates and created a notion ticket for it).

image

2) I use `isFirstOrLastTick` in the `tickPlacements` function as well to avoid this overlap:
Screen Shot 2020-05-28 at 8 42 12 AM

danielgavrilov:
:+1: Looks good! 

I would even say we can leave out the "is first tick" check, since it's not really confusing if December is shown, and it's not shown in most cases anyway. But happy if it stays for technical correctness! 

@breck7 Feel free to merge if you are happy with it

PR #463, opened 2020-06-02
Add new styling for log linear toggle

breck7:
Before & Afters:

Screen Shot 2020-06-02 at 12 50 08 PM

### DIFF
--- a/charts/Controls.tsx
+++ b/charts/Controls.tsx
@@ -509,7 +509,13 @@ export class Controls {
 
     @observable isShareMenuActive: boolean = false
     @observable isSettingsMenuActive: boolean = false
-    @observable paddingTop: number = 0
+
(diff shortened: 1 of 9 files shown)


charts/Controls.tsx
```diff
@@ -734,28 +724,29 @@ export class ProjectionChooser extends React.Component<{
...
     children: JSX.Element
+    paddingTop?: number
 }> {
     static contextType = ChartViewContext
     context!: ChartViewContextType
 
-    get controlOverlays() {
-        return this.context.chartView.overlays
+    @action setOverlay() {
+        this.context.chartView.overlays[this.props.id] = this
```
danielgavrilov, line 733: Nice! I didn't realise we can just use `this`

PR #464, opened 2020-06-04
Chart table

shaahmad:
This is a work in progress protected by a feature flag. Moving code to master in the spirit of continuous integration. Will continue development there with feature flag disabled on `live`.

This PR has three main components:
1. Feature flag to ensure new DataTable code only runs when enabled. Off by default on `live`.
2. Refactoring of existing DataTable code (majority of changes).
3. A few new DataTable features.

Live on Nightingale with feature flag disabled:
https://nightingale-owid.netlify.app


danielgavrilov:
🙌 Awesome, I like the functional components!

PR #465, opened 2020-06-04
Quantitative color legend on ScatterPlot

danielgavrilov:
Roughly:
- Created a reusable `ColorScale` class which implements the existing Map color scale
- Migrated ScatterPlot, SlopeChart and StackedBar to use this new color scale instead of previous scale that supported categorical values only
- Implemented a `MultiColorPolyline` to support colors changing over time on ScatterPlot

There is a lot of refactoring in this PR as I tried to standardize the color scales as much as possible between charts. Also need to do a pass to standardize the color legends, which I haven't done, and will probably do in a later PR when we try to implement continuous color scales.

### Context

The ScatterPlot color legends in the past only supported **categorical** values, meaning that the **color** dimension could only contain **string values**.

In almost all cases, this was only used to color entities by continent: 
co-emissions-per-capita-vs-gdp-per-capita-international- (2)

---

This PR implements color legends that support numeric values by splitting them into bins, as maps do:

daily-new-confirmed-cases-of-covid-19-tests-per-case (1)


admin/client/EditorColorScaleSection.tsx
```diff
@@ -0,0 +1,430 @@
...
+    @action.bound onEqualSizeBins(isEqual: boolean) {
+        this.props.scale.config.equalSizeBins = isEqual ? true : undefined
+    }
+
+    render() {
+        const { scale } = this.props
+        return (
+            <Section name="Legend">
+                <Toggle
+                    label="Disable visual scaling of legend bins"
```
breck7, line 54: I understand this option for the map color legend, but does it make sense in the scatter color legend?

charts/ColorLegendView.tsx
```diff
@@ -0,0 +1,247 @@
...
+const FOCUS_BORDER_COLOR = "#111"
+
+export interface ColorLegendViewProps {
+    legend: ColorLegend
+    onMouseOver: (d: ColorScaleBin) => void
+    onMouseLeave: () => void
+}
+
+@observer
+export class ColorLegendView extends React.Component<ColorLegendViewProps> {
```
breck7, line 23: Question: could you have both a numeric and categorical legend on the same chart? If not, would it be clearer to do something like `legend = numericLegend || categoryLengend`

charts/ColorLegendView.tsx
```diff
@@ -0,0 +1,247 @@
...
+            props.legend.width,
+            props.legend.height
+        )
+    }
+
+    @computed get legend(): NumericColorLegend {
+        return this.props.legend
+    }
+
+    @action.bound onMouseMove(ev: MouseEvent | TouchEvent) {
```
breck7, line 97: nit: might be helpful to add some explanatory comments in this method. I'm not 100% what is going on.

charts/ColorSchemes.ts
```diff
@@ -1319,7 +1319,7 @@ export const ColorSchemes = (() => {
         [
             [
                 "#98064A", // Africa
-                "#AEC7E8", // Antarctica
+                "#5887A1", // Antarctica
```
breck7, line 1322: (verified that this is a cold ice blue)

charts/MultiColorPolyline.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export type MultiColorPolylineProps = Omit<
+    React.SVGProps<SVGPolylineElement>,
+    "fill" | "stroke" | "points" | "strokeLinecap"
+> & {
+    points: MultiColorPolylinePoint[]
+}
+
+@observer
+export class MultiColorPolyline extends React.Component<
```
breck7, line 77: This is a really cool new feature!

charts/ScatterPlot.tsx
```diff
@@ -89,9 +89,15 @@ export class ScatterPlot extends React.Component<{
                 return that.chart.baseFontSize
             },
             get colorables() {
-                return that.transform.colors.colorables.filter(c =>
-                    that.colorsInUse.includes(c.color)
-                )
+                return that.transform.colorScale.legendData
+                    .filter(bin => that.colorsInUse.includes(bin.color))
```
breck7, line 93: A slight orthogonal bug with this in the explorer that if you click a region and then all of those countries are selected, because they are then assigned fixed colors and no country then is using the "region" color, that region will disappear from the legend. Not something to fix here, just wanted to mention it in case you had any solution ideas.

db/migration/1590583070171-ExtractMapColorScale.ts
```diff
@@ -0,0 +1,138 @@
...
+        ...colorScale
+    }
+}
+
+async function transformAllCharts(
+    queryRunner: QueryRunner,
+    getNewConfig: (oldConfig: any) => any
+) {
+    const charts = (await queryRunner.query(
+        "SELECT charts.id AS id, charts.config AS config FROM charts"
```
breck7, line 102: Do we need to worry about "Revisions"/"Restore", or is that rarely used?

charts/ColorScaleBin.ts
```diff
@@ -0,0 +1,121 @@
...
+        return this.props.label
+    }
+    get text() {
+        return this.props.label || ""
+    }
+    get isHidden() {
+        return false
+    }
+
+    contains(value: string | number | undefined): boolean {
```
breck7, line 49: nit: maybe would be a good candidate for unit tests

admin/client/EditorColorScaleSection.tsx
danielgavrilov: It doesn't, but I'm hoping it will when we create better scatter legends. 

Right now I decided to get this live quicker and completely reuse the old scatter legend with its old, unchanged API, which has multiple issues (e.g. you have to write the labels yourself currently, otherwise you get unlabeled boxes by default). We can improve it if we feed the full `ColorScale#legendData` but I think we can do that later.

Also, it's not uncommon to have checkboxes that do nothing in the Grapher. Our internal tools have a pretty low bar for usability.

charts/ColorLegendView.tsx
danielgavrilov: Technically yes, if a variable contains both numerical and non-numerical values.

Categorical legends are automatically used for `string` values, numerical legends for `number` values. (This is after they are parsed, since in the database it's all `VARCHAR`)

db/migration/1590583070171-ExtractMapColorScale.ts
danielgavrilov: Good point! I didn't think of that. It's rarely used as far as I'm aware, but I'll try to migrate those too

charts/ScatterPlot.tsx
danielgavrilov: Ah right, we would have to disable the automatic color assignment happening in the country selector. Will do it in this PR if I can find an easy way.

charts/ColorLegendView.tsx
danielgavrilov: I have not even looked at this, it's just code that's unchanged from before, just moved to a different file.

danielgavrilov:
Thanks for the comments @breck7! 

I am generally not happy with the legends and haven't refactored them in this PR, just renamed and split them up into different files. I hope we can standardize them better, and share the logic between map & scatter legends – right now they are completely separate.

Having said that, I can see that I named this PR pretty badly – it's more about reimplementing the scales (transforms) rather than the legend (views).

admin/client/EditorColorScaleSection.tsx
danielgavrilov: Hid this for now anyway!

db/migration/1590583070171-ExtractMapColorScale.ts
danielgavrilov: Now added those to the migration

charts/ScatterPlot.tsx
danielgavrilov: Ok I just understood a bit better what you mean.

The current legend matches bins to "point groups" by the `group.color` property. I think that property should contain the original value rather than the derived color from the value.

So e.g. it should contain the tests per case (`6`) instead of the derived color from that (`#c32623`). Those values should be converted to colors at the very end, when rendering the components.

This is how the map legend works – it matches bins by the numeric value instead of the color value. But `ScatterColorLegend` is still looking at derived colors. 

I would like to change it but it includes changing the hover/focus behaviour or Scatter, StackedBar, SlopeChart, and we also don't yet have a good idea for what the legend should ideally look like, and how continuous legends could work.

PR #470, opened 2020-06-08
ControlsOverlay: Support padding in any direction

danielgavrilov:
Very fiddly implementation! I hope I never have to touch this code again. 😄 

The immediate thing this fixes is support for `paddingBottom` in order to fix [Log/lin toggle covers x-axis footnote sometimes](https://www.notion.so/owid/Log-lin-toggle-covers-x-axis-footnote-sometimes-6893edce6a30450f84adc9bbff6c12a5).

@breck7 It would be great if you can look through and let me know if you notice anything weird or anything that should be explained.

I drew this diagram to try to understand what is going on, but sadly we live in an era where code is written in plaintext so I can't embed it:

IMG_1751

### DIFF
--- a/charts/ChartLayout.tsx
+++ b/charts/ChartLayout.tsx
@@ -135,25 +135,21 @@ export class ChartLayoutView extends React.Component<{
         return (
             <React.Fragment>
                 <HeaderHTML chart={layout.props.chart} header={layout.header} />
-                {/* The "chart plot area" div helps highlight the overlay controls on hover,
-                as we don't want to show them when the cursor is over the tabs */}
-                <div className="ChartPlotArea">
-                    <ControlsOverlayView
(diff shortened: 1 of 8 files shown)


breck7:
That code is finicky! I wonder if there's some type of automated unit testing we could do for layouts to see if we have unintentional collisions.

I did a quick scan of the code and makes sense.

The y axis scale toggle for line charts needs some more space:

Screen Shot 2020-06-08 at 8 38 34 AM


And probably the slope toggle as well:
Screen Shot 2020-06-08 at 8 40 10 AM

danielgavrilov:
Ah I didn't realise you created an option to see all charts with log switches! Awesome.

I repositioned the toggles and also noticed another bug caused by using margins, so now it uses padding as before.

PR #471, opened 2020-06-09
DataExplorer DataTable fixes

shaahmad:
All of the changes in DataTableTransform after line 194 are Prettier formatting auto-fixes -- ignore.

**Before:**
image

**After:**
image

### DIFF
--- a/charts/DataTab.tsx
+++ b/charts/DataTab.tsx
@@ -117,8 +117,7 @@ export class DataTab extends React.Component<{
                 {DATA_TABLE ? (
                     <div
                         style={{
-                            maxWidth: "100%",
-                            minWidth: "100%",
+                            width: "100%",
                             height: "100%",
                             overflow: "auto"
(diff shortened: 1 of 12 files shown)


charts/DataTab.tsx
```diff
@@ -118,6 +118,7 @@ export class DataTab extends React.Component<{
                     <div
                         style={{
                             maxWidth: "100%",
+                            minWidth: "100%",
```
breck7, line 121: nit: is having these 2 lines the same as having "width: 100%"?

charts/DataTableTransform.ts
```diff
@@ -190,142 +190,154 @@ export class DataTableTransform {
     }
 
     @computed get dimensionsWithValues(): Dimension[] {
```
breck7, line 192: I think this one would benefit from some unit tests in a future commit.

PR #488, opened 2020-06-23
Prominent/Responsive Download Button

shaahmad:
**Before:
image

After:
image

Before:
image

After:**
image

### DIFF
--- a/charts/Controls.tsx
+++ b/charts/Controls.tsx
@@ -36,7 +36,6 @@ import { TimeBound } from "./TimeBounds"
 import { Bounds } from "./Bounds"
 import { MapProjection } from "./MapProjection"
 import { asArray, getStylesForTargetHeight } from "utils/client/react-select"
-import classNames from "classnames"
 
 @observer
 class EmbedMenu extends React.Component<{
@@ -1092,7 +1091,7 @@ export class ControlsFooterView extends React.Component<{
 
         return (
             <div
-                className={classNames("ControlsFooter")}
+                className={"ControlsFooter"}
                 style={{ height: props.controls.footerHeight }}
             >
                 {timelineElement}
--- a/charts/Controls.tsx
+++ b/charts/Controls.tsx
@@ -913,7 +913,7 @@ export class ControlsFooterView extends React.Component<{
                     })}
                     <li
                         className={
-                            "tab clickable icon download" +
+                            "tab clickable icon download-tab-button" +
                             (chart.tab === "download" ? " active" : "")
                         }
                         data-track-note="chart-click-download"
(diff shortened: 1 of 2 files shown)


MarcelGerber:
This gets ugly in the Covid Data Explorer at iPad size:
image

MarcelGerber:
@shaahmad I've experimented with flex stlyes for the footer a bit, and have come up with a solution which I quite like in here: https://github.com/owid/owid-grapher/commit/26dc3f15a7eaef492671a24a17e16bb979a87845
The commit (and branch) is based off your branch, so feel free to use it / merge it into this PR if you also like it.

It accomplishes a nice split between inline controls and the nav footer, has enough space for inline controls on an iPad, and has equal left & right padding for all tabs. And I find having a flex solution nicer than providing explicit widths anyway.

image

charts/client/chart.scss
```diff
@@ -249,6 +254,10 @@ $zindex-IndicatorDropdown: 50;
             border-left: 1px solid #e1e1e1;
             display: table-cell;
             height: $footerRowHeight;
+
+            &.download {
```
breck7, line 258: nit: what's the rationale behind 30%? I'm not quite sure. also, maybe name this downloadButton or something a little more descriptive?

charts/client/chart.scss
```diff
@@ -182,6 +182,7 @@ $zindex-IndicatorDropdown: 50;
     --------------------------------------------- */
 
     $footerRowHeight: 40px;
+    $footerRowHeightMobile: 30px;
```
breck7, line 185: Just wondering why we shrink the footer height on mobile? Is this related or unrelated to adding download text?

shaahmad: Requested by Ernst to increase space on mobile. I believe related to adding the download text.
https://www.notion.so/owid/Users-can-t-find-DATA-download-in-Chart-9df5fea8564b43c996f2b82a07f73936

shaahmad:
@MarcelGerber 
Thanks for taking a look! I tried merging your branch, but am getting this on iPad emulation:
image



Also iPhone X emulation seems to have tabs disappear:
image


I'll look into using flex styles here! Let me know if I'm missing something with your branch here.

charts/client/chart.scss
shaahmad: 30% looked nice and dynamically resized better than an explicit pixel count.
Will rename to `download-tab-button`!

shaahmad:
I went ahead used and some flex stying. Looks good across all my device emulations:

**Desktop Before:**
image

**Desktop After:**
image



**iPhone 11 Before:**
image

**iPhone 11 After:**
image



**iPad Before:**
image

**iPad After:**
image

charts/Controls.tsx
```diff
@@ -1091,7 +1092,7 @@ export class ControlsFooterView extends React.Component<{
 
         return (
             <div
-                className="ControlsFooter"
+                className={classNames("ControlsFooter")}
```
MarcelGerber, line 1095: nit: don't really need that `classNames` here any more (as well as the import)

PR #492, opened 2020-06-25
CSVGenerator Refactor

shaahmad:
Refactoring of CSVGenerator as promised. @breck7

### DIFF
--- a/charts/CSVGenerator.tsx
+++ b/charts/CSVGenerator.tsx
@@ -161,7 +161,7 @@ export class CSVGenerator {
     }
 
     @computed get dataRows() {
-        const { indexingYears } = this.csvParams
+        const { indexingYears } = this.timeUnitDependentParams
         const chartEntities = this.chart.sortedUniqueEntitiesAcrossDimensions
 
         const rows: (string | number)[][] = []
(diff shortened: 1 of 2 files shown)


charts/CSVGenerator.tsx
```diff
@@ -1,91 +1,200 @@
...
+            const yearAndValue = this.yearAndValueForSingleYearDimension(
+                dim,
+                entity
+            )
+            if (yearAndValue !== null) {
+                values.push(...yearAndValue)
+                rowHasSomeValue = true
+            } else values.push("", "")
+        })
+        return rowHasSomeValue ? values : []
```
breck7, line 149: nit: Would be a little less code and probably wouldn't be slower to ditch `rowHasSomeValue` and return something like `values.some(v => v !== "") ? values : []`

charts/CSVGenerator.tsx
```diff
@@ -1,91 +1,200 @@
...
+            return [latestYear, latestValue]
+        } else return null
+    }
+
+    private dimensionsValues(entity: string, year: number) {
+        const values: (string | number)[] = []
+        let rowHasSomeValue = false
+
+        this.allValueDimensions.map(dim => {
+            const value = this.valueForDimensionEntityYear(dim, entity, year)
```
breck7, line 131: nit: why the extra method and not just `dim.valueByEntityAndYear.get(entity)?.get(year)`

charts/CSVGenerator.tsx
shaahmad: Good idea -- removed it.

charts/CSVGenerator.tsx
shaahmad: Thought it might help with readability.

shaahmad: Will push as is right now to enable review of other PR, but curious to hear your thoughts via Slack.

PR #493, opened 2020-06-25
CSV uses x-axis override

shaahmad:
Resolves https://www.notion.so/owid/CSV-download-joins-years-variable-with-latest-year-instead-of-specified-year-4d39021fc4e6411c8d4fd0dd93845b54


charts/CSVGenerator.tsx
```diff
@@ -108,18 +108,24 @@ export class CSVGenerator {
...
-
-        const latestYearValue = dim.yearAndValueOfLatestValueforEntity(entity)
-
-        if (latestYearValue) {
-            latestYear = dim.formatYear(first(latestYearValue) as number)
-            latestValue = last(latestYearValue)
+        let year = undefined
+        let value = undefined
+
+        if (dim.targetYear && dim.tolerance) {
```
danielgavrilov, line 114: `dim.tolerance` can be `0`, in which case it is falsy. Is this intentional? Shouldn't the targetYear be joined anyway, even if the tolerance is `0`?

charts/CSVGenerator.tsx
```diff
@@ -108,18 +108,24 @@ export class CSVGenerator {
...
+            const dataValues = valuesByEntityAtYears(
+                dim.valueByEntityAndYear,
+                [dim.targetYear],
+                dim.tolerance
+            ).get(entity)
```
danielgavrilov, line 119: This computation is repeated for each entity, which isn't great. But it also seems like it might be some work to try to refactor, and the code wouldn't be as elegant, so I think it's fine to leave it as it is.

Potentially a DataFrame or similar abstraction will make this more efficient and simpler in the future – having the concept of a "join" is exactly what we need between the "single value" and "all values" data frames.

danielgavrilov: Actually, I pushed a branch ([`csv_uses_x_axis_override_optimization`](https://github.com/owid/owid-grapher/commit/1c0cd51a9521533dfa94e739bbaf0ccfdfa69f79)) that extracts a `valuesAtYears()` utility function:

https://github.com/owid/owid-grapher/blob/f1b92b9cd9158f0bda4a0f81578b71dbcca9618b/charts/Util.ts#L591-L609

shaahmad: Ah, that was a heavy function! Very good catch, merged it in.

PR #496, opened 2020-06-30
Support iframeless embeds of CovidDataExplorer

danielgavrilov:
Notion issue: https://www.notion.so/owid/Support-iframeless-explorer-embeds-29e27e393abc421fa9e1d66514bbf717

Turns explorer `<iframe>`s into iframeless `<figure>`s. 

This means we can now embed multiple/many explorers, with or without controls. Authors can start embedding views of the explorer that function and look exactly like charts, with the `hideControls=true` query parameter.

In another PR, we will redirect old charts to explorer views, so that:
- authors don't have to manually replace all the old charts
- existing embeds used externally switch to using the explorer


site/client/figures/Figure.ts
```diff
@@ -0,0 +1,13 @@
+import { GlobalEntitySelection } from "../global-entity/GlobalEntitySelection"
+
+export abstract class Figure {
+    abstract get isLoaded(): boolean
```
breck7, line 4: should `isLoaded`, `container`, `boundingRect` be abstract methods or implemented in this base class? seems like the implementations are the same in the derived classes

charts/covidDataExplorer/CovidDataExplorer.tsx
```diff
@@ -1063,6 +1071,32 @@ export class CovidDataExplorer extends React.Component<{
         )
     }
 
+    private observeGlobalEntitySelection() {
```
breck7, line 1074: I'm gonna try and refactor the selection code this week hopefully.

site/client/figures/Figure.ts
danielgavrilov: They probably should, but it just feels inelegant and harder to understand.

Have to pass a subset of props to `super()` then store a prop on the abstract class (can't name it `props` or `container` because those are used in the children classes). You end up following a chain of names in order to get to the fact that `container` is simply passed from the `props`. 

I think repeating the implementation makes the code easier to understand. I've now turned `Figure` into an interface for that reason.

Also, I think @shaahmad can use this abstraction to change the prominent links to include the countries from the floating country select – then the `isLoaded` method would always return `true` for those I imagine, since we don't need to observe them on scroll and load on demand (observing lots of elements makes scrolling expensive).

danielgavrilov:
Thanks a lot for the quick feedback @breck7! I am also happy to resolve the conflicts with the table refactoring, let me know when it's ready!

PR #498, opened 2020-07-03
Map tooltip location fix

shaahmad:
w/ some minor refactoring

**Before:**
image

**After:**
image

**Before:**
ezgif com-video-to-gif

**After:**
ezgif com-video-to-gif (1)

### DIFF
--- a/charts/MapTab.tsx
+++ b/charts/MapTab.tsx
@@ -211,7 +211,7 @@ class MapWithLegend extends React.Component<MapWithLegendProps> {
         return (
             <g ref={this.base} className="mapTab">
                 <ChoroplethMap
-                    bounds={bounds.padBottom(this.mapLegend.height + 15)}
+                    bounds={bounds.padBottom(mapLegend.height + 15)}
                     choroplethData={choroplethData}
                     projection={projection}
                     defaultFill={defaultFill}
--- a/charts/MapTab.tsx
+++ b/charts/MapTab.tsx
@@ -41,7 +41,6 @@ interface MapWithLegendProps {
     projection: MapProjection
     defaultFill: string
     mapToDataEntities: { [id: string]: string }
-    headerHeight: number
 }
 
 @observer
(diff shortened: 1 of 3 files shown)


charts/MapTooltip.tsx
```diff
@@ -71,6 +76,51 @@ export class MapTooltip extends React.Component<MapTooltipProps> {
...
+        return colorScale.singleColorScale && !colorScale.isCustomColors
+            ? this.darkestColorInColorScheme
+            : undefined
+    }
+
+    @computed get tooltipHeightByType() {
+        return {
+            sparkBar: -134,
+            dataValue: -102,
+            noData: -65
```
danielgavrilov, line 102: This is ok, but it feels _hacky_ – the tooltip already knows exactly how big it is: https://github.com/owid/owid-grapher/blob/b3b70fbb8eb7fad2d02935d416effe562a88e133/charts/Tooltip.tsx#L76-L81

So ideally we shouldn't hardcode it. Different browsers can also render it differently, and result in a different height.

As I understand, the problem is that we want the tooltip to have a fixed bottom position, and extend as far as it needs to the top?

We already solve this horizontally for LineChart and AreaChart tooltips – they are "anchored" to an X position, but flip left or right depending on where space is available: https://github.com/owid/owid-grapher/blob/b3b70fbb8eb7fad2d02935d416effe562a88e133/charts/Tooltip.tsx#L36-L44

Could we achieve the same for these map tooltips? Maybe we can add a parameter to `Tooltip` to prefer to show up on top by default, and only swap to bottom if there isn't enough space? (e.g. like tippy's [`placement` parameter](https://atomiks.github.io/tippyjs/#placement))

Let me know if I'm misunderstanding the problem!

charts/MapTab.tsx
```diff
@@ -288,6 +294,7 @@ export class MapTab extends React.Component<MapTabProps> {
                         defaultFill={map.data.colorScale.noDataColor}
                         mapToDataEntities={map.data.mapToDataEntities}
                         formatYear={map.data.formatYear}
+                        headerHeight={layout.header.height}
```
danielgavrilov, line 297: The map ideally shouldn't need to know how big the header above it is. Are we passing it because the tooltip location is relative to the chart bounding box instead of the boundary of the map?

We started using `<ControlOverlay>` to get around this problem – the overlay is always relative to the plot, not the chart box.

shaahmad:
@danielgavrilov I really hated passing those layout params down as well! Thanks for the pointer -- I've pushed a far simpler fix.

shaahmad:
Going to merge it as-is -- please let me know if further changes are needed.

PR #501, opened 2020-07-07
"Related Charts" dynamic links

shaahmad:
This PR:
- syncs "Related Charts" links on pages with the floating country selector list of entities, adding to a link if the link already specifies entities on its own.
- renames `entitiesToQueryParams` to `entitiesToQueryParam` as the return value is a single query parameter
- creates a `splitURLintoPathAndQueryString` function
- makes Charts with disabled "Add country" buttons ignore entity query parameters

Check it out live at: https://nightingale-owid.netlify.app/covid-deaths

Note, we will eventually need to isolate out a `RelatedChartLink` component out of the `ProminentLink` component on the Wordpress, especially if the floating country selector expands to appear across the website.

### DIFF
--- a/site/client/blocks/ProminentLink/ProminentLink.tsx
+++ b/site/client/blocks/ProminentLink/ProminentLink.tsx
@@ -82,15 +82,10 @@ class ProminentLink extends React.Component<{
     }
 
     render() {
-        const updatedURL = this.originalURLPath.includes(
-            "total-covid-deaths-region"
-        )
-            ? this.props.originalURL
-            : this.updatedURL
         return (
             <a
                 dangerouslySetInnerHTML={{ __html: this.props.innerHTML ?? "" }}
-                href={updatedURL}
+                href={this.updatedURL}
             />
         )
     }
--- a/site/client/blocks/ProminentLink/ProminentLink.tsx
+++ b/site/client/blocks/ProminentLink/ProminentLink.tsx
@@ -23,12 +23,12 @@ class ProminentLink extends React.Component<{
         return splitURLintoPathAndQueryString(this.props.originalURL).path
     }
 
-    @computed get originalURLQueryString(): string | undefined {
+    @computed private get originalURLQueryString(): string | undefined {
         return splitURLintoPathAndQueryString(this.props.originalURL)
             .queryString
     }
 
-    @computed get originalURLQueryParams(): QueryParams | undefined {
+    @computed private get originalURLQueryParams(): QueryParams | undefined {
         const { originalURLQueryString } = this
 
         return originalURLQueryString
@@ -36,7 +36,7 @@ class ProminentLink extends React.Component<{
             : undefined
     }
 
-    @computed get originalURLEntityCodes(): string[] {
+    @computed private get originalURLEntityCodes(): string[] {
         const originalEntityQueryParam = this.originalURLQueryParams?.[
             "country"
         ]
@@ -50,13 +50,13 @@ class ProminentLink extends React.Component<{
             : []
     }
 
-    @computed get entitiesInGlobalEntitySelection(): string[] {
+    @computed private get entitiesInGlobalEntitySelection(): string[] {
         return Grapher.globalEntitySelection.selectedEntities.map(
             entity => entity.code
         )
     }
 
-    @computed get updatedEntityQueryParam(): string {
+    @computed private get updatedEntityQueryParam(): string {
         const newEntityList = union(
             this.originalURLEntityCodes,
             this.entitiesInGlobalEntitySelection
@@ -65,7 +65,7 @@ class ProminentLink extends React.Component<{
         return EntityUrlBuilder.entitiesToQueryParam(newEntityList)
     }
 
-    @computed get updatedURLParams(): QueryParams {
+    @computed private get updatedURLParams(): QueryParams {
         const { originalURLQueryParams, updatedEntityQueryParam } = this
 
         return {
@@ -76,7 +76,7 @@ class ProminentLink extends React.Component<{
         }
     }
 
-    @computed get updatedURL() {
+    @computed private get updatedURL() {
         return this.originalURLPath + queryParamsToStr(this.updatedURLParams)
     }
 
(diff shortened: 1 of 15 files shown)


utils/client/url.ts
```diff
@@ -65,3 +65,10 @@ export function setWindowQueryStr(str: string) {
         window.location.pathname + str + window.location.hash
     )
 }
+
+export function splitURLintoPathAndQueryString(
+    url: string
+): { path: string; queryString: string | undefined } {
+    const [path, queryString] = url.split(/\?/)
+    return { path: path, queryString: queryString }
```
breck7, line 73: nit: `{ path, queryString }`

site/client/blocks/index.ts
```diff
@@ -3,6 +3,7 @@ import {
     render as renderAdditionalInformation
 } from "./AdditionalInformation/AdditionalInformation"
 import { render as renderHelp } from "./Help/Help"
+import { render as renderProminentLink } from "./ProminentLink/ProminentLink"
```
breck7, line 6: nit: just rename the export to `renderProminentLink` and then can import `renderProminentLink

site/client/blocks/ProminentLink/ProminentLink.tsx
```diff
@@ -0,0 +1,112 @@
...
+import {
+    strToQueryParams,
+    queryParamsToStr,
+    splitURLintoPathAndQueryString,
+    QueryParams
+} from "utils/client/url"
+import { EntityUrlBuilder } from "charts/ChartUrl"
+import { union, isEmpty } from "charts/Util"
+
+const PROMINENT_LINK_CLASSNAME = "wp-block-owid-prominent-link"
```
breck7, line 15: nit: can we export this or import it from somewhere since it's used in many places?

site/client/blocks/ProminentLink/ProminentLink.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+@observer
+class ProminentLink extends React.Component<{
+    originalURL: string
+    innerHTML: string | null
+}> {
+    @computed get originalURLPath() {
```
breck7, line 22: make anything not intended to be externally called "private"?

site/client/blocks/ProminentLink/ProminentLink.tsx
```diff
@@ -0,0 +1,112 @@
...
+        return splitURLintoPathAndQueryString(this.props.originalURL).path
+    }
+
+    @computed get originalURLQueryString(): string | undefined {
+        return splitURLintoPathAndQueryString(this.props.originalURL)
+            .queryString
+    }
+
+    @computed get originalURLQueryParams(): QueryParams | undefined {
+        const { originalURLQueryString } = this
```
breck7, line 32: nit: if `originalURLQueryString` is only used here, maybe ditch the method and do it inline?

site/client/blocks/ProminentLink/ProminentLink.tsx
```diff
@@ -0,0 +1,112 @@
...
+        .querySelectorAll<HTMLElement>(`.${PROMINENT_LINK_CLASSNAME}`)
+        .forEach(el => {
+            const anchorTag = el.querySelector("a")
+            if (!anchorTag) return
+
+            const originalUrl = anchorTag.href
+            const innerHTML = anchorTag.innerHTML
+
+            const rendered = (
+                <ProminentLink
```
breck7, line 104: Should this be a React component, or would it suffice to have a find/replace of the `href` attribute with entities inserted into query string? If we replace the full `a` tag, what happens if authors had specified other attributes on the a tag?

utils/client/url.ts
shaahmad: Yes!! I was trying to get the syntax for this right, but couldn't figure it out. Thanks!

site/client/blocks/ProminentLink/ProminentLink.tsx
shaahmad: It's a React component so it can react dynamically to changes in the floating country selector.

Ah, good point! I can bring over all attributes on the anchor tag, plus the original div.

site/client/blocks/ProminentLink/ProminentLink.tsx
shaahmad: So I'm curious about your philosophy here. I sometimes prefer to separate the name from the access logic to maximize readability and keep each function simple. The downside is a bit of indirection as noted here: https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection

With modern code editors/instant previews of function definitions, I find most of the cost can basically be avoided, and that the primary benefits actually accrue to the reader (they can understand the point of the code rather than the access logic). What do you think?

site/client/blocks/ProminentLink/ProminentLink.tsx
shaahmad: I'm not sure I understand. This constant is only used once on line 95.

breck7: Sorry, I meant it's used in 3-4 places in other files besides this one.

site/client/blocks/ProminentLink/ProminentLink.tsx
breck7: Either way is fine. I was thinking if it's only used once, maybe `const originalURLQueryString = splitURLintoPathAndQueryString(this.props.originalURL).queryString` is slightly better because fewer tokens and smaller API surface (the latter can also be achieved by making the method private). But super nit and I think either way is fine and your implementation is better if we were to need that logic twice.

site/client/blocks/ProminentLink/ProminentLink.tsx
breck7: Ah gotcha, forgot for a second that this was dynamic!

breck7: Yes preserving existing attributes will probably be important.

PR #502, opened 2020-07-07
Replace Grapher charts with Explorer charts

danielgavrilov:
In order to enable the redirects, we need to populate the `chartExplorerRedirects` in `site/server/bakeCovidExplorerRedirects.tsx`.

Right now there is a single example entry which is commented out. I've deployed this branch with the entry on a staging server: https://playfair-owid.netlify.app/covid-cases#what-is-the-total-number-of-confirmed-cases (you should see the cumulative confirmed cases chart as an explorer embed)

This PR:
- Replaces the matched Grapher embeds with Explorer embeds (at the baking step)
- Creates a client-side redirect from matched Chart pages to Explorer pages (while preserving any user-specified queries, e.g. if a user has selected a bunch of countries on the chart, that selection will be passed onto the explorer too)

There are still some issues to resolve in the explorer in order to be able to replace maps (like [Improve the automatic binning algorithm for legend bins](https://www.notion.so/owid/Improve-the-automatic-binning-algorithm-for-legend-bins-d6090ee49df5461aa6fd4fddc44d34a6) and [Y-axis and map legends should start at zero](https://www.notion.so/owid/Y-axis-and-map-legends-should-start-at-zero-e62f0497a2f0430883fc4b24cd6540ac)).


charts/covidDataExplorer/CovidDataExplorer.tsx
```diff
@@ -88,6 +91,25 @@ export class CovidDataExplorer extends React.Component<{
         )
     }
 
+    static async redirectAndBootstrap(
```
breck7, line 94: Nit: This method is just doing a replaceState right? Maybe better to not have `redirect` in the name?

breck7:
Which chart on the playfair server has the embed? Sorry I'm not seeing one.

charts/covidDataExplorer/CovidDataExplorer.tsx
danielgavrilov: Yeah it's a bad name for what it does. Should it be `replaceStateAndBootstrap` maybe?

danielgavrilov:
> Which chart on the playfair server has the embed? Sorry I'm not seeing one.

I had overwritten them with another deploy, they are temporarily still here (this might change): https://playfair-owid.netlify.app/covid-cases

Some of them don't load for some reason, trying to figure out why now.

danielgavrilov:
I didn't manage to figure this out and will have to leave it for next week.

PR #508, opened 2020-07-14
Editor option to control DataTable column selection

shaahmad:
This PR allows one to select whether a chart's DataTable should show Absolute and/or Relative columns for individual dimensions. This behavior extends to the source charts for the Data Explorer.

The controls can be found on the Basic tab under the individual dimension dropdowns.

image

### DIFF
--- a/charts/covidDataExplorer/CovidDataExplorer.tsx
+++ b/charts/covidDataExplorer/CovidDataExplorer.tsx
@@ -150,7 +150,7 @@ export class CovidDataExplorer extends React.Component<{
                 available: true,
                 label: "Confirmed cases",
                 checked: this.constrainedParams.casesMetric,
-                onChange: value => {
+                onChange: () => {
                     params.setMetric("cases")
                     this.renderControlsThenUpdateChart()
                 }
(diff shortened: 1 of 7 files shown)


charts/covidDataExplorer/CovidDataExplorer.tsx
```diff
@@ -751,22 +753,47 @@ export class CovidDataExplorer extends React.Component<{
...
 
+    private _updateTableDisplaySettings() {
+        const tableDisplaySettingsByDimensionProperty: {
+            [dimensionProperty: string]:
+                | OwidVariableTableDisplaySettings
+                | undefined
+        } = {}
+
+        // get table display settings for every dimension on source chart
+        this.sourceChart?.dimensions.forEach(
```
breck7, line 784: Ah, so right now we are just extracting color styles from the "sourcecharts". The variables on the "sourceCharts" are not equal to the columns in the Explorer. So the dimensions don't necessarily align. For the Explorer we'd probably want to hard code the tableDisplaySettings into the column specs.

shaahmad: :( Done.

shaahmad:
Added the hard-codings, switched to opt-out "hide column", and some very minor code cleanup in the CDE. Should be ready to merge.

PR #510, opened 2020-07-17
Timeline Methods Refactor

shaahmad:
I find myself spending a lot of time trying to remember the difference between `startYear()`, `minTime()`, `timeDomain[0]`, `timelineMinTime`, `minTimelineYear`, etc., etc. This _starts_ the process of separating these concepts into hopefully more clear names.

I've just renamed a few methods in this case to try it out, but I think we'd have to do some kind of database migration to change "minTime" and "maxTime" in the DB. Not sure how difficult this is -- I can also do some kind of live conversion in the code.

Leaving this as a draft PR until I know whether the DB migration can be done fairly easily. The code breaks as-is until we figure that out.

### DIFF
--- a/admin/client/EditorCustomizeTab.tsx
+++ b/admin/client/EditorCustomizeTab.tsx
@@ -82,13 +82,6 @@ class TimelineSection extends React.Component<{ editor: ChartEditor }> {
         return this.props.editor.chart
     }
 
-    @computed get minTime() {
-        return this.chart.props.minTime
-    }
-    @computed get maxTime() {
-        return this.chart.props.maxTime
-    }
-
     @computed get timelineMinTime() {
         return this.chart.props.timelineMinTime
     }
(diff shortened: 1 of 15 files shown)


breck7:
>  I can also do some kind of live conversion in the code.

Could be nice if we swapped out the Chart backend with a Git backed system. Then a sitewide migration is a simple commit.

Maybe do the live conversion for now and if we migrate the system we can upgrade all the charts then?

danielgavrilov:
I definitely agree all the time terms are confusing and it would be great to rename them!

Re migrations: I would be in favour of propagating the changes through the database for consistency, rather than doing translation. The chart pages contain the config, and they are regenerated on every deploy, so this switch would happen atomically. (@breck7 one issue with a separate GitHub repo is maybe that we have to be careful to make migrations like these atomic?)

You can create a migration with:

```
yarn typeorm migration:create -n MigrationName
```

And then populate the created file.

You can see an example chart config migration here: https://github.com/owid/owid-grapher/blob/32cb84a8f5f49c7519ced5d647bed29585bd0534/db/migration/1591128415343-NewChartColorScale.ts#L1-L109

One thing to watch out for in migrations is they are kind of frozen in time, so they should avoid importing any current definitions from the codebase, instead everything should ideally be copied.

shaahmad:
Going to close this for now. I'm investing the Timeline for unrelated perf issues, and I'll have a better idea of what methods map to what concepts after I address that.

PR #512, opened 2020-07-17
Dev: add keyboard shortcuts

breck7:
Should we add Mousetrap? I've used it for many years and found it to be a joy. It only adds a few KB.

We could just use it for our own custom dev and debugging needs (and maybe add some features for advanced users).

Thoughts?


jasoncrawford:
I used Mousetrap for something a long time ago. Seems good to me

PR #513, opened 2020-07-21
Data Table Automatic Time Selection

shaahmad:
With this PR, DataTables automatically select a start year where x>=50% of the entities have values. This is intended to reduce tables with an initial view of sparse data. This auto-selection will not trigger if the user has modified the timeline, the authors have specified a default timeline start year, or if the start year is set to the "latest" data.

Live on [Nightingale](https://nightingale-owid.netlify.app/).

~~Note: the Covid Data Explorer already has some lag when switching tabs that is now slightly exacerbated (when Table->Chart) by this code. I'm adding this as an issue [here](https://www.notion.so/owid/Lag-when-switching-tabs-in-CDE-491b7e8577104fd6bc722d163160e208) that I think we should address separately.~~ Fixed.

**Before:**
image

**After:**
image

### DIFF
--- a/charts/DataTableTransform.ts
+++ b/charts/DataTableTransform.ts
@@ -11,7 +11,7 @@ import {
 import { ChartConfig } from "./ChartConfig"
 import { ChartDimensionWithOwidVariable } from "./ChartDimensionWithOwidVariable"
 import { TickFormattingOptions } from "./TickFormattingOptions"
-import { getTimeWithinTimeRange, Time } from "./TimeBounds"
+import { isUnbounded, getTimeWithinTimeRange, Time } from "./TimeBounds"
 import { ChartTransform } from "./ChartTransform"
 
 // Target year modes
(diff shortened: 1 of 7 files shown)


charts/DataTableTransform.ts
```diff
@@ -108,15 +110,65 @@ function getHeaderUnit(unit: string) {
...
+
+        if (
+            this.chart.userHasSetTimeline ||
+            this.initialTimelineStartYearSpecified ||
+            !this.loadedWithData
+        )
+            return undefined
+
+        const numEntitiesInTable = this.entities.length
+
```
breck7, line 139: You could do an ~O(N) algo along the lines of this: 
`groups = groupBy(this.chart.table.unfilteredRows, ["year", dataTableColumn])
sorted = sortBy(groups, group => group[0].year)
entityCount = this.chart.table.entities // there is some computed here with that
autoSelectedStartYear = sorted.find(group => group.length /entityCount > AUTO_SELECTION_THRESHOLD_PERCENTAGE `

breck7: There's probably even a cleaner way to do it, and perhaps a general method we could add to the table class for these sorts of pivots.

shaahmad: Hmm, wouldn't that break on the Data Explorer (or at least require a special case) because the underlying Table includes data not used by the DataTable?

breck7: Yes as I wrote it above. Looking at yours again I'm not sure if algo speed is too important, I was thinking there were a lot of dimensions but usually just a few.

Could be fine as is, but I think this would be a good candidate for a method on the Table class, which can then be easily unit tested. Maybe something like:

```
testCase = `year,gdp,country
2009,123,brazil
2010,456,brazil
2009,122,usa
2011,333,france`

const table = new Table(testCase)
const counts = table.getCountsByYear()
console.log(counts)
/*
{2009: {gdp: 2},
2010: {gdp: 1},
2011: {gdp: 1}
*/
```

charts/DataTableTransform.ts
```diff
@@ -108,15 +110,65 @@ function getHeaderUnit(unit: string) {
...
+        return this.dimensions.length > 0
+    }
+
+    private readonly AUTO_SELECTION_THRESHOLD_PERCENTAGE: number = 0.5
+
+    /**
+     * If the user or the editor hasn't specified a start, auto-select a start year
+     *  where AUTO_SELECTION_THRESHOLD_PERCENTAGE of the entities have values.
+     */
+    @computed get autoSelectedStartYear(): number | undefined {
```
breck7, line 128: This function looks like it would benefit greatly from unit tests to me (not only for preventing regressions, but also would probably point to a more general purpose query API on the table class), but I guess as it is now it is very easy to swap out (since it's just a method returning a number), so the impact of a bug in it wouldn't be too bad. I'd make a "todo" note somewhere to add tests :)

shaahmad: Will add a to-do, I agree!

PR #515, opened 2020-07-24
Turn on Mobx enforceActions flag

breck7:
Often I hit runtime Mobx errors because observables are being changed by regular methods (or computeds). It seems it is recommended to enable the global 'enforceActions' (https://mobx.js.org/refguide/api.html#enforceactions) so an error is thrown if an observable is being updated not in a function with an `@action` decorator. My hope is that this will help make it easier to debug Mobx issues.

At first I tried the "always" setting, but that throws errors when default properties are initialized on object construction (https://github.com/mobxjs/mobx/issues/1702 if curious). So the "observed" setting makes more sense.

### DIFF
--- a/admin/client/Admin.tsx
+++ b/admin/client/Admin.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import * as ReactDOM from "react-dom"
-import * as _ from "lodash"
-import { observable, computed } from "mobx"
+import * as lodash from "lodash"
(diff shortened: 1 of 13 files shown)


shaahmad:
👌 👌 👌

breck7:
I disabled it for now. Still catching lots of them at runtime. Much easier to fix them when unit tests throw, so as we increase our coverage we could go back and reenable.

PR #517, opened 2020-07-28
Sorting UI/UX improvements

shaahmad:
This PR:
- changes the Explorer country picker icon to an alphanumeric icon when sorting by location, and sorts by largest quantities first
- sorts the DataTable columns by largest quantity first and improves icon selection

**Before:**
gif2

**After:**
gif1

**Before:**
gif4

**After:**
gif3

### DIFF
--- a/charts/covidDataExplorer/CovidCountryPicker.tsx
+++ b/charts/covidDataExplorer/CovidCountryPicker.tsx
@@ -90,7 +90,6 @@ export class CountryPicker extends React.Component<{
 
     set metric(metric) {
         this.props.covidDataExplorer.props.params.countryPickerMetric = metric
-        this.sortOrder = metric === "location" ? SortOrder.asc : SortOrder.desc
     }
 
     @computed get sortOrder(): SortOrder {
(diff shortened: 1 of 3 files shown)


charts/covidDataExplorer/CovidCountryPicker.tsx
```diff
@@ -90,6 +90,7 @@ export class CountryPicker extends React.Component<{
 
     set metric(metric) {
         this.props.covidDataExplorer.props.params.countryPickerMetric = metric
+        this.sortOrder = metric === "location" ? SortOrder.asc : SortOrder.desc
```
breck7, line 93: Should this be a computed? What happens if someone changes the sort order?

danielgavrilov: I think I would put this in the `onChange` handler, where the `metric` setter is also called: 

https://github.com/owid/owid-grapher/blob/62185f015e8fff2919a1ada77dc2500aba50f3d7/charts/covidDataExplorer/CovidCountryPicker.tsx#L402-L408

This is just because I try to avoid setters with side effects (though I have probably written some in the past).

A concrete example: Maybe sometime down the line there's this implementation:

```js
this.sortOrder = ...
this.metric = ...
```

Now the order in which you assign these matters, and it's not obvious.

shaahmad: @breck7 not sure if it should be computed, as we need an observable that responds to user interaction. The code should just run if the user changes the metric and sets the "default" sort state. `sortOrder` can also be modified by the user clicking the change sort button.

@danielgavrilov Good idea, I moved it into the onChange handler. I'm not sure what the concern with the order is -- as I use the value of the new metric rather than state. However, if you're referring to potential inconsistent state, I also wrapped it in a runInAction transaction.

charts/covidDataExplorer/CovidCountryPicker.tsx
```diff
@@ -400,11 +400,17 @@ export class CountryPicker extends React.Component<{
                             onChange={option => {
```
breck7, line 400: nit: I think this lambda would now be better as an @action.bound method, as it's gotten decently long. Then you could drop the runInAction.

shaahmad: Good idea, done!

PR #518, opened 2020-07-28
Add weekly, biweekly, and weekly/biweekly change options to Explorer

breck7:
This change adds 8 more chart views + 8 more map views to the Explorer.

It also changes the Interval control to a dropdown to accommodate the new options.

It changes the dailyFreq=true, totalFreq=true query param booleans to a single enum type instead.

New:
Screen Shot 2020-07-28 at 1 17 47 PM

### DIFF
--- a/charts/covidDataExplorer/CovidChartUrl.ts
+++ b/charts/covidDataExplorer/CovidChartUrl.ts
@@ -320,9 +320,20 @@ export class CovidQueryParams {
         let interval: string = this.interval
         if (interval === "smoothed") interval = "daily"
         if (this.isWeekly || this.isBiweekly) interval = "weeklys"
-        return [this.metricName, interval, this.perCapita ? "per_capita" : ""]
+        return [
+            this.metricName,
+            interval,
+            this.perCapita ? "per_capita" : "",
+            this.intervalChange ? "change" : ""
+        ]
             .filter(i => i)
             .join("_")
+    }
+
+    get intervalChange() {
+        if (this.interval === "weeklyChange") return 7
+        else if (this.interval === "biweeklyChange") return 14
+        return undefined
     }
 }
 
@@ -380,12 +391,6 @@ export class CovidConstrainedQueryParams extends CovidQueryParams {
         if (this.isWeekly) return 7
         else if (this.isBiweekly) return 14
         return 1
-    }
-
-    get intervalChange() {
-        if (this.interval === "weeklyChange") return 7
-        else if (this.interval === "biweeklyChange") return 14
-        return undefined
     }
 
     get isWeeklyOrBiweeklyChange() {
--- a/charts/covidDataExplorer/CovidChartUrl.ts
+++ b/charts/covidDataExplorer/CovidChartUrl.ts
@@ -1,4 +1,4 @@
-import { computed, observable } from "mobx"
+import { computed, observable, action } from "mobx"
 import { ObservableUrl } from "../UrlBinding"
 import { ChartUrl, EntityUrlBuilder } from "../ChartUrl"
 import {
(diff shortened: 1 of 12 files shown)


charts/covidDataExplorer/CovidExplorerTable.ts
```diff
@@ -258,12 +260,7 @@ export class CovidExplorerTable {
         return filtered.concat(continentRows, euRows)
     }
 
-    private static buildCovidVariableId(
-        name: MetricKind,
-        perCapita: number,
-        rollingAverage?: number,
-        daily?: boolean
-    ): number {
+    private static buildCovidVariableId(params: CovidQueryParams): number {
```
shaahmad, line 263: Hmm, this makes it more difficult to override specific settings, and will make it a bit more difficult to build views (like the multimetric table) that differ from the URL.

charts/covidDataExplorer/CovidChartUrl.ts
```diff
@@ -322,16 +372,38 @@ export class CovidConstrainedQueryParams extends CovidQueryParams {
...
     }
 
-    @computed get allowEverything() {
+    get allowEverything() {
         return false
     }
 
-    @computed get available() {
+    get rollingMultiplier() {
+        if (this.isWeekly) return 7
```
shaahmad, line 380: Would maybe be nice to isolate a "DataExplorerOptionsState" class from the URL class and have all these derivations in there and keep the URL class simple with just the state of the query params.

charts/covidDataExplorer/CovidChartUrl.ts
```diff
@@ -58,20 +73,19 @@ export class CovidQueryParams {
...
         ]
-        const timelines = ["totalFreq", 7, "dailyFreq"]
         const perCapita = [true, false]
         const aligned = [true, false]
         const yScale = ["log", "linear"] // todo
         const tab = ["map", "chart"] // todo
         const combos: any = []
         metrics.forEach(metric => {
-            timelines.forEach(timeline => {
+            intervalOptions.forEach(interval => {
```
shaahmad, line 82: Nice to get rid of this "timeline" naming collision! Now we just need to resolve our two `Tables`... 😮

charts/covidDataExplorer/CovidExplorerTable.ts
breck7: Will change URL Params to ExplorerOptions in a later commit.

charts/covidDataExplorer/CovidChartUrl.ts
breck7: Where do you see two tables? I count at least 4 😮

garas:
Could it be option _Per million people_ available for _Weekly_ and _Biweekly_ options?

EU counties often uses _14-day cumulative number of COVID-19 cases per 100 000_ for evaluating the situation. _Biweekly per million people_ would be same value, just 10x, and for all world countries.

https://www.ecdc.europa.eu/en/cases-2019-ncov-eueea

breck7:
Hi @garas,

Currently we disable certain combinations of options because either 1) the combination is hard to interpret (i.e. tests per case per million people) or 2) we haven't yet reviewed and adjusted all of the settings for a view (i.e. adjusting the map color scheme for a combination). For the weekly figures we haven't yet adjusted all of the settings for the map colors.

In the interim, I've added an option for advanced users to skip our constraints. Adding `everythingAvailable=true` to the query string will allow you to generate weekly/biweekly per capita charts (Example: https://ourworldindata.org/coronavirus-data-explorer?zoomToSelection=true&casesMetric=true&interval=weekly&perCapita=true&allowEverything=true&smoothing=0&country=USA~GBR~CAN~BRA~AUS~IND~DEU~MEX~ZAF~COL~KOR~NOR~NGA&pickerMetric=location&pickerSort=asc&everythingAvailable=true). Note that these charts aren't "supported" so the color schemes and so forth might not be ideal.

We don't yet support adjusting the per capita divisor to 100,000, but that would be something we could do at some point via a query string parameter.

Cheers,
Breck

PR #520, opened 2020-08-01
Switchr

breck7:
This is a *draft* PR just to provide a heads up that for the CO2 Explorer I need to generalize CountryPicker, Explorer Controls, and the general explorer layout. So if you are working on those maybe let's chat to avoid some merge nightmares next week.

I'm adding more tests and storybooks for everything I change so hopefully will catch any regressions before pushing and will be in a better place than before.


jasoncrawford:
What's up with this PR? Been open a while @breck7

breck7:
@jasoncrawford thanks! closing now and will be reopening a non-draft one today with the explorer.

PR #521, opened 2020-08-02
CO2 Country Profiles and Search Box

shaahmad:
This PR:
- creates dynamic CO2 country profiles
- generalizes the country profile search box.
- replaces some manual bake-step caching code with lodash `memoize`.

**View live on Nightingale [here](https://nightingale-owid.netlify.app/co2-and-greenhouse-gas-emissions-landing-page#co2-and-greenhouse-gas-emissions-country-profiles).**


site/client/SearchCountry.tsx
```diff
@@ -24,21 +24,26 @@ const CovidSearchCountry = () => {
...
     )
 }
 
-export function runCovidSearchCountry() {
-    const elements = Array.from(
-        document.querySelectorAll(".wp-block-covid-search-country")
-    )
-    elements.forEach(element => {
-        ReactDOM.render(<CovidSearchCountry />, element)
+export function runSearchCountry() {
```
breck7, line 35: I know we don't do this yet, but I was thinking for all of these `run` methods, maybe we could have unit tests (or storybooks) to ensure we have some execution of these code paths and also to have some live documentation. Also would be nice to have some infra setup when things go wrong for regression testing (like what happened with multiple data explorers on the same page).

site/server/views/SiteSubnavigation.tsx
```diff
@@ -2,6 +2,8 @@ import * as React from "react"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import { faChevronLeft } from "@fortawesome/free-solid-svg-icons/faChevronLeft"
 
+export const co2CountryProfilePath = "/co2-country-profile"
```
breck7, line 5: nit: should this be in the constants file?

site/server/devServer.tsx
```diff
@@ -122,10 +123,25 @@ devServer.get(`/${covidDashboardSlug}`, async (req, res) => {
 devServer.get(
     `/${covidCountryProfileRootPath}/:countrySlug`,
     async (req, res) => {
-        res.send(await covidCountryProfileCountryPage(req.params.countrySlug))
+        res.send(
+            await countryProfileCountryPage(
+                countryProfileSpecs.find(
```
breck7, line 128: nit: should you store this in a var like you do below for `co2profile` (or vice versa?)

site/server/siteBaking.tsx
```diff
@@ -333,70 +331,55 @@ export async function feedbackPage() {
...
-            cachedCovidCountryProfilePostFormatted,
-            cachedCovidCountryProfilePostFormattingOptions
-        ]
-    }
-
-    // Get formatted content from generic covid country profile page.
-    const genericCountryProfilePostApi = await wpdb.getPostBySlug(
-        covidCountryProfileSlug
-    )
+const getCountryProfilePost = memoize(
```
breck7, line 334: Nothing to do now but I'm wondering if this should go in a separate file, and how we could refactor Baker a bit. I think it would be nice to ditch the global `settings.ts` as much as possible, to make a more functional codebase. That would also allow us to do easier partial baking and baking things to specific locations.

site/client/CountryProfileConstants.tsx
```diff
@@ -0,0 +1,37 @@
...
+    pageTitle: string
+    genericProfileSlug: string
+    landingPageSlug: string
+    selector: string
+    rootPath: string
+}
+
+export const co2CountryProfileRootPath = "co2/country"
+
+export const countryProfileSpecs: CountryProfileSpec[] = [
```
breck7, line 20: Nit: instead of an array a mapped type might be better (you wouldn't have to do the `finds` later on). Something like:

```typescript
export type CountryProfileProject = "coronavirus" | "co2"

export interface CountryProfileSpec {
    project: CountryProfileProject
    pageTitle: string
}

export const countryProfileSpecs: { [key in CountryProfileProject]: CountryProfileSpec } = {
    coronavirus: {
        project: "coronavirus",
        pageTitle: "Coronavirus Pandemic"
    },
    co2: {
        project: "co2",
        pageTitle: "CO2"
    }
}
```

shaahmad: Ah ok, I had played around with using an object, but wanted iterability. ES6 Maps offer both so that's perhaps the better structure.

site/client/SearchCountry.tsx
shaahmad: Definitely something we need. Input would be an HTML page with selectors.

site/server/siteBaking.tsx
shaahmad: I definitely considered it. The baker needs some organization. This PR touches too many files for my liking. Not surprised there's a memory leak :)

PR #523, opened 2020-08-03
Timeline Years Reflect Selections

shaahmad:
This PR makes the following changes to the Timeline:
- The date on the left of the timeline will always show the currently selected start year rather than the minimum year in the data.
- The date on the right of the timeline will show the currently selected end year rather than the maximum year in the data. If there is only one slider (i.e. start year == end year), then this date reflects the maximum year in the data (like before).

**View it live on Staging [here](https://staging-owid.netlify.app//grapher/broadband-penetration-by-country)**

**Before:**
ezgif com-video-to-gif (3)

**After:**
ezgif com-video-to-gif (2)

### DIFF
--- a/charts/HTMLTimeline.tsx
+++ b/charts/HTMLTimeline.tsx
@@ -57,7 +57,7 @@ export class Timeline extends React.Component<TimelineProps> {
 
     disposers!: IReactionDisposer[]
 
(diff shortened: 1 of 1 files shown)


shaahmad:
Closing this for now until we reach a design decision.

PR #529, opened 2020-08-05
DataTable: shortened dates

shaahmad:
**Before:**
image

**After:**
image


danielgavrilov:
We would need to revert this in December this year? (in 4 months basically) Otherwise dates can potentially be ambiguous.

Possibly, there may be cases in the projection datasets where this already makes it ambiguous.

One alternative way may be to exclude the year only if it's the current year? Though not sure if that's reasonable either.

danielgavrilov:
Nevermind, @vanwoerden pointed out that the year is in the column header. 🙂

PR #530, opened 2020-08-05
DataTable: option to hide small countries

shaahmad:
**Before:**
image

**After:**
image

### DIFF
--- a/charts/ChartConfig.tsx
+++ b/charts/ChartConfig.tsx
@@ -696,8 +696,10 @@ export class ChartConfig {
             this.props.tab === "table" &&
             value === "chart" &&
             !this.userHasSetTimeline
-        )
+        ) {
             this.revertAutoSelectedStartYear()
+            this.props.minPopulationFilter && this.toggleMinPopulationFilter()
+        }
 
         if (value === "chart" || value === "map" || value === "table") {
             this.props.tab = value
--- a/charts/ChartConfig.tsx
+++ b/charts/ChartConfig.tsx
@@ -699,12 +699,9 @@ export class ChartConfig {
 
     /** TEMPORARY: Needs to be replaced with declarative filter columns ASAP */
     set tab(value) {
-        if (this.props.tab === "chart" && value === "table")
+        if (this.props.tab === "chart")
             this.chartMinPopulationFilter = this.props.minPopulationFilter
-        if (
-            this.props.tab === "table" &&
-            (value === "chart" || value === "map")
-        )
+        if (this.props.tab === "table" && value !== "table")
             this.revertDataTableSpecificState()
 
         if (value === "chart" || value === "map" || value === "table") {
--- a/charts/ChartConfig.tsx
+++ b/charts/ChartConfig.tsx
@@ -678,28 +678,34 @@ export class ChartConfig {
         return defaultTo(this.props.entityTypePlural, "countries")
     }
 
-    /** If the start year was autoselected in the DataTable, revert. */
+    /** TEMPORARY: Needs to be replaced with declarative filter columns ASAP */
+    private chartMinPopulationFilter?: number = undefined
     @action
-    revertAutoSelectedStartYear() {
-        this.timeDomain = [
-            this.initialProps.minTime ?? TimeBoundValue.unboundedLeft,
-            this.timeDomain[1]
-        ]
+    revertDataTableSpecificState() {
+        /** If the start year was autoselected in the DataTable, revert. */
+        if (!this.userHasSetTimeline)
+            this.timeDomain = [
+                this.initialProps.minTime ?? TimeBoundValue.unboundedLeft,
+                this.timeDomain[1]
+            ]
+
+        /** Revert the state of minPopulationFilter */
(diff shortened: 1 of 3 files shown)


charts/ChartConfig.tsx
```diff
@@ -696,8 +696,10 @@ export class ChartConfig {
             this.props.tab === "table" &&
             value === "chart" &&
             !this.userHasSetTimeline
-        )
+        ) {
             this.revertAutoSelectedStartYear()
```
breck7, line 700: Would it be better to isolate this `this.revertAutoSelectedStartYear()` method in the data table class? Not sure how exactly, but maybe in something like `component...`methods?

charts/ChartConfig.tsx
```diff
@@ -696,8 +696,10 @@ export class ChartConfig {
...
+            this.props.minPopulationFilter && this.toggleMinPopulationFilter()
```
breck7, line 701: It doesn't feel right to call toggleMinPopulationFilter (a user command/action) in a setter for something else.

Unless I'm misreading this it says "if the pop filter is on, turn it off, and update the URL". Is that what we want?

charts/Controls.tsx
```diff
@@ -599,7 +599,7 @@ export class Controls {
     @computed get hasInlineControls(): boolean {
         const { chart } = this.props
         return (
-            chart.tab === "chart" &&
+            (chart.tab === "chart" || chart.tab === "table") &&
```
breck7, line 602: What about the idea of having fewer conditionals and passing in controls?

charts/__tests__/DataTable.test.tsx
```diff
@@ -124,4 +124,19 @@ describe(DataTable, () => {
             expect(notice.prop("targetYear")).toBe("2016")
         })
     })
+
+    describe("when you try to hide countries", () => {
```
breck7, line 128: Nice!

Not sure if it would be easy, but for tests like these I like a before and after assertion.

Something like:

```typescript
expect(view.find("tbody tr")).toHaveLength(200)
// add filter
expect(view.find("tbody tr")).toHaveLength(187)
```

Also, will be nice someday when we can create simple charts on the fly like:

```typescript
const chart = new Chart({data: `country,pop,gdp
usa,123,123`, minPopulationFilter: 1e6})
...
```

charts/ChartConfig.tsx
shaahmad: Currently, the "hide countries" checkbox is only available on scatterplots on the Chart tab. This PR causes the checkbox to be visible across _all_ tables, not just Scatterplot tables. So a user might hide small countries on the Table, return to Chart, and no longer have a checkbox to uncheck and see small countries again.

I agree it can be improved. One way to is to remember the option state on Chart and restore it rather than just disable the filter. This way, if the option was already enabled, it will stay enabled on switching back.

Re. using the toggle method -- for some reason I had thought `minPopulationFilter` was marked private, but looking back I see it isn't, so I can just manually change that as well.

charts/ChartConfig.tsx
shaahmad: Yes, I'd also like to get some isolation here, but I'm not sure I understand what you mean by `component...` methods.

charts/Controls.tsx
shaahmad: I tried it, but realized the same conditionals would just move up into ChartView, which isn't necessarily a better place for them to be.

charts/__tests__/DataTable.test.tsx
shaahmad: Ah, so that test actually does already exist up at:
 https://github.com/owid/owid-grapher/blob/2e728d79f510019fd05c5338ddbf9984aae1759a/charts/__tests__/DataTable.test.tsx#L41-L43. I can copy-paste it to test the transition.

charts/ChartConfig.tsx
breck7: Gotcha, thanks. Maybe we should just show the hideCountries toggle on all charts then (keeping in place the condition that we check to see if there are countries that would be hidden first)? Whatever you think is best.

Ideally we can get a good transform architecture in soon and make our filters simpler.

charts/__tests__/DataTable.test.tsx
breck7: Oh nice, I didn't see that context.

charts/ChartConfig.tsx
shaahmad: Since non-scatters rarely show many entities and we're starting to run into places where we actually have too many controls in the footer, I'm actually in favor of hiding this control unless necessary.

charts/__tests__/DataTable.test.tsx
shaahmad: Copied it over to make it more explicit.

charts/ChartConfig.tsx
shaahmad: Pushing some temporary imperative code to remember the state. Would like to move to transform-specific filter columns ASAP, as I messaged you about.

PR #537, opened 2020-08-11
Dockerize

abitrolly:
I found these lying on my disk as I was experimenting with self-sufficient containers where the database and application are shipped together. I didn't find a way to run MySQL process in a background (yet), and maybe it will be easier to just use SQLite for now.

As the last resort. the `docker-compose` can help to run this for development locally. I didn't want to switch to `docker-compose`, because it requires a repository checkout to try and see `owid-grapher`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ ENV DB_NAME=owid
 
 # Installing Node.js 12.x and Yarn
 RUN set -x \
-      && apt-get update && apt-get install -y curl jq \
+      && apt-get update && apt-get install -y curl jq git make g++ \
       && NODE_VERSION=$(jq -r .engines.node /app/package.json) \
       && DEB_FILE="nodejs_${NODE_VERSION}-1nodesource1_amd64.deb" \
       && curl -sLO "https://deb.nodesource.com/node_12.x/pool/main/n/nodejs/${DEB_FILE}" \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM mysql:5.7
+
+ENV MYSQL_ALLOW_EMPTY_PASSWORD=yes
+ENV MYSQL_DATABASE=owid
+ENV DB_NAME=owid
+
+ADD https://files.ourworldindata.org/owid_metadata.sql.gz /docker-entrypoint-initdb.d/00-metadata.sql.gz
+ADD https://files.ourworldindata.org/owid_chartdata.sql.gz /docker-entrypoint-initdb.d/01-data.sql.gz
+# https://github.com/docker-library/mysql/issues/624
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,7 +19,29 @@ RUN ["sed", "-i", "s/exec \"$@\"/echo \"skipping $@\"/", "/entrypoint.sh"]
...
+      && NODE_VERSION=$(jq -r .engines.node /app/package.json) \
+      && DEB_FILE="nodejs_${NODE_VERSION}-1nodesource1_amd64.deb" \
+      && curl -sLO "https://deb.nodesource.com/node_12.x/pool/main/n/nodejs/${DEB_FILE}" \
+      && apt-get install -y ./"${DEB_FILE}" && rm "${DEB_FILE}" \
+      && curl -sL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+      && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+      && apt-get update && apt-get install -y yarn \
+      && rm -rf /var/lib/apt/lists/*
+
+#RUN cd app && yarn
```
masaez, line 46: Hi @abitrolly! I am using owid-grapher and planning on dockerizing - glad to see your PR.  Why is this line commented out?

abitrolly: @masaez the line is commented out, because before running the app, there should be database running. I haven't found an example of running database in the same container. Examples recommend using `docker-compose` to start db in another container, but I wanted `owid-grapher` to be self-sufficient.

masaez: I would also try to have the db in a separate containers (I am actually using the DB out of containers at all). Anyway, my comment was around the idea that running yarn would just resolve dependencies, not actually run the project. You would still need to run yarn dev or similar.

abitrolly: As this PR got stalled, feel free to pick it to fix what is needed. I sent you an invite, so you should be able to commit to the fork.

larsyencken:
Thanks for suggesting this way back when. We're likely to roll ahead with a `docker-compose` solution for developing just Grapher, although really we want to split that component out and isolate it more from the DB requirement. Closing this for now.

abitrolly:
Understood. `docker-compose` would still need Dockerfile.

abitrolly:
@larsyencken am I right that all DB calls are isolated in https://github.com/owid/owid-grapher/blob/master/adminSiteServer/apiRouter.ts so ? So that it is the only file to fix to use SQLite instead of MySQL?

larsyencken:
That file indicates the use of two databases (`db` for the charting admin, `wdb` for wordpress). You only need `db` working (e.g. ported to sqlite) if you want to get an editing environment for charts.

PR #540, opened 2020-08-11
CDE Fix: Transform per capita after computing rolling averages

shaahmad:
[Link to bug.](https://www.notion.so/owid/CDE-with-Per-Million-and-Unaligned-Outbreaks-doesn-t-show-any-data-or-axes-ee0dc032294441aab024cd44d2d6a469)

Live at [Nightingale](https://nightingale-owid.netlify.app/coronavirus-data-explorer?yScale=log&zoomToSelection=true&casesMetric=true&interval=smoothed&perCapita=true&smoothing=7&country=USA~GBR~CAN~BRA~AUS~IND~DEU~MEX~ZAF~COL~KOR~NOR~UGA~NGA&pickerMetric=location&pickerSort=asc).

**Before:**
image

**After:**
image

### DIFF
--- a/charts/covidDataExplorer/CovidExplorerTable.ts
+++ b/charts/covidDataExplorer/CovidExplorerTable.ts
@@ -338,22 +338,20 @@ export class CovidExplorerTable {
                 columnName === "positive_test_rate") &&
             smoothing === 7
 
-        // todo: have perCapita column derived from regular column.
-        if (perCapita > 1) {
-            const originalRowFn = rowFn
-            rowFn = row => {
-                const value = originalRowFn(row)
-                if (value === undefined) return undefined
-                const pop = row.population
-                if (!pop) {
-                    console.log(
-                        `Warning: Missing population for ${row.location}. Excluding from perCapita`
-                    )
-                    return undefined
-                }
-                return perCapita * (value / pop)
-            }
-        }
+        const perCapitaTransform =
+            perCapita > 1
+                ? (value: any, row: any) => {
+                      if (value === undefined) return undefined
+                      const pop = row.population
+                      if (!pop) {
+                          console.log(
+                              `Warning: Missing population for ${row.location}. Excluding from perCapita`
+                          )
+                          return undefined
+                      }
+                      return perCapita * (value / pop)
+                  }
+                : undefined
 
         if (smoothing && !alreadySmoothed)
             table.addRollingAverageColumn(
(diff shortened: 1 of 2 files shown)


breck7:
I'm not sure why computing rolling average after per capita or vice versa causes different floating point behavior. Is that just a coincidence? If you changed the numbers, could the opposite be true?

I think we want a minimum regression test here that generates that 5e-19 number or similar, and then maybe update the function that generates it maybe with some rounding and a disclaimer about precision.

shaahmad:
> I'm not sure why computing rolling average after per capita or vice versa causes different floating point behavior. Is that just a coincidence? If you changed the numbers, could the opposite be true?

The theory is that dividing a small case count by a large population results in extremely small numbers with large decimal expansions (possibly irrational numbers). Because of the limited size of floating point numbers, you lose many (if not infinite) digits. When you then compute a rolling average, pre division numbers like `4 cases` and `-4 cases` which would have easily summed to 0 average cases are now very tiny imprecise numbers like 8e-6 and -7e-5 which sum to something very close to, but not exactly, zero. 

Basically, when you do per capita first, you have to do math with irrational numbers. If you compute the rolling average first, you get to do math with integers, and just transform them into tiny irrational numbers at the end.

MarcelGerber:
@breck7 Here's a quick, somewhat arbitrary, example of how JS numbers lose precision when we divide:
image


Turns out we're doing a very similar thing when we first compute per-capita values and then average them.

To put it simply, this happens because division can be lossy in JS / the IEEE 754 64-bit floating point representation it uses.
Integer plus/minus arithmetic, however, is exact (between `MIN_SAFE_INTEGER` and `MAX_SAFE_INTEGER`, that is) and so if we get a zero value there, we won't have to deal with division imprecision since we already know the result will be zero.

breck7:
Thanks guys, very clear explanation of the issue. I think it's fine if we have a temporary fix of changing the order of operations, but probably shouldn't do that for functions that should be commutative. Though I'm not a fan of that rollingAverage method anyway, which has grown to do a lot, so maybe we can get rid of that whole thing at some point.

jasoncrawford:
@breck7, why don't you like this fix? Being smart about the order of operations makes sense to me, when it comes to floating-point arithmetic… ?

breck7:
@jasoncrawford Per capita and rolling average seem to be two of the most common operations researchers want to perform. As we move toward allowing them to generate their own columns and transform pipelines, I think we don’t want to have them require commutative ops in a certain order. (Or maybe we do? But that adds complexity). I think it may be better to just add a second param to compute rolling average of something like rounding=1e9 (not sure what is common, would have to search), and then add a couple unit tests. The other issue with just changing the order, is I think it’s realistic that one of us in two months time makes a change that changes the order again, and we don’t catch that regression.

shaahmad:
@breck7 added a comment for now, but I'm thinking in the long-term to add rounding or sigfigs to `initRollingAverageColumn` method or ideally `addComputedColumn`, and will write a test at that level.

PR #544, opened 2020-08-13
Add ckmeans & quantiles binning strategies for color scales

danielgavrilov:
Notion: https://www.notion.so/owid/Improve-the-automatic-binning-algorithm-for-legend-bins-d6090ee49df5461aa6fd4fddc44d34a6

- Migrates a bunch of the color scale props that weren't named consistently
- Migrates all old charts to the `equalInterval` binning strategy
- Implements `ckmeans` and `quantiles` binning strategies
- Sets `ckmeans` as the new default binning strategy
- Allows configuring how many automatically-generated bins should be created
- Allows using the automatically-generated bins as a starting point (in the past, the bin brackets were hidden if automatic classification was selected)

Screenshot 2020-08-13 at 14 23 34

### DIFF
--- a/admin/client/EditorColorScaleSection.tsx
+++ b/admin/client/EditorColorScaleSection.tsx
@@ -15,7 +15,7 @@ import { faMinus } from "@fortawesome/free-solid-svg-icons/faMinus"
 
 import { ColorScale } from "charts/ColorScale"
 import { ColorScaleBin, NumericBin, CategoricalBin } from "charts/ColorScaleBin"
-import { clone, noop } from "charts/Util"
+import { clone, noop, last } from "charts/Util"
 import { Color } from "charts/Color"
 import { asArray } from "utils/client/react-select"
 
@@ -408,7 +408,9 @@ class NumericBinView extends React.Component<{
         populateManualBinValuesIfAutomatic(scale)
 
         if (index === customNumericValues.length - 1)
-            customNumericValues.push(currentValue + scale.binStepSize)
+            customNumericValues.push(
+                last(scale.sortedNumericValues) ?? currentValue
+            )
         else {
             const newValue = (currentValue + customNumericValues[index + 1]) / 2
             customNumericValues.splice(index + 1, 0, newValue)
--- a/admin/client/EditorColorScaleSection.tsx
+++ b/admin/client/EditorColorScaleSection.tsx
@@ -181,18 +181,21 @@ class ColorsSection extends React.Component<{
             : scale.baseColorScheme
     }
 
-    render() {
-        const { scale } = this.props
-
-        const binningStrategyOptions = Object.entries(
-            binningStrategyLabels
-        ).map(([value, label]) => ({
+    @computed get binningStrategyOptions() {
+        return Object.entries(binningStrategyLabels).map(([value, label]) => ({
             label: label,
             value: value as BinningStrategy
         }))
-        const currentBinningStrategyOption = binningStrategyOptions.find(
-            option => option.value === scale.config.binningStrategy
+    }
+
+    @computed get currentBinningStrategyOption() {
+        return this.binningStrategyOptions.find(
+            option => option.value === this.props.scale.config.binningStrategy
         )
+    }
+
+    render() {
+        const { scale } = this.props
 
         return (
             <Section name="Color scale">
@@ -227,9 +230,9 @@ class ColorsSection extends React.Component<{
                     <div className="form-group">
                         <label>Binning strategy</label>
                         <Select
-                            options={binningStrategyOptions}
+                            options={this.binningStrategyOptions}
                             onChange={this.onBinningStrategy}
-                            value={currentBinningStrategyOption}
+                            value={this.currentBinningStrategyOption}
                             components={{
                                 IndicatorSeparator: null
                             }}
--- a/admin/client/EditorColorScaleSection.tsx
+++ b/admin/client/EditorColorScaleSection.tsx
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { action, IReactionDisposer, reaction, computed } from "mobx"
+import { action, IReactionDisposer, reaction, computed, runInAction } from "mobx"
 import { observer } from "mobx-react"
 import Select, { ValueType } from "react-select"
 
@@ -334,10 +334,12 @@ class BinLabelView extends React.Component<{
 }
 
 function populateManualBinValuesIfAutomatic(scale: ColorScale) {
-    if (scale.config.binningStrategy !== ColorScaleBinningStrategy.manual) {
-        scale.config.binningStrategy = ColorScaleBinningStrategy.manual
-        scale.config.customNumericValues = scale.autoBinMaximums
-    }
+    runInAction(() => {
+        if (scale.config.binningStrategy !== ColorScaleBinningStrategy.manual) {
+            scale.config.customNumericValues = scale.autoBinMaximums
+            scale.config.binningStrategy = ColorScaleBinningStrategy.manual
+        }
+    })
 }
 
 @observer
--- a/admin/client/EditorColorScaleSection.tsx
+++ b/admin/client/EditorColorScaleSection.tsx
@@ -199,14 +199,6 @@ class ColorsSection extends React.Component<{
                     label="Minimum value"
                     auto={scale.autoMinBinValue}
                 />
-                {!scale.config.isManualBuckets && (
-                    <BindAutoFloat
-                        label="Step size"
-                        field="binStepSize"
-                        store={scale.config}
-                        auto={scale.binStepSizeDefault}
-                    />
-                )}
                 {scale.config.isManualBuckets && (
                     <ColorSchemeEditor scale={scale} />
                 )}
@@ -350,7 +342,7 @@ class NumericBinView extends React.Component<{
         const currentValue = colorSchemeValues[index]
 
         if (index === colorSchemeValues.length - 1)
-            colorSchemeValues.push(currentValue + scale.binStepSizeDefault)
+            colorSchemeValues.push(currentValue + scale.binStepSize)
         else {
             const newValue = (currentValue + colorSchemeValues[index + 1]) / 2
             colorSchemeValues.splice(index + 1, 0, newValue)
--- a/admin/client/EditorColorScaleSection.tsx
+++ b/admin/client/EditorColorScaleSection.tsx
@@ -1,5 +1,11 @@
 import * as React from "react"
-import { action, IReactionDisposer, reaction, computed, runInAction } from "mobx"
+import {
+    action,
+    IReactionDisposer,
+    reaction,
+    computed,
+    runInAction
+} from "mobx"
 import { observer } from "mobx-react"
 import Select, { ValueType } from "react-select"
 
@@ -232,12 +238,22 @@ class ColorsSection extends React.Component<{
                         />
                     </div>
                 </FieldsRow>
-                <BindAutoFloat
-                    field="customNumericMinValue"
-                    store={scale.config}
-                    label="Minimum value"
-                    auto={scale.autoMinBinValue}
-                />
+                <FieldsRow>
+                    <BindAutoFloat
+                        field="customNumericMinValue"
+                        store={scale.config}
+                        label="Minimum value"
+                        auto={scale.autoMinBinValue}
+                    />
+                    {!scale.isManualBuckets && (
+                        <BindAutoFloat
+                            field="binningStrategyBinCount"
+                            store={scale.config}
+                            label="Number of bins"
+                            auto={scale.numAutoBins}
+                        />
+                    )}
+                </FieldsRow>
                 <ColorSchemeEditor scale={scale} />
             </Section>
         )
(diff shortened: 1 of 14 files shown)


charts/ColorScaleConfig.ts
```diff
@@ -33,8 +75,10 @@ export class ColorScaleConfigProps {
         [key: string]: true | undefined
     } = {}
 
-    @observable.ref legendDescription?: string = undefined
-    @observable.ref binStepSize?: number = undefined
+    // Other
+    // =====
+
+    @observable legendDescription?: string = undefined
```
MarcelGerber, line 81: I actually don't know what this does, may be good to also add a comment there?

charts/ColorScale.ts
```diff
@@ -130,32 +139,28 @@ export class ColorScale {
...
 
     @computed get singleColorScale(): boolean {
         return this.colorScheme.singleColorScale
     }
 
     @computed get autoMinBinValue(): number {
-        const minValue = Math.min(0, this.valuesWithoutOutliers[0])
-        const magnitude = Math.floor(Math.log(minValue) / Math.log(10))
-        return Math.min(0, round(minValue, -magnitude))
+        const minValue = Math.min(0, this.sortedNumericValuesWithoutOutliers[0])
```
MarcelGerber, line 150: if `sortedNumericValuesWithoutOutliers` is an empty array, `Math.min(0, undefined)` will return `NaN` which is nasty, so better handle that case if it ever comes up

charts/ColorScale.ts
```diff
@@ -130,32 +139,28 @@ export class ColorScale {
...
+        return Math.min(0, roundSigFig(minValue, 1))
```
MarcelGerber, line 151: nit: We probably don't need `Math.min(0, ...)` twice?

charts/ColorScale.ts
```diff
@@ -166,20 +171,43 @@ export class ColorScale {
...
+        } else {
+            // Equal-interval strategy by default
+            const { binStepSize, numAutoBins, minBinValue } = this
+            const bucketMaximums = []
+            let nextMaximum = minBinValue + binStepSize
+            for (let i = 0; i < numAutoBins; i++) {
+                bucketMaximums.push(nextMaximum)
+                nextMaximum += binStepSize
+            }
+            return bucketMaximums
```
MarcelGerber, line 205: I know this is old code, but what about rewriting this like so:

```suggestion
            return range(1, numAutoBins + 1).map(v => 
            	minBinValue + v * binStepSize
            )
```

charts/ColorScale.ts
```diff
@@ -212,38 +240,36 @@ export class ColorScale {
...
+            numAutoBins,
+            minBinValue,
+            sortedNumericValuesWithoutOutliers
+        } = this
+        if (!sortedNumericValuesWithoutOutliers.length) return 10
 
         const stepSizeInitial =
-            (valuesWithoutOutliers[valuesWithoutOutliers.length - 1] -
-                minBinValue) /
+            (last(sortedNumericValuesWithoutOutliers)! - minBinValue) /
```
MarcelGerber, line 266: I guess we also need some handling for empty `sortedNumericValuesWithoutOutliers` here?

admin/client/EditorColorScaleSection.tsx
```diff
@@ -136,29 +152,48 @@ class ColorsSection extends React.Component<{
...
-        return scale.isCustomColors ? "custom" : scale.baseColorScheme
+        return scale.customNumericColorsActive
+            ? "custom"
+            : scale.baseColorScheme
     }
 
     render() {
         const { scale } = this.props
 
+        const binningStrategyOptions = entries(
```
MarcelGerber, line 187: nit: probably better to go with `Object.entries` here?

MarcelGerber:
I hit a weird bug here that I'm able to consistently repro:

image

steps:
1. go to chart http://localhost:3030/admin/charts/1403/edit
2. first, change the binning strategy to "quantiles"
3. then, change the binning strategy to "ckmeans"

result: the two legends / labels will be overlaid over each other

MarcelGerber:
This is not an issue per se, just wondering if we can improve the situation.

For a chart with many 0 values, the "quantiles" strategy creates many labels for the value 0, which is not necessarily helpful.
Maybe we can do better by filtering out 0 values?

Example (http://localhost:3030/admin/charts/4309/edit):
image

charts/ColorScale.ts
danielgavrilov: Much more elegant!

charts/MapTooltip.tsx
```diff
@@ -89,7 +89,8 @@ export class MapTooltip extends React.Component<MapTooltipProps> {
 
     @computed get barColor() {
         const { colorScale } = this.chart.map.data
-        return colorScale.singleColorScale && !colorScale.isCustomColors
+        return colorScale.singleColorScale &&
+            !colorScale.customNumericColorsActive
```
shaahmad, line 93: It'd be nice to keep a "isCustomColors" function, so that the implementation detail of numeric vs categorical doesn't have to be considered for the developer just wondering if the color scales are automatic or manually overridden.

charts/ColorScaleConfig.ts
```diff
@@ -2,28 +2,70 @@ import { observable } from "mobx"
...
-    @observable.ref customColorsActive?: true = undefined
+    /** The strategy for generating the bin boundaries */
+    @observable binningStrategy: ColorScaleBinningStrategy =
+        ColorScaleBinningStrategy.ckmeans
+    /** The *suggested* number of bins for the automatic binning algorithm */
+    @observable binningStrategyBinCount?: number
 
-    // e.g. ["#000", "#c00", "#0c0", "#00c", "#c0c"]
+    // Iff the binningStrategy is `manual`, then overrides are specified below.
+    // Otherwise, the overrides are ignored.
```
shaahmad, line 25: Should overrides be grouped together into a `manualOverrides` object?

admin/client/EditorColorScaleSection.tsx
shaahmad: Does it make sense to generate this object on every render? Maybe better as a `computed`?

admin/client/EditorColorScaleSection.tsx
```diff
@@ -304,6 +349,15 @@ class BinLabelView extends React.Component<{
     }
 }
 
+function populateManualBinValuesIfAutomatic(scale: ColorScale) {
```
shaahmad, line 352: nit: mark the entire function as `@action` and ditch the `runInAction`?

charts/MapTooltip.tsx
danielgavrilov: Yeah I was looking at this – I only renamed `isCustomColors`, the implementation is the same. For maps, in 99% of cases we use a numerical legend ([this](https://ourworldindata.org/grapher/continents-according-to-our-world-in-data) is what a categorical legend looks like).

But, there is one categorical bin always, even on a numerical legend – the "No data" bin. I thought we'd like to exclude that from the logic here anyway, it makes sense to only consider numerical bins when determining the bar color.

Let me know if you disagree!

danielgavrilov: If we want to have a property that considers both we can add this:

```ts
    @computed get hasCustomColors(): boolean {
        return (
            (this.customNumericColorsActive &&
                this.customNumericColors.some(color => color !== undefined)) ||
            Object.values(this.customCategoryColors).some(
                color => color !== undefined
            )
        )
    }
```

But I don't think we should add it unless we use it in some case.

charts/Util.ts
```diff
@@ -255,6 +255,11 @@ export function precisionRound(num: number, precision: number) {
     return Math.round(num * factor) / factor
 }
 
+export function roundSigFig(num: number, sigfigs: number = 1) {
+    const magnitude = Math.floor(Math.log10(Math.abs(num)))
```
MarcelGerber, line 259: Actually just noticed that `Math.log10(0)` is `-Infinity`, so we might also want a case for good ol' zero.

admin/client/EditorColorScaleSection.tsx
danielgavrilov: I don't think decorators can be used on functions in TS?

charts/Util.ts
danielgavrilov: Good catch!

admin/client/EditorColorScaleSection.tsx
```diff
@@ -107,21 +119,25 @@ class ColorsSection extends React.Component<{
         this.disposers.push(
             // When the user disables automatic classification,
             // populate with automatic buckets.
+
+            // TODO extract this into a function with a `wipe` param.
+            // Here the `wipe` param is false.
+            // When the user edits a section while in auto mode, `wipe` is true.
             reaction(
-                () => scale.config.isManualBuckets,
+                () => scale.isManualBuckets,
```
shaahmad, line 127: So I'm not 100% sure about this and this is mostly pre-PR code, but I _think_ reactions modifying observables might be anti-pattern and such reactions should be replaced with `computeds`. The rough reasoning is mobx can build a proper graph with `computed` and do proper state management, whereas `reactions` are asynchronous, untracked, etc. So when you use these kinds of reactions, you can build undetectable cycles in the dependency graph, etc. This is something I've encountered with the Timeline.

charts/ColorScaleConfig.ts
danielgavrilov: I've edited the comments now, only 2 of the 3 properties are actually affected – values and labels. Minimum value is still used even if the strategy is non-manual.

But noticed an existing bug looking into this – the custom numeric labels were applied even when automatic binning is used.

Re nesting objects: we usually avoid it in the rest of the config, and I don't feel it's necessary here. Basically anything beginning with `custom` is a manual override, which are most properties in this config object.

admin/client/EditorColorScaleSection.tsx
danielgavrilov: Thanks! fixed now

admin/client/EditorColorScaleSection.tsx
danielgavrilov: Seems like they can't:

> A _Decorator_ is a special kind of declaration that can be attached to a [class declaration](https://www.typescriptlang.org/docs/handbook/decorators.html#class-decorators), [method](https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators), [accessor](https://www.typescriptlang.org/docs/handbook/decorators.html#accessor-decorators), [property](https://www.typescriptlang.org/docs/handbook/decorators.html#property-decorators), or [parameter](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators).

https://www.typescriptlang.org/docs/handbook/decorators.html#decorators

charts/MapTooltip.tsx
shaahmad: I think this is actually the use case for that function. `barColor` is basically asking, "Is this a non-divergent/multi-hue color scale?", which cannot be answered with certainty for any color scheme with a manual override. (I manually vetted all the ColorBrewer, etc. color schemes 😄   to specify `singleColorScheme`)

admin/client/EditorColorScaleSection.tsx
shaahmad: Ahh, right. Sad.

admin/client/EditorColorScaleSection.tsx
danielgavrilov: Yeah I totally agree! And I feel you on the `Timeline` 😄 

I found a way to [get rid of the reaction](https://github.com/owid/owid-grapher/pull/544/commits/7423aa480bc6554dcf6845c507de6af4a3068374) by hiding the "Manual" option.

danielgavrilov:
Thanks _a lot_ for reviewing & comments @MarcelGerber & @shaahmad!

I think everything is now fixed except the issue that `quantiles` generates some unnecessary bins. I've just deduplicated the bins so that the zeroes are not repeated, but I don't think it's worth trying to improve this unless there are some very easy wins. (One approach may be to remove values that are repeated more than `length / binCount` times?...)

But anyway, the authors will very likely continue to define manual brackets in almost every case – these strategies are just intended to help them have better starting points.

admin/client/EditorColorScaleSection.tsx
shaahmad: 👌 👌 .

Maybe we can do this with the "Color scheme" and "Custom" too, eventually!

charts/MapTooltip.tsx
danielgavrilov: But we can't create a bar plot with categorical values (categorical means they are type `string`), so it doesn't matter if those have custom colors? And as I mentioned, _No data_ is also a categorical bin that always exists, but we want to ignore it even if it has a custom color? So any categorical bins don't matter in this case?

shaahmad: Ah, ok. Ideally I'd be able to do something like 
```
renderBars(): 
    return colorScale.isNumeric

barColor():
    return colorScale.hasCustomColors ? NEUTRAL_COLOR : colorScale.darkestColor
```
All of which would ignore the "No Data" bin.

But this is just a preference so happy to leave this as-is for now!

danielgavrilov: :+1: The implementation here is unchanged apart from the renamed prop, we can make the change outside this PR.

PR #545, opened 2020-08-17
Custom Data Explorers

breck7:
This generalizes the Covid Data Explorer shell, Explorer Controls and Country Picker and adds a new type of general purpose Explorer with author editing tools. The new explorers simply do "chart switching" for now but allow the authors to fully control the options (and each chart view). In the future we could also allow transformations as the action taken when a user switches controls and then migrate the Covid explorer to this system allowing behaviors other than chart switching (and delete a decent amount of code).

I hate massive PRs like this but for this big feature not sure if there was a better way.

Play with it here: https://jefferson.owid.cloud/
https://jefferson.owid.cloud/admin/explorers/preview/co2-data-explorer
https://jefferson.owid.cloud/admin/explorers

**Screenshots of the new features:**

User facing explorers:
Screen Shot 2020-08-16 at 2 01 25 PM

Author editing experience:
Screen Shot 2020-08-16 at 2 01 03 PM

Author management tool:
Screen Shot 2020-08-16 at 2 00 31 PM

**Technical changes of note**

Most of this work was generalization of the current explorer components but there are five higher level things that I would love people's thoughts on.

1. `gitCms/` - The Explorer admin writes files to a Git repo instead of MySQL. I thought this might be a good chance to test this paradigm to see if it's something we might want to do more of in the future. It's probably a dumb idea so I also made it easy to drop this approach and stick to MySQL.
2. Column types. In OwidTable, I start to expand on the idea of Column Types. I think this is potentially a good way to build up our ontology as we work toward the new data API stuff. At the very least should allow authors to get a lot of default formatting by selecting one type.
3. ExplorerDSL. The `ExplorerProgram.ts` class implements a DSL for authors to create explorers. I took Hannah's spreadsheet of the CO2 Explorer and just turned that into a DSL. It's a hybrid of JSON + TSV. You can have key/values, but also drop in TSV content without escaping things. Easy to cut and paste between spreadsheets. If we like it we can make it more robust, name it, and use it in more places. If we don't like it we can just switch to JSON.
4. Handsontable - I added this just in the backend, but I think it's something we might want to explore using in the charts as well. For example, if we want to enable authors to add an annotation to a specific row in a dataset, they could look at the raw data in HOT and simply add their annotation in a cell.
5. Subproject folders. Similar to `gitCms/`, all the explorer code is in `explorer/`. I think there are a few other places where we should move code to project folders like that for modularity, and keep linkages to a minimum. After this I'll probably try and put all the OwidTable, Covid Code, etc, in one place. Finish what we talked about in this PR (https://github.com/owid/owid-grapher/pull/328).


mlbrgl:
Adding an incorrect chart ID resulted in an uncaught exception and wasn't immediately obvious to the end-user in the table view. The culprit cell is now colored in red.

Screenshot 2020-08-19 at 11 26 56

(implementation is a bit convoluted. For instance, the error message is more of a hack to get the component to re-render after validation. Maybe someone with more mobx experience can find a more elegant way to do it)

@breck7 is versioning of explorer configs the main reason for experimenting with this git architecture?

breck7:
> Adding an incorrect chart ID resulted in an uncaught exception and wasn't immediately obvious to the end-user in the table view. The culprit cell is now colored in red.

Nice! 

> (implementation is a bit convoluted. For instance, the error message is more of a hack to get the component to re-render after validation. Maybe someone with more mobx experience can find a more elegant way to do it)

I like the idea of error highlighting throughout. Will look at it.

> @breck7 is versioning of explorer configs the main reason for experimenting with this git architecture?

Yes, I think great document versioning is the main benefit of git vs SQL. For "document" items like these, and things like the current CovidConstants, where we don't gain too many benefits from a relational structure, and would have between 1-100 rows, we could either setup a catch all `documents` table with timestamps and versions and all our own versioning logic, or just write to disk and put it in Git/GitHub.

I could see the Explorer DSL evolving a bit too, so doing it in SQL I'd expect would lead to many migrations. With this structure you can do a simple find/replace and single commit for a migration.

Syncing this folder during dev would also just be a fast git pull.

Finally, I could see by making our content more accessible could lead to a number of unpredicted benefits.

breck7:
Closing this as I squash merged on the other PR. Will readd validation from @mlbrgl

mlbrgl:
@breck7 thanks for expanding on this, very interesting (and useful) to know the rationale behind those architecture decisions!

PR #554, opened 2020-08-20
Move test files next to the class being tested

breck7:
This commit would drop the `__tests__` folders and just put the test files alongside the class being tested.

So instead of this:

```
charts/
 ChartUrl.ts
 __tests__/
  ChartUrl.test.ts
```

We'd have this:

```
charts/
 ChartUrl.ts
 ChartUrl.test.ts
```

**What's wrong with `__test__`?**

Nothing wrong with it and that looks to be the standard jest way.

**What's good about \*.test.ts?**

I don't think it makes that much of an impact, but I've personally found it to be slightly better because it makes **tests more of a first class citizen**. It's easy to see at a glance which files are untested. It's slightly easier to open the test file and add a new test. If someone wants to see how to use a class the test file is right there. It's also the pattern used by goog (like in closure (https://github.com/google/closure-library/tree/master/closure/goog/proto2) or tensorflowjs (https://github.com/tensorflow/tfjs/tree/master/tfjs-core/src)).

Anyway, it's sort of bike shedding so if someone has a strong objection no problem. Otherwise I can merge in.


explorer/client/covidDataExplorer/CovidParams.ts
```diff
@@ -1,12 +1,12 @@
...
-import { EntityUrlBuilder } from "../ChartUrl"
+import { EntityUrlBuilder } from "../../../charts/ChartUrl"
 import {
     QueryParams,
     strToQueryParams,
     queryParamsToStr
 } from "utils/client/url"
 import { SortOrder } from "charts/SortOrder"
-import { oneOf, uniq, intersection } from "../Util"
+import { oneOf, uniq, intersection } from "../../../charts/Util"
```
danielgavrilov, line 9: Might be better to use`"charts/..."` rather than `"../../../charts/.." ?

danielgavrilov:
I think this is much much better! Big :+1: from me!

danielgavrilov (approved):
I think this is much much better! Big :+1: from me!

explorer/client/covidDataExplorer/CovidParams.ts
breck7: Yes. Sometimes when VS Code auto renames imports after a file move it does funky things :)

breck7: There were a few of these. Thanks

PR #558, opened 2020-08-21
Dev: import lodash as lodash not _

breck7:
A very small nit, but GPT-4 can train better with more informative identifier names.


shaahmad:
😄 I must veto this on behalf on Nick Bostrom.

MarcelGerber (approved):
👍 i agree, this makes it easier to search for lodash usages

PR #566, opened 2020-08-23
Dev: draft reorg of site folder

breck7:
This moves most of the `site/server/` scripts to the folder `baker/`. Then `site/server/views` are merged with `site/client` into the flattened `site/` folder. I couldn't see the distinction between a "site/server/view" and a "site/client" component. The former is mostly pages but not entirely pages, so I thought it made sense to lump them together.

This is a broken build but wanted to see if this was alright before going through and updating all the links and references.


breck7:
Closing this for now. Can do this one later.

PR #578, opened 2020-08-25
Dev: make node default test environment

breck7:
Tests run in the JSDOM environment seem to be about 2x-3x slower. JSDOM has been slow for years, in my experience. So the more of our logic we can test in pure node, the better.

This change makes `.test.ts` files run in node and if you need JSDOM add the `.jsdom.test.ts` suffix. It also moves over some tests from JSDOM to node that didn't require changes.

I've also added a shortcut to package.json for running Jest in watch mode.

[1] (https://medium.com/@kevinsimper/how-to-disable-jsdom-in-jest-make-jest-run-twice-as-fast-a01193f23405)


MarcelGerber (approved):
Great!

PR #579, opened 2020-08-26
Dev: refactor LineChart so it is testable

breck7:
In order to add faceting, I think it's best to clean up all the chart code so the components of each chart can be tested independently. Then when we show many charts in one screen it will be a lot easier and more reliable.

This is a step in that direction. It ain't pretty yet, but it's slightly prettier than our current code.

The main things to be done are:
- remove globals like Context
- create interfaces with only the needed properties/methods instead of requiring chartConfig
- create stories and/or unit tests for each component so that we can sanity check the APIs

Before:
[No unit tests or stories]

After:
Screen Shot 2020-08-25 at 8 52 50 PM

### DIFF
--- a/charts/areaCharts/StackedArea.tsx
+++ b/charts/areaCharts/StackedArea.tsx
@@ -24,10 +24,6 @@ import { Tooltip } from "charts/core/Tooltip"
 import { select } from "d3-selection"
 import { easeLinear } from "d3-ease"
 import { rgb } from "d3-color"
-import {
-    ChartViewContext,
-    ChartViewContextType
-} from "charts/core/ChartViewContext"
 import { EntityDimensionKey } from "charts/core/ChartConstants" // todo: remove
 
 export interface StackedAreaValue {
@@ -226,9 +222,6 @@ export class StackedArea extends React.Component<{
 }> {
     base: React.RefObject<SVGGElement> = React.createRef()
 
-    static contextType = ChartViewContext
-    context!: ChartViewContextType
-
     @computed get chart(): ChartConfig {
         return this.props.chart
     }
@@ -302,7 +295,7 @@ export class StackedArea extends React.Component<{
     @observable hoverKey?: string
     @action.bound onLegendClick(key: EntityDimensionKey) {
         if (this.chart.showAddEntityControls) {
-            this.context.chartView.isSelectingData = true
+            this.chart.isSelectingData = true
         }
     }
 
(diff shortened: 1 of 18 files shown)


charts/areaCharts/StackedArea.tsx
```diff
@@ -351,6 +344,7 @@ export class StackedArea extends React.Component<{
 
         return (
             <Tooltip
+                tooltipOwner={this.props.chart}
```
breck7, line 347: `tooltipOwner` is just a temporary step to something cleaner. Passing this param is a slight improvement over the current global (and allows the story to work) but we can clean that up later.

charts/controls/Controls.tsx
```diff
@@ -750,11 +750,14 @@ export class ControlsOverlay extends React.Component<{
     context!: ChartViewContextType
 
     componentDidMount() {
-        this.context.chartView.overlays[this.props.id] = this
+        // todo: remove context
```
breck7, line 753: I'd like to remove context here, and will do in a future commit.

charts/core/Tooltip.tsx
```diff
@@ -92,21 +79,34 @@ export class TooltipView extends React.Component {
...
-    static contextType = ChartViewContext
-    context!: ChartViewContextType
-
     componentDidMount() {
         this.componentDidUpdate()
     }
 
     componentDidUpdate() {
-        runInAction(() => (this.context.chart.tooltip = this.props))
+        runInAction(() => (this.props.tooltipOwner.tooltip = this.props))
```
breck7, line 105: We can probably delete this whole class and just have the TooltipView one

charts/lineCharts/LineChart.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { parseDelimited } from "charts/utils/Util"
+import { OwidTable } from "owidTable/OwidTable"
+
+export default {
+    title: "LineChart",
+    component: LineChart
+}
+
+export const Default = () => {
+    const props = {
```
breck7, line 15: Ideally in a week or two the API for this will be a lot shorter and simpler.

charts/lineCharts/LineChart.tsx
```diff
@@ -225,30 +216,28 @@ class Lines extends React.Component<LinesProps> {
...
         const base = this.base.current as SVGGElement
-        this.container = base.closest("svg") as SVGElement
-
-        this.container.addEventListener("mousemove", this.onCursorMove)
-        this.container.addEventListener("mouseleave", this.onCursorLeave)
-        this.container.addEventListener("touchstart", this.onCursorMove)
-        this.container.addEventListener("touchmove", this.onCursorMove)
-        this.container.addEventListener("touchend", this.onCursorLeave)
-        this.container.addEventListener("touchcancel", this.onCursorLeave)
+        const container = base.closest("svg") as SVGElement
```
breck7, line 222: We've got another global here and should probably clean that up in a later commit.

PR #589, opened 2020-09-01
Fix: relative mode toggle in StackedAreas and add tests

breck7:


charts/areaCharts/StackedAreaTransform.ts
```diff
@@ -163,6 +165,42 @@ export class StackedAreaTransform extends ChartTransform {
...
+        const yValues = this.allStackedValues.map(d => d.y)
+        return [0, max(yValues) ?? 100]
+    }
+
+    @computed get yAxis() {
+        const { isRelativeMode, yDimensionFirst } = this
+        const { chart, yDomainDefault } = this
+
+        const axis = chart.yAxisOptions.toVerticalAxis()
+        if (isRelativeMode) axis.domain = [0, 100]
```
breck7, line 189: This is the actual fix. If is relative mode we always want to set the domain to [0, 100], even if there were other hard codes mins and maxes.

charts/areaCharts/StackedAreaChart.tsx
```diff
@@ -280,48 +277,6 @@ export class StackedAreaChart extends React.Component<{
         })
     }
 
-    // todo: Refactor
```
breck7, line 283: I moved this stuff to Transform for testability

PR #602, opened 2020-09-04
Fix ContainerOptions getting saved with Chart

breck7:
I have a bigger cleanup coming of ChartConfig saving, but this fixes the urgent `containerOptions` bug.

I will add a comment to the actual test and fix.

This commit is a little bigger than that, because the current `charts/core/ChartConfig.jsdom.test.ts` file was actually tests for the Indicator Explorer, so I put that code in the correct place.


charts/core/ChartConfig.jsdom.test.ts
```diff
@@ -0,0 +1,17 @@
...
+describe("ChartConfig", () => {
+    it("can serialize chartConfig for saving", () => {
+        const config = new ChartConfig(new ChartScript())
+        const json = config.json
+        expect(Object.keys(json).length).toBe(57) // It should be 0. Setting it to 57 for just 1 commit.
+    })
+
+    it("container options are not serialized", () => {
+        const config = new ChartConfig(new ChartScript())
+        expect(config.json.xAxis.containerOptions).toBe(undefined)
```
breck7, line 15: The test for the fix.

charts/core/ChartConfig.jsdom.test.ts
```diff
@@ -0,0 +1,17 @@
+#! /usr/bin/env yarn jest
+
+import { ChartConfig } from "./ChartConfig"
+import { ChartScript } from "./ChartScript"
+
+describe("ChartConfig", () => {
+    it("can serialize chartConfig for saving", () => {
+        const config = new ChartConfig(new ChartScript())
+        const json = config.json
+        expect(Object.keys(json).length).toBe(57) // It should be 0. Setting it to 57 for just 1 commit.
```
breck7, line 10: This is something to cleanup in a future commit. When a brand new empty chart is saved it currently it has 57 keys!

charts/core/ChartConfig.tsx
```diff
@@ -883,6 +883,12 @@ export class ChartConfig {
             json.isAutoSlug = true
         }
 
+        if (json.xAxis && json.xAxis.containerOptions)
```
breck7, line 886: This is the fix

PR #604, opened 2020-09-05
Grapher rename

breck7:
Following up on the conversation earlier:

ChartConfig => Grapher
ChartScript (Previously ChartConfigProps) aka what the authors edit => GrapherScript

Should clarify things going forward.


shaahmad:
I'm a bit curious behind the use of the word "script" here. Doesn't a script refer to a sequence of instructions, while this is more a "persisted state"?

breck7:
> I'm a bit curious behind the use of the word "script" here. Doesn't a script refer to a sequence of instructions, while this is more a "persisted state"?

I was trying to distinguish between the runtime and the persistent state. In the PR I'm about to submit, I add an interface of `Persistable`. Grapher just implements that, so we don't need a notion of GrapherScript.

That being said, I think the direction we are heading is more toward a Scripting level => fetch this data, group it by X, filter Y, facet it on Z, and show line charts, and a map tab, and a table tab with X, etc.

I don't think that day is today, but at some point I think we move more toward a more composable DSL.

shaahmad:
Great idea!

PR #609, opened 2020-09-07
Deploy status page in Admin

danielgavrilov:
As described on Notion issue: https://www.notion.so/owid/Create-deploy-status-page-in-Grapher-Admin-2e38fc965531418db65234297056c3d4

http://localhost:3030/admin/deploys

Screenshot 2020-09-07 at 17 51 25

It helps have an easily accessible page for authors (and developers) to check whether their changes have been deployed. It's awkward to bug a colleague about this – authors would usually wait quite some time before reaching out to us.

This is just a first step – exposing the information that's already in the `.queue` and `.pending` files. Ideally, it would be **visible on the chart page itself** what the deploy status of the chart is – whether it's already published, pending or waiting.

We could also add more deploy info on this page, e.g. what commit is the admin server running currently, pull and parse commits from GitHub, etc.

### DIFF
--- a/adminSite/client/DeployStatusPage.tsx
+++ b/adminSite/client/DeployStatusPage.tsx
@@ -1,6 +1,7 @@
 import * as React from "react"
 import { observer } from "mobx-react"
 import { observable, runInAction } from "mobx"
+import { format } from "timeago.js"
 
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import { faCheckCircle } from "@fortawesome/free-solid-svg-icons/faCheckCircle"
@@ -28,6 +29,7 @@ export class DeployStatusPage extends React.Component {
                                     <th>Status</th>
                                     <th>Note</th>
                                     <th>Author</th>
+                                    <th>Time</th>
                                 </tr>
                             </thead>
                             <tbody>
@@ -44,6 +46,15 @@ export class DeployStatusPage extends React.Component {
                                             </td>
                                             <td className="cell-author">
                                                 {change.authorName}
+                                            </td>
+                                            <td className="cell-time">
+                                                {change.timeISOString
+                                                    ? format(
+                                                          Date.parse(
+                                                              change.timeISOString
+                                                          )
+                                                      )
+                                                    : ""}
                                             </td>
                                         </tr>
                                     ))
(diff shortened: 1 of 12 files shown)


adminSite/client/DeployStatusPage.tsx
```diff
@@ -0,0 +1,94 @@
...
+                                        <tr key={`${deploy.status}-${i}`}>
+                                            <td
+                                                className={`cell-status cell-status--${deploy.status}`}
+                                            >
+                                                {deploy.status}
+                                            </td>
+                                            <td className="cell-message">
+                                                {change.message}
+                                            </td>
+                                            <td className="cell-author">
```
breck7, line 45: Would one of those "5 minutes agos" columns be worthwhile to add?

deploy/queue.ts
```diff
@@ -4,6 +4,7 @@ import {
     DEPLOY_PENDING_FILE_PATH
 } from "serverSettings"
 import { deploy } from "./deploy"
```
breck7, line 6: It may be worth it to add a few simple unit tests to the queue as I could see more features being added in the future.

breck7:
Is each deploy independent? Ie, if one deploy is currently pending, and then 3 authors make 3 changes, will those 3 changes be merged into the next deploy? Or will they proceeed serially?

breck7 (approved):
Nice! This will be super helpful.

danielgavrilov:
> Is each deploy independent? Ie, if one deploy is currently pending, and then 3 authors make 3 changes, will those 3 changes be merged into the next deploy? Or will they proceeed serially?

They get merged. We can't freeze the database (I think?), so these queue items are not necessarily in the deploy they are recorded in, e.g. deploy starts, authors make changes to some charts, then deploy reaches stage where it bakes charts – the changes would be included in the pending deploy, but recorded as if they were made in the subsequent deploy.

deploy/queue.ts
danielgavrilov: Yeah good idea, will do!

adminSite/client/DeployStatusPage.tsx
danielgavrilov: Yeah – we don't record a time in the messages but it would be easy to add

PR #616, opened 2020-09-08
Column slugs

breck7:
This:

- Starts adding tests for ChartDimension and flattens the Config into it. Makes it a Persistable
- Starts moving MapConfig to accepting a columnSlug, preserving variableId for legacy for now
- Turns the VariableId into the columnSlug in OwidTable, so that we can easily migrate from variableIds to column slugs.
- Fixes a NoData scss regression and cleans up the "GrapherView" story
- Moves the CSV generating logic to closer to the Table level. Should be easier to test and match the data 1:1


grapher/chart/ChartDimension.ts
```diff
@@ -82,32 +72,49 @@ export class ChartDimensionSpec implements ChartDimensionInterface {
...
+            display: Object.keys(this.display).length
+                ? this.display
+                : undefined,
+        })
     }
 }
 
-export class ChartDimension {
-    spec: ChartDimensionSpec
+export class ChartDimension extends PersistableChartDimension {
```
danielgavrilov, line 99: This means we have to be careful about property name clashes?

danielgavrilov: What are your thoughts on whether these are specified directly on the first level of the object, rather than having them nested under a `spec` or `props` property?

Some points where this matters:
- Sometimes we define the same prop on the "host" object that exists in the persisted. Mostly we do this for validation or setting defaults, e.g. if title is undefined set it to empty string.
- When we redefine a prop like that, we would have to create a setter that modifies the prop on the `super`class.
- When coding, it's easy to know when you are editing persisted props if they are under some conventional property name.

breck7: I played with a few patterns. What I think works best is something like this:

```
export SomeClass implements Persistable {
    @observable title = undefined
    displayTitle() {
      return this.title ?? "Some default title"
    }
   ...
}
```

If you want to set the title, just set it normally. If you want to read the title, read it normally. If you want to read the title or a default, call `displayTitle` (or whatever the method might be).

I don't like the ` ChartDimension extends PersistableChartDimension` stuff and can flatten then, just taking small steps.

breck7: My general thought is that we should not have "props" objects, instead setting the props directly on the object itself. This cuts the # of classes in half, and reduces potential for indirection bugs. Then we have validated getters for external classes that need a valid version of a prop. (It also helps clarify when you are accessing the actual prop, vs the validated version).

I do agree that it's nice to see at a glance whether a prop is being persisted, and the `props` member provided some hints of that. But it wasn't a guarantee, and sometimes props were not persisted. I think the Interface provides some help here, but maybe the real long term solution is to use a decorator strategy (which would probably mean switching to serializr).

PR #623, opened 2020-09-09
Dev: prettify grapher pages while preserving baking version logic

breck7:
With the new Deploy Status page, the benefits from having clean diffs in `owid-static` are even clearer. This hopefully handles the Grapher config part of the page. It's probably also worth it to prettify the HTML. Then it will be really easy to see exactly what changed in each deploy.

This sticks the JSON config on the Grapher page between delimiters so it can be easily parsed without parsing the whole HTML and JS. Adds tests.

Before:
Screen Shot 2020-09-09 at 10 23 45 AM

After:
Screen Shot 2020-09-09 at 10 23 52 AM


danielgavrilov (approved):
Great idea! Looks good to me :+1:

PR #624, opened 2020-09-09
Condense controls

shaahmad:
Currently breaking off to a separate branch to implement a UI component useful for this. Opening this as a draft PR to help avoid big merge conflicts when I return to this branch. Very much WIP.


grapher/controls/ControlsRow.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+        return controls
+    }
+
+    render() {
+        return (
+            <div className="controlsRow">
+                <ul>
+                    {this.controlsToRender.map((control) => (
+                        <li key={control.type} className="control">
```
breck7, line 117: Nice! A lot cleaner.

grapher/controls/ControlsRow.tsx
```diff
@@ -0,0 +1,125 @@
...
+        if (grapher.tab === "chart") {
+            const yAxis = grapher.activeTransform.yAxis
+            yAxis &&
+                yAxis.scaleTypeOptions.length > 1 &&
+                controls.push(
+                    <ScaleSelector scaleTypeConfig={yAxis} inline={true} />
+                )
+
+            grapher.canAddData &&
+                !grapher.hasFloatingAddButton &&
```
breck7, line 44: Okay, we can probably get to the point where we just pass in these controls (or a props object with these flags). Not sure when that will happen but shouldn't be too hard to merge in.

shaahmad: How about something like:
```
interface ChartType {
  ...
  controls: React.ReactElement[]
}

class LineChart extends ChartType {...}
class ScatterPlot extends ChartType {...}

class ControlsRow extends React.Component<> {
  render() {
     return <FancyUI items={grapher.activeChart.controls} />
     // or pass in as prop to avoid using grapher
  }
}
```

breck7: Maybe even just using react children prop?

PR #626, opened 2020-09-11
Collapsible List

shaahmad:
This PR adds a UI component necessary for condensing controls according to @vanwoerden's design specs:
image

Doesn't look like much yet, but will be wrapped in pretty formatting once combined with my draft PR #624 in a bigger component I will call `ControlsRow`.

The desired behaviors of this "Collapsible List" are:
- show the maximum number of items on the screen
- expose invisible items via a dropdown accessible via the "More" button
- allow arbitrary React Components to be used as items, so that we are not limited in what our controls can look like

Couldn't find any good library that met these specs, so had to roll my own.

Added a starter test (JSDOM 😢)  and Story to try it out.


grapher/controls/CollapsibleList.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default {
+    title: "CollapsibleList",
+    component: CollapsibleList,
+}
+
+export const CollapsibleListComponent = () => {
+    return <CollapsibleList>{collapsibleListSampleItems}</CollapsibleList>
+}
+
+export const MoreButtonComponent = () => {
```
breck7, line 14: Nice! It works.

Screen Shot 2020-09-10 at 2 49 46 PM

grapher/controls/CollapsibleList.jsdom.test.tsx
```diff
@@ -0,0 +1,18 @@
+#! /usr/bin/env yarn jest
+
+import * as React from "react"
+import { shallow } from "enzyme"
+import { CollapsibleList } from "./CollapsibleList"
+import { collapsibleListSampleItems } from "./CollapsibleList.sampleInput"
+
+// configure({ adapter: new Adapter() })
+describe(CollapsibleList, () => {
```
breck7, line 9: A jsdom test is a great start! Maybe at somepoint we'll find a way to make these faster automagically

grapher/controls/CollapsibleList.sampleInput.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react"
```
breck7, line 1: File naming and organization is great.

Should we put these in one folder? Either nested in Controls/ or at the same depth?

Not just for this one, but all the components. As at first you have a component, and then a stylesheet, and then a test file, and then a story file, etc.

grapher/controls/CollapsibleList.sampleInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+                    checked={this.checked}
+                    onChange={this.onToggle}
+                />
+                {` checkbox ${this.props.id}`}
+            </label>
+        )
+    }
+}
+
+const items: React.ReactElement[] = []
```
breck7, line 27: Nit: `const collapsibleListSampleItems = range(0,12).map(i => <SampleCheckBox key={i} id={i} />`

grapher/controls/CollapsibleList.scss
```diff
@@ -0,0 +1,15 @@
+.collapsibleList {
```
breck7, line 1: I think we had a team conversation about CSS class naming a while back, and went with camelCase, but I'm wondering if we should revisit that convo as a team and maybe switch to `CSSClassName === TypeScriptClassName`

grapher/controls/CollapsibleList.stories.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react"
+import { CollapsibleList, MoreButton } from "./CollapsibleList"
+import { collapsibleListSampleItems } from "./CollapsibleList.sampleInput"
+
+export default {
+    title: "CollapsibleList",
+    component: CollapsibleList,
```
breck7, line 7: Total sidenote, and not sure if it would even be helpful here.

Storybook also has the cool thing called "argTypes", when then let's you have controls for quickly testing various scenarios with your components. I've got some basic ones going for FacetChart. Honestly it's kind of a pain to use because it's weakly typed and hard to figure out what the options are, but when you get it going it's kind of neat.

Thought it might be useful to have a range to play around with size behavior.

grapher/controls/CollapsibleList.tsx
```diff
@@ -0,0 +1,131 @@
...
+            React.Children.map(this.props.children, (child, i) => {
+                return {
+                    index: i,
+                    child,
+                }
+            }) ?? []
+        )
+    }
+
+    private updateOuterContainerWidth() {
```
breck7, line 42: Hooray for privates!

grapher/controls/CollapsibleList.tsx
```diff
@@ -0,0 +1,131 @@
...
+            this.outerContainerRef.current?.clientWidth ?? 0
+    }
+
+    private calculateItemWidths() {
+        this.outerContainerRef.current
+            ?.querySelectorAll(".list-item")
+            .forEach((item) => this.itemsWidths.push(item.clientWidth))
+    }
+
+    private get numItemsVisible() {
```
breck7, line 53: I would make a tiny little method in Utils for this one (that way you can easily create some unit tests for it).

breck7:
Awesome. This will allow us to finally scale up the user controls. And some very clean code.

Screen Shot 2020-09-10 at 3 08 03 PM

breck7 (approved):
I would add a unit test for the visible items logic other than that just some nits.

grapher/controls/CollapsibleList.stories.tsx
shaahmad: Very interesting! I'll look into it.

grapher/controls/CollapsibleList.scss
shaahmad: I agree, I think it makes to just use the TypeScriptClassName. Will change when we decide that.

grapher/controls/CollapsibleList.sampleInput.tsx
shaahmad: Excellent! Exactly what I was looking for.

grapher/controls/CollapsibleList.sampleInput.tsx
shaahmad: done

grapher/controls/CollapsibleList.tsx
shaahmad: Done!
cc854e775f0423834126721be7b46b4f4a5ca8dd

PR #633, opened 2020-09-15
Condense Controls

shaahmad:
This PR moves inline controls and the vertical axis log/lin selector into a responsive `ControlsRow`.
In my next PR, I will move `ControlsRow` out of `ChartLayout` and into `LineChart`, etc.

**Live on [Nightingale](https://nightingale-owid.netlify.app/grapher/stunting-vs-level-of-prosperity-over-time).**
image

### DIFF
--- a/grapher/axis/AxisViews.tsx
+++ b/grapher/axis/AxisViews.tsx
@@ -132,22 +132,6 @@ export class VerticalAxisComponent extends React.Component<{
     verticalAxis: VerticalAxis
     isInteractive: boolean
 }> {
-    @computed get controls() {
-        const { bounds, verticalAxis } = this.props
-        const showControls =
-            this.props.isInteractive && verticalAxis.scaleTypeOptions.length > 1
-        if (!showControls) return undefined
-        return (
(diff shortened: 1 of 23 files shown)


shaahmad:
Ready for review!
Note I'm going to do a big `Controls` refactor up next. On my to-do, and to save you from having to mention it here:
- `ControlsRow` won't have its own file and will be a pure functional component
- same with `Add Entities` Button
- as mentioned above, `ControlsRow` will probably be rendered by individual ChartTypes rather than by `ChartTab`, so the `controlsRowControls` function in `ChartTab` will disappear.
- I will add tests to make sure the refactor doesn't cause any regressions
- Not a refactor, but I'm also aware that the ControlsRow is wonky on SlopeCharts with log/lin... as this is a small % of our views and it's not catastrophic, I think this is worth fixing soon but not necessarily in this PR.

On net, this PR should significantly reduce wonky controls layouts, so I think it's worth merging sooner rather than later.

I want to put this refactor in a separate PR as this one is already `+342 −401` and 40 commits.

PR #634, opened 2020-09-15
Remove mocks, more stories, and more Years to Times

breck7:
This removes all the "mock.json" files. Now tests generally are passed only the parameters needed to run the test at hand, so  when a test fails, a dev does not have to ruffle through extraneous setup details. Also they were all "any" typed.

I also added basic stories for Map and DataTable.

And removed "minYear" and "maxYear" in Grapher. We still have "minTime" (author set) and "startTime" (user set), minYear/maxYear was largely duplicative. (NB: There are still a few timeline regressions, working on it)

And continued to rename "Years" => "Times" when encountered.


grapher/dataTable/DataTable.jsdom.test.tsx
```diff
@@ -82,62 +112,101 @@ describe(DataTable, () => {
...
 
-        it("renders a tolerance notice when data is not from targetYear", () => {
-            const notice = view
-                .find("tbody tr td.dimension")
-                .first()
-                .find(ClosestYearNotice)
-            expect(notice.prop("closestYear")).toBe("2013")
-            expect(notice.prop("targetYear")).toBe("2016")
-        })
+        // todo: reenable
```
breck7, line 184: @shaahmad I was struggling a bit to figure out tolerance and what combination of setup I needed to get this notice to appear. Any tips?

shaahmad: DM-ing

PR #635, opened 2020-09-16
Refactor Transform Code, Modularize Charts, and Add Faceting

breck7:
I've still got one more day of work to do on this but posting work in progress. (EDIT: this time was estimate _slightly_ optimistic)

The goal of this branch is to add Faceting in a stable way. The problem is the codebase has grown tangled over time and needed to be untangled. Step one of taking it apart was relatively easy, but the hard part (still in progress) is putting it back together in a well engineered way without losing functionality. So lots of tests have been added alongside the refactoring. This isn't all improvements, but hopefully it's only one step back and seven or eight steps forward.

This branch does the following:

- **Adds** Faceting. You can now create Facet charts with 4 different strategies and unlimited smaller charts. Some design and UX details still to be worked out.
Screen Shot 2020-10-01 at 10 40 08 PM
- **Adds** helper methods to generate testing tables. For example, `SynthesizeGDPTable()` generates something like:
```csv
Entity,Code,Population,GDP,Year
Brunei,BRN,863.12M,$861.88 billion,1950
Brunei,BRN,51.03M,$42.41 billion,1951
...
```
- **Simplifies** OwidTable to be Immutable. Now all transforms create a chain of immutable tables, dplyr style. Transforms now look more like: `table.filterSmallCountries().filterByTimeline(1900, 1920).addRelativeColumn()`. You can call `.explain()` on a table to see the operations that created it.
- **Moves** "formatValue" functions to the column level. Now instead of passing around dozens of `formatTick` and `formatY` callbacks, just call the relevant format method directly on the column. There are ~16 column types including Numeric, Currency, and Ratio.
- **Removes** ChartTransforms.  These were all currently tightly coupled with each other and Grapher. I moved most of the common code to the CoreTable level to make it more reusable and testable and now charts can do transforms on tables. 
- **Adds** "ChartManager" interfaces for all of our chart types. This is like the traditional options/model/store object you'd find in another charting library, except the "Manager" name is there to remind you to pass it around directly so you can "dot in" to the properties for Mobx reactivity. All Charts now take one of these instead of taking a Grapher (but Grapher implements these interfaces). You can now easily create a chart without Grapher in a oneliner:
```tsx
<svg width={600} height={600}><LineChart manager={{ table: SynthesizeGDPTable() }} /></svg>
```
-  **Removes** EntityDimensionKey. AFAICT >99% of our charts can make do with just selecting by Entity Name. For the <30 charts that have some special entity + dimension selection requirement, I will figure out some legacy conversion function and/or just update those in the DB. This gets rid of a ton of complexity. In the future we can easily add any sort of selection functions we want "select these entity names" or "select by this continent" or "select countries with large populations", etc. 
- **Separates** EntitySelection from custom Colors. Right now those are tightly coupled which makes a lot of things harder. Right now if the author applies a color applied to a country, then the user deselects/reselects it, that color is gone (and that goes for programmatic things as well). Separating these should make things simpler. Now custom colors are added as a simple column. (ColorScales work as before).
- **Removes** ChartDimension class. What makes more sense is simply a `columnSlug` and a property. The rest of that functionality is now in Columns. (Not entirely gone in the short term for legacy reasons)
- **Removes** that = this pattern. Now uses interfaces.
- **Removes** VariableID in most places, in favor of column slug (Not entirely gone in the short term for legacy reasons)
- **Minimizes** saving of Grapher Configs. Defaults are no longer saved. So a blank chart gets saved as {} instead of 50 keys.
- **Removes** ChartLayout and GrapherView
- **Removes** React.context and ControlsOverlay
- **Adds** ~10 new test suites and ~80 new tests.
- **Adds** about 25 new stories
- **Flattens** unit tests that didn't need to be nested
- **Converts** typed unions to Enums throughout the code base
- **Removes** mock tests. We had a lot of tests that had big blobs of extraneous data. This removes those.


breck7:
I am archiving this PR

PR #640, opened 2020-09-22
Set unbounded value when dragging beyond timeline start/end

danielgavrilov:
A few months ago we could [set an unbounded value (`"earliest"` or `"latest"`) by dragging past the timeline brackets](https://owid.slack.com/archives/C011F008G6S/p1586599539086300):

Screenshot_2020-04-11_at_11_00_56

It seems we lost this feature at some point – not sure if intentionally or not?

This PR adds back that feature, as long as we want it.


breck7 (approved):
Nice! LGTM

PR #651, opened 2020-10-12
Exclude charts tagged "Private" from test pages

danielgavrilov:
This would hide all charts tagged as "Private", unless the `includePrivate` query param is passed.

### DIFF
--- a/adminSite/server/testPageRouter.tsx
+++ b/adminSite/server/testPageRouter.tsx
@@ -191,6 +191,21 @@ testPageRouter.get("/embeds", async (req, res) => {
         query = query.andWhere(`config->"$.hasChartTab" IS TRUE`)
     }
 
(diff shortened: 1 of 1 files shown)


breck7 (approved):
That is some impressive SQL.

danielgavrilov:
@breck7 You should see Marcel's migration: https://github.com/owid/owid-grapher/blob/chart-time-unit/db/migration/1602158086346-ChartYearToTimeUnit.ts

breck7:
> @breck7 You should see Marcel's migration: https://github.com/owid/owid-grapher/blob/chart-time-unit/db/migration/1602158086346-ChartYearToTimeUnit.ts

@MarcelGerber looks great!

Reminds me we may have a few scenarios where the current legacy conversion code is broken (if you have mixed time units in one chart). I probably haven't sufficiently tested for that. But that's purely client side.

"// There's no going back!"

Amen.

PR #653, opened 2020-10-13
Distinct admin context & admin user flags

danielgavrilov:
Fixes: https://www.notion.so/owid/Fix-isAdmin-Grapher-flag-fa3cc7b0762c43c7be5f3f172d8299dd

There might be a better way to name these?

### DIFF
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -491,11 +491,21 @@ export class Grapher
         this._receiveLegacyData(json)
     }
 
-    @computed get isAdmin() {
+    /**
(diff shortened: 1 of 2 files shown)


grapher/core/Grapher.tsx
```diff
@@ -491,11 +491,12 @@ export class Grapher
         this._receiveLegacyData(json)
     }
 
-    @computed get isAdmin() {
+    @computed get isAdminContext(): boolean {
```
breck7, line 494: What about something boring but descriptive like `get loadDataViaAPI()`?

And then copy/paste the notion comment above it for future ref?

breck7 (approved):

PR #654, opened 2020-10-13
Timeline fixes

danielgavrilov:
A bunch of fixes for the timeline. The goal was to fix [Timeline plays backwards](https://www.notion.so/owid/Timeline-plays-backwards-9e69532e53394bdeb825a9d1f51e3a54).

### DIFF
--- a/grapher/timeline/TimelineController.ts
+++ b/grapher/timeline/TimelineController.ts
@@ -27,6 +27,14 @@ export class TimelineController {
         return this.manager.times
     }
 
+    private get startTime() {
+        return findClosestTime(this.timesAsc, this.manager.startTime)!
+    }
+
+    private get endTime() {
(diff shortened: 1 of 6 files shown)


grapher/timeline/TimelineController.ts
```diff
@@ -165,15 +167,17 @@ export class TimelineController {
         this.updateEndTime(endTime)
     }
 
+    // TODO
```
shaahmad, line 169: Still needed, or Much Todo About Nothing?

danielgavrilov: Haha I really wanted to emphasise it and I forgot it anyway in the end 🤦

danielgavrilov:
I will optimistically merge this because otherwise the timeline on `next` is a bit broken, and it gets in the way of other fixes.

PR #667, opened 2020-10-22
Tolerance-&-timeline-related fixes

danielgavrilov:
Still more to go. But I think these are good to merge?

### DIFF
--- a/coreTable/OwidTable.ts
+++ b/coreTable/OwidTable.ts
@@ -216,8 +216,9 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
     }
 
     filterByEntityNames(names: EntityName[]) {
+        const namesSet = new Set(names)
         return this.filter(
-            (row) => names.includes(row[OwidTableSlugs.entityName]),
+            (row) => namesSet.has(row[OwidTableSlugs.entityName]),
             `Filter out all entities except '${names}'`
         )
     }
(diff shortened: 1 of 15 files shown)


danielgavrilov:
I created isLineChartThatTurnedIntoDiscreteBar in order to add it to `DiscreteBarManager` and assign bar colors based on that. But then I realised they need to be in the `transformTable()` or right after it... and wrote up a [Notion doc](https://www.notion.so/owid/How-to-assign-color-scheme-to-lines-and-bars-409c0ca95b354092a317c49be303680b) about that.

coreTable/OwidTable.ts
```diff
@@ -215,13 +215,11 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
...
-    filterByEntityName(name: EntityName) {
-        // todo; why not a filter by?
-        return new OwidTable(this.rowsByEntityName.get(name) || [], this.defs, {
-            parent: this,
-            tableDescription: `Filter out all entities except '${name}'`,
-            transformCategory: TransformType.FilterRows,
-        })
+    filterByEntityNames(names: EntityName[]) {
+        return this.filter(
+            (row) => names.includes(row[OwidTableSlugs.entityName]),
```
breck7, line 220: Nice!

Do you know if the JS engine will optimize string[].includes ? If not may want to add a one liner above converting names into a set for perf.

breck7 (approved):
💯 💯 💯 💯 💯

PR #673, opened 2020-10-26
Dev: add group and reduce methods to CoreTable. Add perf test harness for CE

breck7:
I am switching off perf for now to work on other issues but wanted to merge this in.

This adds general "groupBy" and "reduce" methods. The benefits of having these general GroupBy/Reduce methods, in addition to fixing this perf, will be that we may find them useful elsewhere for new functionality.

It also separates the (now) slowest part of loading the CE into an `addGroups` method in MegaCsv.ts. Next step would be to use the new methods for that method. My other branch does that but haven't optimized perf yet.

I also turned the plain `boolean[]` filter mask into a tiny class. There's some perf gains that can be had in filtering and just took a step in that direction.

I also turned the table.concat method from accepting a single table to an array of tables which makes more sense.

Finally, I am committing my "perf harness" for the CE to a Story so when I switch back to perf can get some help on figuring it out.

### DIFF
--- a/coreTable/CoreTable.ts
+++ b/coreTable/CoreTable.ts
@@ -48,7 +48,6 @@ import {
     concatColumnStores,
     rowsToColumnStore,
     autodetectColumnDefs,
-    applyFilterMask,
     reverseColumnStore,
     renameColumnStore,
     replaceNonPositives,
@@ -64,8 +63,6 @@ const TransformsRequiringCompute = new Set([
     TransformType.AppendColumns,
     TransformType.UpdateRows,
 ])
-
-type FilterMask = boolean[] // todo: change into bit array
 
 interface AdvancedOptions {
     tableDescription?: string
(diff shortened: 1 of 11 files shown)


coreTable/CoreTable.ts
```diff
@@ -1191,11 +1191,85 @@ export class CoreTable<
...
+            const set = new Set(input as number[])
+            this.mask = range(0, numRows).map((index) =>
+                set.has(index) ? keepThese : !keepThese
+            )
+        } else this.mask = input as boolean[]
+    }
+
+    inverse() {
+        return new FilterMask(this.numRows, this.mask.reverse())
+    }
```
danielgavrilov, line 1264: @breck7 Should this be `reverse()`? `inverse()` I understand as flipping the booleans, from TRUE → FALSE and vice versa.

danielgavrilov:
🙌 This is awesome! Especially having `groupBy`.

coreTable/CoreTable.ts
breck7: Ah yes! Nice catch. I'm guessing my test passes because I have only 2 samples. Will take a look.

breck7: https://github.com/owid/owid-grapher/commit/256bcb325c6a486f5c709caf618456a4788bcecf

PR #675, opened 2020-10-28
ScatterPlots: handle xOverrideTime and apply color scales

danielgavrilov:
Also adds more tests around this. Keeps all behaviour as it was before, haven't yet implemented mixed-time-type table joins...

### DIFF
--- a/adminSite/server/testPageRouter.tsx
+++ b/adminSite/server/testPageRouter.tsx
@@ -199,6 +199,23 @@ testPageRouter.get("/embeds", async (req, res) => {
         tab = "map"
     }
 
+    if (req.query.mixedTimeTypes) {
+        query.andWhere(
+            `
+            (
+                SELECT COUNT(DISTINCT CASE
+                    WHEN variables.display->"$.yearIsDay" IS NULL
+                    THEN "year"
+                    ELSE "day"
+                END) as timeTypeCount
+                FROM variables
+                JOIN chart_dimensions ON chart_dimensions.variableId = variables.id
+                WHERE chart_dimensions.chartId = charts.id
+            ) >= 2
+        `
+        )
+    }
+
     if (req.query.ids) {
         query = query.andWhere(`charts.id IN (${req.query.ids})`)
     }
--- a/adminSite/server/testPageRouter.tsx
+++ b/adminSite/server/testPageRouter.tsx
@@ -13,6 +13,7 @@ import * as querystring from "querystring"
 import * as lodash from "lodash"
 import * as url from "url"
 import { getComparePage, svgCompareFormPage } from "svgTester/SVGTester"
+import { parseIntOrUndefined } from "grapher/utils/Util"
 
 const IS_LIVE = ADMIN_BASE_URL === "https://owid.cloud"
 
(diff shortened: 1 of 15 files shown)


coreTable/OwidTable.test.ts
```diff
@@ -496,62 +496,76 @@ describe("time domain", () => {
...
-                    year: 2005,
-                }),
-            ])
-        )
-    })
+    function applyTolerance(table: OwidTable) {
+        return table.interpolateColumnWithTolerance("gdp", 1)
+    }
+
+    // Applying the tolerance twice to ensure operation is idempotent.
```
breck7, line 517: Ah, nice catch! Didn't think of that.

coreTable/CoreTable.ts
```diff
@@ -1061,6 +1061,34 @@ export class CoreTable<
...
+            )
+        }
+
+        const existingColumn = this.get(slug)
+        if (!existingColumn) {
+            throw new Error(`Cannot duplicate column: '${slug}' does not exist`)
+        }
+
+        return this.transform(
+            {
```
breck7, line 1077: I like it. This is maybe the pattern to use for the other transforms (just building a new column store then passing that to transform)

coreTable/CoreTableUtils.ts
```diff
@@ -227,7 +228,8 @@ export const makeKeyFn = (
     columnStore: CoreColumnStore,
     columnSlugs: ColumnSlug[]
 ) => (rowIndex: number) =>
-    columnSlugs.map((slug) => columnStore[slug][rowIndex].toString()).join(" ")
+    // toString() handles `undefined` and `null` values, which can be in the table.
```
breck7, line 231: Makes sense. Relatedly I think it'd be ideal to keep undefineds and nulls out of the table at some point, but haven't nailed that down yet (and not entirely sure it's important).

breck7 (approved):
LGTM but I haven't in my head made the connection yet as to the downstream fix on scatters this has (I know it's something to do with date joins, but forget the exacts).

Is there a before and after photo? Or is this a building step toward the fix?

danielgavrilov:
Ok this now handles the legacy `dimension.targetYear`. We had renamed it to `targetTime` but the database config is still under `targetYear` so we'll have to stick to it for now.

The ScatterPlot seems to receive the right table now, but still not fixed, will do that in a separate PR.

danielgavrilov:
Implemented some ScatterPlot fixes:
- [Scatter: handle xOverrideTime](https://www.notion.so/owid/Scatter-handle-xOverrideTime-6fdb71161d2a478aafe19367c6dff58d)
- [Scatter: fix color scales](https://www.notion.so/owid/Scatter-fix-color-scales-532323e7d0084e229789970c559d7235)

There are still others to go, keeping track in [Scatter: issues & broken charts](https://www.notion.so/owid/Scatter-issues-broken-charts-16577b3c1ea449e4a93b9da815fd86be).

Since this introduces some changes in legacy conversion & other fixes, I think we can merge it now, and fix the other ScatterPlot issues separately.

adminSite/server/testPageRouter.tsx
```diff
@@ -97,7 +103,9 @@ function EmbedTestPage(props: EmbedTestPageProps) {
                 </div>
                 {props.charts.map((chart) => (
                     <div key={chart.slug} className="row">
-                        <div className="chart-id">{chart.id}</div>
+                        <div className="chart-id">
+                            <a href={`?ids=${chart.id}`}>{chart.id}</a>
```
breck7, line 107: Good idea!

PR #681, opened 2020-11-03
Fix: country param ignored in exports

danielgavrilov:
As reported on Slack: https://owid.slack.com/archives/C46U9LXRR/p1604397643008400


breck7 (approved):
lgtm

PR #686, opened 2020-11-04
Unskip Map Tests

shaahmad:
Tiny change -- makes it so that the timelineHandles use the closest data point to the query string `time` param instead of the param itself.

### DIFF
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -457,7 +457,9 @@ export class Grapher
     @observable private selectedEntitiesInQueryParam: string[] = []
 
     @action.bound private setTimeFromTimeQueryParam(time: string) {
-        this.timelineHandleTimeBounds = getTimeDomainFromQueryString(time)
+        this.timelineHandleTimeBounds = getTimeDomainFromQueryString(time).map(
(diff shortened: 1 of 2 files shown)


grapher/core/Grapher.tsx
```diff
@@ -457,7 +457,9 @@ export class Grapher
     @observable private selectedEntitiesInQueryParam: string[] = []
 
     @action.bound private setTimeFromTimeQueryParam(time: string) {
-        this.timelineHandleTimeBounds = getTimeDomainFromQueryString(time)
+        this.timelineHandleTimeBounds = getTimeDomainFromQueryString(time).map(
```
breck7, line 460: Nice! this makes sense to me. The only thing I can think of going wrong is the order of operations (I can't yet wrap my head around when this.times is available :) ).

But the tests passing gives me confidence that times is available.

breck7 (approved):

grapher/core/Grapher.tsx
shaahmad: Great, yeah I had the same concern which is why I did it as a PR. But from my testing it seems to look good.

PR #687, opened 2020-11-04
Fix: Use isMissing instead of undefined column check

danielgavrilov:
There are a bunch in the `CovidExplorer` that I didn't fix, let me know if I should.

Also, some of these checks may be redundant – I wasn't paying attention to that, just trying to get it to the previous state.

Also, no tests added, and it seems it might be a bit of an effort to design test cases for all?

### DIFF
--- a/coreTable/CoreTable.ts
+++ b/coreTable/CoreTable.ts
@@ -477,7 +477,9 @@ export class CoreTable<
     // Todo: remove this. Generally this should not be called until the data is loaded. Even then, all calls should probably be made
     // on the column itself, and not tied tightly to the idea of a time column.
     @imemo get timeColumnFormatFunction() {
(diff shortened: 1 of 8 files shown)


breck7:
Would this pattern work?

```
abstract class AbstractCoreColumn<JS_TYPE extends PrimitiveType> {
// ...
get isMissing(): boolean {
        return this instanceof MissingColumn
    }
// ...
}
```

And then:

```
this.timeColumn.isMissing
this.yColumn.isMissing
mapColumn.isMissing
// etc?
```

breck7:
Also, I didn't realize this was used so much. Thanks for fixing all the places!

danielgavrilov:
yeah was thinking of doing that, will rewrite it tomorrow

PR #692, opened 2020-11-06
Rewrite from row to column transform: toPercentageFromEachColumnForEachEntityAndTime

danielgavrilov:
Not sure if correct, will double check tomorrow.


coreTable/OwidTable.ts
```diff
@@ -281,70 +281,77 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
...
-                        ? (100 * value) / total
-                        : value
-                })
-                return newRow
-            }),
+
+        const columnStore = this.columnStore
+        const columnStorePatch: CoreColumnStore = {}
+
+        const totals = new Array(this.numRows).fill(0).map((_, i) =>
```
breck7, line 327: Is `new Array(this.numRows).fill(0)` the fastest way in JS to create a range? I've been using the lodash `range`) method but figured it's probably not the fastest way.

danielgavrilov: Yeah I should've gone with range, I just thought it's quicker to use vanilla `new Array(length)` but that doesn't work if you wanna map over it.

breck7: You can still map over the indexes.My hunch is `new Array` may be faster but have to check. We could define our own range if so.

breck7: (obviously not a priority, just wondering aloud)

danielgavrilov: https://stackoverflow.com/a/35086350

PR #694, opened 2020-11-06
Handle mismatching columns in concatColumnStores()

danielgavrilov:


coreTable/CoreTableUtils.ts
```diff
@@ -360,14 +362,29 @@ export const appendRowsToColumnStore = (
     return newColumnStore
 }
 
+export const getColumnStoreLength = (store: CoreColumnStore) => {
```
breck7, line 365: nit: does this need to be exported?

coreTable/CoreTableUtils.test.ts
```diff
@@ -258,3 +259,48 @@ describe(sortColumnStore, () => {
...
+    it("concats column stores with missing columns", () => {
+        expect(
+            concatColumnStores([
+                {
+                    a: [1, 2],
+                    b: [6, 7],
+                },
+                {
+                    a: [3, 4],
+                    c: [0, 0],
```
breck7, line 291: So we don't append columns, right? `c` is drop. I think this is reasonable, but maybe we should add a comment to the `concatColumnStores` method or even rename it to indicate that only the first columnStore's columns will be kept?

coreTable/CoreTableUtils.ts
```diff
@@ -360,14 +362,29 @@ export const appendRowsToColumnStore = (
...
+        return store[slug].length
+    }
+    return 0
+}
+
 export const concatColumnStores = (stores: CoreColumnStore[]) => {
+    if (!stores.length) return {}
+
+    const lengths = stores.map(getColumnStoreLength)
+    const slugs = Object.keys(first(stores)!)
```
breck7, line 376: Ah wait, forget about my comment above. What if this was a param?

```
export const concatColumnStores = (stores: CoreColumnStore[], slugsToKeep = Object.keys(first(stores)!)) => {
```

Then it would be clear that only certain slugs will be kept and theoretically you could specify to keep them all

danielgavrilov: Great idea!

PR #701, opened 2020-11-09
ScatterPlot fixes

danielgavrilov:
~~Based on `next-columnfilter-error-values` which should be merged first.~~

~~Missing tests, will add soon.~~ Added tests.

### DIFF
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -498,7 +498,7 @@ export class Grapher
     }
 
     @computed
-    private get tableAfterPopulationFilterAndActiveChartTransform(): OwidTable {
+    get tableAfterPopulationFilterAndActiveChartTransform(): OwidTable {
         const table = this.tableAfterPopulationFilter
         if (!this.isReady || !this.isChartOrMapTab) return table
         return this.chartInstance.transformTable(table)
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -149,6 +149,7 @@ import { ChartComponentClassMap } from "grapher/chart/ChartTypeMap"
 import { ColorSchemeName } from "grapher/color/ColorConstants"
 import { SelectionArray } from "grapher/selection/SelectionArray"
 import { legacyToOwidTableAndDimensions } from "coreTable/LegacyToOwidTable"
+import { ScatterPlotManager } from "grapher/scatterCharts/ScatterPlotChartConstants"
 
 declare const window: any
 
@@ -207,6 +208,7 @@ export class Grapher
         TooltipManager,
         FooterControlsManager,
         DataTableManager,
+        ScatterPlotManager,
         MapChartManager {
     @observable.ref type = ChartTypeName.LineChart
     @observable.ref id?: number = undefined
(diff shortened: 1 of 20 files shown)


grapher/core/Grapher.tsx
```diff
@@ -520,9 +510,23 @@ export class Grapher
     }
 
     @computed
-    private get tableAfterPopulationAndActiveChartAndTimelineFilters() {
+    get tableAfterAuthorTimelineAndActiveChartTransformAndPopulationFilter() {
```
breck7, line 513: I really like these methods names as a stopgap until we discover what the pattern is that we should be using. But then again, I think of all the data science scripts and really not sure if there is a better pattern. Maybe this will be the long term solution too.

grapher/core/Grapher.tsx
```diff
@@ -1240,27 +1244,6 @@ export class Grapher
         )
     }
 
-    // todo: this is only relevant for scatter plots. move to scatter plot class?
-    getEntityNamesToShow(filterBackgroundEntities?: boolean): EntityName[] {
```
breck7, line 1244: woohoo!

grapher/scatterCharts/ScatterPlotChart.tsx
```diff
@@ -79,28 +81,38 @@ export class ScatterPlotChart
     @observable private hoverColor?: Color
 
     transformTable(table: OwidTable) {
+        if (this.manager.excludedEntities) {
+            const excludedEntityIdsSet = new Set(this.manager.excludedEntities)
```
breck7, line 85: sidenote: post merge as part of a grapher config cleanup we should probably change `excludedEntities` from Ids to entityNames.

grapher/scatterCharts/ScatterPlotChart.tsx
```diff
@@ -159,13 +194,22 @@ export class ScatterPlotChart
         if (this.canAddCountry) this.selectionArray.toggleSelection(entityName)
     }
 
-    // Only want to show colors on legend that are actually on the chart right now
-    @computed private get colorsInUse() {
+    // Returns the colors that are used by all points, *across the whole timeline*.
+    // This is why we need the table before the timeline filter is applied.
+    @computed private get colorsInUse(): Color[] {
+        const allValues =
+            this.manager.tableAfterAuthorTimelineAndActiveChartTransformAndPopulationFilter?.get(
```
breck7, line 201: todo post merge: maybe a pattern like `manager.table`, `manager.transformedTable`, and `manager.untransformedTable`, and the latter would just be:

```
Grapher {
computed  get untransformedTable() {
reutrn this.tableAfterAuthorTimelineAndPopulationFilter
}
}
```

grapher/scatterCharts/ScatterPlotChart.tsx
```diff
@@ -595,53 +641,16 @@ export class ScatterPlotChart
         return new Set(seriesNames)
     }
 
-    @computed private get compareEndPointsOnly() {
+    @computed get compareEndPointsOnly() {
         return !!this.manager.compareEndPointsOnly
     }
 
-    private set compareEndPointsOnly(value: boolean) {
+    set compareEndPointsOnly(value: boolean) {
```
breck7, line 648: nit: do we need this method? This is probably fine, but I always get confused when mixing setters and @computeds with mobx. Maybe just:

```
@action.bound private onToggleEndpoints() {
       this.manager.compareEndPointsOnly = !this.compareEndPointsOnly ?? undefined
    }
```

breck7 (approved):
LGTM!

grapher/scatterCharts/ScatterPlotChart.tsx
danielgavrilov: Good idea!

grapher/core/Grapher.tsx
danielgavrilov: Part of the issue is also that it's better performance to have computations chopped up in as many steps as possible, with computed props for each, so that mobx can optimize and only update when an update is necessary.

PR #702, opened 2020-11-09
Create OwidTable#complete() method

danielgavrilov:
The CoreTable#complete() method doesn't copy down entity attributes, see: https://www.notion.so/owid/complete-doesn-t-copy-down-entity-attributes-4a402cdf4c914011b1d460c7c25094ab

The intent is that we would use it in all the places we use `complete()` currently. But it's definitely messy – is there some more generalized way to do it?

Todo:
- [ ] Fill `entityColor` column too
- [ ] Use this in favor of `CoreTable#complete`

### DIFF
--- a/coreTable/OwidTable.ts
+++ b/coreTable/OwidTable.ts
@@ -54,6 +54,8 @@ import {
     toleranceInterpolation,
     replaceDef,
     InterpolationProvider,
+    cartesianProduct,
+    rowsFromMatrix,
 } from "./CoreTableUtils"
 import { CoreColumn, ColumnTypeMap } from "./CoreTableColumns"
 import { OwidSourceProps } from "./OwidSource"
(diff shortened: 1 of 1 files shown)


danielgavrilov:
This is definitely messy. Is there some more generalized way to do it?

breck7:
What about a complete and a left join?

breck7:
Ah nm, not quite. I think maybe join is not filling, though maybe that's how joins are supposed to work (need to look that up). 

```
const csv = `country,year,code
usa,2000,USAAAA
usa,2002,USAAAA
uk,2001,UKKKKK`
    const table = new CoreTable(csv)
    const completed = table.complete(["country", "year"]).dropColumns(["code"])
    completed.leftJoin(table, ["country", "year"]).dump()
```
Gives me:
    ```
┌─────────┬─────────┬──────┬─────────────────────────────┐
    │ (index) │ country │ year │            code             │
    ├─────────┼─────────┼──────┼─────────────────────────────┤
    │    0    │  'usa'  │ 2000 │          'USAAAA'           │
    │    1    │  'usa'  │ 2002 │          'USAAAA'           │
    │    2    │  'uk'   │ 2001 │          'UKKKKK'           │
    │    3    │  'usa'  │ 2001 │ NoMatchingValueAfterJoin {} │
    │    4    │  'uk'   │ 2000 │ NoMatchingValueAfterJoin {} │
    │    5    │  'uk'   │ 2002 │ NoMatchingValueAfterJoin {} │
    └─────────┴─────────┴──────┴─────────────────────────────
```

**Comment by danielgavrilov** [comment]
Yeah, but good idea to explore!

Maybe create a table `[entityId, entityName, entityCode]` and join it with the completed on `[entityId]`.

File: coreTable/OwidTable.ts [hunk 1]
```diff
@@ -499,6 +501,51 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
...
+        const toAdd = product
+            .filter((row) => !index.has(row.join(" ")))
+            .map((row) => {
+                const entityName = row[0]
+                return [
+                    ...row,
+                    this.entityNameToIdMap.get(entityName),
+                    this.entityNameToCodeMap.get(entityName),
+                ]
+            })
```
danielgavrilov, line 518: @breck7 This is effectively a LEFT JOIN of the missing rows (`[entityName, time]`) with the table of unique entities (`[entityName, entityId, entityCode]`) ON `[entityName]`.

Could make it slightly neater, but it would be overhead to create a CoreTable with `[entityName, entityId, entityCode]` only to join it later?

breck7:
This one is probably fine, but I'm still not quite sure.

How would this handle this test case:

```
time Country Color
2000 USA Red
2001 USA White
2002 CAN Blue
```
And we ran `complete` on `[time, Country]`? What colors would we use for CAN in 2000 and 2001? What color would we use for USA in 2002?

I think maybe the bigger problem we have is that we want that multi-table solution, where we could just pass in multiple tables to a chart (like one for population, color, annotations, etc), and do joins/maps to the appropriate series down there?

Not sure.

danielgavrilov:
Closing this – it's messy and we likely need to rethink the key-value column relationship. Eventually we need to support multiple entity columns, so need to find a better approach.

PR #705, opened 2020-11-11
Add ShortHash to Explorer URLS and deprecate country param

breck7:
Don't be fooled by the LOC, this is actually removing code (the high LOC is just because I've dumped the ~300 entity codes from the DB into a file so we can do sync legacy url upgrading).

This:

- Adds the Git Short Hash to Explorer URLS
- Introduces the simplified "selection" param which **only** operates on EntityNames, and upgrades legacy URLs to the new thing
- Fixes a number of Selection & URL related bugs

Once this is deployed to Jefferson, I will update the existing explorers' defaultView to use `selection` instead of country.


explorer/client/Explorer.tsx
```diff
@@ -45,91 +48,60 @@ import { EntityPicker } from "grapher/controls/entityPicker/EntityPicker"
...
+            this.props.queryString ?? program.defaultView
         )
+        return program
     }
 
-    @observable hideControls = false
     @observable entityPickerMetric?: ColumnSlug =
-        strToQueryParams(this.explorerProgram.queryString ?? "").pickerMetric ??
-        undefined
+        strToQueryParams(this.props.queryString ?? "").pickerMetric ?? undefined
```
breck7, line 90: Ah, just remembered I need to put pickerMetric into query string (And I'll add typing for this). There is a notion issue already

PR #707, opened 2020-11-14
Dev: reduce usage of dangerouslySetInnerHTML

breck7:
Sometimes I'll do a "Find All References" and TypeScript will miss certain references that are in strings.

I'm wondering if we could move away from the pattern of using `dangerouslySetInnerHTML`, but I don't know enough about the webpack build and bundling process, etc, to know if that's too much of a pain.

Regardless I think it may be nice to reduce usage of untyped strings, and thought that this is one thing we could do.

Any other ideas?

### DIFF
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -76,7 +76,6 @@ import {
 import {
     strToQueryParams,
     queryParamsToStr,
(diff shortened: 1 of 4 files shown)


breck7:
I'm going to close for now and take a fresh look later when we can come up with a better approach

PR #708, opened 2020-11-16
New ScatterPlot interpolation: a value can belong only to a single point

danielgavrilov:
The tolerance used to be applied independently to X and Y column, which meant e.g. a constant Y value from a single year would be used as X changes over time, creating some "staircases":

Before:

Screenshot 2020-11-12 at 16 38 04

Now:

Screenshot 2020-11-16 at 17 48 51

Related discussion on Slack: https://owid.slack.com/archives/C01D295TM4K/p1605193993004700

To-do:
- [x] Resolve `TODO` and `TO-DISCUSS` 
- [x] Add ScatterPlot tests
- ~~Drop any duplicate points? This needs to happen after the timeline filter. But it needs to minimize the difference between `timeValue` and `time.x`.~~ The `endPointsOnly` transform also gets more complex if we drop duplicates, so I decided to leave duplicates unless they cause problems. No problems spotted so far...

### DIFF
--- a/coreTable/OwidTable.ts
+++ b/coreTable/OwidTable.ts
@@ -165,26 +165,13 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
         return [min(mins), max(maxes)]
     }
 
-    // TO-DISCUSS
-    // Do we want to have two separate functions, or possibly a parameter: "any" | "all"?
-    // Do we want to access the column methods, or do it like this through the columnStore?
-    timeDomainForIntersection(
+    originalTimeDomainFor(
         slugs: ColumnSlug[]
     ): [Time | undefined, Time | undefined] {
-        let minTime: Time | undefined = undefined
-        let maxTime: Time | undefined = undefined
-        const { columnStore, timeColumn } = this
-        for (let index = 0; index < this.numRows; index++) {
-            if (
-                // The difference between above and this is .every() vs .some()
-                slugs.every((slug) => isNotErrorValue(columnStore[slug][index]))
-            ) {
-                const time = columnStore[timeColumn.slug][index] as Time
-                if (minTime === undefined || minTime > time) minTime = time
-                if (maxTime === undefined || maxTime < time) maxTime = time
-            }
-        }
-        return [minTime, maxTime]
+        const cols = this.getColumns(slugs)
+        const mins = cols.map((col) => min(col.originalTimes))
+        const maxes = cols.map((col) => max(col.originalTimes))
+        return [min(mins), max(maxes)]
     }
 
     filterByEntityNames(names: EntityName[]) {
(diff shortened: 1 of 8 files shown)


grapher/utils/Util.ts
```diff
@@ -1045,3 +1068,86 @@ export const splitArrayIntoGroupsOfN = (arr: any[], maxPerGroup: number) => {
...
+        const highCandidateIndexB =
+            sortedTimesB[closestIndexInB] >= timeA ? closestIndexInB : undefined
+
+        if (
+            lowCandidateIndexB !== undefined &&
+            highCandidateIndexB !== undefined &&
+            timeA - sortedTimesB[lowCandidateIndexB] <= maxDiff &&
+            timeA - sortedTimesB[lowCandidateIndexB] <
+                sortedTimesB[highCandidateIndexB] - timeA
+        ) {
```
breck7, line 1115: _super_ nit: maybe drop the {}? I prefer no "{}" when the body of a block is only one line. It tells me that the block is as simple as can be. but suppper nit

grapher/scatterCharts/ScatterPlotChart.test.ts
```diff
@@ -708,3 +704,131 @@ describe("scatter plot with xOverrideTime", () => {
         )
     })
 })
+
+describe("x/y tolerance", () => {
+    const table = new OwidTable(
+        [
+            [
+                "entityId",
```
breck7, line 712: nit: you can drop entityId and entityCode now I think, not required anymore.

coreTable/OwidTable.ts
```diff
@@ -615,6 +631,154 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
         )
     }
 
+    interpolateColumnsByClosestTimeMatch(
```
breck7, line 634: This fn is huge, and I'm sure there are ways to break it up, but seems like it is just a complex set of operations and more importantly the tests look good so we can always break it up later.

breck7:
The Devil's Chart! LGTM. I find it pretty clear. I think maybe in the future we could figure out smaller intermediate steps but gonna be realllly nice to have tests for all this scatter interpolation!!

grapher/utils/Util.ts
breck7: (obviously no need to fix any of these nits before merignG)

danielgavrilov: I think no `{}` works well when the full statement is in a single line, but otherwise find `{}` better. 

In a multi-line, you might add a line above/below and it would compile without errors

danielgavrilov:
Yeah it would be great to split up the huge function and test the chunks independently. It ran into many bugs implementing it, always a pain to debug.

PR #710, opened 2020-11-17
Apply Admin Dimension-level Changes

shaahmad:
Tried a few different approaches, but the simplest and most generalizable one in the admin context seemed to be to rebuild the input table on any dimension-level changes. The table-rebuild is triggered `onBlur` and perf seems to be zippy, even on `localhost` with Docker.

### DIFF
--- a/adminSite/client/DimensionCard.tsx
+++ b/adminSite/client/DimensionCard.tsx
@@ -68,6 +68,16 @@ export class DimensionCard extends React.Component<{
         )
     }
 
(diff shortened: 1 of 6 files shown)


danielgavrilov:
Awesome! I didn't expect it to be so quick.

Should this support adding dimensions too? I'm still getting errors when trying to create a new chart or when adding a dimension to an existing one.

shaahmad:
Good catch @danielgavrilov! That is fixed now.

PR #713, opened 2020-11-18
Fix: Use % axes for ScatterPlot 'average annual change' transform

danielgavrilov:
Instead of passing a column to the axis, it passes a formatting function only.

The slightly worrying thing with this is that the `.formatForTick` method needs to be bound to the instance. Right now only one exists, but I can see how it could go wrong if we change its definition.

Anything clever we can do, or should I create an intermediate arrow function every time?

```ts
axis.formatTick = (value, options) => column.formatForTick(value, options)
// instead of the current:
axis.formatTick = column.formatForTick
```

### DIFF
--- a/coreTable/CoreTableColumns.ts
+++ b/coreTable/CoreTableColumns.ts
@@ -31,6 +31,7 @@ import { getOriginalTimeColumnSlug } from "./OwidTableUtil"
 import { imemo } from "./CoreTableUtils"
 import moment from "moment"
 import { OwidSource } from "./OwidSource"
+import { bind } from "decko"
 
 interface ColumnSummary {
     numErrorValues: number
@@ -186,7 +187,8 @@ abstract class AbstractCoreColumn<JS_TYPE extends PrimitiveType> {
         return this.formatValue(value)
     }
 
-    formatForTick(value: any, options?: TickFormattingOptions) {
+    // Needs to be bound to instance to pass it to Axis without creating an intermediate function
+    @bind formatForTick(value: any, options?: TickFormattingOptions) {
         return this.formatValueShort(value, options)
     }
 
--- a/grapher/axis/Axis.ts
+++ b/grapher/axis/Axis.ts
@@ -7,12 +7,12 @@ import {
     uniq,
     sortBy,
     maxBy,
+    anyToString,
 } from "grapher/utils/Util"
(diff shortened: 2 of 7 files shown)


grapher/scatterCharts/ScatterPlotChart.tsx
```diff
@@ -748,23 +750,34 @@ export class ScatterPlotChart
         return this.domainDefault("y")
     }
 
+    readonly relativeAxisTickFormatter: (
```
breck7, line 753: If it's a relative mode, the column type should be a percentage already and should have the correct format fn, right?

grapher/axis/Axis.ts
```diff
@@ -26,7 +26,10 @@ interface Tickmark {
 abstract class AbstractAxis {
     config: AxisConfig
     @observable.ref domain: ValueRange
-    @observable formatColumn?: CoreColumn // Pass the column purely for formatting reasons. Might be a better way to do this.
+    @observable tickFormatter?: (
```
breck7, line 29: I actually prefer passing the column, the more I think about it. But seems like we should pass it as `column` instead of `formatColumn` (which jus sounds weird).

The really nice thing about having all formating done in one place (the column level), is hat i s really easy o test.

coreTable/CoreTableColumns.ts
```diff
@@ -186,7 +187,8 @@ abstract class AbstractCoreColumn<JS_TYPE extends PrimitiveType> {
         return this.formatValue(value)
     }
 
-    formatForTick(value: any, options?: TickFormattingOptions) {
+    // Needs to be bound to instance to pass it to Axis without creating an intermediate function
+    @bind formatForTick(value: any, options?: TickFormattingOptions) {
```
breck7, line 191: nit: I've found `@bind` to be a code smell. I think passing the column around might be a better approach, and now that i hink about it we may be able to utilize having column inside the axis class more, beyond formatting

grapher/scatterCharts/ScatterPlotChart.tsx
danielgavrilov: Yeah you're right, seems like all these transforms should actually be done on the table. I think it might not be too much work to rewrite them. Will try it out

PR #715, opened 2020-11-20
Color scales on columns

breck7:
Working toward pulling colorscales into Explorer builder to finish Covid parity.

Screen Shot 2020-11-19 at 9 25 05 PM

Want to rev it a little so just putting it as a draft, but stable enough if we want to merge and not diverge.

### DIFF
--- a/adminSite/client/EditorColorScaleSection.tsx
+++ b/adminSite/client/EditorColorScaleSection.tsx
@@ -27,10 +27,8 @@ import {
     BindString,
 } from "./Forms"
 import { ColorSchemeOption, ColorSchemeDropdown } from "./ColorSchemeDropdown"
-import {
-    BinningStrategy,
-    binningStrategyLabels,
-} from "grapher/color/BinningStrategies"
+import { binningStrategyLabels } from "grapher/color/BinningStrategies"
+import { BinningStrategy } from "grapher/color/BinningStrategy"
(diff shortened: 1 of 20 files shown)


danielgavrilov:
🙌 Awesome!

I think we also want:
- `equalSizeBins` boolean to disable visual scaling, very very often used by authors
- `customNumericMinValue` to control the start of the first bin. Useful if there might be any negatives in the data that we want to discard, it otherwise defaults to `Math.min(0, minimumNumericValue)` currently.

The binning strategy is used less, it could be automatically set to `"manual"` when authors define custom bins in the DSL? Otherwise they have to always remember to set it themselves.

But if they happen to want other binning strategies, then they'd also want to control the number of bins (`binningStrategyBinCount`) because it's the param you most likely want to tweak it if you use a non-manual binning strategy. (But I think they don't need it. I think all charts have manual bins.)

breck7:
just an update on this: I'm gonna try a full table version, instead of cramming the scheme into one cell. it feels wrong, so will see how it feels as a table.

danielgavrilov:
Closing since it's been (partially) implemented in #793. 

#793 doesn't include [Dev: experimental way to poach color scales from existing graphers](https://github.com/owid/owid-grapher/pull/715/commits/f8bca6641a375a00f3beaef352c530673a171127), but for now using Graphers in Explorers is discouraged, and we might merge Grapher+Explorer to become one thing.

PR #719, opened 2020-11-24
Fix: replace non-numeric cells with errors for charts that don't support them

danielgavrilov:
Notion issue: https://www.notion.so/owid/StackedArea-LineChart-DiscreteBar-crash-when-data-contains-non-numeric-vaue-8b1b59ed3ae74d0fa40b004590c694cc

### DIFF
--- a/coreTable/CoreTable.ts
+++ b/coreTable/CoreTable.ts
@@ -11,6 +11,7 @@ import {
     uniqBy,
     intersectionOfSets,
     isPresent,
+    isNumber,
 } from "grapher/utils/Util"
 import { queryParamsToStr } from "utils/client/url"
 import { CoreColumn, ColumnTypeMap, MissingColumn } from "./CoreTableColumns"
@@ -48,7 +49,6 @@ import {
     autodetectColumnDefs,
     reverseColumnStore,
     renameColumnStore,
-    replaceNonPositives,
     replaceRandomCellsInColumnStore,
     getDropIndexes,
     parseDelimited,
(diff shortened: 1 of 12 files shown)


breck7:
code LGM but could we have a test case and/or stories? I'm having trouble visualizing how we'd hit that scenario where we get into this situation.

also, is this a temporary thing, where we only need this until we clean up our datasets? maybe add a "// todo: remove once we eliminate mixed type columns in our datasets", to the method? OR do we want to always have charts support passing in poorly typed data? If the latter case I definitely think we want stories for examples

danielgavrilov:
> could we have a test case and/or stories?

Yeah we should, I was trying to round some corners

> is this a temporary thing, where we only need this until we clean up our datasets?

Yes, temporary until we clean our datasets but I don't think we'll ever really do that, so most likely a more permanent version of temporary.

I wrote more in the Notion issue:

> Ok [Find and correct all variables which are used in charts & contain non-numeric data](https://www.notion.so/Find-and-correct-all-variables-which-are-used-in-charts-contain-non-numeric-data-4343e74d516c4f75802887819066acf8) is very tedious, and at the end of it we would've cleaned up the database for the variables that happen to be used in charts but not others.
> 
> It's a pain for us to clean the data, and it's a pain for authors to go back and reupload the data. I'm more in favour of adding a transform to replace string cells with empty cells in LineChart, StackedArea & DiscreteBar, which would match the behaviour on master.
> 
> Charts to test with: https://docs.google.com/spreadsheets/d/1JytPrwBzqGDjiyLBg5uQrtGa4xi-9X9vui7XDlxHBWI/edit#gid=1086325742
> 
> On some LineCharts, the chart crashes with non-numeric values, on others it's plotted as 0. The issue is that it's very easy to miss this, especially since the data upload process doesn't have checks against it.
> 
> So yes we should clean the data, but thoroughly, and also catch these issues at the upload step. That is all too much work, so matching the behaviour to master is probably better for now

breck7:
Makes a ton of sense. Maybe just one test case, to repro the behavior, and a todo comment?

breck7:
it looks beautiful

danielgavrilov:
Tests are great to have, as long as I don't have to write them

PR #720, opened 2020-11-24
Show helpful message on single-year ScatterPlot avg annual change

danielgavrilov:
Before:

Screenshot 2020-11-24 at 22 27 47

After:

Screenshot 2020-11-24 at 22 27 31

### DIFF
--- a/grapher/noDataModal/NoDataModal.scss
+++ b/grapher/noDataModal/NoDataModal.scss
@@ -4,6 +4,7 @@
     display: flex;
     flex-direction: column;
     justify-content: center;
+    height: 100%;
 
     .action {
         background-color: $controls-color;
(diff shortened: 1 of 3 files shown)


breck7:
nice! love how we are getting nicer error messages

shaahmad:
This is great! Is it possible to vertically center-align?

MarcelGerber (approved):
Nice!
Can maybe change the text to something like
```
Please select two points on the timeline below to display the average annual change between these points
```
to make it more clear _why_ two points have to be selected? not sure.

danielgavrilov:
> Can maybe change the text to something like

Not sure – shorter is more likely to be read, longer makes it clearer what the chart shows.

Should we A/B test it?

danielgavrilov:
^ I'm joking, I'm not sure that it's even worth the brain cycles 🙂

I'll merge, feel free to tweak it

PR #721, opened 2020-11-25
Fix: custom dimension colors in admin

shaahmad:
Dimension-level colors can now be edited in the admin. Can't merge until I finish column re-ordering early tomorrow.
image


shaahmad:
Turns out column reordering wasn't the problem, the save/restore buttons just weren't working. Fixed now.

PR #722, opened 2020-11-26
Embed cleanup

breck7:
Just trying to cleanup the "isEmbed" style properties a bit, and remove any general css classes like "chart" and "iframe". Ideally we can get our SCSS and TypeScript linking to the same identifiers at some point.


breck7:
To provide a bit more context: I was getting "embed" vs "iframe" confused, as well as a few others like "isExporting", so attempted to clarify those more. I wouldn't be surprised if I broken one or 2 display things, but looks okay to me.

PR #731, opened 2020-12-09
Column Reordering in Admin "Basic" Tab

shaahmad:
ezgif-1-4c85927366b2


MarcelGerber:
Great, I like it!

Maybe it would be more obvious that dimensions can be dragged if we include an arrow as drag handle, like we do in the "Data" tab? (can also do this later on)
image

Also, all these different variables (`dimensionsOrderedAsInPersistedSelection`, `dimensionsOrderedAsDisplayed`, `legacySelectionOrderedAsDisplayed`) are quite involved, can we maybe trim them down in some way? If not, then that's also alright with me.

shaahmad:
@MarcelGerber I added the arrows and fixed the layout too. Thought hard about renaming the functions/variables, but couldn't think of a way yet that would simplify without reducing the amount of information in the name. I'll merge as-is and keep thinking about it.

PR #738, opened 2020-12-12
Dev: make deploy queue server a class

breck7:
Just going through all the files like "ExplorerBaker" and "GitCMSServer" and this one and cleaning things up a little bit my moving from singletons/bags of functions to a class and instance. This was a In general this one makes testing a bit easier and can utilize instance members for paths and reduce globals and params.


package.json
```diff
@@ -5,7 +5,7 @@
         "node": ">=12.13.1"
     },
     "scripts": {
-        "buildAndDeploySite": "node ./itsJustJavascript/baker/bakeAndDeploySite.js",
+        "buildAndDeploySite": "node ./itsJustJavascript/baker/buildAndDeploySite.js",
```
breck7, line 8: oh yeah and fixed a typo.

PR #753, opened 2020-12-17
Lists in Grid Langs

breck7:
This is a minor tweak that adds a new Grammar flag "isHorizontalList". If you set that to true on a cell definition, then that cell takes a list of strings, not just one. This fixes the ugly hack where authors had to do "Canada~France" for default selection, instead of "Canada\tFrance". It also will allow better autocomplete, type checking, and work better with Patch.

Before:
Screen Shot 2020-12-16 at 3 36 44 PM

After:
Screen Shot 2020-12-16 at 3 36 12 PM

I also renamed a few existing keywords and added some more comments to clean up as I go.


breck7:
wrong branch

PR #774, opened 2021-01-05
Remove Timeline#snapTimes()

danielgavrilov:
I _think_ it was introduced when we had the timeline dragging freely without snapping. Now that it always snaps to available years, it's not necessary.

The fact that it snaps on `≤1` also causes this bug where you can't select a range that's only 1 time step apart: 

https://user-images.githubusercontent.com/1308115/103649389-48acd480-4f56-11eb-8095-91597c521afc.mov

Reported here: https://owid.slack.com/archives/C46U9LXRR/p1609849260000400


breck7 (approved):
nice!

PR #778, opened 2021-01-13
Dev: don't use autorun on choice change

shaahmad:
[Continuing](https://github.com/owid/owid-grapher/pull/544#discussion_r470149021) my crusade against `reaction` and `autorun`.

Should help avoid some bugs in the future.


breck7 (approved):
Agreed!

PR #779, opened 2021-01-13
fix: each grapher gets its own query params

shaahmad:
Should fix [After changing an explorer to log scale, url doesn't revert when you change it back to linear](url) by giving each grapher its own query parameters.

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -28,6 +28,7 @@ import {
     SlideShowManager,
 } from "../grapher/slideshowController/SlideShowController"
 import {
+    ExplorerChoice,
     ExplorerContainerId,
     EXPLORERS_PREVIEW_ROUTE,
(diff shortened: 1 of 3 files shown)


explorer/Explorer.tsx
```diff
@@ -196,6 +187,11 @@ export class Explorer
         )
     }
 
+    private persistedQueryParamsByGrapher: Map<
```
breck7, line 190: oooh I like this idea. if grapher switching ends up being a long term thing then doing this right and minimizing url noise could be worth it. is there a way we can add tests? this code has been terrible to work with (my fault) and if we could get tests in there could be good.

shaahmad: yes, tests are included on the new PR to be merged with Next!

shaahmad: jk turns out I can change the target of this PR

PR #786, opened 2021-01-21
Generalize redirects for all explorers & implement LegacyCovidExplorer redirects

danielgavrilov:
Redirects are no longer Grapher-specific, we can add redirects from any path on the site (this is how `/coronavirus-data-explorer` is redirected).

The redirects are all in a CSV (inside a .ts file, could be extracted in future), and one of the columns is `migrationId`. Currently the only `migrationId` we have is `legacyToGridCovidExplorer` (see variable `explorerUrlMigrationsById`). ~Will implement CO₂ and Energy explorer redirects next (I haven't looked at them at all, I don't even know if they need redirects, just assuming).~

We could also use this migrations approach to implement backwards-compatible URLs for future Explorer grid programs.

I find it hard to name these things, currently using:

- "migration" to refer to a URL transform that happens client-side. This is usually just query string manipulation.
- "redirect" to refer to a URL transform that happens server-side. This can include a path rewrite.

The distinction is so that the (potentially) long list of redirects doesn't get serialized into the client-side bundle. It instead gets baked into the page that redirects: https://github.com/owid/owid-grapher/pull/786/files#diff-61dd6f57638281a44abcd9a7642d50bab29a2609ba7675c321ce1a7ff73ad811R79-R81

### Todo

- [x] Add more tests

Will do separately:

- Implement CO₂ explorer redirects 
- Implement Energy explorer redirects


danielgavrilov:
Decided to do Energy & CO₂ explorer redirects separately ([Notion issue](https://www.notion.so/owid/Create-redirect-map-for-CO-Energy-explorers-9fed5db5881c4205bcb9ea26997e79ed)) and get this out quickly.

Most of the work on the explorer redirects is figuring out which params need to be migrated and how, rather than coding.

urls/readme.md
```diff
@@ -0,0 +1,76 @@
+# `Url` a utility class to parse & manupulate URLs
```
breck7, line 1: Nice! I really like this method of working.

Even though there are loads of url libs out
there, by rolling your own here as a separate
project, you can really solve our specific needs,
understand the problem domain better, and then
if you come across an external lib that ends up
being good enough, dropping this sub project
and hooking up that lib is a breeze.

urls/readme.md
```diff
@@ -0,0 +1,76 @@
...
+We created this utility because most existing URL parsing libraries assume you pass a whole URL – in many cases in our code, we need to parse query strings only.
+
+Keeping query strings and full URLs in one class makes it easier to write URL migrations.
+
+## Creating a `Url`
+
+### `Url.fromURL(url: string)`
+
+```ts
+let url = Url.fromURL(
```
breck7, line 14: Very simple and clean API!

urls/readme.md
```diff
@@ -0,0 +1,76 @@
...
+console.log(url.hash) // ""
+
+console.log(url.queryParams) // { stackMode: "relative", country: "USA~DEU" }
+```
+
+## Updating a `Url`
+
+### `Url.update(props)`
+
+A **`Url`** object is **immutable** – updating it returns a different `Url` object, the original is left unchanged.
```
breck7, line 43: Love it! My next child will be named Immutable

danielgavrilov: You mean `ImmutableNotCodeButAHumanName`

urls/readme.md
danielgavrilov: Yeah definitely! I was wondering if it should go in `clientUtils` but I liked that `Patch` is its own thing and has no OWID-specific functionality.

PR #788, opened 2021-01-26
Recover Explorer selection from URL params on load

danielgavrilov:
Notion issue: https://www.notion.so/owid/selection-param-not-correctly-decoded-by-Explorer-518f9a5fdf2e46d7a8ca20146eb06321

@shaahmad Does that the one-line change for `persistedParams` look reasonable? Not sure if it might break something else, it was a shot in the dark 🎯

### DIFF
--- a/explorer/Explorer.jsdom.test.tsx
+++ b/explorer/Explorer.jsdom.test.tsx
@@ -1,4 +1,4 @@
-#! /usr/bin/env jest
+#! yarn testJest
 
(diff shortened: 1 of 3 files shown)


danielgavrilov:
Will do an optimistic merge, but feel free to change on `next`

shaahmad:
Cleaned up the code a bit in https://github.com/owid/owid-grapher/pull/790 so it's less of a shot in the dark.

PR #790, opened 2021-01-27
Dev: refactor reactToUserSelection method

shaahmad:
While reviewing https://github.com/owid/owid-grapher/pull/788, I realized some of the code in this method worked by accident and not design. This should partially fix that.


breck7:
What do you think about:
https://github.com/owid/owid-grapher/tree/switch-from-mobx-to-command-pattern

Should we try and merge that command pattern in?

PR #800, opened 2021-02-05
Global Selection Refactor + Fixes

shaahmad:
This PR makes GlobalSelection more functional and testable, and fixes a variety of bugs:
- lazy-loaded charts now have the proper selection
- selection is saved in URL
- prominent links now use selection again

Architecturally, the `MultiEmbedderSingleton` object now maintains a selection, and passes it down as a prop to Explorers/Graphers as well the `GlobalEntitySelector`. That way, Graphers/Explorers don't know anything about a GlobalSelection, which is a purely `site` concept.

The MultiEmbedder also maintains a list of graphers/explorers to update, so that we don't need the global variable `GlobalEntityRegistry`.


site/multiembedder/MultiEmbedder.tsx
```diff
@@ -118,13 +140,18 @@ const figuresFromDOM = (
 export const shouldProgressiveEmbed = () =>
     !isMobile() ||
     window.screen.width > 680 ||
-    pageContainsGlobalEntityControl()
+    pageContainsGlobalEntitySelector()
+
+const pageContainsGlobalEntitySelector = () =>
+    document.querySelector(`[${GLOBAL_ENTITY_SELECTOR_DATA_ATTR}]`) !== null
```
MarcelGerber, line 146: what about this?
```suggestion
    globalEntitySelectorSelector() !== null
```

site/multiembedder/MultiEmbedder.tsx
```diff
@@ -118,13 +140,18 @@ const figuresFromDOM = (
...
 
-const pageContainsGlobalEntityControl = () =>
-    document.querySelector(`[${GLOBAL_ENTITY_CONTROL_DATA_ATTR}]`) !== null
+const globalEntitySelectorSelector = () =>
```
MarcelGerber, line 148:
```suggestion
const globalEntitySelectorElement = () =>
```
may be a better name?

site/multiembedder/MultiEmbedder.tsx
```diff
@@ -200,6 +232,29 @@ class MultiEmbedder {
...
     }
+
+    setUpGlobalEntitySelectorForEmbeds() {
+        const element = globalEntitySelectorSelector()
+        if (!element) return
+
+        const selectionParam = getWindowQueryParams().selection
+        const defaultCountry = element.getAttribute(
+            GLOBAL_ENTITY_SELECTOR_DEFAULT_COUNTRY
+        )
```
MarcelGerber, line 243: We should probably have some way to pass _multiple_ countries as the default selection, no?
[Charlie requested this just last week.](https://owid.slack.com/archives/C46U9LXRR/p1612279381009700) (related commit on `master`: https://github.com/owid/owid-grapher/commit/c9916e2bf36ee40b6696384b5214f48b9a2fa558)

MarcelGerber (approved):
Nicely done! I've had a few comments here and there, but overall this looks good.
I also noticed some related issues when playing around with it, which I will address in comments below.

MarcelGerber:
I've deployed this PR to _tufte_, so you can play around with it there.

So, one big problem here is that by default, **all charts on the page start out with an empty selection** (see video!).

https://user-images.githubusercontent.com/2641501/107220362-8c41a500-6a12-11eb-8687-afc6740d5648.mp4

Relatedly, when working on this, you should also make sure that when a user un-selects all countries from the global selection, the charts should revert back to their "default" view.

MarcelGerber:
Another problem, and this will probably spin out into another issue but I want to at least mention it here, is that country pages are generated with entity codes (e.g. USA, DEU) rather than the full country names (e.g. United States, Germany) at this point, and we need to switch that over at some point.
Currently, the selection in country profiles is wrong and all charts are blank.

site/owid.entry.ts
```diff
@@ -58,9 +58,13 @@ window.runSiteFooterScripts = () => {
     runCookiePreferencesManager()
     runCovid()
     runFootnotes()
+    renderProminentLink()
```
danielgavrilov, line 61: Seems like both branches below call this function anyway?

danielgavrilov:
> [...] country pages are generated with entity codes (e.g. USA, DEU) rather than the full country names (e.g. United States, Germany) at this point, and we need to switch that over at some point.

Yes, and there would be external links with the old `country=[code]` query param, so we might need to handle the legacy param? Agree this can go on a separate issue.

explorer/Explorer.tsx
```diff
@@ -191,15 +192,21 @@ export class Explorer
     componentDidMount() {
         this.setGrapher(this.grapherRef!.current!)
         this.updateGrapherFromExplorer()
-        this.grapher?.populateFromQueryParams(this.initialPatchObject)
+
+        const initialPatchObject = {
+            ...this.initialPatchObject,
+            selection: this.props.selection?.hasSelection
+                ? this.props.selection.asParam
+                : this.initialPatchObject.selection,
```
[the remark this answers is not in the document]
shaahmad: I do this to have a const variable I can use in-scope in line 202 regardless of the conditional. Is there a way to do that with an if-statement?

grapher/controls/globalEntitySelector/GlobalEntitySelector.tsx
```diff
@@ -376,23 +378,20 @@ export class GlobalEntityControl extends React.Component<{
             </div>
         )
     }
-
-    static singleton() {
-        return new GlobalEntityControl({}).selection
-    }
 }
 
 // todo: add analytics back
```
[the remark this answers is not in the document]
shaahmad: nope, good catch! deleted

explorer/Explorer.tsx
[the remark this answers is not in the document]
MarcelGerber: Even if the object is const, its entries are not. So you could still do this in an if statement without losing out on `const` ;)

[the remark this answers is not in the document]
danielgavrilov: I don't find it weird actually – it's kind of nice that it doesn't mutate the original prop but creates a copy?

I've used this pattern too: https://github.com/owid/owid-grapher/blob/2b1d8d99a9a0af4bf1ef50a99c6d8ac3416ece5f/explorer/urlMigrations/ExplorerUrlMigrationUtils.ts#L10-L19

site/multiembedder/MultiEmbedder.tsx
shaahmad: done!

site/multiembedder/MultiEmbedder.tsx
shaahmad: ah yes, that was a temp placeholder -- this is better

site/owid.entry.ts
shaahmad: perfect -- fixed

explorer/Explorer.tsx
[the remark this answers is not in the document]
shaahmad: Yep I quite liked [how it looked](https://github.com/owid/owid-grapher/pull/800/commits/5383375e0c00138fc2762134f3a0734eb7a217e0) but reverted it as it was mutating the original prop.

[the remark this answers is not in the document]
MarcelGerber: ah, right! didn't think about that. then let's just leave it as-is.

site/multiembedder/MultiEmbedder.tsx
shaahmad: done

shaahmad:
> Relatedly, when working on this, you should also make sure that when a user un-selects all countries from the global selection, the charts should revert back to their "default" view.

This is the behavior on prod, and I raised concerns about [something](https://owid.slack.com/archives/C0149NKLZAM/p1600297716057400) very similar some months ago, but people disagreed. Given our policy of keeping to prod, I'll leave it as-is for now, and we can address it in a small PR later.

shaahmad:
> Relatedly, when working on this, you should also make sure that when a user un-selects all countries from the global selection, the charts should revert back to their "default" view.

This is the behavior on prod, and I raised concerns about [something](https://owid.slack.com/archives/C0149NKLZAM/p1600297716057400) very similar some months ago, but people disagreed. Given our policy of keeping to prod, I'll leave it as-is for now, and we can address it in a small PR later.

MarcelGerber:
> This is the behavior on prod, and I raised concerns about [something](https://owid.slack.com/archives/C0149NKLZAM/p1600297716057400) very similar some months ago, but people disagreed. Given our policy of keeping to prod, I'll leave it as-is for now, and we can address it in a small PR later.

I'm not quite sure we're talking about the same thing.
To clarify: What I mean is that, in the video below, the charts should show their _default_ configuration both at the start (where no global selection is present) _and at the end_ (where the global selection was completely cleared out).

https://user-images.githubusercontent.com/2641501/107426672-d0bb6680-6b20-11eb-935e-b9fc10732211.mp4

This is not currently happening on tufte, but that might very well be because showing the default config at the start doesn't work yet. But once it works at the start, let's also make sure it works at the end :)

shaahmad:
Right, totally understand! The starting bug is related to the country-code vs. entity name thing, for which I'll have a fix soon. Definitely, all charts should start populated with a selection.

But the end-stage behavior on prod is that when the global entity selection is emptied, all the _Explorers_ on the page have no selection, not even the selection specified in the embed URL. On a closer look, Graphers maintain their default selection. Now I understand where we differed -- I've been looking at /covid-cases, which is mostly Explorers, whereas /covid-vaccinations is all Graphers.

In the Slack thread I shared, I had asked whether emptying an individual grapher selection should take you to the default selection or to the "no matching data/no country selected" page, and most people seemed to prefer the no country selected page. It's similar here on pages like /covid-cases, where you start with a global selection. When emptied, according to the preferences from that Slack thread, all charts should show "no country selected", not a default selection. In addition, when the user adds a global selection and then empties it, we should probably show "no country selected". Of course, if the global selection starts empty, defaults should be shown.

shaahmad:
Backwards compatibility with old URLs is addressed. Thank you for the thorough reviews @danielgavrilov @MarcelGerber!

PR #803, opened 2021-02-08
Update README to use NVM

shaahmad:
NVM seems to be the preferred way to install Node, as opposed to Brew. e.g.
https://www.reddit.com/r/node/comments/a9mwti/should_node_be_installed_with_homebrew_or/
https://jamesauble.medium.com/install-nvm-on-mac-with-brew-adb921fb92cc

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,26 +24,40 @@ The owid-grapher visualization frontend code can run isomorphically under node t
     ```sh
     brew tap homebrew/services
     ```
+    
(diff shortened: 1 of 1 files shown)


mlbrgl (approved):
yes, given we already have NVM as a rather strongly recommended dependency, that makes sense.

danielgavrilov:
:+1: Also at some point maybe we should update our Node version... 😐

PR #815, opened 2021-03-03
fix: add types to settings

danielgavrilov:
Seems like a good idea?


danielgavrilov:
Closing in favour of #817 which also adds types.

PR #820, opened 2021-03-08
Serialize all choice params in Explorer URL

danielgavrilov:
Notion issue: [Include all choice params in Explorer URL](https://www.notion.so/Include-all-choice-params-in-Explorer-URL-87b675a53a01460899edcc61b149b99a).

> Is it a good idea to include all Explorer params (metric, interval, align outbreaks, etc) in the URL every time, regardless whether they are set to their default value?
> That way, if we change the default, which is highly likely, we won’t end up with unexpected views.
> This is the behaviour of the current (live) CovidExplorer.
> On the Next explorer, we omit params that are set to their defaults. This can be problematic if we change the default view (which we do every now and then).

> An example: I want to embed a cases, daily, per million chart.
> Because all these are defaults, they’re not in the URL. If we change the defaults, you will get a completely different chart in your embed.

### DIFF
--- a/explorer/Explorer.jsdom.test.tsx
+++ b/explorer/Explorer.jsdom.test.tsx
@@ -39,4 +39,16 @@ describe(Explorer, () => {
         const explorer = element.instance() as Explorer
         expect(explorer.selection.selectedEntityNames).toEqual(["Ireland"])
     })
+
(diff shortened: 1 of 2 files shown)


MarcelGerber (approved):
Looks good to me 👍

PR #864, opened 2021-04-20
fix: CoreTable#complete() does not append missing rows

danielgavrilov:
- Adds a new `comparisonUrl` param to embed test pages, so for example you can compare _local_ vs _jefferson_, instead of always _local_ vs _live_. Example: http://localhost:3030/admin/test/embeds?ids=306&comparisonUrl=https://jefferson-owid.netlify.app

- Fixes `complete()` being a noop in certain cases. Added a ScatterPlot test case that was affected by this. 

- There was a map test case that I got wrong some time ago, now fixed

- Makes performance on charts that apply tolerance much worse

### DIFF
--- a/adminSiteServer/testPageRouter.tsx
+++ b/adminSiteServer/testPageRouter.tsx
@@ -12,14 +12,18 @@ import {
     BAKED_GRAPHER_URL,
     BAKED_BASE_URL,
 } from "../settings/serverSettings"
-import * as querystring from "querystring"
-import * as lodash from "lodash"
-import * as url from "url"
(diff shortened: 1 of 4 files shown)


danielgavrilov:
With this change, https://github.com/owid/owid-grapher/pull/830 is almost fully reverted I think. So we are back to slower (but more correct) charts.

MarcelGerber (commented):
I finally found the time to review this all - looks good!

Great idea with the comparison URL, that can come in very handy at times 👍 

> Makes performance on charts that apply tolerance much worse
Is that only due to the missing `parent`, or is there also other stuff going on that impacts performance? And how much worse is it?

PR #873, opened 2021-05-06
Fix line charts with projection

danyx23:
This fixes the issue described in https://www.notion.so/owid/Broken-chart-ea6e5f3f9dcf49d4859b74dad585d6ca). The problem is that autoDetectSeriesStrategy checks if there is more than one variable involved and if so it sets the strategy to column. When there is a projection we have two variables but logically this is shown as one and entity is still the right strategy.

This is my first code commit so please let me know extensively what you think. It might also be that the entire approach is wrong, but this is what I came up with to resolve the issue :).

### DIFF
--- a/grapher/lineCharts/LineChart.tsx
+++ b/grapher/lineCharts/LineChart.tsx
@@ -49,6 +49,7 @@ import {
     autoDetectSeriesStrategy,
     autoDetectYColumnSlugs,
     getDefaultFailMessage,
+    getSeriesKey,
     makeClipPath,
     makeSelectionArray,
 } from "../chart/ChartUtils"
@@ -165,9 +166,7 @@ class Lines extends React.Component<LinesProps> {
         return this.backgroundLines.map((series, index) => (
             <g key={index}>
                 <path
-                    key={`${series.seriesName}-${series.color}-${
-                        series.isProjection ? "projection" : ""
-                    }-line`}
+                    key={getSeriesKey(series, "line")}
                     strokeLinecap="round"
                     stroke="#ddd"
                     d={pointsToPath(
@@ -380,11 +379,7 @@ export class LineChart
                                 : series.color
                             return (
                                 <tr
-                                    key={`${series.seriesName}-${
-                                        series.color
-                                    }-${
-                                        series.isProjection ? "projection" : ""
-                                    }`}
+                                    key={getSeriesKey(series)}
                                     style={{ color: textColor }}
                                 >
                                     <td>
@@ -565,11 +560,7 @@ export class LineChart
 
                             return (
                                 <circle
-                                    key={`${series.seriesName}-${
-                                        series.color
-                                    }-${
-                                        series.isProjection ? "projection" : ""
-                                    }`}
+                                    key={getSeriesKey(series)}
                                     cx={horizontalAxis.place(value.x)}
                                     cy={verticalAxis.place(value.y)}
                                     r={4}
--- a/grapher/lineCharts/LineChart.tsx
+++ b/grapher/lineCharts/LineChart.tsx
@@ -11,7 +11,6 @@ import {
     exposeInstanceOnWindow,
 } from "../../clientUtils/Util"
 import { computed, action, observable } from "mobx"
-import { some } from "lodash"
 import { observer } from "mobx-react"
 import { select } from "d3-selection"
 import { easeLinear } from "d3-ease"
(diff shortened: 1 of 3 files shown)


grapher/lineCharts/LineChart.tsx
```diff
@@ -554,7 +561,11 @@ export class LineChart
 
                             return (
                                 <circle
-                                    key={`${series.seriesName}-${series.color}`}
+                                    key={`${series.seriesName}-${
+                                        series.color
+                                    }-${
+                                        series.isProjection ? "projection" : ""
+                                    }`}
```
danielgavrilov, line 568: Rather than appending `"projection"` multiple times, it maybe could be added to `seriesName` when the series are constructed? (Not sure if this might have some unintended effects, I _think_ not.)

https://github.com/owid/owid-grapher/blob/1c87608d84528a4788536b03039478483df4c2fb/grapher/lineCharts/LineChartUtils.ts#L13-L18

grapher/lineCharts/LineChart.tsx
```diff
@@ -622,7 +633,13 @@ export class LineChart
     }
 
     @computed get seriesStrategy(): SeriesStrategy {
-        return autoDetectSeriesStrategy(this.manager)
+        const hasNormalAndProjectedSeries =
+            some(this.yColumns, (col) => col.isProjection) &&
+            some(this.yColumns, (col) => !col.isProjection)
```
danielgavrilov, line 638: You can safely use `Array#some` directly without using lodash, [a very tiny amount of browsers don't support it](https://caniuse.com/?search=Some).

(Related to that, we are most limited in the browsers we support by our use of Mobx 5, which requires support for [Proxy](https://caniuse.com/?search=Proxy).)

danielgavrilov (commented):
Amazing that you figured this out so quickly! This is a problem we have introduced in our big refactoring project. 

The assumption that >1 column => column series might go a bit deeper than the line chart, I think we need to review if it affects other charts (e.g. DiscreteBar). 

(The projections themselves are also quite ad-hoc, potentially this is a good excuse to check if we can refactor some parts to find a better implementation for them.)

danyx23:
@danielgavrilov this is ready to merge from my side. Aside from the intended change this also slightly changes the way some line charts are labeled. Some of these are clearly better (e.g. the label changes from "Something -" to "Something") but for around 10 charts we now prepend the variable name where it wasn't used before - e.g. the [Peaking farmland](https://ourworldindata.org/grapher/peaking-farmland-global-arable-and-permanent-crop-area-and-ausubel-2013-projections-of-peak-farmland?country=~OWID_WRL) chart is currently labelled "Ausbel et al (2013) Projection" and would now be labelled "World - Ausbel et al (2013) Projection". I think that this is fine but please let me know what you think. I can also collect all charts that are different now and we go through them one by one and see if this warrants adding another flag on the charter admin or similar

grapher/lineCharts/LineChartUtils.ts
```diff
@@ -10,12 +10,20 @@ export const columnToLineChartSeriesArray = (
     const { isProjection, owidRowsByEntityName } = col
     const entityNames = Array.from(owidRowsByEntityName.keys())
     return entityNames.map((entityName) => {
-        const seriesName =
-            seriesStrategy === SeriesStrategy.entity
-                ? entityName
-                : canSelectMultipleEntities
-                ? `${entityName} - ${col.displayName}`
-                : col.displayName
+        let seriesName = ""
```
danielgavrilov, line 13: nit: if you use `let seriesName: string` without assigning it, TypeScript will check that the variable is being assigned in all branches

grapher/lineCharts/LineChartUtils.ts
```diff
@@ -10,12 +10,20 @@ export const columnToLineChartSeriesArray = (
...
+        if (seriesStrategy === SeriesStrategy.entity) {
+            if (col.isProjection && col.displayName) {
+                seriesName = `${entityName} - ${col.displayName}`
```
danielgavrilov, line 16: Do you have an example chart where this applies? Can we get away with not appending `${col.displayName}`?

danielgavrilov: I found these: http://localhost:3030/admin/test/embeds?ids=682,688,958,1006,1133,1193,1194,1480,2822,3265

Did you make this change with the intention of matching previous behaviour (currently on live)?

I think it might be better not to keep the "- Projection" suffix when `SeriesStrategy.entity`? (I think this is how it was before Next)

grapher/lineCharts/LineChartUtils.ts
danyx23: Ah nice, I didn't know this - ML languages don't let you do this :)

danielgavrilov (approved):

PR #883, opened 2021-05-12
style(grapher): Add return type annotations

danyx23:
This PR re-enables several TS lint rules project wide. 

The two warnings that create the most noise are explicit function return type and explicit module boundary types. These are disabled on the project level for now.

For three folders these two warnings have been enabled (with a per-directory .eslint.yaml file): db, clientUtils and coreTable. In these, almost all return type annotations have been added (sans a few that need new interface types).

As is, running eslint as a command line app on the entire grapher directory sans wordpress (this is not directly included and thus gives an error when run this way) leads to 260 warnings of various minor things scattered in the codebase. I think these could be fixed slowly as we go.

There are two further big steps that I would like to do after this PR: 
1. add return type annotations for all the other folders, probably in 2-3 additional PRs (~3500 lines to annotate)
2. turn on no-explicit-any and fix all occurences (replacing any with at least unknown or more specific types where possible)
3. maybe turn on no-non-null-assertions as well as this can easily be a cause for subtle bugs but is of course sometimes justified as the TS codebase can't be fully annotated with correct nullability tracking

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,30 +20,29 @@ module.exports = {
     ],
     rules: {
         // These rules are only preliminary to ease the migration from tslint, many of them might make sense to have enabled!
-        "@typescript-eslint/camelcase": "off",
-        "@typescript-eslint/explicit-function-return-type": "off",
-        "@typescript-eslint/explicit-module-boundary-types": "off",
-        "@typescript-eslint/no-empty-function": "off",
+        "@typescript-eslint/explicit-function-return-type": "off", // This rule is enabled on a folder by folder basis and should be enabled project wide soon
+        "@typescript-eslint/explicit-module-boundary-types": "off", // This rule is enabled on a folder by folder basis and should be enabled project wide soon
+        "@typescript-eslint/no-empty-function": "warn",
         "@typescript-eslint/no-empty-interface": "warn",
-        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/no-explicit-any": "off", // TODO: enable this rule
         "@typescript-eslint/no-for-in-array": "warn",
-        "@typescript-eslint/no-inferrable-types": "off",
-        "@typescript-eslint/no-namespace": "off",
-        "@typescript-eslint/no-non-null-assertion": "off",
-        "@typescript-eslint/no-this-alias": "off",
+        "@typescript-eslint/no-inferrable-types": "off", // If this is on the ESLint automatic fixing removes type annotations in assignments like "somenumber: number = 4"
+        "@typescript-eslint/no-namespace": "warn",
+        "@typescript-eslint/no-non-null-assertion": "off", // TODO: enable this rule
+        "@typescript-eslint/no-this-alias": "warn",
         "@typescript-eslint/no-unused-vars": "warn",
-        "@typescript-eslint/no-use-before-define": "off",
-        "@typescript-eslint/no-var-requires": "off",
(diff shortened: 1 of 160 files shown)


danyx23:
I just saw that some of the type annotations create a problem when running tsc - will fix that now

adminSiteClient/AdminLayout.tsx
```diff
@@ -22,32 +22,32 @@ export class AdminLayout extends React.Component<{
     static contextType = AdminAppContext
     context!: AdminAppContextType
 
-    @observable private showFAQ: boolean = false
-    @observable private showSidebar: boolean = false
+    @observable private showFAQ = false
+    @observable private showSidebar = false
```
danielgavrilov, line 26: Why are these no longer explicit?

.eslintrc.js
```diff
@@ -20,30 +20,30 @@ module.exports = {
     ],
     rules: {
         // These rules are only preliminary to ease the migration from tslint, many of them might make sense to have enabled!
-        "@typescript-eslint/camelcase": "off",
+        // "@typescript-eslint/camelcase": "warn",
```
danielgavrilov, line 23: Is this deprecated now? https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md

Should we outright remove it?

clientUtils/PromiseSwitcher.test.ts
```diff
@@ -1,11 +1,11 @@
 import { PromiseSwitcher } from "./PromiseSwitcher"
 
-const delayResolve = (result: any, ms: number = 10) =>
+const delayResolve = (result: any, ms = 10): Promise<void> =>
     new Promise((resolve) => {
         setTimeout(() => resolve(result), ms)
     })
 
-const delayReject = (error: any, ms: number = 10) =>
+const delayReject = (error: any, ms = 10): Promise<void> =>
```
danielgavrilov, line 8: Like class properties, seems parameters with default values have the explicit types removed? Automated tool or intentional?

adminSiteClient/AdminLayout.tsx
danyx23: This is an artifact of an attempt at automating this - I removed almost all of these again but some where still in there and if the file had some otherwise beneficial changes I left them in since I will tackle them soon anyhow

clientUtils/PromiseSwitcher.test.ts
danyx23: Ah this one I should have seen - will fix it

.eslintrc.js
danyx23: Let's remove it. We can consider doing another round of adding style rules later.

adminSiteClient/AdminLayout.tsx
danielgavrilov: :+1: Did you miss these on the 2nd pass or intending to fix later?

adminSiteClient/ChartEditor.ts
```diff
@@ -83,7 +83,7 @@ export class ChartEditor {
     @observable.ref tab: EditorTab = "basic"
     @observable.ref errorMessage?: { title: string; content: string }
     @observable.ref previewMode: "mobile" | "desktop"
-    @observable.ref savedGrapherJson: string = ""
+    @observable.ref savedGrapherJson = ""
```
danielgavrilov, line 86: this one too?

adminSiteClient/AdminLayout.tsx
danielgavrilov: I can see more of them – is it fine if I push a commit to revert the removal of types from initialized properties? Since I'm reviewing it all, it's fairly easy.

PR #884, opened 2021-05-13
Add return type annotations in grapher - part 1

danyx23:
Because grapher is such a big folder I am proceeding in increments of a few folders per PR inside grapher. This PR adds return type annotations for axis, chart, color and core folders. 

I try to do only primitive type annotations and leave adding new interface types for later but in Axis for TickPlacement it felt like a reasonable thing to add right away.

I also started accidentally working on MapChart.tsx and included this as well - will continue in a later PR


danielgavrilov (approved):
It's awesome to have these! Good to 🚢

PR #893, opened 2021-05-19
chore: Drive dev environment with `make`

larsyencken:
Automate more of the development environment with `make`, heavily reusing existing `yarn` commands and `lando`. The idea is that the environment should come up cleanly right out of the box with minimal intervention.

- Run `make` for available commands
- `make start` brings up background services (nginx, mysql), does one-time db imports (grapher, wordpress), and runs foreground services in watch mode (admin, webpack)
- Reduces MySQL services from two to one in lando
- Unifies database settings to work out of the box with lando defaults
- Updates `.env.example` and copies `.env.example` to `.env` on first run

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,6 +11,9 @@ DB_PASS=''
 DB_HOST='127.0.0.1'
 
 WORDPRESS_DB_NAME='wordpress'
+WORDPRESS_URL=http://owid.lndo.site
 
 # Overrides of anything else can go here.
 # See clientSettings.ts and serverSettings.ts for available settings.
+WORDPRESS_API_USER=
+WORDPRESS_API_PASS=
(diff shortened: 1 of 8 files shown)


danyx23:
Looks good and cool to see a makefile in action like this, I haven't touched those in many years and then only a little bit :). My changes to the readme that I have recently merged create a minor merge conflict but I think it is good to go

danyx23 (approved):

larsyencken:
It turns out there's a wordpress problem in the setup, might have to re-add the second DB spun up by Lando.

larsyencken:
Since this doesn't break anyone's existing dev environment, I'm going to merge it now, and update it with better Wordpress config with Matthieu in a separate PR.

PR #894, opened 2021-05-19
fix: tolerance not applied on ScatterPlot with mixed time units

danielgavrilov:
Notion: [ScatterPlot tolerance broken with mixed time units](https://www.notion.so/ScatterPlot-tolerance-broken-with-mixed-time-units-ed771026c1184fa9866f1f83cc9b30e1)

### TODO

- [x] Create a (failing) test case
- [ ] Fix the bug


danielgavrilov:
With the test skipped, the small improvements here would still be useful. Will leave this in drafts for a bit longer.

danielgavrilov:
Closing this – the whole tolerance implementation needs to be changed, for all charts. Will be fixed in #1085.

PR #897, opened 2021-05-24
enhance: Shrink explorer title to reduce wasted space

larsyencken:
In a previous design, the title could take more space because there were more controls to the right of it. Now that there are fewer, there is blank space left.

This change reduces the title so as to remove the need for blank space in the panel to its right.

See: https://www.notion.so/owid/Condense-explorer-title-d5bbac463dd54470a60cac48c1c2dee2

### DIFF
--- a/explorer/Explorer.scss
+++ b/explorer/Explorer.scss
@@ -30,8 +30,8 @@ html.IsInIframe #ExplorerContainer {
     justify-content: space-around;
 
     .ExplorerTitle {
-        font-size: max(16px, min(1.5vw, 19px));
-        line-height: max(18px, min(1.7vw, 23px));
+        font-size: clamp(16px, 1.5vw, 19px);
+        line-height: clamp(18px, 1.7vw, 23px);
         font-weight: bold;
         font-family: "Lato";
     }
--- a/explorer/Explorer.scss
+++ b/explorer/Explorer.scss
@@ -30,10 +30,10 @@ html.IsInIframe #ExplorerContainer {
     justify-content: space-around;
 
     .ExplorerTitle {
-        font-size: 28px;
+        font-size: 19px;
         font-weight: bold;
-        line-height: 32px;
-        font-family: "Playfair Display";
+        line-height: 23px;
+        font-family: "Lato";
     }
 
     .ExplorerSubtitle {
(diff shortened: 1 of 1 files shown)


larsyencken:
## Before

Screen Shot 2021-05-24 at 14 52 21

## After

Screen Shot 2021-05-24 at 14 52 01

explorer/Explorer.scss
```diff
@@ -30,10 +30,10 @@ html.IsInIframe #ExplorerContainer {
     justify-content: space-around;
 
     .ExplorerTitle {
-        font-size: 28px;
+        font-size: max(16px, min(1.5vw, 19px));
+        line-height: max(18px, min(1.7vw, 23px));
```
MarcelGerber, line 34: Quick tip: You can also use the [`clamp()` CSS function](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp()) for such a term.

> `clamp(MIN, VAL, MAX)` is resolved as `max()(MIN, min()(VAL, MAX))`

It also has [the same browser support](https://caniuse.com/css-math-functions) as `min` and `max`.

larsyencken: Way more readable! I'll change it.

explorer/Explorer.scss
```diff
@@ -30,10 +30,10 @@ html.IsInIframe #ExplorerContainer {
     justify-content: space-around;
 
     .ExplorerTitle {
-        font-size: 28px;
+        font-size: clamp(16px, 1.5vw, 19px);
+        line-height: clamp(18px, 1.7vw, 23px);
```
danielgavrilov, line 34: One potential thing to check here – for browsers that [don't support `clamp`](https://caniuse.com/css-math-functions), should we have a fallback? Or are those browsers not supported anyway (due to JavaScript requirements)?

I think fallbacks are defined by setting multiple properties, e.g. 

```css
font-size: 19px;
font-size: clamp(16px, 1.5vw, 19px);
```

Then if the bottom one is not supported, the top one is used?

larsyencken: Ahh... do we have an official list of browser ranges we support? 😄

MarcelGerber: Not really - I guess the unofficial rule is "as many as we can _easily_ support".

However, I checked what this looks like in Firefox 53 (which doesn't have support for `clamp`) and it looks alright. In my opinion we don't absolutely need to have fallback styles _in this case_.

But good thinking and something to keep in mind for the next time around!

image

PR #898, opened 2021-05-25
chore: Drive dev environment with make

larsyencken:
Make it simpler to bring up the full grapher+wordpress stack using Lando and Docker.

Reintroduction of this reverted PR, with additional fixes: https://github.com/owid/owid-grapher/pull/893

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,9 +5,15 @@ TZ=utc
 SECRET_KEY='fejwiaof jewiafo jeioa fjieowajf isa fjidosajfgj'
 
 # Database settings
-DB_NAME='owid'
+DB_NAME='grapher'
 DB_USER='root'
 DB_PASS=''
+DB_HOST='127.0.0.1'
+
+WORDPRESS_DB_NAME='wordpress'
+WORDPRESS_URL=http://owid.lndo.site
 
 # Overrides of anything else can go here.
 # See clientSettings.ts and serverSettings.ts for available settings.
+WORDPRESS_API_USER=
+WORDPRESS_API_PASS=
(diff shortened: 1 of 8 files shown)


larsyencken:
## Migrating from an existing dev environment

### Without lando/docker

You can continue to develop grapher without using Lando's dockerized MySQL. In this case, you can continue to bring up your environment manually without `make`.

If you need to switch to a full stack with Lando, you should:

- Stop your local running MySQL instance
- Edit your `.env` to look more like `.env.example` (`DB_NAME`, `DB_HOST`, etc.)
- Run `make start` to kick it off

You can switch back to doing it without Lando by stopping all docker services, and re-editing your `.env` file to match.

### Using lando/docker

If you don't care about the contents of your local database, no special care is needed when switching to this branch. Just destroy and start again your environment:

```bash
cd owid-grapher
make destroy
make start
```

If you do care about the contents, you should `mysqldump` all databases you want to keep, then load them again on the other side. For example:

````
# take snapshots of our local environment before tearing it down
mysqldump -u root -h 127.0.0.1 --port=3307 owid >grapher.sql
mysqldump -u root -h 127.0.0.1 wordpress >wordpress.sql

make destroy

# lie and say we've already imported data
# (prevent a one-time db import from the live server)
touch .full-snapshot-imported

# start mysql and other services
make start

# import the data manually from the local data
msyql -u root -h 127.0.0.1 -e 'CREATE DATABASE grapher'
msyql -u root -h 127.0.0.1 grapher <grapher.sql
msyql -u root -h 127.0.0.1 -e 'CREATE DATABASE wordpress'
msyql -u root -h 127.0.0.1 wordpress <wordpress.sql
```

larsyencken:
Withdrawing this PR for now, due to difficulties having Lando on MacOS run its part reliably.

PR #905, opened 2021-05-27
Svg tester V1

danyx23:
This is a first version of the Svg tester (see also https://www.notion.so/owid/Set-up-SVG-Tester-as-automated-Github-action-968caff6b0c841a3bffe0a9945b9e1ed). This does not yet include automation to run this as a github action but the basic code of the 3 svg tester parts are working. See the readme for a description of how this works.

Feedback is very welcome - both on the style of the TS and on how the scripts do their work.

My idea is to merge this when you are all happy with it and then tackle the automation setup and any changes to the scripts that might come with that in another PR.

This PR also makes 2 slight changes to the Grapher itself. One is to add the ability to reset the stateful id variable to make ids more deterministic when they are generated in a different order.

The other is in Bounds.ts where there used to exist a cache the bounds calculated for strings. After talking with @danielgavrilov  I decided to remove this cache. We use a library that uses precomputed pixel widths for each character of a small number of fonts which is very fast. On the other hand our cache used to put the full string literal into the hash table key which could end up using substantial amounts of memory (I have not profiled the actual memory use though). I did not see an increase in the runtime of the svg generation because of this change but I have not benchmarked interactive performance for any changes. I would not know how to do the latter and I doubt that this could be a big impact but please let me know if you disagree.


devTools/svgTester/readme.md
```diff
@@ -1,11 +1,31 @@
...
+
+-   The idea is to use dump-data to create an authorative data dump to be used by many verify runs. It should be done from a current production database if possible and is the only step that needs a database connection.
+
+-   Then export-graphs should be run to create the reference svgs and the csv with the md5 hashes.
+
+Steps to do on every PR etc:
+
+-   Run verify-graphs to check if there are any differences to the reference export. You can use partitioning to run this only on a subset of graphs and you can run multiple verification processes in parallel to get better cpu/io utilization.
+
+For all tools use the verbose flag if you want to see what the tool is doing, otherwise there is no output to stdout except for failing graph ids in the verify-graphs script for easy bash collection of failing graphs across multiple parallel verify-graphs runs
```
larsyencken, line 31: The headings describe what's available, but they'd read better if they instead described how it fits in the flow of what you're trying to achieve. Suggest something like:

```
## Overview

## Make a reference set of SVGs

### 1. Generate a reference dataset

### 2. Generate reference SVGs

## Check against reference SVGs
```

clientUtils/Bounds.ts
```diff
@@ -88,25 +88,23 @@ export class Bounds {
...
-            if (bounds.x === x && bounds.y === y - bounds.height) return bounds
-            return bounds.extend({ x: x, y: y - bounds.height })
-        }
 
-        if (str === "") bounds = Bounds.empty()
-        else {
+        const isBold = fontWeight >= 600
+
+        let bounds = Bounds.empty()
+        if (str !== "") {
```
larsyencken, line 95: `if (!str) {` will test for truthiness, just as good I think

clientUtils/Bounds.ts
```diff
@@ -88,25 +88,23 @@ export class Bounds {
     ): Bounds {
         // Collapse contiguous spaces into one
         str = str.replace(/ +/g, " ")
-        const key = `${str}-${fontSize}`
-        let bounds = this.textBoundsCache[key]
-        if (bounds) {
-            if (bounds.x === x && bounds.y === y - bounds.height) return bounds
-            return bounds.extend({ x: x, y: y - bounds.height })
-        }
```
larsyencken, line 96: I'm cool with removing the bounds cache from a correctness perspective. What exactly will it make slower though? The site baking?

baker/GrapherImageBaker.tsx
```diff
@@ -104,6 +114,52 @@ export async function bakeGrapherToSvg(
...
+    height: number
+) {
+    const fileKey = grapherSlugToExportFileKey(slug, queryStr)
+    const outFilename = `${fileKey}_v${version}_${width}x${height}.svg`
+    return outFilename
+}
+
+export function buildSvgOutFilepath(
+    slug: string,
+    queryStr: string,
```
larsyencken, line 144: Default to empty string

baker/GrapherImageBaker.tsx
```diff
@@ -104,6 +114,52 @@ export async function bakeGrapherToSvg(
...
+        manuallyProvideData: true,
+        queryStr,
+    })
+    grapher.isExportingtoSvgOrPng = true
+    return grapher
+}
+
+export function buildSvgOutFilename(
+    slug: string,
+    queryStr: string,
```
larsyencken, line 132: Default to empty string

baker/GrapherImageBaker.tsx
```diff
@@ -38,27 +38,39 @@ export async function bakeGraphersToPngs(
...
+            graphersBySlug.set(row.slug, grapher)
+        }
     }
 
     return graphersBySlug
 }
 
-export async function getPublishedGraphersBySlug() {
+export async function getPublishedGraphersBySlug(
+    includePrivate: boolean = true
```
larsyencken, line 59: Seems like the default should be `false`, otherwise you're now baking private graphers into the site.

baker/GrapherImageBaker.tsx
```diff
@@ -38,27 +38,39 @@ export async function bakeGraphersToPngs(
...
     const { graphersBySlug, graphersById } = await getPublishedGraphersBySlug()
 
     const redirectQuery = db.queryMysql(
         `SELECT slug, chart_id FROM chart_slug_redirects`
     )
 
     for (const row of await redirectQuery) {
-        graphersBySlug.set(row.slug, graphersById.get(row.chart_id))
+        const grapher = graphersById.get(row.chart_id)
+        if (grapher !== undefined) {
```
larsyencken, line 50: `if (grapher) {` just as good?

larsyencken (changes requested):
A nice approach! Overall, would suggest a neater parallelisation strategy if possible. Otherwise, one or two possible bugs, rest is just readability suggestions.

clientUtils/Bounds.ts
danielgavrilov: Have not tested, but my intuition is that the difference won't be measurable, or at least insignificant. The text bounds are calculated by concat-ing the bounds of every letter. The bounds of every letter are in a lookup table in the `string-pixel-width` library: https://github.com/adambisek/string-pixel-width/blob/master/src/widthsMap.js

danyx23: Yes I wrote some of the rationale in the commit message of the commit that does this but I realize now that this does not show up as context in the usual Github workflow - here is what I wrote then: 
    The method we use to calculate text bounds is already a fast heuristic and the memory overhead of keeping a bounds cache that includes the full string to be measured as a key is probably not worth it.
I checked that for the full svg generation this does not make a noticeable runtime difference. The memory use looked a bit lower in my totally non-scientific looking at the process explorer ;). At some point we should try to set up a memory benchmark for the grapher BTW

devTools/svgTester/export-graphs.ts
```diff
@@ -0,0 +1,64 @@
...
+        partition
+    )
+    const svgRecords: utils.SvgRecord[] = []
+    for (const dir of directoriesToProcess) {
+        const svgRecord = await utils.renderSvgAndSave(dir, outDir)
+        svgRecords.push(svgRecord)
+    }
+
+    await utils.writeResultsCsvFile(outDir, svgRecords)
+}
```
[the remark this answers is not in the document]
danyx23: Nice, multiprocessing seems like exactly what I wanted. Will try to use that

devTools/svgTester/readme.md
danyx23: Good idea, thanks!

larsyencken:
Totally agreed, multiprocessing for checking is the most important part. Giving it one more look through now and then it should be good to go.

devTools/svgTester/verify-graphs.ts
```diff
@@ -0,0 +1,110 @@
...
+        console.warn(
+            `${
+                differences.length
+            } graphs had differences: ${differences.join()}`
+        )
+        for (const id of differences) {
+            console.log("", id) // write to stdout one grapher id per file for easy piping to other processes
+        }
+        process.exitCode = differences.length
+    }
```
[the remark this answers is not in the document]
larsyencken: The function's looking better now, I would still suggest splitting lines 59-79 out into their own `displayResult` function, so that here it only looks like:

```
const ok = displayResults(validationResults)

process.exit(0 ? ok : 1)
```

larsyencken (changes requested):
Good work, very nearly there!

One fix to use `path.join()` is necessary, then there's some stylistic feedback. It will take no time to go through, then it will be good to ship.

devTools/svgTester/utils.ts
```diff
@@ -0,0 +1,447 @@
...
+        chartId: newSvgRecord.chartId,
+        startIndex: firstDiffIndex,
+        referenceSvgFragment: preparedReferenceSvg.substr(
+            firstDiffIndex - 20,
+            40
+        ),
+        newSvgFragment: preparedNewSvg.substr(firstDiffIndex - 20, 40),
+    })
+}
+export async function decideDirectoriesToVerify(
```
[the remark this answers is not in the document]
danyx23: Makes sense - I have the references and implementation code lenses enabled which introduce some visual white space above functions in the editor so I don't see this as clearly and didn't notice it but I'll add a pass now to do this

danyx23:
Good feedback, thanks! Merging now, further work will go into a new PR

PR #907, opened 2021-05-28
fix: hide timeline for Source or Download tabs

larsyencken:
When you click on the Source or Download tabs of a chart, the timeline currently stays rendered, despite logic that is meant to hide it.

This is because the check is done on `this.tab`, but actually Source and Download are overlays on the tab; when you click on them, `this.tab` stays at wherever it was last.

This fix runs the same check, but on `this.currentTab` instead, which picks up any overlay (such as Source or Download) if they are available. This has the effect of correctly hiding the timeline when the user clicks on those tabs.

See: https://www.notion.so/owid/Hide-timeline-on-Sources-and-Download-tabs-8766e5c1c13f431b978b8d579b92ff65

### DIFF
--- a/grapher/core/Grapher.jsdom.test.ts
+++ b/grapher/core/Grapher.jsdom.test.ts
@@ -207,6 +207,17 @@ describe("hasTimeline", () => {
         grapher.tab = GrapherTabOption.table
         expect(grapher.hasTimeline).toBeTruthy()
     })
+
+    it("source and download tabs do not have a timeline", () => {
+        const grapher = new Grapher(legacyConfig)
+        grapher.type = ChartTypeName.LineChart
+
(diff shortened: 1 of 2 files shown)


MarcelGerber (approved):
Good call, I wasn't really aware of this distinction [back when I wrote that piece of code](https://github.com/owid/owid-grapher/pull/814/files).

danielgavrilov:
Ah the good old `overlayTab` that [we should get rid of](https://owid.slack.com/archives/CQQUA2C2U/p1602073183059800) (Notion issue: [Change downloads overlay tab to popover with equal options](https://www.notion.so/Change-downloads-overlay-tab-to-popover-with-equal-options-44c5138fbef442a5b7ddde430addeb89)).

PR #925, opened 2021-06-02
fix(grapher): Facet graphs controlled by the timeline

larsyencken:
Currently, facet charts totally ignore the timeline control. This is because they are generated from the `inputTable` before transformations, which include filtering the data to only the selected range.

This fix builds the facet charts instead from the `transformedData` that has been correctly filtered this way.

### DIFF
--- a/grapher/facetChart/FacetChart.test.ts
+++ b/grapher/facetChart/FacetChart.test.ts
@@ -12,7 +12,29 @@ it("can create a new FacetChart", () => {
         selection: table.availableEntityNames,
     }
     const chart = new FacetChart({ manager })
+
+    // default to country facets
     expect(chart.series.length).toEqual(2)
+
+    // switch to column facets
     manager.facetStrategy = FacetStrategy.column
     expect(chart.series.length).toEqual(3)
+})
+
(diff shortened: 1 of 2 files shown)


larsyencken:
Note for review: this fix is for country facets, I also manually tested column facets and they don't seem to suffer from this bug.

Good graphs for manual comparison:

- `covid-vaccine-willingness-and-people-vaccinated-by-country`: supports both types of facets and has a single-point timeline
- `life-expectancy`: has a slider supporting a time range

danyx23:
Looks good to me. Now that facetted charts obey the timeline you can collapse them to a single time point and then they become facetted discreet bar charts with suboptimal labelling/spacing. Should we create a followup issue for this?

danyx23 (approved):

larsyencken:
The switch to a bar chart when you only pick one time point is pretty elegant behaviour, even if when it's faceted the spacing is not great. I think that's a separate thing to optimise. In that case, disabling faceting gives you a nice bar chart again.

@danielgavrilov Gonna merge and ship this PR now just to get it out the door, but happy to patch anything you think needs it next week.

PR #931, opened 2021-06-08
enhance(grapher): add relative yaxis mode in facets

larsyencken:
Currently, many charts support facets by typing `f` when viewing. The y-axis of every facet has the same range, so that you can compare scale across entities.

This PR adds a "Uniform y-axis" toggle when in facet mode which allows you to switch to relative y-axis mode, good for comparing the shape across entities. The default mode can be set to relative by an author, set via the "Facets have uniform y-axis" checkbox in the "Customize" tab. This controls the `yAxis.facetAxisRange` grapher config key.

See: https://www.notion.so/owid/Adjusted-vs-uniform-y-axis-scale-b88534934da04681b753742627d24ced

### DIFF
--- a/adminSiteClient/EditorCustomizeTab.tsx
+++ b/adminSiteClient/EditorCustomizeTab.tsx
@@ -22,6 +22,7 @@ import { ColorSchemeDropdown, ColorSchemeOption } from "./ColorSchemeDropdown"
 import { EditorColorScaleSection } from "./EditorColorScaleSection"
 import { ColorSchemeName } from "../grapher/color/ColorConstants"
 import { TimeBoundValue } from "../clientUtils/TimeBounds"
+import { FacetScaleMode } from "../grapher/core/GrapherConstants"
 @observer
 class ColorSchemeSelector extends React.Component<{ grapher: Grapher }> {
     @action.bound onChange(selected: ColorSchemeOption) {
@@ -307,6 +308,26 @@ export class EditorCustomizeTab extends React.Component<{
                                             (yAxisConfig.canChangeScaleType =
                                                 value || undefined)
                                         }
+                                    />
+                                </FieldsRow>
+                                <FieldsRow>
+                                    <Toggle
+                                        label={`Facets have uniform y-axis`}
+                                        value={
+                                            // default is absolute
+                                            [
+                                                FacetScaleMode.absolute,
+                                                undefined,
+                                            ].includes(
+                                                yAxisConfig.facetScaleMode
+                                            )
+                                        }
+                                        onValue={(value) => {
+                                            // remap absolute back to undefined
+                                            yAxisConfig.facetScaleMode = value
+                                                ? undefined
+                                                : FacetScaleMode.relative
+                                        }}
                                     />
                                 </FieldsRow>
                             </React.Fragment>
--- a/adminSiteClient/EditorCustomizeTab.tsx
+++ b/adminSiteClient/EditorCustomizeTab.tsx
@@ -22,7 +22,7 @@ import { ColorSchemeDropdown, ColorSchemeOption } from "./ColorSchemeDropdown"
 import { EditorColorScaleSection } from "./EditorColorScaleSection"
 import { ColorSchemeName } from "../grapher/color/ColorConstants"
 import { TimeBoundValue } from "../clientUtils/TimeBounds"
-import { FacetScaleMode } from "../grapher/core/GrapherConstants"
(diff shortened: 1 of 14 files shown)


grapher/core/GrapherConstants.ts
```diff
@@ -42,6 +42,11 @@ export enum FacetStrategy {
     column = "column", // One chart for each Y column
 }
 
+export enum FacetScaleMode {
+    relative = "relative", // all facets have their own y range
+    absolute = "absolute", // all facets share the same y range
+}
```
danielgavrilov, line 48: Hm, the relative/absolute terms don't make the right neurons fire in my head. 

Are there some other terms that could describe the same thing? Just throwing some terms: `individual`, `none`, `independent` maybe? And `uniform`, `global`, `identical`, `shared`... not exactly sure. 

Could also go for something longer and more descriptive. The comments describe it really well.

adminSiteClient/EditorCustomizeTab.tsx
```diff
@@ -309,6 +310,26 @@ export class EditorCustomizeTab extends React.Component<{
...
+                                    <Toggle
+                                        label={`Facets have uniform y-axis`}
+                                        value={
+                                            // default is absolute
+                                            [
+                                                FacetScaleMode.absolute,
+                                                undefined,
+                                            ].includes(
+                                                yAxisConfig.facetScaleMode
+                                            )
```
danielgavrilov, line 323: Is it possible to have this "absolute is default" logic somewhere more "central"? Potentially like how `fontSize` is defined on the axis: https://github.com/owid/owid-grapher/blob/9bc2033bbdc0e834d0d8589a080232d602e6f84e/grapher/axis/AxisConfig.ts#L61-L63

grapher/core/Grapher.tsx
```diff
@@ -544,6 +546,14 @@ export class Grapher
...
     }
 
+    get facetYAxisMode(): FacetScaleMode {
+        return this.yAxis.facetScaleMode || FacetScaleMode.absolute
+    }
+
+    set facetYAxisMode(mode: FacetScaleMode) {
+        this.yAxis.facetScaleMode = mode
+    }
+
```
danielgavrilov, line 556: Instead of creating these in order to implement the manager interface, could the `FacetYAxisToggleManager` be implemented by the `AxisConfig` instead? I think the `ScaleSelectorManager` is implemented similarly: https://github.com/owid/owid-grapher/blob/9bc2033bbdc0e834d0d8589a080232d602e6f84e/grapher/axis/AxisConfig.ts#L27

Not sure if it's definitely a good idea, since maybe only the Y axis should be toggle-able for now. But maybe for ScatterPlots we would like to control both, at some point? (Again, something along the same lines as the scale selector (log/linear switch))

grapher/core/GrapherConstants.ts
larsyencken: Agreed. I like `independent` and `shared`, I'll switch to those. Note that `uniform` comes from the visible toggle label "Uniform y-axis". I'll leave the label text the same.

grapher/core/Grapher.tsx
larsyencken: Good call, would prefer more stuff isn't implemented directly onto Grapher. I'll look at grafting it across.

adminSiteClient/EditorCustomizeTab.tsx
larsyencken: Moved it to `AxisConfig` 👍

grapher/core/Grapher.tsx
larsyencken: Done

grapher/core/Grapher.tsx
```diff
@@ -1446,6 +1448,18 @@ export class Grapher
...
+        return this.facetYScale === FacetAxisRange.shared
+    }
+
+    get facetYScale(): FacetAxisRange {
+        return this.yAxis.facetAxisRange
+    }
+
+    set facetYScale(mode: FacetAxisRange) {
+        this.yAxis.facetAxisRange = mode
+    }
```
danielgavrilov, line 1461: Seems like the getter is only used in the computed, and the setter not at all. Do we really need them? The original property is only ~6 characters more to access. 

As a developer the getter/setter also make me think I should be using this interface to control the axis. Is that a good idea? Isn't controlling directly through `.yAxis` better?

larsyencken: Yeah, agreed also. Cleaned up now.

PR #934, opened 2021-06-09
fix(admin): always show "Tolerance" input in DimensionCard

danielgavrilov:
Quite a lot of charts support tolerance now.

The field in `DimensionCard` was hidden for some charts, this PR makes it always shown.


danielgavrilov:
Figured this out because [someone reported United States becoming hidden in StackedDiscreteBar](https://owid.slack.com/archives/C3RAU3BJ9/p1623246108022400) when changing the selection.

PR #937, opened 2021-06-14
Add first version of chart approval tool to admin UI

bnjmacdonald:
Adds a UI to the admin site for approving/rejecting chart revisions that have been suggested by the `ChartRevisionSuggester` class from the `importers` repository (added in [this PR](https://github.com/owid/importers/pull/5)).

Notion: [Chart revision quality assurance tool v1.0](https://www.notion.so/Chart-revision-quality-assurance-tool-v1-0-d72fdd073d414871b66abebe24bc315e)

The purpose of this PR is to provide a layer of quality assurance for our charts that are updated by automated scripts (e.g. following a bulk dataset update). This PR creates a new MySQL table (`suggested_chart_revisions`) for housing suggested chart revisions, alongside an admin UI for approving/rejecting the suggested revisions. 

A "suggested chart revision" is an amended OWID chart, but where the amendments have not yet been applied to the chart in question. If the suggested chart revision gets approved, then the amendments are applied to the chart (which overwrites and republishes the chart).

Deployed on `nightingale` with suggested chart revisions for World Bank World Development Indicator charts (to update data to current version). Try it out:

-   [Chart approval tool](https://nightingale.owid.cloud/admin/suggested-chart-revisions/review)
-   [List of suggested chart revisions](https://nightingale.owid.cloud/admin/suggested-chart-revisions)


adminSiteClient/SuggestedChartRevisionApproverPage.tsx
```diff
@@ -0,0 +1,1367 @@
...
+    VisionDeficiencySvgFilters,
+    VisionDeficiencyDropdown,
+    VisionDeficiencyEntity,
+} from "./VisionDeficiencies"
+
+@observer
+export class SuggestedChartRevisionApproverPage extends React.Component<{
+    suggestedChartRevisionId?: number
+}> {
+    @observable.ref suggestedChartRevision?: any
```
danyx23, line 39: I suggest to add an interface type for this field

adminSiteClient/SuggestedChartRevisionApproverPage.tsx
```diff
@@ -0,0 +1,1367 @@
...
+        }
+        this.decisionReasonInput = this.suggestedChartRevision.decisionReason
+            ? this.suggestedChartRevision.decisionReason
+            : ""
+        this.rerenderGraphers()
+    }
+
+    @action.bound async rerenderGraphers() {
+        this._isGraphersSet = false
+        setTimeout(() => {
```
danyx23, line 165: Why is this an imperative function that uses setTimeout? Could this not be solved with a computed property? I haven't followed the dependency chain to see if this would be difficult, just wondering about this as I read the code

adminSiteClient/SuggestedChartRevisionApproverPage.tsx
```diff
@@ -0,0 +1,1367 @@
...
+
+    renderReadme() {
+        return (
+            <div className="collapsible">
+                <h3>
+                    README
+                    <button
+                        className="btn btn-outline-dark"
+                        type="button"
+                        onClick={() => (this.showReadme = !this.showReadme)}
```
danyx23, line 410: AFAIK when using the MobX library that we use for state management then all modification of state should happen in functions that are marked as action (as you are doing on named functions). Alternatively you can also wrap this event handler call with the action function - see https://mobx.js.org/actions.html#wrapping-functions-using-action . Note that I am relatively new to mobx as well and maybe this does not apply here for some reason

adminSiteClient/SuggestedChartRevisionApproverPage.tsx
```diff
@@ -0,0 +1,1367 @@
...
+    renderSettings() {
+        return (
+            <div className="collapsible">
+                <h3>
+                    Settings
+                    <button
+                        className="btn btn-outline-dark"
+                        type="button"
+                        aria-expanded={this.showSettings}
+                        onClick={() => (this.showSettings = !this.showSettings)}
```
danyx23, line 712: See above concerning actions

danyx23 (changes requested):
Nice work! I left some feedback that I think would be good to tackle before merging. Otherwise I think merging this soon is more important than perfecting various areas as we might want to do things a bit differently anyway once we improved things on the data infrastructure side and have more historic information about variables.

adminSiteClient/SuggestedChartRevisionApproverPage.tsx
danielgavrilov: Yes, we should be doing this, and the existing codebase is a bad example. But technically, actions I think actions only make a difference when multiple observable properties are edited.

danielgavrilov:
Hi Bobbie! As we mentioned, this is impressive! 

I have made some changes that I pushed on a separate branch: https://github.com/owid/owid-grapher/compare/feature/admin-suggested-chart-revision-approver...feature/admin-suggested-chart-revision-approver-tweaks

The diff is bigger because I split up `explorerAdmin` into two modules (`explorerAdminServer` and `explorerAdminClient`) to avoid circular dependencies, because I defined some chart revision types in `adminSiteClient` which are used in the database entity. (There are other ways to organise the modules to avoid circular dependencies, but this seemed cleanest, and mirrors the `adminSite` setup.)

I will leave a bit more detailed comments tomorrow, but there isn't anything on my side that needs to be changed in order to merge and use this.

adminSiteClient/SuggestedChartRevisionListPage.tsx
```diff
@@ -0,0 +1,264 @@
...
+    static contextType = AdminAppContext
+    context!: AdminAppContextType
+
+    @observable searchInput?: string
+    @observable maxVisibleCharts = 50
+    @observable suggestedChartRevisions: SuggestedChartRevisionListItem[] = []
+    @observable numTotalRows?: number
+
+    @observable sortBy: string = "updatedAt"
+    @observable sortDirection: string = "DESC"
```
danielgavrilov, line 38: We already have a `SortOrder` type that would be good to use here: https://github.com/owid/owid-grapher/blob/06853e84ea16ac460ae6fed4d31081080b04f4a8/clientUtils/owidTypes.ts#L13-L16

bnjmacdonald: resolved in c9aedc1

adminSiteClient/SuggestedChartRevisionApproverPage.tsx
bnjmacdonald: resolved in 6387f7e2d8a96149102d8b663ffcd1aa47e3b40d

adminSiteServer/apiRouter.ts
```diff
@@ -539,6 +540,281 @@ apiRouter.delete("/charts/:chartId", async (req: Request, res: Response) => {
...
+                    suggestedChartRevisionId,
+                ]
+            )
+            if (status === "approved") {
+                await saveGrapher(
+                    res.locals.user,
+                    suggestedChartRevision.suggestedConfig,
+                    suggestedChartRevision.existingConfig
+                )
+            } else if (
```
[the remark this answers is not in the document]
bnjmacdonald: added in e7ac70fff74125100770fb96a92da8ed43854eca

db/migration/1618193628982-CreateSuggestedChartRevisionsTable.ts
```diff
@@ -0,0 +1,36 @@
...
+                \`suggestedReason\` varchar(512), 
+                \`decisionReason\` varchar(512), 
+                \`createdAt\` datetime NOT NULL, 
+                \`updatedAt\` datetime NOT NULL, 
+                \`originalVersion\` int GENERATED ALWAYS AS (\`originalConfig\`->>'$.version') NOT NULL,
+                \`suggestedVersion\` int GENERATED ALWAYS AS (\`suggestedConfig\`->>'$.version') NOT NULL,
+                CHECK (\`status\` IN ('approved', 'rejected', 'pending', 'flagged')), 
+                PRIMARY KEY (\`id\`),
+                CONSTRAINT FOREIGN KEY (\`chartId\`) REFERENCES \`charts\` (\`id\`),
+                UNIQUE KEY (\`chartId\`, \`originalVersion\`, \`suggestedVersion\`, \`suggestedReason\`)
```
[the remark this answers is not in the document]
bnjmacdonald: Good point, I like this idea. Added in fbc1b415c5c06fd7d03144385e3411347cad88d2 and 367f032d5c0f30943204705c6ea023721a07b77c

adminSiteClient/SuggestedChartRevisionApproverPage.tsx
bnjmacdonald: I initially tried getting this to work with a computed property but was having trouble getting the `<Grapher />` component to re-render when the `suggestedChartRevision` state changed. Not sure what the component lifecycle problem is here.

This is similar to an issue in the chart editor where clicking on the "mobile" vs. "desktop" view does not induce a re-render.

bnjmacdonald (commented):
@danyx23 @danielgavrilov This is all set to be merged, with one possible exception that you may want to review:

In commit 1dd0fc4, I refactor `saveGrapher` slightly so that it receives a db TransactionContext as a parameter instead of initializing a TransactionContext within the funcion itself. I've played around in the chart editor UI to check that its expected behavior has not changed, but it might be good for at least one of you to confirm that this change is acceptable. 

This refactor solves a problem in the POST "/suggested-chart-revisions/:suggestedChartRevisionId/update" API method where `saveGrapher` was being called within a db transaction (thereby initializing a nested transaction). Passing a TransactionContext to saveGrapher as a parameter to avoid the nested transaction seemed like the easiest solution.

danielgavrilov (approved):
Awesome work Bobbie! 🙌 

(There are CodeQL errors that I introduced and can be ignored.)

PR #942, opened 2021-06-21
Remove a few any in admin client and client utils

danyx23:
This PR removes a few any annotations from adminClient and client utils 

There is one section in clientUtils/util.s that is currently commented out where I would like some input - it used to be that SVGElement was redefined to any in this file, thus disabling all type checks on this type. If I treat the element as used as an Element then I can't access any of the properties in the commented out code. Should there be a runtime type check and if so we do this? I am unclear why asking the SVGElement for a single point would be desirable in the first place...


MarcelGerber:
What's the state of this? Has this been reviewed, is it ready to be merged?

danyx23:
@MarcelGerber yes I think this could be merged but I don't know if @danielgavrilov did a full PR review or just looked at the one location I was having issues with. Do you want to give it a brief look and then I'll merge it if you ok it?

danyx23:
I just rebased to master to bring it up to date (no other changes though)

MarcelGerber:
Some change in here causes `hoverNearbyFeature` of map charts to not work correctly, and I don't know why:

image

https://github.com/owid/owid-grapher/blob/9544a173a6ad3aff26a9a1daf169c956f3f66a18/grapher/mapCharts/MapChart.tsx#L787-L815

MarcelGerber:
Now fixed!

danyx23:
Thanks a lot @MarcelGerber !

PR #944, opened 2021-06-23
enhance(grapher): Offer a dropdown menu for facet strategy

larsyencken:
Add a dropdown menu for faceting, which is enabled by pressing "f", allowing you to choose what faceting strategy to use, when there are multiple available.

See: https://www.notion.so/owid/Plot-entities-together-or-metrics-together-f47c2de5ca554294a78a540dc383597f

## This PR

- [x] Add a faceting dropdown control to the grapher
- [x] Update "f" hotkey to just show/hide faceting controls
- [x] Do not offer faceting by entity for StackedDiscreteBar
- [x] Fix interaction with relative control and faceting

## In a follow-up PR

- Turn the dropdown into the full hovering design

### DIFF
--- a/grapher/core/GrapherConstants.ts
+++ b/grapher/core/GrapherConstants.ts
@@ -38,8 +38,8 @@ export enum StackMode {
 export const BASE_FONT_SIZE = 16
 
 export enum FacetStrategy {
-    together = "together", // No facets
-    country = "country", // One chart for each country
+    none = "none", // No facets
+    entity = "entity", // One chart for each country/entity
     column = "column", // One chart for each Y column
 }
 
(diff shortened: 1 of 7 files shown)


grapher/core/GrapherConstants.ts
```diff
@@ -38,6 +38,7 @@ export enum StackMode {
 export const BASE_FONT_SIZE = 16
 
 export enum FacetStrategy {
+    together = "together", // No facets
```
danielgavrilov, line 41: Is `none` maybe more appropriate? Since faceting is technically disabled.

larsyencken: Agreed, fixed!

PR #946, opened 2021-06-25
fix(grapher): Facets use a chart y min if available

larsyencken:
For example, in the Covid explorer, the min is set to zero on all axes, but the facets did not share this minimum.

Now facets share the one y min when it is configured.

See: https://www.notion.so/owid/yAxisMin-explorer-setting-should-be-respected-in-facets-f992da1cbd944de18301b3bd26915555

## TODO

- [x] Use shared y min for country strategy
- [x] Use shared y min for column strategy
- [ ] Fix uniform mode for column strategy
- [ ] Use min(shared y min, detected y min) in uniform mode

### DIFF
--- a/grapher/facetChart/FacetChart.tsx
+++ b/grapher/facetChart/FacetChart.tsx
@@ -157,7 +157,10 @@ export class FacetChart
     }
 
     @computed private get columnFacets(): FacetSeries[] {
-        return this.yColumns.map((col) => {
+        const yAxis = new AxisConfig()
+        yAxis.min = this.manager.yAxis!.min
+
+        return this.yColumns.map((col: CoreColumn) => {
             return {
                 seriesName: col.displayName,
                 color: facetBackgroundColor,
@@ -166,6 +169,7 @@ export class FacetChart
                     yColumnSlug: col.slug,
                     yColumnSlugs: [col.slug], // In a column facet strategy, only have 1 yColumn per chart.
                     seriesStrategy: SeriesStrategy.entity,
+                    yAxis,
                 },
             }
         })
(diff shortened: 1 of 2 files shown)


grapher/facetChart/FacetChart.tsx
```diff
@@ -131,12 +133,12 @@ export class FacetChart
                 this.manager.yAxis!.facetAxisRange == FacetAxisRange.shared
                     ? {
                           max: sharedYDomain[1],
-                          min: sharedYDomain[0],
+                          min: yAxisMin ?? sharedYDomain[0],
```
danielgavrilov, line 136: I think potentially the logic we want here is: 

```suggestion
                          min: Math.min(yAxisMin, sharedYDomain[0]),
```

Otherwise, I think, but am not sure, that the axes can be mismatched across charts. The `min` setting I think only extends the axes, it doesn't crop them (when data exists beyond the domain).

larsyencken: It appears that the chart is already configured to not crop data, i.e. that `yAxis.min` is a guideline that can be ignored if needed to show the data.

larsyencken:
Apologies, this is not done yet, marking it as draft mode. The TODO at the top indicates what's left.

grapher/facetChart/FacetChart.tsx
danielgavrilov: Right, but it doesn't seem straightforward what the behaviour should be?

Let's say we facet by entity, and one entity has range `[-10, 10]` and all the others have `[0,10]`, and `yAxisMin = 0`. 

Should the shared axis be `[-10, 10]` or `[0, 10]`? If we go for the latter (I think what the PR currently does), one chart will have an axis not matching the others.

larsyencken: Ah, good call. Given the graph's flexible use of `yAxis.min` and `yAxis.max`, we need to cap them the way you're describing, otherwise the y axis will not be the same across facets. Good catch! 🙌

danielgavrilov:
Closing this in favour of https://github.com/owid/owid-grapher/pull/960 which the issues listed.

PR #963, opened 2021-07-20
enhance(grapher): Improve UI for facet selection

larsyencken:
Convert the standard drop-down for selecting facets into a more visual and user friendly design. The expected behaviour is that:

- The "f" key enables the chart control, but only for charts that support faceting (e.g. not scatterplots)
- The control looks like a normal dropdown, but shows hovering options when you click on it
- You should have the option "All together" and either "Split by entity" or "Split by country", depending on the chart

Refactors the faceting variables in `Grapher.tsx` to make clear what the desired faceting strategy is vs the actual rendered strategy, which has to take into account what is actually possible given the data.

See: https://www.notion.so/owid/Plot-entities-together-or-metrics-together-f47c2de5ca554294a78a540dc383597f

## Todo

- [x] Basic hovering UI dialog for facet selection
- [x] Enable both types of faceting for the multi-country, multi-metric case
- [x] Remove tippy arrow
- [x] Left-align hovering UI with control
- [x] Grapher config key for strategy selection
- [x] Add cursor: pointer
- [x] Admin UI for strategy selection
- [x] Explorer setting for strategy selection
- [ ] ~~Harmonize dropdown to existing components~~
- [x] Shrink width to longest label

### DIFF
--- a/explorer/ExplorerProgram.ts
+++ b/explorer/ExplorerProgram.ts
@@ -32,7 +32,10 @@ import { ColumnGrammar } from "./ColumnGrammar"
 import { DecisionMatrix } from "./ExplorerDecisionMatrix"
 import { CoreColumnDef } from "../coreTable/CoreColumnDef"
 import { PromiseCache } from "../clientUtils/PromiseCache"
-import { FacetAxisDomain } from "../grapher/core/GrapherConstants"
+import {
+    FacetAxisDomain,
+    FacetStrategy,
+} from "../grapher/core/GrapherConstants"
 
 export const EXPLORER_FILE_SUFFIX = ".explorer.tsv"
 
@@ -48,6 +51,7 @@ interface ExplorerGrapherInterface extends GrapherInterface {
     yScaleToggle?: boolean
     yAxisMin?: number
     facetYDomain?: FacetAxisDomain
+    selectedFacetStrategy?: FacetStrategy
 }
 
 const ExplorerRootDef: CellDef = {
--- a/explorer/ExplorerProgram.ts
+++ b/explorer/ExplorerProgram.ts
@@ -32,10 +32,7 @@ import { ColumnGrammar } from "./ColumnGrammar"
 import { DecisionMatrix } from "./ExplorerDecisionMatrix"
 import { CoreColumnDef } from "../coreTable/CoreColumnDef"
 import { PromiseCache } from "../clientUtils/PromiseCache"
-import {
-    FacetAxisDomain,
-    FacetStrategy,
-} from "../grapher/core/GrapherConstants"
+import { FacetAxisDomain } from "../grapher/core/GrapherConstants"
 
 export const EXPLORER_FILE_SUFFIX = ".explorer.tsv"
 
--- a/explorer/ExplorerProgram.ts
+++ b/explorer/ExplorerProgram.ts
@@ -51,7 +51,6 @@ interface ExplorerGrapherInterface extends GrapherInterface {
     yScaleToggle?: boolean
     yAxisMin?: number
     facetYDomain?: FacetAxisDomain
-    selectedFacetStrategy?: FacetStrategy
 }
 
 const ExplorerRootDef: CellDef = {
--- a/explorer/GrapherGrammar.ts
+++ b/explorer/GrapherGrammar.ts
@@ -144,6 +144,16 @@ export const GrapherGrammar: Grammar = {
             cssClass: "",
         })),
     },
+    selectedFacetStrategy: {
+        ...EnumCellDef,
+        keyword: "selectedFacetStrategy",
+        description: "Whether the chart should be faceted or not",
+        terminalOptions: Object.keys(FacetStrategy).map((keyword) => ({
+            keyword,
+            description: "",
+            cssClass: "",
+        })),
+    },
     baseColorScheme: {
         ...EnumCellDef,
         keyword: "baseColorScheme",
--- a/explorer/GrapherGrammar.ts
+++ b/explorer/GrapherGrammar.ts
@@ -148,7 +148,7 @@ export const GrapherGrammar: Grammar = {
         ...EnumCellDef,
         keyword: "selectedFacetStrategy",
         description: "Whether the chart should be faceted or not",
-        terminalOptions: Object.keys(FacetStrategy).map((keyword) => ({
+        terminalOptions: Object.values(FacetStrategy).map((keyword) => ({
             keyword,
             description: "",
             cssClass: "",
--- a/grapher/captionedChart/CaptionedChart.tsx
+++ b/grapher/captionedChart/CaptionedChart.tsx
@@ -255,7 +255,10 @@ export class CaptionedChart extends React.Component<CaptionedChartProps> {
         if (manager.showZoomToggle)
             controls.push(<ZoomToggle key="ZoomToggle" manager={manager} />)
 
-        if (manager.showFacets) {
+        if (
+            manager.showFacets &&
+            manager.availableFacetStrategies!.length > 1
+        ) {
             controls.push(
                 <FacetStrategyDropdown
                     key="FacetStrategyDropdown"
--- a/grapher/captionedChart/CaptionedChart.tsx
+++ b/grapher/captionedChart/CaptionedChart.tsx
@@ -257,7 +257,7 @@ export class CaptionedChart extends React.Component<CaptionedChartProps> {
 
         if (
             !manager.hideFacetControl &&
-            manager.availableFacetStrategies!.length > 1
+            manager.availableFacetStrategies.length > 1
         ) {
             controls.push(
                 <FacetStrategyDropdown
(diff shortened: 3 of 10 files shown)


grapher/controls/Controls.scss
```diff
@@ -13,11 +13,118 @@ $zindex-ControlsFooter: 2;
...
+.selected .FacetStrategyPreview-none-child {
+    background: rgb(217, 234, 251);
+}
+
+.FacetStrategyPreview-none-child {
+    background: #dde1e7;
+}
+
+.selected .FacetStrategyPreview-split-child {
+    background: rgb(217, 234, 251);
```
larsyencken, line 123:
```suggestion
    background: #d9eafb;
```

grapher/controls/Controls.scss
```diff
@@ -13,11 +13,118 @@ $zindex-ControlsFooter: 2;
...
+
+.FacetStrategyPreview-split-child:nth-of-type(1),
+.FacetStrategyPreview-split-child:nth-of-type(2),
+.FacetStrategyPreview-split-child:nth-of-type(4),
+.FacetStrategyPreview-split-child:nth-of-type(5) {
+    margin-right: 2px;
+}
+
+.selected .FacetStrategyPreview-none-child {
+    background: rgb(217, 234, 251);
```
larsyencken, line 115:
```suggestion
    background: #d9eafb;
```

grapher/core/Grapher.tsx
```diff
@@ -1864,9 +1861,16 @@ export class Grapher
     }
 
     private disableAutoFaceting = true // turned off for now
+
+    // the actual facet setting used by a chart, potentially overriding selectedFacetStrategy
     @computed get facetStrategy(): FacetStrategy {
-        if (this.facet && this.availableFacetStrategies.includes(this.facet))
-            return this.facet
+        if (!this.showFacets) return FacetStrategy.none
```
danielgavrilov, line 1867: An author might want to set a faceting strategy but not allow readers to edit it?

We do this for example for the timeline – an author can change the timeline, then hide it (`hideTimeline`), and it stays applied for readers.

grapher/core/Grapher.tsx
```diff
@@ -297,7 +297,9 @@ export class Grapher
     @observable.ref annotation?: Annotation = undefined
 
     @observable showFacets?: boolean = false
-    @observable lastFacet: FacetStrategy = FacetStrategy.none
+
+    // the desired faceting strategy, which might not be possible if we change the data
+    @observable selectedFacetStrategy: FacetStrategy = FacetStrategy.none
```
danielgavrilov, line 302: Maybe we want to add this to the list of keys that get serialized into the JSON for the database? https://github.com/owid/owid-grapher/blob/c8826790803ad20a99c93cee3b628754fb7ccbff/grapher/core/GrapherInterface.ts#L117-L169

grapher/controls/Controls.scss
```diff
@@ -13,11 +13,118 @@ $zindex-ControlsFooter: 2;
...
     width: 160px;
     font-family: "Lato";
-    font-size: 10px;
+    font-size: 12px;
     font-weight: bold;
+    border: 1px solid #afb7c1;
+    border-radius: 5px;
+    padding-left: 10px;
+    padding-right: 10px;
+}
```
danielgavrilov, line 28: Can add `cursor: pointer` to give a hint the component is interactive.

danielgavrilov (commented):
Awesome! I left some code comments.

Some potential design improvements:

- The dropdown could close after selection
- Could use `cursor: pointer` to indicate that the dropdown is clickable 
- The icons aren't clear enough I think. We should probably use some hand-crafted SVG icons to communicate better that a chart get split into multiple.

Good to hear what @vanwoerden thinks.

grapher/captionedChart/CaptionedChart.tsx
```diff
@@ -252,7 +252,10 @@ export class CaptionedChart extends React.Component<CaptionedChartProps> {
         if (manager.showZoomToggle)
             controls.push(<ZoomToggle key="ZoomToggle" manager={manager} />)
 
-        if (manager.showFacets) {
+        if (
+            manager.showFacets &&
+            manager.availableFacetStrategies!.length > 1
```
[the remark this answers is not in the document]
larsyencken: Miraculously, this one was easy. The other parameters could not be made required without affecting a huge number of places, because of the interface inheritance pattern.

grapher/core/Grapher.tsx
larsyencken: Agreed, we want to support that. I renamed the flag to `hideFacetControl` and made the faceting stick even if the control is hidden.

grapher/core/Grapher.tsx
larsyencken: Done

grapher/controls/Controls.scss
larsyencken: Done 👍

explorer/GrapherGrammar.ts
```diff
@@ -144,6 +144,16 @@ export const GrapherGrammar: Grammar = {
...
+    selectedFacetStrategy: {
+        ...EnumCellDef,
+        keyword: "selectedFacetStrategy",
+        description: "Whether the chart should be faceted or not",
+        terminalOptions: Object.keys(FacetStrategy).map((keyword) => ({
+            keyword,
+            description: "",
+            cssClass: "",
+        })),
+    },
```
larsyencken, line 156: @danielgavrilov As a heads up, it's not clear to me that selecting one faceting strategy globally for an explorer is the right way to go here. It might be enough to get us started.

explorer/ExplorerProgram.ts
```diff
@@ -48,6 +51,7 @@ interface ExplorerGrapherInterface extends GrapherInterface {
     yScaleToggle?: boolean
     yAxisMin?: number
     facetYDomain?: FacetAxisDomain
+    selectedFacetStrategy?: FacetStrategy
```
danielgavrilov, line 54: There is already `facet` on `GrapherInterface`, should that be removed?

grapher/controls/Controls.tsx
```diff
@@ -237,52 +235,103 @@ export class FilterSmallCountriesToggle extends React.Component<{
...
+                        <div className="FacetStrategyPreview-split-child"></div>
+                        <div className="FacetStrategyPreview-split-child"></div>
+                        <div className="FacetStrategyPreview-split-child"></div>
+                        <div className="FacetStrategyPreview-split-child"></div>
+                        <div className="FacetStrategyPreview-split-child"></div>
+                    </>
+                )
+            return (
+                <div
+                    className={"FacetStrategyOption" + selected}
```
danielgavrilov, line 314: There's a `classnames` library we use that allows you to do something like this:

```suggestion
                    className={classnames({
                        "FacetStrategyOption": true,  
                        "selected": value === this.facetStrategy 
                    })}
```

grapher/controls/Controls.tsx
```diff
@@ -237,52 +235,103 @@ export class FilterSmallCountriesToggle extends React.Component<{
...
-    @action.bound onChange(
-        selected: ValueType<{ label: string; value: string }>
-    ): void {
-        this.props.manager.facet = asArray(selected)[0].value as FacetStrategy
+    // A hovering visual display giving options to be selected from
+    @computed get content(): JSX.Element {
+        const parts = this.strategies.map((value: FacetStrategy) => {
+            // There are only ever two strategies, none + (country OR entity).
+            const label = facetStrategyLabels[value]
+            const selected = value == this.facetStrategy ? " selected" : ""
```
danielgavrilov, line 296: There is something not going quite right with the reactivity, the selected option doesn't correctly propagate: 

https://user-images.githubusercontent.com/1308115/128192216-23e73e63-ec6f-422d-a7e3-dee5179e8348.mov

danielgavrilov (approved):
Apart from some small things, this is looking great! I like the new name `selectedFacetStrategy`, and the fact that you left an "Auto" option on the Admin.

Switching to a show-control-by-default would mean that a lot of past charts are affected, so this is maybe something to discuss before launch, if it hasn't been discussed yet.

danielgavrilov:
Another thing related to this: Instead of [Split by **country**], the option should really be [Split by **`entityType`**], which is defined here:

https://github.com/owid/owid-grapher/blob/e70959ccbed54ad7c72eba8621d6f6a32107063f/grapher/core/Grapher.tsx#L261-L262

grapher/captionedChart/CaptionedChart.tsx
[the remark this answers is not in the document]
larsyencken: Sorry, removed! I actually wish a redundant ! was a linting error.

explorer/ExplorerProgram.ts
larsyencken: Turns out this property wasn't needed on `ExplorerGrapherInterface`, removed it.

grapher/controls/Controls.tsx
larsyencken: Done

grapher/controls/Controls.tsx
larsyencken: Turns out I forgot to mark the dropdown class with `@observer`, doing that fixes it.

PR #964, opened 2021-07-20
enhance: Introduce grapher json schema

danyx23:
This is an initial base for a json schema for the grapher config. See this notion page for details: https://www.notion.so/owid/Grapher-json-schema-791109f371254a32999cd03b74d7056c

### DIFF
--- a/grapher/core/schema.json
+++ b/grapher/core/schema.json
@@ -216,15 +216,6 @@
                     "maximum": 32767,
                     "minimum": 0
                 },
-                "defaultYear": {
-                    "oneOf": [
-                        {
-                            "type": "string",
-                            "enum": ["latest"]
-                        },
-                        { "type": "number", "minimum": 0 }
-                    ]
-                },
                 "variableId": {
                     "type": "integer",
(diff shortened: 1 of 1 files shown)


grapher/core/schema.json
```diff
@@ -0,0 +1,731 @@
...
+                },
+                "defaultYear": {
+                    "oneOf": [
+                        {
+                            "type": "string",
+                            "enum": ["latest"]
+                        },
+                        { "type": "number", "minimum": 0 }
+                    ]
+                },
```
larsyencken, line 227: Remove it altogether, we no longer have code that even looks at it

```suggestion
```

larsyencken (approved):
Looks good. I reckon we could auto-annotate any property that hasn't been used in 1y as "probably-deprecated" in the "description" field.

I also think we could refactor out sub-objects and use references to them in the main object, it will help it to be slightly more readable.

In future data migrations, I'd love to also get rid of properties that are there but with the value `null`, it seems a waste to store that.

PR #966, opened 2021-07-22
chore: Prune and annotate the JSON schema

larsyencken:
Remove fields that do not exist in our codebase and auto-annotate as deprecated fields that have not been used in the past year.

See Jupyter notebook for details of auto-annotation/marking for deletion: https://gist.github.com/larsyencken/9f157a3db5a4d5dcd26262d5e72a51c5

### DIFF
--- a/grapher/core/schema.json
+++ b/grapher/core/schema.json
@@ -12,7 +12,8 @@
             "properties": {
                 "legendOrientation": {
                     "type": ["string", "null"],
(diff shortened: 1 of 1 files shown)


danyx23:
I didn't have time to do an in-depth look at every one of these in turn. From a birds eye view this looks good and can be merged. We might want to go back and retrieve one of those properties again but I think it makes sense to try and get to a modern version of this schema and only include legacy stuff if we have a good reason.

PR #970, opened 2021-07-26
WIP: unfinished attempt to fix cookie notice

danielgavrilov:
The cookie policy implementation uses "date policy was accepted" to decide whether to show or hide the policy, by comparing it to the "date policy was last updated". 

This has issues:

- A policy update doesn't happen instantaneously at the beginning of a day, so if people accept the policy on a day, they may have accepted the old one. This means we should set the "policy last updated" date at least one day in the future. This brings up another issue:
- We use non-UTC dates, so users in more western timezones might be a day behind. This results in them not being able to get the banner away, because for them the condition holds even after accepting: "last accepted date" < "last policy update date".

I've attempted to fix this by using the policy date as an identifier, and storing that in the cookie. But have not finished this, and have still not gotten my head around the whole implementation. At the moment I'm also on an old laptop which doesn't have Wordpress set up to test this.

So I've left it half-finished. 🏳️  We can update the policy date with these bugs in place for now.

### DIFF
--- a/site/CookieNotice.tsx
+++ b/site/CookieNotice.tsx
@@ -3,15 +3,13 @@ import { useEffect, useState } from "react"
 import classnames from "classnames"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import { faCheck } from "@fortawesome/free-solid-svg-icons/faCheck"
-import { Action, getTodayDate } from "./CookiePreferencesManager"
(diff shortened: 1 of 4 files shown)


danielgavrilov:
I'm closing this since it's a tiny issue and this PR is messy and incomplete. The comment above is maybe the most useful thing from it – describing the problem.

PR #981, opened 2021-08-04
enhance(grapher): Add URL parameters for faceting

larsyencken:
Add two new URL parameters to preserve faceting settings. They are:

- `facet`: can be (`none`, `entity`, `metric`)
- `uniformYAxis`: can be (`1`, `0`)

Refactoring

- Renamed `FacetStrategy.column` to `FacetStrategy.metric`
- Make `yAxis.facetDomain` to `undefined` by default (but default it to `FacetAxisDomain.shared`)

Expected behaviour

- The URL parameters can override the default settings
- They get auto-added if you vary from the default settings, and auto-removed if you go back to any default setting

https://www.notion.so/owid/URL-parameter-support-for-facets-c7fb9b2a1e864f5e85137de71a736fca

### DIFF
--- a/grapher/axis/AxisConfig.ts
+++ b/grapher/axis/AxisConfig.ts
@@ -29,7 +29,7 @@ class AxisConfigDefaults implements AxisConfigInterface {
     @observable.ref maxTicks?: number = undefined
     @observable.ref compactLabels?: boolean = undefined
     @observable.ref scaleType?: ScaleType = ScaleType.linear
-    @observable.ref facetDomain?: FacetAxisDomain = FacetAxisDomain.shared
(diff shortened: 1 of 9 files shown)


larsyencken:
@danielgavrilov When reviewing, note that this branch is on top of `facet-pivot-nice-ui`. This means you should only review the last two commits. It will be rebased on top of master before merging.

danielgavrilov:
> * `facet`: can be (`none`, `entity`, `metric`)

Ah, sorry, did you by any chance rename `column` → `metric` because I mentioned that in a Notion comment? It seems that `column` is the more consistent name with the rest of the codebase after all.

PR #983, opened 2021-08-09
fix(grapher): add exception to default source name shortening

bnjmacdonald:
Solves a minor problem where the Grapher is shortening source names on display when this behavior is not desired for a number of source names in the latest World Bank World Development Indicators dataset.

- e.g. the source name "World Bank and OECD" should not be shortened to "World Bank" by default.
- e.g. the source name "World Bank based on UN Population Division" should not be shortened to "World Bank" by default.

This fix alters the default source name shortening behavior to NOT shorten source names on display when the source name starts with a `majorSource` (e.g. "World Bank") followed by either " and" or " based on".

### DIFF
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -1302,7 +1302,16 @@ export class Grapher
                 "World Bank",
                 "ILOSTAT",
             ]) {
-                if (sourceName.startsWith(majorSource)) return majorSource
+                if (
+                    sourceName.startsWith(majorSource) &&
+                    !sourceName.match(
+                        new RegExp(
+                            "^" + majorSource + "\\s+(based on|and)",
+                            "gi"
+                        )
+                    )
+                )
+                    return majorSource
             }
             return sourceName
         })


MarcelGerber (approved):
Makes sense to me.

danielgavrilov:
Could we get rid of this `majorSources` exception altogether? As the code comment maybe suggests, it's fragile to have something like this at runtime: https://github.com/owid/owid-grapher/blob/9c8c22b2ced0dac0e2a3f30ce7f6cc6c6a104b63/grapher/core/Grapher.tsx#L1298

I imagine it tripped Bobbie up and it was unexpected to figure out that the Grapher has this special case? I didn't expect it to be there, despite seeing it before. 

This was implemented before my time, so at this point I would assume only Max has context behind this decision. I think probably best to get rid of it, and fix things on the source/database level if it's an issue?

danielgavrilov:
Sorry, I totally missed this notification. Maybe @larsyencken or @danyx23 can merge and deploy this – it all looks good to me!

PR #985, opened 2021-08-09
fix(DataTable): show correct entity type in table header

danielgavrilov:
Now gets `entityType` from manager, not from the `OwidTable`. 

This is a lapse from Project Next – the entity type isn't ever defined on the table, at least yet. It's messier right now to define it there, even though it _feels_ like it belongs there.

https://www.notion.so/owid/Grapher-table-view-labels-entity-as-Country-always-209c27d7a8aa45b593f0b754153db901


larsyencken:
Looks good 👍

PR #988, opened 2021-08-10
enhance: Improve faceting boundary cases

larsyencken:
Constrain the faceting experience so that we offer faceting only when it makes sense, and so that faceting interacts better with relative mode in the StackedDiscreteBar case.

Note that we change the relative mode state when switching to faceting. The ideal solution would not do this, just override it, but that would take a big refactor. So this is the compromise approach.

- [X] Disable facet-by-entity if there is only one data point
- [X] Disable facet-by-metric if there is only one data point
- [x] Disable facet-by-metric when one metric is a projection of the other
- [x] Disable relative mode when turning on faceting
- [x] Disable relative mode control during faceting

https://www.notion.so/owid/Ensure-all-possible-faceting-combinations-make-sense-for-v1-launch-5006ee5a119844c9bca5007b1e42e598

### DIFF
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -1912,7 +1912,12 @@ export class Grapher
             strategies.push(FacetStrategy.metric)
         }
 
-        if (this.selection.numSelectedEntities > 1) {
+        if (
+            // multiple entities
+            this.selection.numSelectedEntities > 1 &&
+            // more than one data point per entity
+            this.transformedTable.numRows > this.selection.numSelectedEntities
+        ) {
             strategies.push(FacetStrategy.entity)
         }
 
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -1908,7 +1908,12 @@ export class Grapher
     @computed get availableFacetStrategies(): FacetStrategy[] {
         const strategies: FacetStrategy[] = [FacetStrategy.none]
 
-        if (this.hasMultipleYColumns) {
+        if (
+            // multiple metrics
+            this.hasMultipleYColumns &&
+            // more than one data point per metric
+            this.transformedTable.numRows > 1
+        ) {
             strategies.push(FacetStrategy.metric)
         }
 
--- a/grapher/core/Grapher.tsx
+++ b/grapher/core/Grapher.tsx
@@ -1908,9 +1908,12 @@ export class Grapher
     @computed get availableFacetStrategies(): FacetStrategy[] {
         const strategies: FacetStrategy[] = [FacetStrategy.none]
 
+        const numNonProjectedColumns = this.yColumns.filter(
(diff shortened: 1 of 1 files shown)


grapher/core/Grapher.tsx
```diff
@@ -1485,7 +1485,7 @@ export class Grapher
                 this.yScaleType !== ScaleType.log
             )
 
-        if (this.facetStrategy === FacetStrategy.metric) return false
+        if (this.facetStrategy !== FacetStrategy.none) return false
```
danielgavrilov, line 1488: Hm, this is tricky, because there are some charts (like [CO₂ by source](https://ourworldindata.org/grapher/co2-by-source?stackMode=relative&facet=entity&country=SWE~AUS~DEU~GBR)) where this is valid, but then others where it isn't. 

Or more generally: We can have relative stacked charts, as long as the stacking strategy is not the same as the facet strategy. 

But the issue is that, again, like in many other cases developing this, we need information to flow up from the chart to the facets.

And then there is the additional complication that "relative" means different things for different charts, e.g. LineChart: https://ourworldindata.org/grapher/fossil-fuel-consumption-by-type?stackMode=relative&facet=entity&country=SWE~AUS~DEU~GBR

(This one is actually broken in facet mode, see [what it actually needs to look like](https://ourworldindata.org/grapher/fossil-fuel-consumption-by-type?stackMode=relative&time=1997..latest&facet=none&uniformYAxis=0&country=SWE))

For LineCharts, we shouldn't force disable the setting, because the relative means `value at current time / value at start time`.

I'm just dumping my thoughts for now, not sure which way I'm leaning on this. I was initially thinking that we could favour the safer approach (just disable it and don't worry about edge cases) but in some cases that would be kind of broken.

grapher/core/Grapher.tsx
```diff
@@ -1908,11 +1908,24 @@ export class Grapher
     @computed get availableFacetStrategies(): FacetStrategy[] {
         const strategies: FacetStrategy[] = [FacetStrategy.none]
 
-        if (this.hasMultipleYColumns) {
+        const numNonProjectedColumns = this.yColumns.filter(
+            (c) => !c.display || !c.display.isProjection
```
danielgavrilov, line 1912:
```suggestion
            (c) => !c.display?.isProjection
```

is more elegant

grapher/core/Grapher.tsx
larsyencken: Ah, I forgot that line charts can be in relative mode, since there few cases of it. Thanks for spotting it! I'll look at it before shipping anything.

larsyencken: I've constrained it to only StackedDiscreteBar now just to get something out, but actually even there what we care about is whether we have more than one metric or not. We'll be back here when we do multi-metric selection.

PR #990, opened 2021-08-17
enhance(webpack): limit chunk size to 1MB, create multiple chunks

danielgavrilov:
[Live on playfair](https://playfair-owid.netlify.app/coronavirus-data-explorer)

This hopefully mitigates the issue we have with @netlify, which is that our `commons.js` chunk, which is currently ~1.8MB, is often evicted from their edge cache, and on top of that their origin → edge transfer speeds are abysmal in Europe, leading to very long loading times (8–9 seconds).

This PR splits the `commons.js` into multiple chunks:

Screenshot 2021-08-17 at 21 44 46

Exactly how many is up to webpack to decide, we can only set an upper bound on the file size of a single chunk (with `maxSize`):

> Using `maxSize` tells webpack to try to split chunks bigger than `maxSize` bytes into smaller parts. Parts will be at least `minSize` (next to `maxSize`) in size. The algorithm is deterministic and changes to the modules will only have local impact. So that it is usable when using long term caching and doesn't require records. `maxSize` is only a hint and could be violated when modules are bigger than `maxSize` or splitting would violate `minSize`.

https://v4.webpack.js.org/plugins/split-chunks-plugin/#splitchunksmaxsize

It makes the behaviour between development and live different with regards to splitting chunks, which introduces opportunities to unintentionally break things. Specifically, the way `webpackUrls()` reads the webpack `manifest.json` and pattern matches feels a little bit fragile.


danielgavrilov:
Something to be careful with: [`embedCharts.js`](https://playfair-owid.netlify.app/grapher/embedCharts.js) isn't used anywhere within our site, but it is used on [sdg-tracker.org](https://sdg-tracker.org).

package.json
```diff
@@ -30,8 +30,8 @@
         "startWebpackServer": "webpack-dev-server --no-live-reload --config ./itsJustJavascript/webpack.config.js",
         "startSiteFront": "webpack-dev-server --config ./itsJustJavascript/webpack.config.js",
         "startLocalBakeServer": "http-server ./localBake -p 3000",
-        "testBundlewatch": "ENV=production webpack --config ./itsJustJavascript/webpack.config.js -p && bundlewatch --config .bundlewatch.config.json",
-        "testBundlewatchJson": "ENV=production webpack --config ./itsJustJavascript/webpack.config.js -p --json > itsJustJavascript/webpack/webpack-bundle-stats.json && bundlewatch --config .bundlewatch.config.json",
+        "testBundlewatch": "ENV=production webpack --config ./itsJustJavascript/webpack.config.js -p && cat itsJustJavascript/webpack/commons\\~*.js > itsJustJavascript/webpack/commons.js && bundlewatch --config .bundlewatch.config.json",
+        "testBundlewatchJson": "ENV=production webpack --config ./itsJustJavascript/webpack.config.js -p --json > itsJustJavascript/webpack/webpack-bundle-stats.json && cat itsJustJavascript/webpack/commons\\~*.js > itsJustJavascript/webpack/commons.js && bundlewatch --config .bundlewatch.config.json",
```
danielgavrilov, line 34: Note that in order to get `bundlewatch` to work again, all chunks get concatenated into a single `commons.js` which may potentially not be valid JS, but is useful to estimate total file sizes.

Potentially, we could rename this to something like `commons--merged--bundlewatch-only.js` so that it's more obvious that it isn't supposed to be used.

larsyencken: Leaving this for later 👍

larsyencken:
I played with `minSize` a little now, but it didn't seem to decrease the number of chunks or be respected properly. I'm going to ship this as is, then review in a week or so if there's no issues.

PR #991, opened 2021-08-18
enhance(facets): Add global legend for Line, StackedArea & StackedDiscreteBar

danielgavrilov:
Notion: [Global legend for all facets](https://www.notion.so/Global-legend-for-all-facets-c579c294009b4e72a6f21b801a366c8f)

Ready to review. Not quite sure whether this is the right approach for combining legends across charts, but it could be good enough for v1. Wrote some thoughts below:

Not deployed on a staging server because my staging server is used for another PR. But here are screenshots:

Screenshot 2021-08-18 at 18 41 34
Screenshot 2021-08-18 at 18 42 15

### Questions

**Is it reasonable to assume that a legend with 1 item should be automatically hidden?** 

This applies to the global facet legend, but also the stacked bar legend. There may be cases where it's a multi-metric chart, so the titles don't contain the full information and we need to rely on the legends, but maybe all except one metric happen to be missing for the selected countries. This behaviour would mean that the 1 legend item that is hidden is actually necessary to interpret the chart. 

What we actually want to check for is whether the author, in the original chart config, has selected multiple metrics. Right now we're getting the legend bins from the charts without information whether they are entities or metrics, so we don't have the necessary information on the facet level. 

Do charts maybe need to expose an axis/strategy along which to create a legend? Instead of the way I've done it currently, which is that the legend items are generated by the individual charts and then combined and shown at the facet level, maybe a chart only needs to specify the `legendStrategy`? Does this generalize across charts?

### DIFF
--- a/adminSiteClient/EditorFeatures.tsx
+++ b/adminSiteClient/EditorFeatures.tsx
@@ -55,7 +55,11 @@ export class EditorFeatures {
     }
 
     @computed get hideLegend() {
-        return this.grapher.isLineChart || this.grapher.isStackedArea
+        return (
+            this.grapher.isLineChart ||
+            this.grapher.isStackedArea ||
+            this.grapher.isStackedDiscreteBar
+        )
     }
 
     @computed get stackedArea() {
--- a/grapher/chart/ChartInterface.ts
+++ b/grapher/chart/ChartInterface.ts
@@ -31,5 +31,9 @@ export interface ChartInterface {
     yAxis?: HorizontalAxis | VerticalAxis
     xAxis?: HorizontalAxis | VerticalAxis
 
+    /**
+     * The legend bins that have been hidden from the chart plot and should be shown externally.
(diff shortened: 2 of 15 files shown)


danielgavrilov:
The SVG diffs seem to be due to:
- The `<HorizontalColorLegend>` being omitted in `LineChart` if `hideLegend === true`. Only invisible `<g>` tags are removed.
- In `HorizontalColorLegend`, the sizes of rectangles and slight positioning changed

larsyencken:
Just a note for future PRs, it speeds up review to link to an existing chart in prod that displays the behaviour, rather than just screenshots.

This one seems like a good candidate for this PR: https://ourworldindata.org/grapher/per-capita-energy-stacked?facet=none&country=USA~GBR~OWID_WRL~CHN~IND~FRA~DEU~SWE~ZAF~JPN~BRA

PR #993, opened 2021-08-19
fix(webpack): do not split CSS chunks, fix embedCharts.js

danielgavrilov:
[Live on playfair](https://playfair-owid.netlify.app/covid-hospitalizations)

Currently the CSS is being chunked, and the chunks are out of order.

Had to set a distinct `name` for the JS and the CSS chunk, otherwise they end up merging somehow.


larsyencken:
I would deeply love a better way to test this locally, given that it takes so long to stage it, and mashing `isProduction` on in a bunch of places doesn't seem to be enough.

larsyencken:
I deployed now, but it seemed to break the admin site. It also seems like my force push accidentally picked up other recent side changes by accident. Rolling back...

PR #997, opened 2021-08-25
Simple admin improvements

danyx23:
This PR does two things: it changes the way search works from fuzzySearch that does sublime style search to one that splits by word boundaries and then searches for ALL of the words but accepts them inside longer words (i.e. "fish soup" matches "fishes megasoupmonster")

The second change is that the number of times a variable is used in charts is available as an API endpoint and used in the variable search dialog. This is a bit hacky but I think for a better solution we would need a better way to cache data about graphers on the client and a nicer way to expose API endpoints on the admin server. Let me know if you find this acceptable :)


adminSiteServer/apiRouter.ts
```diff
@@ -948,6 +948,36 @@ apiRouter.get("/variables.json", async (req) => {
...
+            }
+            const chartsForVariable = variablesToChartIds.get(variable)
+            chartsForVariable!.add(row.id)
+        }
+    }
+
+    // JSON.stringify supports neither Map nor Set so we have to convert the Map to an object and the Sets to arrays
+    return _.mapValues(Object.fromEntries(variablesToChartIds), (chartIds) => [
+        ...chartIds,
+    ])
```
MarcelGerber, line 978: Not sure if it's a good idea API-wise, but we could also just return `variablesToChartIds.entries()` and then simply parse it on the other end using `new Map(...)` without needing `parseInt` and the like?

(would still need to convert the Set, of course)

adminSiteServer/apiRouter.ts
```diff
@@ -948,6 +948,36 @@ apiRouter.get("/variables.json", async (req) => {
     return { variables: rows, numTotalRows: numTotalRows }
 })
 
+apiRouter.get("/variables.usages.json", async (req) => {
+    // This is a bit of a workaround - what we would like is to select
+    // all variables and do a group by on the joined charts table with a count(*)
+    // but since the variables are stored inside the json config and mysql 5.7's limited
+    // support for json and arrays we can'd do it that way - so instead we select the arrays
+    // and create a map that we use to "transpose" the table, then serialize this
+    console.log("entering usages")
```
MarcelGerber, line 957: might want to get rid of that :)

clientUtils/search.tsx
```diff
@@ -0,0 +1,207 @@
...
+            { searchTerm: "2", regexTerm: "(2|\u00B2|\u2082)" },
+            { searchTerm: "3", regexTerm: "(3|\u00B3|\u2083)" },
+            { searchTerm: "4", regexTerm: "(4|\u2074|\u2084)" },
+            { searchTerm: "5", regexTerm: "(5|\u2075|\u2085)" },
+            { searchTerm: "6", regexTerm: "(6|\u2076|\u2086)" },
+            { searchTerm: "7", regexTerm: "(7|\u2077|\u2087)" },
+            { searchTerm: "8", regexTerm: "(8|\u2078|\u2088)" },
+            { searchTerm: "9", regexTerm: "(9|\u2079|\u2089)" },
+        ]
+    let moreTolerantMatch = escapedString //.replace(/([0-9])/g, "($1|p{No})")
```
MarcelGerber, line 31: get rid of that if it's no longer in use

clientUtils/search.tsx
```diff
@@ -0,0 +1,207 @@
...
+        .map((item) => ({
+            regex: buildRegexFromSearchWord(item),
+            word: item,
+        }))
+    return wordRegexes
+}
+
+/** Given a list of SearchWords constructed with buildSearchWordsFromSearchString
+    and a search field string extractor function, this function returns a filter function
+    that tells you if all search terms occur in any of the fields extracted by the extractor fn.
```
MarcelGerber, line 56: maybe add: ", in any order"

clientUtils/search.tsx
```diff
@@ -0,0 +1,207 @@
...
+                    <span key={`${lastIndex}-start`}>
+                        {text.substring(lastIndex, index)}
+                    </span>
+                )
+                fragments.push(
+                    <span
+                        key={`${lastIndex}-content`}
+                        style={{ color: "#aa3333" }}
+                    >
+                        {word}
```
MarcelGerber, line 105: I feel like it would be better to use the original casing here, as it is in the searchString.
Consider this for example:
image

Imho it should be "Output-side real ***GD***P per capita"

adminSiteServer/apiRouter.ts
```diff
@@ -948,6 +948,36 @@ apiRouter.get("/variables.json", async (req) => {
     return { variables: rows, numTotalRows: numTotalRows }
 })
 
+apiRouter.get("/variables.usages.json", async (req) => {
+    // This is a bit of a workaround - what we would like is to select
+    // all variables and do a group by on the joined charts table with a count(*)
+    // but since the variables are stored inside the json config and mysql 5.7's limited
+    // support for json and arrays we can'd do it that way - so instead we select the arrays
+    // and create a map that we use to "transpose" the table, then serialize this
```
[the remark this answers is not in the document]
danielgavrilov: There _shouldn't_ be a case where `chart_dimensions` is out of sync. If there is, we should probably fix it, because we depend on this foreign key reference to prevent data for charts being deleted.

adminSiteClient/ChartEditor.ts
```diff
@@ -10,6 +10,8 @@ import { Grapher } from "../grapher/core/Grapher"
 import { EditorFeatures } from "./EditorFeatures"
 import { Admin } from "./Admin"
 import { BAKED_GRAPHER_URL } from "../settings/clientSettings"
+import _ from "lodash"
+import { number } from "prop-types"
```
danielgavrilov, line 14: Unintentional `number` import?

adminSiteClient/ChartEditor.ts
```diff
@@ -153,6 +160,18 @@ export class ChartEditor {
             this.database.dataByNamespace.set(namespace, data as any)
         )
     }
+    async loadVariableToGraphersMap(): Promise<void> {
```
MarcelGerber, line 163: The function name here is no longer accurate

adminSiteClient/VariableSelector.tsx
```diff
@@ -297,9 +319,28 @@ export class VariableSelector extends React.Component<VariableSelectorProps> {
...
+                                                                            style={{
+                                                                                fontWeight: 500,
+                                                                                color:
+                                                                                    "#555",
+                                                                            }}
+                                                                        >
+                                                                            {v.usageCount
+                                                                                ? ` (used ${
+                                                                                      v.usageCount ??
+                                                                                      "-"
```
MarcelGerber, line 338: You can get rid of that because `v.usageCount` can never be undefined

adminSiteServer/apiRouter.ts
```diff
@@ -948,6 +948,24 @@ apiRouter.get("/variables.json", async (req) => {
...
+
+    const query = `SELECT variableId, COUNT(DISTINCT chartId) AS usageCount
+FROM chart_dimensions
+GROUP BY variableId
+ORDER BY usageCount DESC`
+
+    const rows = await db.queryMysql(query)
+
+    // JSON.stringify supports neither Map nor Set so we have to convert the Map to an object and the Sets to arrays
```
MarcelGerber, line 965: No longer accurate

PR #1020, opened 2021-09-28
Marimekko improvements

danyx23:
This PR fixes a bunch of smaller issues and adds a few quality of life improvements (see https://www.notion.so/owid/Marimekko-improvements-21afcc0a6ff3493d9becd9a0e2373bbe)

- [x]  Tooltips ATM show in the wrong place if the topmost series does not have a value for an entry
- [ ]  ~~There is one medium screen size where Y axis labels are pushed a bit into the chart~~
- [x]  Bug: on some charts there is a data merge error that crashes grapher
- [ ]  Detail: bars don't line up to exactly 100%?
- [x]  Relative mode does not take exlcuded entities into account!
- [x]  Show the relative toggle by default
- [x]  Add toggle to hide no-data values instead of showing it as a gray area
- [x]  Add toggle to invert base selection (from all entities to none) and make exclude work as an include. This is useful if you only want to select a few world regions in a dataset that includes all countries
- [x]  Allow fixing of max extent of X axis (so that e.g. for electricity/capita on Y and population on X (both including projections into 2100) the area expected in 2100 could be the full area and as you zoom through the years both the used population and the used energy/capita grows while the axis remain fixed
- [x]  Change relative scale to 10% increments instead of 20%


danyx23:
Changes are [live on snow](https://snow.owid.cloud/admin/charts) except for the last two commits - will update snow again once I have better internet connection on Wednesday

grapher/stackedCharts/MarimekkoChart.tsx
```diff
@@ -665,9 +644,16 @@ export class MarimekkoChart
...
-        const axis = config.toHorizontalAxis()
-        if (manager.isRelativeMode) axis.domain = [0, 100]
-        else axis.updateDomainPreservingUserSettings(xDomainDefault)
+        let axis = config.toHorizontalAxis()
+        if (manager.isRelativeMode) {
+            // MobX and classes  interact in an annoying way here so we have to construct a new object via
+            // an object copy of the AxisConfig class instance to be able to set a property without
+            // making MobX unhappy about a mutation originating from a computed property
+            axis = new HorizontalAxis(
+                new AxisConfig({ ...config.toObject(), maxTicks: 10 })
```
danyx23, line 653: I wondered if the second constructor option (FontSizeManager) would be an issue and indeed all the SVGs are slightly different now. I'll fix this so it works as expected

danyx23:
Sorry for the confusion with the suggested reviews but I just noticed that a change around how Axis are constructed led to all svgs shifting ever so slightly - will fix this, then ask for review again

grapher/stackedCharts/MarimekkoChart.tsx
```diff
@@ -211,13 +212,22 @@ function MarimekkoBarsForOneEntity(props: MarimekkoBarsProps): JSX.Element {
 
     let content = undefined
     if (bars.length) {
-        const allButLast = bars.slice(0, -1)
-        const last = bars[bars.length - 1]
+        const lastNonZeroBarIndex =
+            bars.length -
+            bars.reverse().findIndex((bar) => bar.yPoint.value > 0) -
+            1
```
danielgavrilov, line 218: nit: don't have to but can use [`findLastIndex`](https://lodash.com/docs/4.17.15#findLastIndex) from lodash (just needs to be exposed in `Util.ts`)

adminSiteClient/EditorMarimekkoTab.tsx
```diff
@@ -83,9 +87,25 @@ export class EditorMarimekkoTab extends React.Component<{ grapher: Grapher }> {
...
-                        placeholder="Select an entity to exclude"
+                        label={
+                            invertExcludedEntitiesList
+                                ? "Include individual entities"
+                                : "Exclude individual entities"
+                        }
+                        placeholder={
+                            "Select an entity to " +
+                            (invertExcludedEntitiesList ? "include" : "exclude")
+                        }
```
danielgavrilov, line 108: This is great and it's very likely useful for other charts. Should we extract it as a reusable component/transform? Marcel [suggested this in the previous PR](https://github.com/owid/owid-grapher/pull/965#discussion_r676579614), and I think it's a good idea if there is no rush to merge this. If there is a rush to launch, then it's fine to leave it as it is. 

Regardless, I would be happy to implement it next week, I have some time. I [checked the database](https://github.com/owid/owid-grapher/pull/965#discussion_r677195472) and only ScatterPlot and Marimekko had `excludedEntities` populated, so it's safe to enable for all charts. We could move it to the Customize tab and have it available for all chart types.

grapher/stackedCharts/MarimekkoChart.tsx
```diff
@@ -211,13 +212,22 @@ function MarimekkoBarsForOneEntity(props: MarimekkoBarsProps): JSX.Element {
...
+        const predecessors = bars.slice(0, lastNonZeroBarIndex)
+        const lastNonZero = bars.slice(
+            lastNonZeroBarIndex,
+            lastNonZeroBarIndex + 1
+        )
```
MarcelGerber, line 223: Why is this not just as below? or is it preferable to use arrays for consistency?
```ts
const lastNonZero = bars[lastNonZeroBarIndex + 1]
```

grapher/stackedCharts/MarimekkoChart.tsx
```diff
@@ -75,6 +75,7 @@ import {
     LabelCandidateWithElement,
     MarimekkoBarProps,
 } from "./MarimekkoChartConstants"
+import { ErrorValue } from "../../coreTable/ErrorValues"
```
MarcelGerber, line 78: nit: there are some imports (including this) that are not used

grapher/core/Grapher.tsx
```diff
@@ -302,6 +303,7 @@ export class Grapher
     @observable selectedEntityColors: { [entityName: string]: string } = {}
     @observable selectedEntityIds: EntityId[] = []
     @observable excludedEntities?: number[] = undefined
+    @observable invertExcludedEntitiesList?: boolean = undefined
```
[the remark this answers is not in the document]
MarcelGerber: I feel like it's tough if there's both `includedEntities` and `excludedEntities` set - what should the behaviour be in this case?

I totally agree that what we currently have is not great - especially the naming - but at least it goes with the convention to make illegal states unrepresentable.

grapher/stackedCharts/MarimekkoChart.tsx
danyx23: Yeah the idea was to make it more uniform below but it's probably more confusing than is warranted

grapher/stackedCharts/MarimekkoChart.tsx
danyx23: Nice!

adminSiteClient/EditorMarimekkoTab.tsx
danyx23: Good idea - I created a task to tackle this separately so we can publish marimekko without this: https://www.notion.so/owid/Add-include-exclude-entities-to-all-charts-724e198dfab54d4cb3512fed1dda82bc

grapher/core/Grapher.tsx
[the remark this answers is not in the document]
danyx23: Bit of a tough call. I agree that invertExcludedEntitiesList is not great and might cause some confusion. I think I will change this to an includedEntities list instead that is kept in addition (also in the UI). The logic would then always be: 
```
selection = []
if includedEntities:
  selection = includedEntites
else:
  selection = all
selection = selection - excludedEntities
``` 
Of course setting excludedEntities when includedEntities are set is not terribly useful (why not just modify includedEntities in the first place?) but I think it is well defined and will cause less confusion

grapher/stackedCharts/MarimekkoChart.tsx
```diff
@@ -75,7 +74,7 @@ import {
     LabelCandidateWithElement,
     MarimekkoBarProps,
 } from "./MarimekkoChartConstants"
-import { ErrorValue } from "../../coreTable/ErrorValues"
+import { findLastIndex } from "lodash"
```
MarcelGerber, line 77: 🚨 bundle size detective here! you know the drill

danyx23: Oh sorry, I thought this way it can do tree shaking (as opposed to import * from lodash).

MarcelGerber:
A nice-to-have, but not absolutely necessary just now:
It would be nice if there was a "Clear all" button for both the "Exclude" and "Include" sections.
This would be especially handy if we choose to unify the Marimekko and the Scatter tab.
image

Other than this this all looks good from my side, very nice work there!

MarcelGerber (approved):

danielgavrilov (approved):

PR #1029, opened 2021-10-21
enhance: Remove specific numbers from country profiles

larsyencken:
The country profile numbers are poorly formatted, have too many decimal places, and no units. We would prefer not to show them at all. So this PR prunes them from the output.


larsyencken:
@mlbrgl I added an explainer note, but not a future promise or work in progress sign -- they always end up lasting way longer than you intended, which looks bad.

PR #1040, opened 2021-11-16
feat: non-redistributable datasets

danielgavrilov:
Notion issue: [Implement config fields to show "not allowed to download dataset" notice](https://www.notion.so/Implement-config-fields-to-show-not-allowed-to-download-dataset-notice-ac27ec5ac7b143c0b788465272bf7f64)
Project: [Remove data download from IHME charts/explorers](https://www.notion.so/Remove-data-download-from-IHME-charts-explorers-9b2b3edab0d24dc28675942e14050a2e)

This PR:
- Adds a **`nonRedistributable` column** in the `datasets` database table
- Passes the flag [**down to the variable level**](https://github.com/owid/owid-grapher/blob/1f3f177f8c6565636df7dddc53a9302e00a3c56b/clientUtils/OwidVariable.ts#L64) to make it available in Grapher, and to make it possible to define it in explorers.
- If **at least one variable** in the chart is non-redistributable, the full data download is prevented by showing a notice:
   
   Screenshot 2021-11-16 at 16 48 53

- Some refactoring to improve server-side types (moved variable types from `grapher` to `clientUtils` in order to make them available on the server)


danielgavrilov:
Potentially, we might want to enable downloads of the dataset with the non-redistributable variables dropped. I didn't do this for now in order to save time (our "appetite" for this project is low).

If we had an explorer with some IHME here and there, that would likely be a good reason to implement the above. But until then, this should be enough?

danyx23 (approved):
I haven't run it but I looked over the changes and it LGTM

danielgavrilov:
Thanks for the review @danyx23! 

I now realised I could've made your job easier by splitting this PR into two: refactor (up to https://github.com/owid/owid-grapher/pull/1040/commits/082de5d9ef398887c304554b10f1550851724852)  + feature PRs... will be more conscious of that next time!

PR #1044, opened 2021-11-19
feat(admin): add bulk chart revision tool

bnjmacdonald:
Adds two experimental features to the admin UI:

1. A "Bulk downloads" page, with a link to download all charts in csv format.
2. A "bulk chart revisions" tool, which allows OWID staff to make changes to a large number of charts by uploading a csv file containing chart revisions (where each row in the csv file is a chart and each column is a field, such as title, subtitle, etc). 
    - Once the csv is uploaded, the chart revisions contained within are queued up in the chart approval tool for review.

Both features are deployed on `nightingale`: 
1. [Bulk downloads page](https://nightingale.owid.cloud/admin/bulk-downloads)
2. [Import tool for bulk chart revisions](https://nightingale.owid.cloud/admin/suggested-chart-revisions/import)


adminSiteServer/apiRouter.ts
```diff
@@ -664,6 +730,284 @@ apiRouter.get(
...
+                )
+
+            // some chart configs do not have an `id` field, so we check for it
+            // and insert the id here as needed. This is important for the
+            // lodash.isEqual condition later on.
+            // for (const [id, config] of Object.entries(originalConfigs)) {
+            //     if (config.id === null || config.id === undefined) {
+            //         config.id = id
+            //     }
+            // }
```
danyx23, line 921: Why did you decide to comment this out? It seems to me that this should be safe to do and it would fix the issue with chart 5327 (and the 160 other charts that don't have the id set correctly)

danyx23 (approved):
Looks good to me, nice work! The only open question is if setting the config.id field to the id if the former is null should be enabled again - I think so, but let's discuss on slack

adminSiteServer/apiRouter.ts
bnjmacdonald: fixed in 82c35fa. I had written it originally as a quick fix, then commented it out thinking it would be better to fix the underlying bug before this PR (which didn't happen, and then I forgot this about this commented block).

PR #1052, opened 2021-12-15
Bulk fastt editor

danyx23:
This is not quite ready to merge but getting close. The main missing items are a few relatively easy usability features (paging, more search fields) so that authors can have a first look.

More structured feedback and discussion can take place on the [notion card for the Bulk FASTT editing project](https://www.notion.so/owid/Bulk-FASTT-editing-c6df808a1e1149a5997581fade8dc972#a877057269934c25a2271f6f788093ad).

I think that the review can already start on this PR. The smaller open issues that make this a draft PR instead of a proper one from my side:
- [x] Paging (currently only the first 50 variables for a search term can be seen)
- [x] Filtering on dataset name at least (in addition to variable name that exists now)
- [x] Searching word by word instead of exact phrase (so it works like it does in other admin tools)
- [ ] Array editors editable? Currently columns that store array values are read only - this includes the country selection that is necessary for line charts to display something. You can set countries with the "add country" button but this in then not persisted
- [ ] ~Use display settings from (bulk imported) variable (though maybe currently these are implicitly take into account - I have to check)~
- [ ] ~Tune preview size? Not sure if the current size of 500x400 px is too small~
- [x] Check if this will work for non-published charts on prod (on staging servers data is only available for variables used in existing charts)


clientUtils/patchHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+            if (json !== undefined) newObject = { ...json }
+            const currentValue = newObject.hasOwnProperty(currentPart)
+                ? newObject[currentPart]
+                : undefined
+            const updatedValue = setValueRecursive(
+                currentValue,
+                pointer.slice(1),
+                newValue
+            )
+            if (updatedValue === null) delete newObject[currentPart]
```
danielgavrilov, line 25: As I understand, setValue with `null` value means delete the property?

Does this come from an external spec, or is it our interpretation?

I'm wondering because `null` is a valid JSON value, and in some exceptional circumstances we may want to use it? On the other hand, `undefined` isn't a valid JSON value, so it _seems_ better to use that to delete properties.

clientUtils/patchHelper.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        pointer: ["a", "0", "b"],
+        newValue: 2,
+        jsonValueBefore: {},
+        jsonValueAfter: { a: [{ b: 2 }] },
+    },
+    {
+        pointer: ["a", "b", "0"],
+        newValue: 2,
+        jsonValueBefore: {},
+        jsonValueAfter: { a: { b: [2] } },
```
danielgavrilov, line 35: If the `pointer` was instead `["a", "b", "1"]`, what would the expected output be? Or is that not allowed?

clientUtils/schemaProcessing.ts
```diff
@@ -0,0 +1,277 @@
...
+        // If the current schema fragment describes a normal object
+        // then do not emit anything directly and recurse over the
+        // described properties
+        if (
+            schema.hasOwnProperty("type") &&
+            schema.type === "object" &&
+            schema.hasOwnProperty("properties")
+        ) {
+            for (const key of keys(schema.properties)) {
+                const newPath = `${pointer}/${key}`
```
danielgavrilov, line 127: Is it safe to construct the path this way? Does it need to be escaped?

danielgavrilov (commented):
Partial review – `VariablesAnnotationPage.tsx` and `apiRouter.ts` left, will do them tomorrow.

danielgavrilov:
There are `.lsp` and `..clj-kondo` directories that seem like they shouldn't be here?

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,928 @@
...
+        const filterOperations = [
+            searchFieldStringToFilterOperations(
+                variableNameFilter ?? "",
+                new SqlColumnName(SQL_COLUMN_NAME_VARIABLE_NAME)
+            ),
+            searchFieldStringToFilterOperations(
+                datasetNameFilter ?? "",
+                new SqlColumnName(SQL_COLUMN_NAME_DATASET_NAME)
+            ),
+        ].filter((item): item is Operation => item !== undefined)
```
danielgavrilov, line 206: nit: there's an `excludeUndefined()` utility that we use in cases like this, it has the type guard too.

danielgavrilov:
Got through all of it now! I have reviewed the code, but haven't played around with it.

Some design concerns:
- Compared to the Admin tabs, it's harder to find all possible settings that can be changed, and figure out what they do.
- Even when you know what fields you're looking for, there's a lot of horizontal scrolling to navigate to them. Not sure what the solution could be – maybe changed columns float to the left, or authors can reorder columns as they like.
- Some of the map chart properties (e.g. selected year or color scales) are duplicated on the chart tab (selected years and color scales also exist for the chart tab). Are authors supposed to keep those in sync? Or do we have default behaviour that defining them in one place is enough?

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,511 @@
...
+export const SQL_COLUMN_NAME_VARIABLE_NAME = "variables.name"
+export const SQL_COLUMN_NAME_DATASET_NAME = "datasets.name"
+export const SQL_COLUMN_NAME_NAMESPACE_NAME = "namespaces.name"
+
+export class SqlColumnName implements Operation {
+    static allowedColumnNamesAndTypes: Map<string, ExpressionType> = new Map([
+        [SQL_COLUMN_NAME_VARIABLE_NAME, ExpressionType.string],
+        [SQL_COLUMN_NAME_DATASET_NAME, ExpressionType.string],
+        [SQL_COLUMN_NAME_NAMESPACE_NAME, ExpressionType.string],
+    ])
```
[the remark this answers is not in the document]
danyx23: Yes this is a temporary solution. I'm adding this comment now to explain my thinking:
// NOTE: this is a temporary solution for the beginning of using S Expressions for filtering.
    //       Once we start using this in more places, declaring the whitelist of columns that are
    //       valid to query needs to come from the piece of code that wants to use the expression.
    //       I think we should not start allowing expressions that traverse foreign keys and instead
    //       create views and whitelist column names in those if we need more complex filters

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,928 @@
...
+import { toStream, fromStream, IStreamListener } from "mobx-utils"
+import { stringifyUnkownError, differenceOfSets } from "../clientUtils/Util"
+import { queryParamsToStr } from "../clientUtils/urls/UrlUtils"
+
+function parseVariableAnnotationsRow(
+    row: VariableAnnotationsResponseRow
+): VariableAnnotationsRow {
+    return {
+        ...row,
+        grapherConfig: JSON.parse(row.grapherConfig),
```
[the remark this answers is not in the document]
danyx23: Yes - we should move the interface type definitions so that at a top level we declare them and then downstream projects can use them all

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,928 @@
...
+    "/dimensions/0/shortUnit",
+]
+
+/** All the parameters we need for making a fully specified request to the /variable-annotations
+    endpoint. When any of these fields change we need to trigger a new request */
+interface FetchVariablesParameters {
+    pagingOffset: number
+    filterQuery: Operation
+    sortByColumn: string // sort is currently ignored but here for future use
+    sortByAscending: boolean // sort is currently ignored but here for future use
```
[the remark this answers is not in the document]
danyx23: Yes I think we will want to add this at some point but I would not add it yet. Should be a straightforward change in the future.

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,928 @@
...
+            ...json,
+            bounds: new Bounds(0, 0, 500, 400),
+            getGrapherInstance: (grapher: Grapher) => {
+                this.grapher = grapher
+            },
+        }
+        if (this.grapherElement) {
+            {
+                this.grapher.setAuthoredVersion(newConfig)
+                this.grapher.reset()
```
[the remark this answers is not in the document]
danyx23: Yes but it is also not the worst way of doing it for now - we want to keep some state the same and not recreate everything here. But it is something to revisit for us and make a bit more clearer (and maybe split reset into two functions catering for different use cases)

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,928 @@
...
+        if (richDataRows === undefined) return
+
+        // Get the grapherConfig of the currently selected row and then
+        // merge it with the necessary partial information (e.g. variableId field)
+        // to get a config that actually works in all cases
+        const config = richDataRows[row].grapherConfig
+        const finalConfigLayer = getFinalConfigLayerForVariable(
+            richDataRows[row].id
+        )
+        const mergedConfig = merge(config, finalConfigLayer)
```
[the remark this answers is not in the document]
danyx23: That is a valid concern - I think in practice the way things are used it shouldn't affect us much and in the mid term I would like to migrate the grapher config structure to a place where it is more "correct by construction" to help with this. E.g. if currently we have 3 arrays that have to be in sync it would be better to have one array with a nested structure of 3 values so that this problem can't come up. Do you think it is better to invest more into verification here that this value is correct?

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,928 @@
...
+    patches: VariableAnnotationPatch[]
+}
+
+const PAGEING_SIZE: number = 50
+const enum Tabs {
+    EditorTab = "EditorTab",
+    FilterTab = "FilterTab",
+}
+
+const ALL_TABS = [Tabs.FilterTab, Tabs.EditorTab]
```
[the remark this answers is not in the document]
danyx23: This was a const enum which doesn't have a runtime representation but probably the upside of that is not worth the risk of ALL_TABS getting out of sync. Switched to Object.values(Tabs) as suggested

clientUtils/schemaProcessing.ts
```diff
@@ -0,0 +1,277 @@
...
+
+export function extractFieldDescriptionsFromSchema(
+    schema: unknown
+): FieldDescription[] {
+    if (isPlainObjectWithGuard(schema)) {
+        const dereferenced = dereference(schema)
+        const fieldDescriptions: FieldDescription[] = []
+        extractSchemaRecursive(dereferenced, "", fieldDescriptions)
+        return fieldDescriptions
+    } else throw "Schema was not an object!"
```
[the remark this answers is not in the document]
danyx23: Thanks!

clientUtils/schemaProcessing.ts
danyx23: The assumption here is that we control the schema so I think not - am I missing something?

clientUtils/patchHelper.test.ts
danyx23: It would be the same expected value - anything beyond the length of the current array is just appended

db/chart-migrations/migrate.ipynb
```diff
@@ -0,0 +1,343 @@
+{
```
MarcelGerber, line 1: We can probably remove this file now right?

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,515 @@
+import parse = require("s-expression")
+import pointer from "json8-pointer"
+import { isArray, tail, without } from "lodash"
+export type SExprAtom = string | String | SExprAtom[] // eslint-disable-line @typescript-eslint/ban-types
```
MarcelGerber, line 4: hm, why do we need `String` (uppercase) in here?

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,515 @@
...
+        super()
+        this.value = v
+    }
+    expressionType = ExpressionType.string
+    value: string
+    escapedValue(): string {
+        return this.value.toString().replace("'", "''")
+    }
+    toSql(): string {
+        return `'${this.escapedValue()}'` // escape single quotes to avoid SQL injection attacks :)
```
MarcelGerber, line 92: Looking at [PHP's `mysqli_real_escape_string`](https://www.php.net/manual/en/mysqli.real-escape-string.php), it says
> Characters encoded are `NUL (ASCII 0), \n, \r, \, ', ", and Control-Z`.

So we should also ensure that there really cannot be any SQL Injection happening.

---
Your current implementation is definitely susceptible to `abc\'; DROP TABLE students; --`

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,515 @@
...
+        else path += `.${fragment}`
+    }
+    return path
+}
+
+export class JsonPointerSymbol implements Operation {
+    // Json pointer is actually less restrictive and allows pretty much all unicode
+    // code points but for the foreseeable future we will only have \w and \d, _ and -
+    // characters in actual use. Tilde is from json pointer to escape / inside field names
+    // We also disallow empty jsonPointer for now even though they are legal (if you)
```
MarcelGerber, line 115: nit:
```suggestion
    // We also disallow empty jsonPointer for now even though they are legal (if you
```

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,515 @@
...
+    // want to enable them then make sure that the parsing decisions outside of here
+    // all correctly allow this)
+    static jsonPointerRegex: RegExp = /^\/[\w\d/-~]+$/
```
MarcelGerber, line 118: Be careful with this regex - `/-~` is a range, allowing every ASCII character between `/` and `~`.
Instead, do `\-/~_` (so we also include `_`).

MarcelGerber (commented):
Not yet done with this review, there's more to come!

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,927 @@
...
+        the "Next page" quickly twice then this field will be 2 paging offsets higher
+        than currentPagingOffset until the request to retrieve that data is complete
+        at which point they will be the same */
+    @observable desiredPagingOffset: number = 0
+    // Sorting fields - not yet used - TODO: implement sorting
+    @observable sortByColumn: string = "id"
+    @observable sortByAscending: boolean = false
+    disposers: Disposer[] = []
+    /** Computed property that combines all relevant filter, paging and offset
+        properties into a FetcHVariablesParameter object that is used to construct
```
MarcelGerber, line 188:
```suggestion
        properties into a FetchVariablesParameter object that is used to construct
```

adminSiteClient/VariablesAnnotationPage.tsx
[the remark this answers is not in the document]
MarcelGerber: Sounds valid to me, and an absolute thumbs up to `it would be better to have one array with a nested structure of 3 values so that this problem can't come up`!

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,927 @@
...
+    ): Promise<void> {
+        const reversedPatches = patches.map((patch) => ({
+            ...patch,
+            newValue: patch.oldValue,
+            oldValue: patch.newValue,
+        }))
+        await this.sendPatches(reversedPatches)
+    }
+
+    /** Performs an undo step - i.e. applies it to the richDataRows object,
```
MarcelGerber, line 341: I find the naming as an UndoStep pretty confusing - it's an undoable step, but it's not being undone at this point.

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,927 @@
...
+            )
+            const fieldDescsMap = new Map(
+                fieldDescsFiltered.map((desc) => [desc.pointer, desc])
+            )
+            const importantDescs = IMPORTANT_COLUMNS.map((pointer) =>
+                fieldDescsMap.get(pointer)
+            )
+            const remainingDescs = differenceOfSets([
+                new Set(fieldDescsMap.keys()),
+                new Set(IMPORTANT_COLUMNS),
```
MarcelGerber, line 548: You can actually incorporate the `HIDDEN_COLUMNS` filtering into this call here.

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,927 @@
...
+        const invalidTypeAssignment = match<
+            [FieldType | FieldType[], string],
+            boolean
+        >([columnDefinition.type, typeof firstNewValue])
+            .with([FieldType.string, "string"], (_) => false)
+            .with([FieldType.number, "string"], (_) =>
+                Number.isNaN(Number.parseFloat(firstNewValue))
+            )
+            .with([FieldType.integer, "string"], (_) =>
+                Number.isNaN(Number.parseFloat(firstNewValue))
```
MarcelGerber, line 660: should be `parseInt` for integer, right?

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,927 @@
...
+        )
+        if (!allChangesSameType) {
+            console.error("Not all changes were of the same type?")
+            return
+        }
+
+        // Check if the type matches the target (e.g. to avoid a string value ending
+        // up in a text field). This is hard to nail down fully but we try to avoid
+        // such mistakes where possible
+        const firstNewValue = changes[0][3]
```
MarcelGerber, line 650: where does the `[0][3]` come from? should add a comment here.

MarcelGerber (commented):
Still not quite finished with the review but here is a bunch of comments again !

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -0,0 +1,927 @@
...
+            this.desiredPagingOffset + PAGEING_SIZE < (this.numTotalRows ?? 0)
+        ) {
+            console.log("increasing desired paging offset")
+            this.desiredPagingOffset = this.desiredPagingOffset + PAGEING_SIZE
+        }
+    }
+
+    renderPagination(): JSX.Element {
+        const { numTotalRows, currentPagingOffset } = this
+        const currentStartLabe = currentPagingOffset + 1
```
MarcelGerber, line 820:
```suggestion
        const currentStartLabel = currentPagingOffset + 1
```

MarcelGerber:
I made it! (except for that I didn't review `schemaProcessing.ts` because my brain apparently went into shutdown when I tried to do that, so sorry bout that!)
This is great, and structured very very well. I also like the functional bits, but I might be a bit biased there :)

Also, I believe for this massive PR it can be a good thing to mark feedback that you have either incorporated or disregarded as "Resolved", just so we can keep track of what's been done and what is left to do. I marked some of Gavrilovs feedback as resolved in the cases where I was sure that it was incorporated already.

---
I share some of @danielgavrilov's UX concerns:
- Navigating horizontally is hard, just because there are so so many columns. One solution could be a filter box for columns, such that one can easily search for `note` for example, and don't need to push their luck trying to find it. A filterable column chooser, basically.
- Also, I fear the names are not always clear to authors. For example, if they see `/map/colorScale/legendDescription` they might not know what that is - I don't know it either. To make things worse, the admin names are sometimes a bit different from the keys in code.
  One solution here could be to include a description for most fields, which could be displayed on hover over the column name.

Also, some suggestions in case you didn't spend enough time on this project yet :)
- The currently-selected row should be highlighted in a different color, so that one can see what dataset they're looking at.
- The dataset name and variable name could be links, linking to the detail pages where the dataset/variable metadata can be changed.
- The EditorTab and FilterTab seem quite unnecessary at the moment, but I guess that's future-proofing for now?

clientUtils/jsonPatchUtils.ts
```diff
@@ -0,0 +1,33 @@
+export type OpPatch =
```
MarcelGerber, line 1: remove this file; it's no longer in use

clientUtils/patchHelper.test.ts
danielgavrilov: Let's add a test case for that?

clientUtils/schemaProcessing.ts
```diff
@@ -0,0 +1,276 @@
...
+        console.error("Unexpected type/object", pointer)
+    }
+}
+
+function objectMap(object: any, mapFn: (result: any) => any): any {
+    return Object.keys(object).reduce(function (result: any, key) {
+        result[key] = mapFn(object[key])
+        return result
+    }, {})
+}
```
danielgavrilov, line 231: Is this `mapValues` in lodash?

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,511 @@
+import parse = require("s-expression")
```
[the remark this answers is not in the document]
danyx23: I looked into this but the typescript fork changes semantics and doesn't allow distinguishing between string literals and symbols. I think it is quite useful to have that distinction so I decided to stay with the untyped one

clientUtils/SqlFilterSExpression.ts
danyx23: Good idea!

clientUtils/SqlFilterSExpression.ts
danyx23: Very nice catch! I think if we add a replace from \ to \\ that should cover all cases though?

package.json
```diff
@@ -139,6 +139,8 @@
         "html-to-text": "^5.1.1",
         "js-base64": "^3.4.5",
         "js-cookie": "^2.2.1",
+        "json2csv": "^5.0.6",
```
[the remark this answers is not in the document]
danyx23: Must have been - thanks for catching!

clientUtils/SqlFilterSExpression.ts
danyx23: Ah that's what you get for not checking again after modifications - I had \d-~ before in which case you don't need to escape the dash. Added your suggestion

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,515 @@
...
+
+export class BinaryLogicOperation extends BooleanOperation {
+    constructor(operator: BinaryLogicOperators, operands: BooleanOperation[]) {
+        super()
+        this.operator = operator
+        this.operands = operands
+    }
+
+    operator: BinaryLogicOperators
+    operands: BooleanOperation[]
```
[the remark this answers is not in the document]
danyx23: Nice, didn't know about those!

clientUtils/SqlFilterSExpression.ts
```diff
@@ -0,0 +1,515 @@
...
+                    new ArithmeticOperation(op, args as NumericOperation[]),
+            },
+        ] as const
+)
+const equalityOperatorInfos = allEqualityOperators.map(
+    (op) =>
+        [
+            op.toString(),
+            {
+                arity: Arity.nary,
```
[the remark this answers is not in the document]
danyx23: Good catch

adminSiteServer/apiRouter.ts
```diff
@@ -1333,6 +1349,79 @@ apiRouter.get("/variables.json", async (req) => {
...
+        let offset = req.query.offset ? Number.parseInt(req.query.offset) : 0
+        if (Number.isNaN(offset) || offset === undefined) offset = 0
+
+        // Note that our DSL generates sql here that we splice directly into the SQL as text
+        // This is a potential for a SQL injection attack but we control the DSL and are
+        // careful there to only allow carefully guarded vocabularies from being used, not
+        // arbitrary user input
+        const whereClause = filterSExpr?.toSql() ?? "true"
+        const resultsWithStringGrapherConfigs =
+            await db.execute(`SELECT variables.id as id, variables.name as name, variables.grapherConfig as grapherConfig, datasets.name as datasetname, namespaces.name as namespace
```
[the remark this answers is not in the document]
danyx23: I switched to it but I don't see any guarantees that this would disallow non-select queries. Am I missing something or is this not implemented yet?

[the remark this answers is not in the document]
MarcelGerber: Ah yes, I was wrong then. I thought that the MySQL server would only handle SELECTs when using `query`, but that is apparently not the case. Sorry for the confusion!

clientUtils/patchHelper.ts
```diff
@@ -0,0 +1,106 @@
...
+                        newArray.splice(currentPartAsNumber, 1)
+                    }
+                    // otherwise ignore it since it is not in the array yet anyhow
+                } else {
+                    if (currentPartAsNumber < newArray.length) {
+                        newArray[currentPartAsNumber] = updatedValue
+                    } else newArray.push(updatedValue)
+                }
+                return newArray
+            } else {
```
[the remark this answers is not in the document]
danyx23: Added comments to clarify

adminSiteClient/VariablesAnnotationPage.tsx
danyx23: Fair enough - I renamed it to Action instead, is that better?

clientUtils/SqlFilterSExpression.test.ts
```diff
@@ -0,0 +1,149 @@
...
+        checkSql("(>= 4 1)", "(4 >= 1)")
+        checkSql("(AND true false false)", "(true AND false AND false)")
+        checkSql("(OR true false false)", "(true OR false OR false)")
+        checkSql("(NOT true)", "(NOT true)")
+        checkSql(
+            '(CONTAINS /map/projection "Europe")',
+            "(JSON_EXTRACT(grapherConfig, \"$.map.projection\") LIKE '%Europe%')"
+        )
+    })
+
```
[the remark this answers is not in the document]
danyx23: Good idea - added some more tests for wrong s-expressions

danyx23:
Thanks a lot @danielgavrilov and @MarcelGerber for your excellent feedback! Merging this now 🎉

PR #1062, opened 2022-01-03
svgTester: prettier SVGs & file names

danielgavrilov:
This makes two changes:
- All SVGs are prettified. This makes the diffs more easily understandable, and the GitHub text diff view becomes a bit more useful.
- The SVG file names now contain only the slug. Before, they contained chart version and dimensions, but this naming scheme is only used for grapher embeds inside OWID articles (it's more of a CMS feature than a Grapher feature). The standalone Grapher SVGs only use the slug (e.g. https://ourworldindata.org/grapher/exports/excess-deaths-cumulative-economist.svg)

Not sure if there are some undesired side effects from both of these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ serverSettings.json
 localBake/
 /grapherData
 /grapherSvgs
+/differentGrapherSvgs
--- a/devTools/svgTester/utils.ts
+++ b/devTools/svgTester/utils.ts
@@ -16,6 +16,7 @@ import * as util from "util"
 import { GrapherInterface } from "../../grapher/core/GrapherInterface"
 import { TESTING_ONLY_reset_guid } from "../../clientUtils/Util"
 import _ from "lodash"
+import prettier from "prettier"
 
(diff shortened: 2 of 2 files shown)


danielgavrilov:
Probably at least one undesired side effect is that it makes it slower. So it's a question whether the performance cost is worth it.

danielgavrilov:
Ok I'm leaning towards the performance impact not being worth it (it has been running for >25min already and not finished). 

Potentially it could off by default, and only enabled with a flag? 

I currently find it useful to ignore all the non-consequential diffs for #1061, but this could also be achieved in other ways. The most robust way might be to do visual diffs – convert to PNG and do pixel comparison, which would take even longer, but if like in this case there are >1,000 diffs, I would not mind waiting for a long time to make sure I'm not breaking something.

danielgavrilov:
I think a [Bulk PNG comparison tool](https://www.notion.so/Bulk-PNG-comparison-tool-fc175efa678346279247659666b4ad84) is a better approach for the problem I was trying to solve, and the changes in this PR are not worth the performance costs.

PR #1063, opened 2022-01-04
fix(legends): omit hidden bins in numeric legend

danielgavrilov:
The "Hide" checkbox otherwise has no effect:

Screenshot_2022-01-04_at_15_07_51


danyx23 (approved):
LGTM

PR #1065, opened 2022-01-05
feat(scatter): plot time on X axis by default if no columns specified

danielgavrilov:
[Charlie needs a ScatterPlot with Time on the X axis](https://owid.slack.com/archives/C5BDCB2R3/p1641411409032800?thread_ts=1641405010.031600&cid=C5BDCB2R3).

Surprisingly, setting the default column (if none is specified) to be `time` _just works_:
Screenshot 2022-01-05 at 19 47 35

And svgTester reports no differences.

What do you think about the approach?

### DIFF
--- a/grapher/scatterCharts/ScatterPlotChart.test.ts
+++ b/grapher/scatterCharts/ScatterPlotChart.test.ts
@@ -58,6 +58,8 @@ it("doesn't show 'No data' bin when there is no color column", () => {
 it("can remove points outside domain", () => {
     const manager: ScatterPlotManager = {
         table: SynthesizeFruitTable(undefined, 2),
+        yColumnSlug: SampleColumnSlugs.Fruit,
+        xColumnSlug: SampleColumnSlugs.Vegetables,
     }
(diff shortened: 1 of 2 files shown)


MarcelGerber (approved):
Oh this is surprisingly easy, great. I'm all for it!

danyx23:
Interesting :). I played a bit with this and for many metrics this then becomes another kind of line chart which is interesting. 

    If a scatter plot is vs time and you connect the dots, is there any difference to a line chart? 
    -- old buddhist proverb

Let's merge it since we have a use case for it :)

danielgavrilov:
I feel a little bit scared by this one, in case it happens that we need to make tweaks (e.g. hide "time" axis label?), because there is no way to control the time column from the Admin, everything has to be ad-hoc in the code.

But on the other hand, we can learn whether, and in which cases, a chart like this is useful. Maybe this is the only chart we ever make.

I think we should treat it as an experiment and not implement features unless they can be implemented in some non-ad-hoc way.

PR #1066, opened 2022-01-05
enhance: retry fetch requests

danielgavrilov:
We have a _lot_ of fetch errors on Bugsnag. I am guessing some good fraction of them are transient network errors (but haven't investigated much), so implementing retries (up to 3 times) would decrease the fetch error rate.

Deployed on [playfair](https://playfair-owid.netlify.app).

Note that `fetch()` doesn't throw when the server responds with an error code, (`404` or `500`):

> -   The Promise returned from `fetch()` **won't reject on HTTP error status** even if the response is an HTTP 404 or 500. Instead, as soon as the server responds with headers, the Promise will resolve normally (with the [`ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok "ok") property of the response set to false if the response isn't in the range 200--299), and it will only reject on network failure or if anything prevented the request from completing.

— [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)

### DIFF
--- a/clientUtils/Util.test.ts
+++ b/clientUtils/Util.test.ts
@@ -204,17 +204,23 @@ describe(retryPromise, () => {
 
     it("resolves when promise succeeds first-time", async () => {
         const promiseGetter = resolveAfterNthRetry(0, "success")
-        expect(retryPromise(promiseGetter, 1)).resolves.toEqual("success")
+        expect(retryPromise(promiseGetter, { maxRetries: 0 })).resolves.toEqual(
+            "success"
+        )
     })
 
     it("resolves when promise succeeds before retry limit", async () => {
         const promiseGetter = resolveAfterNthRetry(2, "success")
-        expect(retryPromise(promiseGetter, 3)).resolves.toEqual("success")
+        expect(retryPromise(promiseGetter, { maxRetries: 2 })).resolves.toEqual(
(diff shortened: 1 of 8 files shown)


MarcelGerber (approved):
Oh awesome 🙌 
I got one of these fetch errors when I was on a train a few days ago and (probably) moved between cell towers, and I bet I'm not the only one at that.
This won't be handling proper disconnects though - where the cell connection goes out for a few seconds - but it's certainly a good start and way better than before. Ideally we would want to have a retry button that the user can click.
But that's all for the future - let's go for this and make the world better one step at a time 🚀

danyx23:
Good idea! We could include an increasing timeout in between, something like `await sleep(300 * numRetries)` in the catch block. The reason for doing it this way would be that if you are getting a network error that throws quickly (e.g. because the wifi is properly down) then retrying a second later *might* already work whereas retrying 3 times with no delay will not add much. I

danyx23 (approved):

danielgavrilov:
Good idea – I added exponentially increasing delay, seemed reasonable? (So 1st retry is `100ms` wait, 2nd is `200ms`, 3rd `400ms`, 4th `800ms`...)

PR #1071, opened 2022-01-18
feat(explorers): allow configuring a related question

danielgavrilov:
Notion issue: [Add related link at the bottom of data explorer](https://www.notion.so/Add-related-link-at-the-bottom-of-data-explorer-a06d10a7389e4541b255475a096bc784)

This came up recently, [Max wanted a related question on an explorer](https://owid.slack.com/archives/CLDM2AWCD/p1641496014000100?thread_ts=1641294116.014600&cid=CLDM2AWCD).

In the Grapher config it's possible to add multiple questions, but in the Explorer only a single one. But there are no Graphers that have defined more than one, so I think this is fine.

<details>
<summary>SQL to check for <code>relatedQuestions</code> config:</summary>
<pre>
SELECT id, config->"$.relatedQuestions", JSON_LENGTH(config->"$.relatedQuestions")
FROM charts
WHERE config->"$.relatedQuestions" IS NOT NULL
</pre>
</details>

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -305,8 +305,15 @@ export class Explorer
         const grapher = this.grapher
         if (!grapher) return
         const grapherConfigFromExplorer = this.explorerProgram.grapherConfig
-        const { grapherId, tableSlug, yScaleToggle, yAxisMin, facetYDomain } =
-            grapherConfigFromExplorer
+        const {
+            grapherId,
+            tableSlug,
+            yScaleToggle,
+            yAxisMin,
+            facetYDomain,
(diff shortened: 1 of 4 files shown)


mlbrgl (approved):
> In the Grapher config it's possible to add multiple questions, but in the Explorer only a single one. But there are no Graphers that have defined more than one, so I think this is fine.

As far as I remember, I thought it'd be better to anticipate storing multiple related questions, even if the grapher admin UI only lets you add one so far. So it's equally good to have that limitation in the explorer UI.

Thanks for narrowing the gap, LGTM!

PR #1117, opened 2022-01-29
Add Gitpod

shafy:
This PR adds a Gitpod configuration (`.gitpod.yml`), which allows opening up this code base in Gitpod with one click. It greatly simplifies contributing to projects by spinning up a dev environment with everything set up in the cloud.

Screenshot 2022-01-28 at 16 00 01

You can start the Gitpod environment by simply prefixing a GitHub URL by `gitpod.io/#`. For example, [this link](https://gitpod.io/#https://github.com/owid/owid-grapher/pull/1117) opens this branch/PR in Gitpod. If you're not familiar with Gitpod, check out their [website](https://gitpod.io). It's similar to GitHub codespaces, but open-source and more user-friendly (in my mind).

One caveat with this setup is that I chose MySQL 8.0 (default for Ubuntu 20.04), not 5.7 as recommended in the readme. I can switch to 5.7, but this will complicate the setup a bit and seeing as MySQL reaches EOL end of next year, I thought this should be fine. Seems to work fine. Let me know if this is not good.

I've also added `/owid-content` to `.gitignore`, since that's the folder where `https://github.com/owid/owid-content` is cloned into. I'm happy to remove it again if I'm missing something.

If you want, I can add a short section to the readme that explains this development setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ serverSettings.json
 localBake/
 /grapherData
 /grapherSvgs
+/owid-content
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,45 @@
+image: gitpod/workspace-mysql
+
+ports:
+    # express server
+    - port: 3030
+      onOpen: open-browser
+    # mysql server
+    - port: 3306
+      onOpen: ignore
+    - port: 33060
+      onOpen: ignore
+    # webpack server
+    - port: 8090
+      onOpen: ignore
+
+tasks:
+    # set everything up
+    - init: |
+          gp await-port 3306 && ./db/downloadAndCreateDatabase.sh
+          mysql -e "ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '';"
+          nvm install
+          npm install -g yarn
+          git clone https://github.com/owid/owid-content
+          yarn install
+          cp .env.example .env && echo $'\n' >> .env
+          tsc -b
+          gp sync-done initial-setup
+      # setting env variables required for Gitpod, because it doesn't run on localhost
+      command: |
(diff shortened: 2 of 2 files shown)


larsyencken (approved):
Hey Can, nice work! It's surprisingly little code in the end to be able to set up this kind of development environment. We'll give it a try.

danyx23:
Hi @shafy! We are restructuring our documentation at the moment and I tried added an a short section [here](https://github.com/owid/owid-grapher/tree/developer-docs-improvements#initial-development-setup) on running this with Gitpod. For this I just tried running it but I didn't see how to access the admin at port 3030 (it says "not served"). I used the direct link to master for gitpod, i.e. https://gitpod.io/#https://github.com/owid/owid-grapher . Is this the way this is supposed to be used? I think this would be a really neat setup to support so if you have any tips that would be great. Thanks!

danyx23:
Me again - I was just too impatient 😅 - the database just took a while longer to load than on a local machine and the admin UI wasn't up yet. All good, I'll mention this in the docs!

shafy:
Hey @danyx23 I'm glad you figured it out. But in general, what takes long is downloading the dataset to the machine. So Gitpod does this in the prebuild stage. If you start with Gitpod from an existing branch, it should be ready to go (if this is not the case, let me know and I'll have another look).  If you just made a merge or PR, it will re-run the rebuild.

Regarding URL, I guess you figured it out, too. But the general form is `https://<port>-<URL of current cloud machine>`. For example: `https://3030-subdomain.gitpod.io/`.

Doc looks good to me. I guess people who don't know what Gitpod but want to use it can always have a look at the Gitpod docs for basic things (like the port/url thing).

Let me know if you need anything else.

PR #1120, opened 2022-01-31
Docker compose setup

danyx23:
This PR creates a fully working setup of mysql, the grapher and wordpress mysql databases, php running wordpress and nginx to handle other files on the wordpress side. It is currently set up to be an easier alternative to Lando. I haven't tried using it for development of our OWID wordpress plugin or similar so for those cases Lando might still be preferrable but for simply getting a full dev environment running this should be an easier alternative.


danyx23:
I'm merging this as is an additive change for now but it already helps in my local dev setup. Happy for any comments though still :)

Issue #1108, opened 2022-01-27
Bulk FASTT column management


danyx23:
The bulk FASTT editor has a a high number of columns. Usually you only need a few of those at any given moment. It should therefore be possible to:

- [x] Show/hide columns
- [x] Reorder them
- [x] Mark the first N columns as frozen (so they stay visible when scrolling)
- [x] Filter columns using a text box (i.e. type map and only columns containing that text fragments are shown in the selection list)
- [x] Have a drop down with common groupings (e.g. title, subtitle, footer; or color scale related things in both the map and the normal chart; or axis related columns like min and max value)

The implementation will most likely be a list like control in the sidebar that shows all columns below each other with a text filter box. Ever row has an eye icon to toggle visibility and drag handle for reordering. A numeric input is used to freeze the first N columns.

PR #1131, opened 2022-02-03
Bulk fastt columns editor

danyx23:
First draft for column editing in the bulk FASTT editor


danyx23:
This is work to implement #1108. The basic functionality is working but the current mechanism of showing and hiding columns by changing the HotColumns react components instead of the HiddenColumns plugin seems to have some issues. I'll probably have to switch to the latter.

danyx23:
Switched to using the HiddenColumns plugin, finished the PR

danyx23:
Video showing the feature in action is here: https://owid.slack.com/archives/CQQUA2C2U/p1643977990210499

danyx23:
This is live on staging (minus the latest little commit that improves ordering when making columns visible)

danielgavrilov:
I think it would help with both reviewing and merging this if it was based on top of #1134? (It seems it already is? Can just click Edit at the top and change the base.)

danielgavrilov (approved):
Looks good! Super helpful to have the video too 👌

danielgavrilov:
By the way – not sure if any are introduced in this PR, but there are 26 `no-unused-vars` warnings in `VariablesAnnotationPage.tsx`. Would be good to either resolve them or add an exception in eslint so they don't get shown (e.g. maybe we always want to ignore `no-unused-vars` for variables starting with underscore `_`?). Doesn't need to be addressed in this PR though, might be cleaner to do it separately.

danyx23:
Thanks @danielgavrilov! I removed the unused variables - I was just so used to having _ be ignored from F# :)

PR #1134, opened 2022-02-04
chore: Unify our default SelectField and SelectGroupsField

danyx23:
Select used to have two array properites, one for values and one for labels; the group version had one array which is an object of value and optional label. The latter is better as the two arrays can't accidentally not be aligned...
This unifies the SelectField to work like the SelectGroupField.


danielgavrilov (approved):
🙌

adminSiteClient/EditorMapTab.tsx
```diff
@@ -44,8 +44,8 @@ class VariableSection extends React.Component<{
                 </section>
             )
 
-        const projections = Object.keys(MapProjectionLabels)
-        const labels = Object.values(MapProjectionLabels)
+        // const projections = Object.keys(MapProjectionLabels)
+        // const labels = Object.values(MapProjectionLabels)
```
danielgavrilov, line 48: nit: these lines can be deleted?

MarcelGerber (approved):
Looks good! Just remove the two lines Gavrilov mentioned, and then we're good

PR #1152, opened 2022-02-11
fix: consistent grapher schema URLs

danielgavrilov:
We are currently linking to files in `nyc3.digitaloceanspaces.com` which is harder to redirect from if we moved the files in the future. Instead I think it's better to link to `files.ourworldindata.org` which we have complete control over. (We could also use a different domain for these things, but this one is already set up.)

Also there were some dead links linking to `grapher-schema.v001.json` instead of `grapher-schema.001.json`.

### Todo

- [ ] Make sure files.ourworldindata.org supports CORS


danyx23:
I haven't forgotten this, just need to find some quiet time to put the cloudflare worker in place

danielgavrilov:
Since this now has conflicts, it's probably better to do a fresh rename. I'm closing this PR and tracking this as an issue instead: #1439

Issue #1106, opened 2022-01-27
Filtering in Bulk FASTT editor


danyx23:
Filtering is a pretty central operation for a spreadsheet like edit view that operates on > 100K variables. The naive approach is to just create a list of filters where each filter is (column + operator + value) and use a logical AND to connect them. This is [how Datasette does this](https://owid-datasette-y43qr.ondigitalocean.app/owid/posts?_sort_desc=id&content__contains=DALY&_sort_desc=id).

But this has the downside of not being able to express a lot of legitimate queries. E.g. "All variables that have in the name either 'Mortality' or 'Death' and have a targetYear > 2015" is perfectly valid. It is obvious that just being able to select between AND and OR is not enough and you quickly need parenthesis to express even relatively simple queries.

After some consideration I am now leaning towards trying the [React awesome query builder](https://github.com/ukrbublik/react-awesome-query-builder) control. It would look like this for a simple query: 
![Screenshot 2022-01-27 at 12-20-45 React Awesome Query Builder - Example](https://user-images.githubusercontent.com/2772449/151349194-4992012e-cb1c-47e5-b194-f2bbb07ae392.png)

and handles parenthesis and different logic operators nicely.

PR #1153, opened 2022-02-11
Bulk FASTT filtering

danyx23:
This PR implements #1106

### DIFF
--- a/adminSiteClient/VariablesAnnotationPage.tsx
+++ b/adminSiteClient/VariablesAnnotationPage.tsx
@@ -18,6 +18,7 @@ import {
     isEmpty,
     isEqual,
     isNil,
+    isPlainObject,
     merge,
 } from "lodash"
 
@@ -56,12 +57,23 @@ import {
 import { faEye } from "@fortawesome/free-solid-svg-icons/faEye.js"
 import { faEyeSlash } from "@fortawesome/free-solid-svg-icons/faEyeSlash.js"
 import {
+    allComparisonOperators,
     BinaryLogicOperation,
     BinaryLogicOperators,
     BooleanAtom,
+    ComparisonOperator,
+    EqualityComparision,
+    EqualityOperator,
+    JsonPointerSymbol,
+    Negation,
+    NumberAtom,
+    NumericComparision,
+    Operation,
     SqlColumnName,
     SQL_COLUMN_NAME_DATASET_NAME,
     SQL_COLUMN_NAME_VARIABLE_NAME,
+    StringAtom,
+    StringContainsOperation,
 } from "../clientUtils/SqlFilterSExpression.js"
 import {
     parseVariableAnnotationsRow,
@@ -85,6 +97,8 @@ import { Query, Builder, Utils as QbUtils } from "react-awesome-query-builder"
 // types
 import {
     JsonGroup,
+    JsonTree,
+    JsonItem,
     SimpleField,
     Config,
     ImmutableTree,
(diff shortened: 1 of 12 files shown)


danyx23:
A video with a short demo is here: https://owid.slack.com/archives/CQQUA2C2U/p1644854600073719

adminSiteClient/VariablesAnnotationTypesAndUtils.tsx
```diff
@@ -183,3 +273,244 @@ export const getItemStyle = (
...
+                    (op) => {
+                        const operator = getNullCheckOperator(op as string)!
+                        const field = getFieldSymbol(
+                            filterTree.properties.field!
+                        )
+                        return new NullCheckOperation(operator!, field)
+                    }
+                )
+                .with("is_empty", "is_not_empty", (operator) => {
+                    const field = getFieldSymbol(filterTree.properties.field!)
```
MarcelGerber, line 432: Can pull the `getFieldSymbol` out before the `match`, since it's used in all cases

clientUtils/SqlFilterSExpression.ts
```diff
@@ -143,20 +143,45 @@ export class JsonPointerSymbol implements Operation {
     }
 }
 
-export const SQL_COLUMN_NAME_VARIABLE_NAME = "variables.name"
-export const SQL_COLUMN_NAME_DATASET_NAME = "datasets.name"
-export const SQL_COLUMN_NAME_NAMESPACE_NAME = "namespaces.name"
-
+export enum WHITELISTED_SQL_COLUM_NAMES {
```
MarcelGerber, line 146: typo in `COLUM` (missing an n)

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -711,14 +743,13 @@ class VariablesAnnotationComponent extends React.Component {
...
             }))
-        const readonlyReorderItems: ColumnReorderItem[] =
-            VariablesAnnotationComponent.readOnlyColumnNamesFields.map(
-                ([label, id]) => ({
-                    key: id,
-                    visible: false,
-                    description: label,
-                })
-            )
+        const readonlyReorderItems: ColumnReorderItem[] = [
```
MarcelGerber, line 746: Just flagging this here, and it's not new, but I have absolutely no clue why this all is called `reorder`. What are we reordering in the first place?

adminSiteClient/VariablesAnnotationTypesAndUtils.tsx
```diff
@@ -183,3 +273,244 @@ export const getItemStyle = (
...
+    const fieldType = match(column.type)
+        .with("string", () => "text")
+        .with("number", () => "number")
+        .with("datetime", () => "datetime")
+        .exhaustive()
+
+    return [
+        column.key,
+        {
+            label: column.key,
```
MarcelGerber, line 461: Could also use `column.label` here.
image

package.json
```diff
@@ -171,6 +173,7 @@
         "randomstring": "^1.1.5",
         "react": "^16.13.1",
         "react-animate-height": "^2.0.17",
+        "react-awesome-query-builder": "4",
```
MarcelGerber, line 176: Version 5 has come out in the meantime and an update right away might be worthwhile if it's not too much pain.
If only because they have renamed `Like` to `Contains` in the dropdown, which I like better :)

clientUtils/SqlFilterSExpression.ts
```diff
@@ -143,20 +143,45 @@ export class JsonPointerSymbol implements Operation {
...
+    SQL_COLUMN_NAME_VARIABLE_NAME = "variables.name",
+    SQL_COLUMN_NAME_DATASET_NAME = "datasets.name",
+    SQL_COLUMN_NAME_NAMESPACE_NAME = "namespaces.name",
+    SQL_COLUMN_NAME_VARIABLE_DESCRIPTION = "variables.description",
+    SQL_COLUMN_NAME_VARIABLE_CREATED_AT = "variables.createdAt",
+    SQL_COLUMN_NAME_VARIABLE_UPDATED_AT = "variables.updatedAt",
```
MarcelGerber, line 152: Should also contain `id`

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -1070,6 +1125,140 @@ class VariablesAnnotationComponent extends React.Component {
...
+        if (isEmpty(filterPanelConfigFields)) return undefined
+
+        const fieldsObject = Object.fromEntries(filterPanelConfigFields)
+        const config = {
+            ...filterPanelInitialConfig,
+            fields: fieldsObject,
+        }
+        // Hide operators in the UI that we don't have a good equivalent for
+        // in the S-Expressions
+        const operatorsToDrop = new Set([
```
MarcelGerber, line 1195: I would be a little more comfortable with a whitelist rather than a blacklist.

It will be a little bit longer, but not by too much.

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -1070,6 +1125,140 @@ class VariablesAnnotationComponent extends React.Component {
...
+            "not_like",
+            "proximity",
+            "starts_with",
+            "ends_with",
+            "between",
+            "not_between",
+            "select_any_in",
+            "select_not_any_in",
+            "multiselect_equals",
+            "mutliselect_not_equals",
```
MarcelGerber, line 1205: typo: should be `multiselect`

adminSiteClient/VariablesAnnotationPage.tsx
```diff
@@ -1070,6 +1125,140 @@ class VariablesAnnotationComponent extends React.Component {
...
+        ])
+        config.types.text.widgets.field.operators =
```
MarcelGerber, line 1207: I have quickly tried to get rid of this godawful bunch of code, and the following seems to work:
`config.operators = omit(config.operators, operatorsToDrop) as any` (where omit comes from lodash)

MarcelGerber (approved):
I had a lot to say but nothing of substance!

This is all great, and you found an awesome tool for building queries 🎈

adminSiteClient/VariablesAnnotationTypesAndUtils.tsx
```diff
@@ -29,6 +60,9 @@ export interface VariableAnnotationsRow {
     grapherConfig: GrapherInterface
     datasetname: string
     namespacename: string
+    description: string
```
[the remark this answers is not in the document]
danyx23: I'd like to do that but this one can't use GrapherInterface yet in the current dependecy chain. When we create a separate folder/tsproject for the types we could move all of these there and then they could reuse the same type definition. Let me know if I missed something though :)

adminSiteClient/VariablesAnnotationTypesAndUtils.tsx
```diff
@@ -183,3 +273,244 @@ export const getItemStyle = (
...
+): Operation | undefined {
+    if (filterTree.type === "group") {
+        // If we have a group then we need to decide
+        // on the operator and build the list of children recursively
+        const logicOperator = getLogicOperator(
+            filterTree.properties?.conjunction ?? "AND"
+        )
+
+        let children: Operation[] = []
+        if (isArray(filterTree.children1))
```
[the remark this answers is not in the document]
danyx23: I think you are right but the type definition claims otherwise. This is probably an error in the type definition though. I'm tempted to leave it as is but add a comment, what do you think?

adminSiteClient/VariablesAnnotationTypesAndUtils.tsx
```diff
@@ -183,3 +273,244 @@ export const getItemStyle = (
...
+                        const field = getFieldSymbol(
+                            filterTree.properties.field!
+                        )
+                        if (
+                            filterTree.properties.value.length === 0 ||
+                            filterTree.properties.value[0] === undefined
+                        )
+                            return undefined
+                        const val = getValueAtom(filterTree.properties.value[0])
+                        return new NumericComparision(operator!, [field, val!])
```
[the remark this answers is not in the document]
danyx23: Thanks!

danielgavrilov (approved):
It’s very awesome to see how you architected this from the beginning and how nicely it scales!

And great find with that query builder library!

adminSiteClient/VariablesAnnotationPage.tsx
danyx23: I renamed this to ColumnInformation - the point of this data structure is to be kept in an array that signifies the user specified order, which is why I called it ColumnReorderItem. But ColumnInformation is prbably the better name for now. If you have an even better one let me know :)

package.json
```diff
@@ -39,6 +39,7 @@
         "testJest": "jest"
     },
     "dependencies": {
+        "@ant-design/icons": "^4.7.0",
```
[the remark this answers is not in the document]
danyx23: Yes - react awesome query builder has a few UI libraries it can use and I liked this one quite a better than their react bootstrap integration which also seemed quite heavy. But I could try the bootstrap version as well if you think it's worth it?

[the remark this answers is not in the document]
MarcelGerber: No, that's not what I'm saying - it's just that I think we can drop the `@ant-design/icons` dependency, keep the `antd` dependency, and everything will still work :)

adminSiteClient/VariablesAnnotationTypesAndUtils.tsx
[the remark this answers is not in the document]
MarcelGerber: Either is fine with me - either leaving it in as-is or going with a `console.error("This code path is not supposed to happen")` instead.

package.json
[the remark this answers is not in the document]
danyx23: Ah gotcha :)

danyx23:
@danielgavrilov @MarcelGerber thanks a lot for your review and esp all your feedback Marcel!

ukrbublik:
Thanks for using my lib in your project!
Can you please share demo video with me, maybe send on email?
I'd like to add it [here](https://github.com/ukrbublik/react-awesome-query-builder/wiki/Used-in-projects) if you don't mind, as a motivation for future development.
Btw, if you need some fixes/feature requests, let me know ;)

PR #1159, opened 2022-02-15
Map sparklines #1: Prep work

danielgavrilov:
This is prep work for implementing https://github.com/owid/owid-grapher/issues/1087.

### Axis

- ⭐ **`ticks` is a config property allowing you to override the automatic tick generation and use manual ticks (it allows showing just the start & end ticks in sparklines). If a tick value is `-Infinity` or `Infinity`, it gets replaced with the minimum or maximum value in the data.**
- `hideGridlines` is now a config property instead of a mutable property of `Axis` (it made things cleaner)

### CoreTable

- ⭐ **Time columns now have prepositions: `in` used for years & quarters, `on` for dates.** (the only place they will be used are sparklines, but we can add them in other places after this is merged)
- `replaceCells` used to be a utility function private to CoreTable, now it's a public method on CoreTable. I had used it for one aspect but ended up not using it in the final version.

### All chart types

- ⭐ **`disableIntroAnimation` GrapherInterface property that disables the animation shown on load**
- Tooltips are more opaque, from 92% to 95%. This is to reduce noise in the background of sparklines, which are harder to interpret because it's a multi-colored line.
- `colorScaleOverride` and `colorScaleColumnOverride` are supported on all chart types. They are currently only used in faceted charts (we want the color scale to be consistent across all facets) and in sparklines (we want the sparkline to use the same color scale as the map chart)

### LineChart

- Fix `lineStrokeWidth` config property being ignored
- Various design details to make it work better as a sparkline, without noticeable differences with full-sized line charts.

### DIFF
--- a/coreTable/CoreTable.ts
+++ b/coreTable/CoreTable.ts
@@ -56,7 +56,6 @@ import {
     sortColumnStore,
     emptyColumnsInFirstRowInDelimited,
     truncate,
-    replaceCells,
 } from "./CoreTableUtils.js"
 import {
     ErrorValueTypes,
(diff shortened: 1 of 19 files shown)


danyx23 (approved):
Looks great! If I understand correctly then the two axis related fields could also be set via the config now, right? If so then please add these two new fields to the schema.

And thanks for breaking it up this way, this makes it a lot nicer to review :)

danielgavrilov:
> If I understand correctly then the two axis related fields could also be set via the config now, right? If so then please add these two new fields to the schema.

Ah, yes, and a bunch more are also not there. All of these cannot be specified through the Admin currently, but can technically be manually added in the Grapher config, although I'm not sure if it's a good idea to encourage that. 

Let's discuss in a separate issue? #1174

PR #1160, opened 2022-02-15
Map sparklines #2: Feature work

### LINKED ISSUE #1087: Turn sparkbars into sparklines on map hover panel

danielgavrilov:
Fixes #1087, continuation of and based on top of #1159.

https://user-images.githubusercontent.com/1308115/154110274-e1a8fe4f-901a-4984-a399-182d7df1e131.mp4

### Preview on playfair
- [Life expectancy map](https://playfair-owid.netlify.app/grapher/life-expectancy) (interesting because UK data starts at 1543 but others much later)
- [Energy embedded in traded goods map](https://playfair-owid.netlify.app/grapher/traded-energy-share-domestic) (diverging color scale, contains negative values)
- [COVID vaccine age map](https://playfair-owid.netlify.app/grapher/covid-vaccine-age) (numerical values are coded to text values)
- [COVID cases map](https://playfair-owid.netlify.app/explorers/coronavirus-data-explorer?tab=map&zoomToSelection=true&time=2021-08-20&facet=none&pickerSort=asc&pickerMetric=location&Interval=7-day+rolling+average&Relative+to+Population=true&Color+by+test+positivity=false&country=~FRA&Metric=Confirmed+cases) (lots of data points, demonstrates potential performance issues)
- [World bank income groups](https://playfair-owid.netlify.app/grapher/world-banks-income-groups?country=ITA~NLD~LIE) (categorical maps have no sparkline)
- [Yaws endemicity and number of cases](https://playfair-owid.netlify.app/grapher/yaws-status-and-case-numbers) (mixed categorical and numeric values, sparkline only shown for numeric, see Cameroon and Indonesia)
- [Excess mortality map](https://playfair-owid.netlify.app/explorers/coronavirus-data-explorer?tab=map&zoomToSelection=true&time=2022-01-31&facet=none&pickerSort=asc&pickerMetric=location&Metric=Excess+mortality+%28%25%29&Interval=Weekly&Relative+to+Population=true&Align+outbreaks=false&country=~FRA) (contains data gaps, faint color scale)

### Design decisions ([discussed on Slack](https://owid.slack.com/archives/C5BDCB2R3/p1644485174789959))

- **Use a colored line**. Sparklines only show the trend, color helps you compare countries by magnitude too (and quickly, without having to read an axis, although less accurate).

- **Show only the minimum & maximum ticks on both the X and Y axes**. 

- **Use `yAxis.min` and `yAxis.max` from the top-level Grapher config** but not any of the other axis config options. The effect of these is that authors often want to show Y=0 line, and it will be shown on sparklines too if that's set up for the top-level chart.

- **Always extend the X axis domain up to the selected time on the timeline (to the past or future)**. This means if the country has data 1990–2000 but 2010 is selected on the timeline, a domain of 1990–2010 will be used for the sparkline (half of the sparkline being blank basically).
   
   Instead of doing this, another option is to keep the X domain of the sparkline identical across all countries. So for the [life expectancy map](http://localhost:3030/grapher/life-expectancy?time=2019) this would mean all sparklines start from 1543, but only UK data start that early, most countries start from 1900s, so I decided not to do it this way.

### Issues

- Because of the way we are plotting the colored line, the coloring can potentially be misleading when you look at it up close. Basically we color the points correctly, but the color transition always happens halfway between points:
  
   multi-color-polyline

   We decided it's not worth addressing for now.

- Charts where the numerical values have been transformed to categorical (using the legend) are a bit unusual and might be confusing for readers:

   Screenshot 2022-02-10 at 18 47 13

  We kind of decided not to address it, but [Max is still in weak favour of visualising these differently](https://owid.slack.com/archives/C5BDCB2R3/p1644595692629799?thread_ts=1644485460.127289&cid=C5BDCB2R3). For now, it seems good enough as it is, but might need to be fixed in the future.

### DIFF
--- a/clientUtils/Util.ts
+++ b/clientUtils/Util.ts
@@ -149,14 +149,29 @@ type OptionalKeysOf<T> = Exclude<
     undefined
 >
 
-type PartialWithoutUndefinedKey<T> = {
+type AllowUndefinedValues<T> = {
     [K in keyof T]: T[K] | undefined
 }
 
+/**
+ * This generic makes every (top-level) optional property in an interface required,
+ * but with `undefined` as an allowed value.
+ *
+ * For example:
+ *     AllKeysRequired<{
+ *         a: number
+ *         b?: number
+ *     }>
+ * becomes:
+ *     {
+ *         a: number
+ *         b: number | undefined
+ *     }
+ */
 // This was tricky to construct.
 // It seems like the initial, elegant approach is:
 //
-// export type NoUndefinedKeys<T> = {
+// export type AllKeysRequired<T> = {
 //     [K in keyof T]-?: T extends Record<K, T[K]> ? T[K] : T[K] | undefined
 // }
 //
@@ -164,7 +179,7 @@ type PartialWithoutUndefinedKey<T> = {
 // make the key required with `-?`. So we have this ugly workaround.
 //
 // -@danielgavrilov, 2022-02-15
-export type NoUndefinedKeys<T> = PartialWithoutUndefinedKey<
+export type AllKeysRequired<T> = AllowUndefinedValues<
     Required<Pick<T, OptionalKeysOf<T>>>
 > &
     Exclude<T, OptionalKeysOf<T>>
--- a/clientUtils/Util.ts
+++ b/clientUtils/Util.ts
@@ -142,6 +142,33 @@ export type NoUndefinedValues<T> = {
     [P in keyof T]: Required<NonNullable<T[P]>>
 }
 
+type OptionalKeysOf<T> = Exclude<
+    {
+        [K in keyof T]: T extends Record<K, T[K]> ? never : K
+    }[keyof T],
+    undefined
+>
+
+type PartialWithoutUndefinedKey<T> = {
+    [K in keyof T]: T[K] | undefined
+}
+
+// This was tricky to construct.
+// It seems like the initial, elegant approach is:
+//
+// export type NoUndefinedKeys<T> = {
+//     [K in keyof T]-?: T extends Record<K, T[K]> ? T[K] : T[K] | undefined
+// }
+//
+// But TypeScript will omit `undefined` from the value type whenever you
+// make the key required with `-?`. So we have this ugly workaround.
+//
+// -@danielgavrilov, 2022-02-15
+export type NoUndefinedKeys<T> = PartialWithoutUndefinedKey<
+    Required<Pick<T, OptionalKeysOf<T>>>
+> &
+    Exclude<T, OptionalKeysOf<T>>
+
 export type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
 
 // d3 v6 changed the default minus sign used in d3-format to "−" (Unicode minus sign), which looks
--- a/grapher/mapCharts/MapChart.tsx
+++ b/grapher/mapCharts/MapChart.tsx
@@ -322,7 +322,7 @@ export class MapChart
         this.mapConfig.projection = value
     }
 
-    @computed private get formatTooltipValue(): (d: PrimitiveType) => string {
+    @computed private get formatTooltipValue(): (d: number | string) => string {
         const { mapConfig, mapColumn, colorScale } = this
 
         return (d: PrimitiveType): string => {
--- a/grapher/mapCharts/MapChart.tsx
+++ b/grapher/mapCharts/MapChart.tsx
@@ -66,7 +66,11 @@ import { NoDataModal } from "../noDataModal/NoDataModal.js"
 import { ColorScaleConfig } from "../color/ColorScaleConfig.js"
 import { CoreColumn } from "../../coreTable/CoreTableColumns.js"
 import { SelectionArray } from "../selection/SelectionArray.js"
-import { Color, HorizontalAlign } from "../../clientUtils/owidTypes.js"
+import {
+    Color,
+    HorizontalAlign,
+    PrimitiveType,
+} from "../../clientUtils/owidTypes.js"
 
 const PROJECTION_CHOOSER_WIDTH = 110
 const PROJECTION_CHOOSER_HEIGHT = 22
@@ -318,10 +322,10 @@ export class MapChart
         this.mapConfig.projection = value
     }
 
-    @computed get formatTooltipValue(): (d: number | string) => string {
+    @computed private get formatTooltipValue(): (d: PrimitiveType) => string {
         const { mapConfig, mapColumn, colorScale } = this
 
-        return (d: number | string): string => {
+        return (d: PrimitiveType): string => {
             if (isString(d)) return d
             else if (mapConfig.tooltipUseCustomLabels) {
                 // Find the bin (and its label) that this value belongs to
@@ -334,8 +338,7 @@ export class MapChart
     }
 
     @computed get series(): ChoroplethSeries[] {
-        const { mapColumn, selectionArray, targetTime, formatTooltipValue } =
-            this
+        const { mapColumn, selectionArray, targetTime } = this
         if (mapColumn.isMissing) return []
         if (targetTime === undefined) return []
 
@@ -346,7 +349,6 @@ export class MapChart
                 if (!color) return undefined
                 return {
                     seriesName: entityName,
-                    displayValue: formatTooltipValue(value),
                     time,
                     value,
                     isSelected: selectionArray.selectedSet.has(entityName),
@@ -590,10 +592,6 @@ export class MapChart
 
         const { projection } = mapConfig
 
-        const tooltipDatum = tooltipTarget
-            ? seriesMap.get(tooltipTarget.featureId)
-            : undefined
-
         return (
             <g ref={this.base} className="mapTab">
                 <ChoroplethMap
@@ -627,13 +625,15 @@ export class MapChart
                 </foreignObject>
                 {tooltipTarget && (
                     <MapTooltip
-                        tooltipDatum={tooltipDatum}
+                        entityName={tooltipTarget?.featureId}
+                        timeSeriesTable={this.inputTable}
+                        formatValue={this.formatTooltipValue}
                         isEntityClickable={this.isEntityClickable(
                             tooltipTarget?.featureId
                         )}
                         tooltipTarget={tooltipTarget}
                         manager={this.manager}
-                        colorScale={this.colorScale}
+                        colorScaleManager={this}
                         targetTime={this.targetTime}
                     />
                 )}
--- a/grapher/mapCharts/MapTooltip.scss
+++ b/grapher/mapCharts/MapTooltip.scss
@@ -28,6 +28,5 @@
     .value-wrapper.no-plot {
         text-align: center;
         padding: 0;
-        text-align: center;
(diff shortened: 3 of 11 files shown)


danielgavrilov:
Things I'm not sure about:

- There are a bunch of improvements we could make to maps using numeric-turned-categorical legends – we could hide the Y axis since it's often misleading, we could flatten the line and only change colors (as Max suggested)

- I didn't write _any_ tests... 😬 A lot of the implementation is constructing config, so didn't find great tests to add (although if I'm honest, I also didn't look very hard)

grapher/mapCharts/MapChart.tsx
```diff
@@ -318,10 +322,10 @@ export class MapChart
         this.mapConfig.projection = value
     }
 
-    @computed get formatTooltipValue(): (d: number | string) => string {
+    @computed private get formatTooltipValue(): (d: PrimitiveType) => string {
```
MarcelGerber, line 325: What is the switch over to using `PrimitiveType` here for? It also includes `boolean`, which I don't think we will ever show in a tooltip at this point?

grapher/mapCharts/MapTooltip.tsx
```diff
@@ -2,113 +2,118 @@ import * as React from "react"
...
+        }
+        return new ColorScale(newManager)
     }
 
-    @computed private get darkestColorInColorScheme(): string | undefined {
-        const { colorScale } = this
-        return colorScale.isColorSchemeInverted
-            ? first(colorScale.baseColors)
-            : last(colorScale.baseColors)
+    @computed private get textColorScale(): ColorScale {
```
MarcelGerber, line 99: Hm, this is exactly the same as `lineColorScale`...
I like that you make the separation explicit, but maybe a simple `return this.lineColorScale` is enough unless and until we actually want them to differ?

grapher/mapCharts/MapTooltip.tsx
```diff
@@ -125,93 +130,165 @@ export class MapTooltip extends React.Component<MapTooltipProps> {
...
             : targetTime
         const displayDatumTime =
-            timeColumn && tooltipDatum
-                ? timeColumn.formatValue(tooltipDatum.time)
-                : tooltipDatum?.time.toString() ?? ""
+            timeColumn && datum
+                ? timeColumn.formatValue(datum?.time)
+                : datum?.time.toString() ?? ""
+        const valueColor: string | undefined = darkenColorForHighContrastText(
+            textColorScale?.getColor(datum?.value) ?? "#333"
```
MarcelGerber, line 210: `textColorScale` has `darkenColorForText` already, and then we're applying `darkenColorForHighContrastText` on top of that. seems a little unnecessary

grapher/mapCharts/MapTooltip.tsx
```diff
@@ -125,93 +130,165 @@ export class MapTooltip extends React.Component<MapTooltipProps> {
...
                     style={{
-                        padding: "0.3em 0.3em",
+                        padding: "0.3em",
                         margin: 0,
                         fontWeight: "normal",
                         fontSize: "1em",
                     }}
                 >
                     {tooltipTarget.featureId ||
                         tooltipTarget.featureId.replace(/_/g, " ")}
```
MarcelGerber, line 233: I know this is not something you touched, but this right here is just extremely weird? The part after `||` is always empty when it is reached?

grapher/mapCharts/MapTooltip.scss
```diff
@@ -1,79 +1,33 @@
...
-            padding-top: 0.3em;
-
-            .count {
-                text-align: center;
-            }
-        }
+    .value-wrapper.no-plot {
+        text-align: center;
+        padding: 0;
+        text-align: center;
```
MarcelGerber, line 31: nit: twice!

MarcelGerber (approved):
This is _great_!

grapher/mapCharts/MapTooltip.tsx
```diff
@@ -2,113 +2,118 @@ import * as React from "react"
...
+    colorScaleManager: ColorScaleManager
+    formatValue: (d: PrimitiveType) => string
+    timeSeriesTable: OwidTable
+    tooltipTarget: { x: number; y: number; featureId: string }
     targetTime?: Time
+    isEntityClickable?: boolean
 }
 
+const SPARKLINE_WIDTH = 140
+const SPARKLINE_HEIGHT = 60
```
MarcelGerber, line 36: One thing I would love to see is for these to be a little bit larger on a big wide screen.
I realize that the current size is a great compromise that works on both mobile and desktop, but on a large screen it can be hard to make out details.
Additionally, the small size leads to information like the x ticks and the min/max y ticks not being shown for lack of space.

grapher/mapCharts/MapChart.tsx
danielgavrilov: It's true, `boolean` is not supported in many parts of our code, but I find it tricky to decide what's best. Considering a lot of stuff doesn't handle `boolean`, probably we should remove it as a possible type in our data tables (basically from `PrimitiveType`).

Right now we don't have many (any?) boolean types in production, even though I think they are possible to define in an Explorer. It's probably worthwhile to have eventually, but not without inspecting all code paths that would need to support it.

I'm thinking about (and slowly moving towards) not making any assumptions about what the incoming values are, because we are intending to make Grapher more reusable. 

As for specific context about why `PrimitiveType` is used specifically here – the type comes from the `datum`:

```ts
    @computed private get datum(): OwidVariableRow<PrimitiveType> | undefined {
        return this.mapTable.get(this.mapColumnSlug).owidRows[0]
    }
```

`PrimitiveType` is the most general type, and if we made no assumptions about input, it _can_ technically be a boolean.

It would either be formatted as `"true"` or `"false"`, unless we implement boolean color scales too. We have quite a lot of "yes/no" map charts where the most reasonable value would be a `boolean` if it was supported, here are just a few where it's encoded as a number instead: https://playfair.owid.cloud/admin/test/embeds?tab=map&ids=118,1190,1203,1571,1675,1677,1678,1679,1680,1681,1682,1806,2025,2098,2099,2105,2108,2114

---

Back on topic: the fix that seems reasonable is to completely drop `boolean` from `PrimitiveType` – do you agree?

The alternative is to assume `boolean` can never be passed to a Map chart, which is not true in our current circumstances, we don't enforce that.

grapher/mapCharts/MapTooltip.tsx
danielgavrilov: Great point – will change to use a single color scale instead, not sure what I was thinking.

grapher/mapCharts/MapTooltip.tsx
danielgavrilov: Yes! It's weird. I think `&&` was maybe intended. But even then, I can't see why we would replace underscores in the tooltip only? I also can't see how `_` would appear in the string, there don't seem to be underscores in `MapTopology.ts`. 

I will go ahead and remove the stuff after `||`, let me know if you have ideas.

clientUtils/Util.ts
```diff
@@ -142,6 +142,33 @@ export type NoUndefinedValues<T> = {
...
+//
+// export type NoUndefinedKeys<T> = {
+//     [K in keyof T]-?: T extends Record<K, T[K]> ? T[K] : T[K] | undefined
+// }
+//
+// But TypeScript will omit `undefined` from the value type whenever you
+// make the key required with `-?`. So we have this ugly workaround.
+//
+// -@danielgavrilov, 2022-02-15
+export type NoUndefinedKeys<T> = PartialWithoutUndefinedKey<
```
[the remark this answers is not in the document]
danielgavrilov: Yeah I find it also hard to write, there is probably a better solution for this.

Does this description make sense:

```ts
/**
 * This generic makes every (top-level) optional property in an interface required,
 * but with `undefined` as an allowed value.
 *
 * For example:
 *     AllKeysRequired<{
 *         a: number
 *         b?: number
 *     }>
 * becomes:
 *     {
 *         a: number
 *         b: number | undefined
 *     }
 */
// This was tricky to construct.
// It seems like the initial, elegant approach is:
//
// export type AllKeysRequired<T> = {
//     [K in keyof T]-?: T extends Record<K, T[K]> ? T[K] : T[K] | undefined
// }
//
// But TypeScript will omit `undefined` from the value type whenever you
// make the key required with `-?`. So we have this ugly workaround.
//
// -@danielgavrilov, 2022-02-15
```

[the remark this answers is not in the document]
MarcelGerber: that's way better, thank you!

grapher/mapCharts/MapChart.tsx
MarcelGerber: I vote for removing it from `PrimitiveType` for now.
We may want to add back a `BooleanColumn` at some point in the future, but once we do that we can reconsider and properly check that everything works as expected.

danielgavrilov: Oops, `Boolean` cells are used in `GrapherGrammar` and `ColumnGrammar`, so we can't drop support in `CoreTable`.

This in turn means these boolean values could leak into chart data too, since that can be specified in an Explorer config.

Do you happen to see any easy way out of this? I think making it possible to define booleans in CoreTable but hardcoding `number | string` in the chart leaves a hole for potential type errors that could be caught. 

One option is to apply a filter in `MapChart#transformTable()` to ensure all booleans are dropped, another option is to leave it as it is.

danyx23: It looks like our data_values never used the boolean type though right? I think we'll have to rethink the architecture of data values a little when we switch to a different data model that avoids putting everything into the data_values mysql. As part of this we will probably also get rid of the terrible ambiguity of whether a variable is numeric or string or mixed. When we do that we will know which type everything is and then we can decide if bools are worth the trouble or not (my guess is not). We should then also maybe rethink how grid grammars work, they are a bit odd alltogehter.

Because we haven't used booleans as data values I am inclined to hardcode to number | string. What we could do though is to add a sanity check into transformTable that only runs in dev builds. This way we don't have overhead in prod, we ourselves would notice if we made a mistake in when we think bools may occur and we can clean this up properly once we restructure our data serving pipeline. What do you think?

danyx23 (approved):
Great work, LGTM!

grapher/mapCharts/MapTooltip.tsx
danielgavrilov: Testing this at 50% zoom level (don't have a big screen right now), the tooltip is still pretty massive?

Screenshot 2022-02-16 at 19 35 09

Compared to the other elements (subtitle font size, legend), it seems a reasonable size?

If we made it much bigger, it would occlude way too big of an area of the map? I already think it's too big. 

Here's how it is currently as a comparison:

Screenshot 2022-02-16 at 19 35 21

danielgavrilov: I'll deploy this (since it's an improvement over what we have today) but happy to continue discussing/tweaking the design later on.

danielgavrilov:
Thanks a lot for the super quick reviews @MarcelGerber & @danyx23! I'll deploy this (to avoid it dragging on for another week) but happy to continue discussing/making design tweaks.

PR #1201, opened 2022-02-18
enhance(map): use levelled sparkline for some charts

danielgavrilov:
The Slack thread to [change the way we handle ordinal variables in sparklines](https://owid.slack.com/archives/C5BDCB2R3/p1645098840053389?thread_ts=1644485460.127289&cid=C5BDCB2R3) is managing to stay alive. I'm not sure if we're gonna change them. Might be a cooldown task.

I'm just dumping some code I had to created **"levelled" line charts (sparklines)**:

Screenshot 2022-02-18 at 10 12 05

As opposed to how they are visualised now:

Screenshot 2022-02-18 at 10 12 49

This PR contains a very crude way to decide when sparklines should be levelled – basically whenever `tooltipUseCustomLabels` is `true`. We likely need something better before we can ship to production.

### Todo

- [ ] Figure out a better way to detect ordinal/categorical variables (or just add `type` as explicit configuration on the variable level?)
- [ ] Add tests for levelled line chart – especially around the color/y changes. It works a bit differently.

### Potential issues

- **Is the midpoint between two points the right place to split up the line?** Or should it be at the start/end?

### DIFF
--- a/grapher/lineCharts/LineChart.tsx
+++ b/grapher/lineCharts/LineChart.tsx
@@ -212,6 +212,7 @@ class Lines extends React.Component<LinesProps> {
                         )}
                     />
                     <MultiColorPolyline
+                        interpolation={this.props.lineInterpolation}
                         points={series.placedPoints}
                         strokeLinejoin="round"
                         strokeWidth={this.strokeWidth}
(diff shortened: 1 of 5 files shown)


danielgavrilov:
If we had types, we would probably want this. Since we don't have types, it's a harder call. I think the logic here will lead to some tooltips being worse. I think there was one that was one SDG map that was "above target" or "below target" in the tooltip text, but the value is continuous, not discrete.

Also another issue is that the tooltip is not covered by the svgTester so this PR needs some tests.

Closing it since it's somewhat stale.

PR #1202, opened 2022-02-19
Fix source text footer and improve source styling

### LINKED ISSUE #1151: Make source text in footer selectable
## Core problem

The source text in Grapher charts is not selectable due to the `.clickable` CSS class being applied : https://github.com/owid/owid-grapher/blob/06f48d54713661adba074015a3da76e4659c7e1d/grapher/footer/Footer.tsx#L244-L250

The CSS class applies `user-select: none`:

https://github.com/owid/owid-grapher/blob/8d7ae579d30f36fcf0c57dc18b76a78a621ff280/grapher/core/grapher.scss#L61-L69

## Proposed solution

We should keep the "clickable" style, meaning `cursor: pointer`, but we should remove `user-select: none` in this particular case. I'm not sure if we should do it in _all_ cases, probably not?

## Context

Screenshot 2022-02-10 at 11 24 16

It's helpful to make the text selectable so that readers can more easily copy and cite the source.

(This issue was brought up by [Max on the #tech-issues Slack channel](https://owid.slack.com/archives/C46U9LXRR/p1644485149805549).)

shafy:
This PR closes #1151 and contains the following:

**Fix grapher footer source text selection**

Users couldn't select the source text, but this is useful to quickly copy and paste the source. I made the text selectable and removed the `onClick` action that would open the Source tab when clicking the source. I think it's a better user experience because a) selecting the source text in order to copy it would open the Source tab (annoying), b) clicking a non-link, non-button text is not expected to do something, and c) the Source tab is just below and well visible anyways.

I also did the following:
- Removed the class `noselect` as it was not being used (or at least I didn't find where)
- Removed `user-select` from `button`and `clickable` - I don't see what this really adds here?

**Improve source tab styling**

While this was not part of issue #1151 it was somewhat related and I made a proposition to improve the styling of the source table and fix a decoding issue with the source name. In this example `&GreaterEqual;` and HTML entities in general were not decoded probably. Furthermore, the table's first row's top border and the source description's bottom border conflicted visually, I added styling to remove the top border from the first row. Finally, I've removed the border spacing between the `td` elements to get rid of the little gap on the rows' top borders.

It now looks like this:
Screenshot 2022-02-19 at 13 23 30

And looked like this before:
Screenshot 2022-02-19 at 13 24 36

Let me know if some of the changes I made don't make sense. Happy to update and implement feedback. I'm new to this project so don't have too much context yet.

### DIFF
--- a/grapher/sourcesTab/SourcesTab.tsx
+++ b/grapher/sourcesTab/SourcesTab.tsx
@@ -2,13 +2,15 @@ import { linkify } from "../../clientUtils/Util.js"
 import React from "react"
 import { computed } from "mobx"
 import { observer } from "mobx-react"
+import { decodeHTML } from "entities"
 import { Bounds, DEFAULT_BOUNDS } from "../../clientUtils/Bounds.js"
 import { faPencilAlt } from "@fortawesome/free-solid-svg-icons/faPencilAlt.js"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome/index.js"
 import { CoreColumn } from "../../coreTable/CoreTableColumns.js"
 import { OwidColumnDef } from "../../coreTable/OwidTableConstants.js"
 
-const formatText = (s: string) => linkify(s).replace(/(?:\r\n|\r|\n)/g, "<br/>")
+const formatText = (s: string): string =>
+    linkify(s).replace(/(?:\r\n|\r|\n)/g, "<br/>")
 
 export interface SourcesTabManager {
     adminBaseUrl?: string
@@ -47,7 +49,7 @@ export class SourcesTab extends React.Component<{
         return (
             <div key={slug} className="datasource-wrapper">
                 <h2>
-                    {column.name}{" "}
+                    {decodeHTML(column.name)}{" "}
                     {editUrl && (
                         <a href={editUrl} target="_blank" rel="noopener">
                             <FontAwesomeIcon icon={faPencilAlt} />
--- a/grapher/sourcesTab/SourcesTab.tsx
+++ b/grapher/sourcesTab/SourcesTab.tsx
@@ -2,7 +2,6 @@ import { linkify } from "../../clientUtils/Util.js"
 import React from "react"
 import { computed } from "mobx"
 import { observer } from "mobx-react"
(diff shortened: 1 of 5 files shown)


danyx23:
Hi @shafy! Thanks a lot for your PR! I'll try to have an in-depth look still this week.

danyx23 (approved):
Thanks a lot @shafy! This looks good to me. I'd just like @danielgavrilov to take a quick look as he has a longer history with the code base and might know about the choice of the styles here. @danielgavrilov if you agree that this is good them please merge it. Thank you!

grapher/sourcesTab/SourcesTab.tsx
```diff
@@ -47,7 +49,7 @@ export class SourcesTab extends React.Component<{
         return (
             <div key={slug} className="datasource-wrapper">
                 <h2>
-                    {column.name}{" "}
+                    {decodeHTML(column.name)}{" "}
```
danielgavrilov, line 52: This is a database issue on our side – we shouldn't be storing HTML entities in variable names, but unicode symbols. This is an old import (from 2018) that wasn't properly cleaned.

I think it's a good idea to drop it from the PR and we can fix it next time we run the importer.

danielgavrilov (changes requested):
Hi @shafy! Sorry this took a _while_ to review. Thank you very much for contributing and writing it up so well! 🙌 

I only have one minor suggestion, happy to merge when it's updated!

grapher/sourcesTab/SourcesTab.tsx
shafy: Yes, that makes more sense. I've reverted this in [c49acfc](https://github.com/owid/owid-grapher/pull/1202/commits/c49acfc69ce9ddbf0195bdb9bc54ede83b300afc).

shafy:
Hey @danielgavrilov no worries! [c49acfc](https://github.com/owid/owid-grapher/pull/1202/commits/c49acfc69ce9ddbf0195bdb9bc54ede83b300afc) should do the job :-)

danielgavrilov (approved):
Awesome, thank you! 🙌

PR #1207, opened 2022-02-21
fix: do not show non-numeric warning for scientific notation

danielgavrilov:
When importing data, there are client-side warnings that check if a value is a number. We have plenty of scientific notation numbers in our database and they are handled fine, so I'm allowing scientific notation in the regex that checks for numbers.

Problem reported by [Pablo on Slack](https://owid.slack.com/archives/C01GAN73R5W/p1645453849936399).

### DIFF
--- a/adminSiteClient/ImportPage.tsx
+++ b/adminSiteClient/ImportPage.tsx
@@ -85,6 +85,9 @@ class EditableDataset {
     }
 }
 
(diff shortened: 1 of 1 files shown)


mlbrgl (approved):
makes sense to me

danyx23 (approved):

PR #1216, opened 2022-02-23
feat(bulk fastt): show only columns where the current row has values

danyx23:
This is a small feature that changes the visibility of the columns when pressing a button to only show those columns that have values in the current row.


MarcelGerber (approved):
Looks good!

PR #1250, opened 2022-03-10
enhance(docker): Attempt unified M1 compatible docker image

danyx23:
This PR switches to a different mysql image that should also exist as an armv8 version so that M1 apples should be able to transparently choose the right flavour.


larsyencken (commented):
MySQL comes up cleanly on M1 \o/

Separate issue, there are missing settings in `.env.example` before the full site comes up clean. I'd suggest renaming `.env.example` to `.env.grapher-only` and making a `.env.wordpress` with the `docker-compose` setup in mind.

PR #1254, opened 2022-03-11
chore: avoid the need for special ssh settings

larsyencken:
We currently assume that everyone has the same `.ssh/config` settings for our work, but that's mainly to hard-code IP addresses for our servers.

Instead, this change simply uses the friendly DNS for each step, avoiding the need for common settings.


mlbrgl (approved):
makes sense to me 👍

PR #1268, opened 2022-03-15
SQL updates for MySQL 8 compatibility

danyx23:
mysql 8 treats json extracted values on IS TRUE or IS FALSE differently from 5.7

This commit switches to the new ->> operatator that unquotes the json value to a string and compares that to false or true. See https://bugs.mysql.com/bug.php?id=99237


marcelgerber (approved):
Looks great!

I also quickly checked that this syntax works fine on MySQL 5.7, too, so there's nothing holding you off from merging.

PR #1277, opened 2022-03-16
Set up grapher-only docker-compose workflow

larsyencken:
Split the `docker-compose` config into two files:

- `docker-compose.grapher.yml`: for the public grapher-only setup
- `docker-compose.full.yml`: for the private setup including Wordpress

Add a `Makefile`, so that someone can type `make up` to get a full working local grapher-only stack run within tmux.


danyx23:
Ah nice, good idea with the tmux cheat sheet :)

PR #1281, opened 2022-03-17
chore: Add "make up.full" and "make down.full" for wordpress

larsyencken:
Add a Makefile flow for working with the full wordpress stack, including downloading the data you need.

- All data required will be downloaded, including uploads
- If your `.env` file is missing wordpress configuration, it will warn you and abort
- If you have no `.env` file, it will use `.env.example-full` instead


larsyencken:
Everything on port 8080 comes up just fine, but I get this on port 3030, which makes me think there's some piece of config that I'm missing. It's a sign of a failed GraphQL query against Wordpress:

Screen Shot 2022-03-17 at 13 24 18

PR #1300, opened 2022-03-21
feat: enable `equalSizeBins` by default

### LINKED ISSUE #1113: Map brackets should always be equal bin by default
## Core problem

Map brackets can either be equally-spaced or not – the author currently sets this option. But, especially for some charts with large ranges, this makes some of the brackets unusable.

Example of [not equally spaced](https://ourworldindata.org/grapher/gdp-per-capita-worldbank) (there are worse examples where the bottom brackets are not hoverable) vs. [equally spaced](https://ourworldindata.org/grapher/gross-domestic-product).

## Proposed solution

We would like map bracket bins to be equally-spaced by default, and for all previous charts to adopt this behaviour. i.e. all maps on OWID should have equal bin spacing.

danielgavrilov:
Fixes #1113. No migration applied to past charts. Since in the past only `true` and `undefined` values were possible for `equalSizeBins`, now `undefined` becomes `true`, so we don't _need_ to do any migrations. 

It might be nice for cleanliness, let me know if you think so!

### DIFF
--- a/adminSiteClient/EditorColorScaleSection.tsx
+++ b/adminSiteClient/EditorColorScaleSection.tsx
@@ -68,7 +68,7 @@ class ColorLegendSection extends React.Component<{
     onChange?: () => void
 }> {
     @action.bound onEqualSizeBins(isEqual: boolean) {
-        this.props.scale.config.equalSizeBins = isEqual ? true : undefined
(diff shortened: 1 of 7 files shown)


marcelgerber (approved):
Great, and surprisingly easy to do. Go for it!

PR #1309, opened 2022-03-22
Mysql update utf8mb4

### LINKED ISSUE #1266: Mysql corrupts some unicode codepoints in all tables - migrate to utf8mb4
## Description

In MySQL utf8 used to mean utf8mb3 - a, frankly, broken and wrong utf8 implementation that only captures utf codepoints not on the complex plane. Many emoji fall into that space as do some other, less often used languages. 

Unfortunately our databases and tables were created with this setting instead of the later, correct utf8 implementation which is called utf8mb4. This leads to bugs when unicode codepoints from this plane are used anywhere in our database as they do not round trip correctly.

To avoid nasty surprises when e.g. emoji were to be used (e.g. flag emoji might be quite relevant for us) we should convert all out tables to utf8mb4.

## Expected behaviour

Unicode text should round trip to the database without being corrupted.

## Steps to reproduce

1. Go to grapher admin and create or edit a chart
2. Change e.g. the title to include an emoji, e.g. 👪 
3. Save
4. Reload -> the emoji is gone and a question mark or similar in its place

## Proposed solution

Write a TS migration with the help of a script [like this](https://dba.stackexchange.com/questions/257962/upgrade-all-mysql-columns-tables-and-databases-from-utf8mb3-to-utf8mb4)

danyx23:
This PR fixes #1266. It basically changes the charset for all grapher tables to be the correct utf8mb4.

I used the [utf8mb4_0900_as_cs](https://dev.mysql.com/doc/refman/8.0/en/charset-mysql.html) collation which is the modern (Unicode 9) case sensitive collation that is roughly as fast as latin1 used to be in mysql 5.7 (this is relevant for string comparisions and orderings and should make ordering and comparing strings with non-latin1 characters behave better than it used to).

Because the actual space in bytes that MySQL needs changes with this, I had to make one string shorter: variables.name used to be a VARCHAR(1000) but that exceeded the maximum index now for a unique index on that field. I shortened it to 750 characters which ensures that the entries fit into the max 3072 bytes.

The data_values migration takes a long time which is why I will only merge this to prod on Friday evening

The upside of this work is that now an emoji like 👪 round trips correctly when used in the chart title 😸


db/migration/1647947484386-UpdateCharaterSet.ts
```diff
@@ -0,0 +1,47 @@
...
+ALTER TABLE \`grapher\`.\`knex_migrations\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`migrations\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`namespaces\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`posts\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`settings\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`sources\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`suggested_chart_revisions\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`tags\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`users\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
+ALTER TABLE \`grapher\`.\`variables\` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;
```
danielgavrilov, line 28: Appears twice?

db/migration/1647947484386-UpdateCharaterSet.ts
```diff
@@ -0,0 +1,47 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class UpdateCharaterSet1647947484386 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        // Note that as a first step we change the variables name column from 1000 chars long to 750 so that
+        // once we use the potentially 4 byte UTF8mb4 encoding the unique index will still stay below the 3072 byte limit
+        const queries = `SET FOREIGN_KEY_CHECKS = 0;
+ALTER TABLE \`grapher\`.\`variables\` modify column name VARCHAR(750);
```
danielgavrilov, line 8: Was just wondering what length our longest names are, apparently these (429 chars):

| id | name |
|-|-|
| 150649 | 16.1.2 - Number of conflict-related deaths (civilians), by sex, age and cause of death (Number) - VC_DTH_TOCVN - All age ranges or no breaks by age - Both sexes - Incendiary; Chemical, biological, radiological, nuclear (CBRN); Electromagnetic weapons; Less lethal weapons; Denial of access to/ destruction of objects indispensable to survival; Accidents related to conflict; Use of objects and other means; Multiple Weapons Used. |
| 179312 | 16.1.2 - Number of conflict-related deaths (civilians), by sex, age and cause of death (Number) - VC_DTH_TOCVN - All age ranges or no breaks by age - Both sexes - Incendiary; Chemical, biological, radiological, nuclear (CBRN); Electromagnetic weapons; Less lethal weapons; Denial of access to/ destruction of objects indispensable to survival; Accidents related to conflict; Use of objects and other means; Multiple Weapons Used. |
| 312587 | 16.1.2 - Number of conflict-related deaths (civilians), by sex, age and cause of death (Number) - VC_DTH_TOCVN - All age ranges or no breaks by age - Both sexes - Incendiary; Chemical, biological, radiological, nuclear (CBRN); Electromagnetic weapons; Less lethal weapons; Denial of access to/ destruction of objects indispensable to survival; Accidents related to conflict; Use of objects and other means; Multiple Weapons Used. |

danielgavrilov:
When I create a new table:

```sql
CREATE TABLE test (name VARCHAR(20));
```

...the default collation is `utf8mb4_0900_ai_ci`.

Should we add a statement to set the default on the database level?

```sql
ALTER DATABASE
    grapher
    CHARACTER SET = utf8mb4
    COLLATE = utf8mb4_0900_as_cs;
```

PR #1310, opened 2022-03-22
enhance: Add view to mysql that exposes the variables used by charts

danyx23:
We had this in datasette and bigquery but now we can also do it as a view with the shiny new mysql 8 json_table function. This might be useful for data update work.

### DIFF
--- a/db/migration/1647967721182-CreateChartVariablesView.ts
+++ b/db/migration/1647967721182-CreateChartVariablesView.ts
@@ -1,0 +1,36 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class CreateChartVariablesView1647967721182
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE OR REPLACE VIEW chart_variables AS (
+-- get the map variables that are not null
+SELECT
+	id as chartId,
+	config ->> '$.map.variableId' as variableId
+FROM
+	charts
+WHERE
+	config ->> '$.map.variableId' is not null
+UNION DISTINCT
+-- and union it together with all the variables hidden in the dimensions json array, extracted with json_table
+SELECT
+	id as chartId,
+	dims.variableId as variableId
+FROM
+	charts AS c
+CROSS JOIN
+JSON_TABLE(
+  c.config,
+	'$.dimensions[*]'
+  COLUMNS(
+    variableId INT PATH '$.variableId'
+  )
+) AS dims);`)
+    }
+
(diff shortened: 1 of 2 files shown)


danielgavrilov:
There is already a `chart_dimensions` table that is used for this purpose?

danyx23:
Ah, right. To be honest I had just missed that this table exists. But looking at it now it seems that the variable used in the map section of the config is not included (and/or maybe there is some other difference?).

Running this query gives a difference between the content of chart_dimensions and this view (copied it as opposed to querying it because I happened to look at it at a branch):
```sql
with v as 
(
SELECT
	id as chartId,
	config ->> '$.map.variableId' as variableId
FROM
	charts
WHERE
	config ->> '$.map.variableId' is not null
UNION DISTINCT
-- and union it together with all the variables hidden in the dimensions json array, extracted with json_table
SELECT
	id as chartId,
	dims.variableId as variableId
FROM
	charts AS c
CROSS JOIN
JSON_TABLE(
  c.config,
	'$.dimensions[*]'
  COLUMNS(
    variableId INT PATH '$.variableId'
  )
) AS dims)
select
	cd.id,
	cd.order,
	cd.property,
	cd.chartId,
	cd.variableId,
	v.chartId as vcid,
	v.variableId as vvid
from
	chart_dimensions cd
right join v on
	cd.chartId = v.chartId
	and cd.variableId = v.variableId
where
	cd.chartId is null
```

I think that since this is probably not a super frequently used query I'm tempted to replace the existing table with the view because it lowers the maintenance burden somewhat (you can't forget to update chart_dimensions in a script e.g. if you modify the json structure). 

What is nice about the chart_dimensions table is the property and order field, I added those to the view as well.

What do you think, should I remove the code that uses chart_dimensions and replace it with the view?

danielgavrilov:
> because it lowers the maintenance burden somewhat

I agree, it would be great to have it as a view. Is there some way to add a foreign-key constraint on the dimensions this way? One of the current uses of `chart_dimensions` is to prevent data being deleted that is used in charts.

Marigold (approved):
I've leave the decision whether to keep this one or chart_dimensions to you guys. Otherwise looks good, I'd love to have it in staging as soon as possible :)

Marigold:
Ok, I'll switch to `chart_dimensions` until we decide what to do with this.

danielgavrilov:
Seems like we abandoned this and will stay with `chart_dimensions` for now? 

Closing this since it's a bit stale, feel free to reopen if we want to keep it!

danyx23:
Yes that makes sense. I deleted the charts_variables view that was accidentally created on the live db by a migration that should not have been run yet.

PR #1311, opened 2022-03-23
Fix TextWrap: do not render unescaped HTML when linkifyText is true

### LINKED ISSUE #1299: TextWrap: do not render unescaped HTML when linkifyText is `true`
## Description

In the subtitle of Grapher charts we currently use `linkifyText: true` which will wrap any strings beginning with `https://` with `<a>` tags. But it also means we are injecting unsanitized HTML, so if an author writes `<0.1mm` in the subtitle, that gets treated as an HTML tag and is omitted by the browser.

## Expected behaviour

When we wrap with `{ linkifyText: true, rawHtml: false }` we should escape HTML.

## Example

For the following chart, we had to replace `<` with `&#60;` to get the subtitle rendering properly, otherwise it breaks:

https://ourworldindata.org/grapher/prevalence-of-vitamin-a-deficiency-in-pregnant-women

## Technical scope

- [ ] Do a data migration of all HTML entities in published charts
- [ ] Fix the escaping in subtitle rendering

PutawanDE:
Fix #1299 by adding escape for symbol `<` and `>` in function renderHtml of TextWrap when `{ linkifyText: true, rawHtml: false }`
This issue only happens in function renderHtml.

Test:

HeaderHtml - uses fn renderHtml for subtitle -- fixed
image

HeaderView - uses fn render for subtitle - work as expected
image

### DIFF
--- a/grapher/text/TextWrap.tsx
+++ b/grapher/text/TextWrap.tsx
@@ -75,11 +75,24 @@ export class TextWrap {
     @computed get lines(): WrapLine[] {
         const { text, maxWidth, fontSize, fontWeight } = this
 
+        const escapeSymbolMap = {
+            "<": "&lt;",
+            ">": "&gt;",
+            "\n": " \n",
+        }
+
         const words = isEmpty(text)
             ? []
-            : // We prepend spaces to newlines in order to be able to do a "starts with"
+            : // We escape html tag and other symbols.
+              // We also prepend spaces to newlines in order to be able to do a "starts with"
               // check to trigger a new line.
-              text.replace(/\n/g, " \n").split(" ")
+              text
+                  .replace(/[<>\n]/g, (char) => {
+                      return escapeSymbolMap[
+                          char as keyof typeof escapeSymbolMap
+                      ]
+                  })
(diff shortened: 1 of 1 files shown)


danielgavrilov:
Hey @PutawanDE, thanks a lot for the contribution!

There's unfortunately some additional work this needs for us to merge it:

- There is HTML content in some of our existing charts that needs to be changed. It's a task you might potentially be able to help with – finding all charts where the subtitle/note are rendered differently because of this change – but you aren't able to make changes to our live database, so it might not make sense for you to spend time on this. 
- It would be very helpful to have some tests, especially in case there are other characters that need escaping (`escape` from lodash seems to include more characters, I haven't yet researched why.)

Since this issue has a workaround (chart authors can use the HTML entities for `<` and `>`, as some charts already have), it's low priority for us to tackle, so this PR might unfortunately be left unmerged for quite some time on our end – sorry about that! 

If you are looking to make more contributions, it might be better to look for an issue in [good first issues](https://github.com/owid/owid-grapher/labels/good%20first%20issue) than to continue the work on this issue.

PR #1312, opened 2022-03-23
enhance(Makefile): add granular validation for env files

ikesau:
Instead of just checking to see if `WORDPRESS` exists in `.env` or not, we can now make sure that `.env` has every variable defined by referring to our example dotfiles.

Undefined variables get mentioned specifically.

```
$ make validate.env.full
==> Validating your .env file for make up
.env valid for make up
==> Validating your .env file for make up.full
ERROR: .env variable WORDPRESS_DB_NAME not set
make[1]: *** [guard-WORDPRESS_DB_NAME] Error 1
ERROR: .env variable WORDPRESS_DB_USER not set
make[1]: *** [guard-WORDPRESS_DB_USER] Error 1
```

It leverages Makefile's wildcard functionality to run, for example, `make guard-DB_NAME`, which tests if the variable exists and has a length via `if [ -z '${${*}}' ]`.

A limitation of this method is that it does not support optional variables being undefined - we'd require some slightly smarter scripting to allow for that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,8 @@ help:
 	@echo '  make down.full    stop any services still running'
 	@echo
 
-up: require validate.env tmp-downloads/owid_chartdata.sql.gz	
+up: require create-if-missing.env tmp-downloads/owid_chartdata.sql.gz
+	@make validate.env
 	@echo '==> Building grapher'
 	yarn install
 	yarn run tsc -b
@@ -46,8 +47,8 @@ up: require validate.env tmp-downloads/owid_chartdata.sql.gz
 		bind Q kill-server \
 		|| make down
 
-up.full: require validate.env.full tmp-downloads/owid_chartdata.sql.gz tmp-downloads/live_wordpress.sql.gz wordpress/web/app/uploads/2022
-	@grep -q WORDPRESS .env || (echo 'ERROR: your .env is missing some wordpress variables'; exit 1)
+up.full: require create-if-missing.env.fulltmp-downloads/owid_chartdata.sql.gz tmp-downloads/live_wordpress.sql.gz wordpress/web/app/uploads/2022
+	@make validate.env.full
 	
 	@echo '==> Building grapher'
 	yarn install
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,6 +89,23 @@ require:
 	@which tmux >/dev/null 2>&1 || (echo "ERROR: tmux is required."; exit 1)
 	@which finger >/dev/null 2>&1 || (echo "ERROR: finger is required."; exit 1)
 
+guard-%:
+	@if [ -z '${${*}}' ]; then echo 'ERROR: .env variable $* not set' && exit 1; fi
```
larsyencken, line 93: For some reason, with single quotes here it doesn't do the check properly, and I get:

```
$ make validate.env
==> Validating your .env file for make up
/bin/sh: line 0: [: too many arguments
.env valid for make up
```

But with double-quotes, it works for me:

```suggestion
	@if [ -z "${${*}}" ]; then echo 'ERROR: .env variable $* not set' && exit 1; fi
```

larsyencken (changes requested):
Super nice! I've been using Makefiles a lot, but it never occurred to me to include the `.env` file. 

I suggest a quoting fix which made it work for me. The only thing that's worse than before is that on first run of `make up` it used to copy the .env file for you. Now it instead fails and asks you to set a lot of variables.

Seems like in a dev setting, you basically never want to edit `.env` yourself, you always want to use the examples.

We could have a rule like:

- If the user is running `make up`, and their `.env` is missing or identical to `.env.example-full`, replace `.env` with `.env.example-grapher`
- If the user is running `make up.full`, and their `.env` is missing or identical to `.env.example-grapher`, replace `.env` it with `.env.example-full`

That would cover all the standard use cases.

ikesau:
Thanks for catching that, @larsyencken! I got the single-quote error a couple of times and then it stopped for some reason before I could debug it.

I'll move the .env tests down into the validate targets, whoops!

To your suggestion: I like the idea of it (assuming you mean identical-by-keys 😅) as long as we make sure it's not going to overwrite people's env files if they're switching branches while working on a feature that involves adding a new env variable.

larsyencken (approved):
Looks good!

I found a way to actually do it all in one without needing you to rerun `make`, basically by recursively calling `make` for the validation step. https://github.com/owid/owid-grapher/commit/59296e5397cabd3ef2cb85bdcfa1244bc238b621

Shipping it!

larsyencken:
> EDIT: I thought about it a bit more, it seems like it would be annoying for my `.env` file to be overwritten because of how I choose to run my server. Especially because the example env files don't actually contain usable values, right?
> 
> Say I'm developing with `make up.full` using my staging DB and my `.env` file is set accordingly. Afterwards, I test with `make up` but now my `.env` file has extra variables that aren't necessary. I still wouldn't want my `.env` file to be reset!
> 
> Additionally, say I'd been developing with `make up` and have customized the values for `DB_NAME`, etc. If I switch to `make up.full`, I wouldn't want my `DB_NAME` value to be overwritten with the default value again.
> 
> So unless I'm misunderstanding your suggestion, I think just validating that the keys are all accounted for is better.

I was suggesting overwriting uncustomised, default `.env` files only. If you customised it, it should not overwrite.

This would suit a suitation where OWID staff switch between two development modes without ever customising `.env`. But it's probably just too confusing to consider, and it's not at all important!

PR #1314, opened 2022-03-23
chore: add ptolemy staging environment

ikesau:
Ptolemy is arguably not a central figure of data visualisation, but all the good ones were taken and I like the name.


marcelgerber:
nicely done!

PR #1317, opened 2022-03-25
fix(staging): update refresh script for staging

larsyencken:
Fixing the database refresh script on staging that was broken for @ikesau and myself.

- Update server paths (live vs live-db)
- Use different database passwords for grapher and wordpress
- Simplify a little


ikesau (approved):
Copied it from halley to test and it worked perfectly!

PR #1319, opened 2022-03-25
enhance(refresh-staging.sh): check env variables are defined

ikesau:
Just one more little sanity check until we improve this 😊 

Outputs:
```
Importing live Grapher metadata database (owid_metadata)
$GRAPHER_DB_NAME is set to ptolemy_grapher
$GRAPHER_DB_HOST is set to localhost
$GRAPHER_DB_USER is set to ptolemy_grapher
$GRAPHER_DB_PASSWORD is set to *******************
Downloading live Grapher chartdata database (owid_chartdata)
```

or 

```
Importing live Grapher metadata database (owid_metadata)
$GRAPHER_DB_NAME is unset
```

and exits the program.

### DIFF
--- a/wordpress/scripts/refresh-staging.sh
+++ b/wordpress/scripts/refresh-staging.sh
@@ -25,7 +25,7 @@ STAGING_SERVER_NAME=$(basename $DIR | cut -d '-' -f1)
 
 check_env() {
   eval value='$'$1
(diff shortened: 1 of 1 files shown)


larsyencken (commented):
Good call! We don't want the script to even run without the environment variables it needs.

Arguably we could move all these checks to right underneath the `.env` loading.

`echo '$'$1 is unset; exit 1` can also be replaced by `bail "\$$1 is unset"`

PR #1359, opened 2022-04-04
improve: expose mysql-server on port 3307 and improve environment variable naming

ikesau:
A bit of scope creep on this ticket, but hopefully worth it.

## Changes
1. The mysql-server container now exposes the its server on port 3307 instead of 3306. This will allow people to run local MySQL instances and prevent errors occurring due to port conflicts

2. Standardizes environment variable naming across our Makefile, docker-compose, and server code.

Now everything is either:

- `GRAPHER_DB_X`
- `WORDPRESS_DB_X`
- `DB_ROOT_X` - a slightly misleading prefix: refers to both the literal root user password as well as the mysql-server host for other containers to access with.

## Questions
- How will we merge and migrate this safely?


settings/serverSettings.ts
```diff
@@ -42,27 +42,26 @@ export const BLOG_POSTS_PER_PAGE: number =
...
-export const DB_HOST: string = serverSettings.DB_HOST ?? "localhost"
-export const DB_PORT: number =
-    parseIntOrUndefined(serverSettings.DB_PORT) ?? 3306
+export const GRAPHER_DB_NAME: string = serverSettings.GRAPHER_DB_NAME ?? "owid"
+export const GRAPHER_DB_USER: string = serverSettings.GRAPHER_DB_USER ?? "root"
+export const GRAPHER_DB_PASS: string = serverSettings.GRAPHER_DB_PASS ?? ""
+export const GRAPHER_DB_HOST: string =
+    serverSettings.GRAPHER_DB_HOST ?? "localhost"
+export const GRAPHER_DB_PORT: number =
+    parseIntOrUndefined(serverSettings.DB_PORT) ?? 3307
```
danyx23, line 51: Maybe for the default here we should try to stay with 3306? For the blessed setting we override this anyhow but I'm not sure if some dev using local mysql somewhere might not have set this and then it's nicer to stay with the upstream default

ikesau:
@danyx23 done!

Very strange how `valdiate.env` was behaving. I've tweaked it to consistently exit, though it will break out of the script at the very first unset variable, as opposed to listing them all and then exiting 😢 

I also realized I wasn't actually _doing_ anything with the `X_DB_PORT` variables, because it was hard-coded in the docker-compose files. I've now made that dynamic too (it uses the Grapher port, as we currently don't support having two separate MySQL servers)

And finally, added a warning if your port is still set to 3306 👍

danyx23:
Nice! Testing the port doesn't work for me yet. If I have the port at 3306 then I get this error message in make and the script pauses until I ack a message in VS code about the process having terminated (weird interaction between make and the VS Terminal there)
```
/bin/sh: 1: [: 3306: unexpected operator
```

I use zsh, maybe that makes it use /bin/sh instead of bash or something?

PR #1360, opened 2022-04-04
Fixed documentation to run Storybook server

BhCh7051:
This PR fixes step description for Storybook server.


danyx23 (approved):
Good catch, thank you!

PR #1376, opened 2022-04-12
fix(formatValue): fix the ordering of +/- symbols when formatting currency values

### LINKED ISSUE #1330: `formatValue` should include +/- before unit prefix, e.g. +$123 instead of $+123
## Description

In the table tab of some charts (and probably in other cases too), when the unit is `$` (therefore a prefix unit), it is prepended before a `+` or `-` sign. [Example](https://ourworldindata.org/grapher/disposable-household-income-by-income-decile-absolute?tab=table&stackMode=relative):

Screenshot_2022-03-29_at_14_47_30

## Expected behaviour

The prefix unit (`$`) should be _after_ the `+` or `-` sign.

ikesau:
Fixes #1330 

I rewrote the `formatValue` function to make it more linear and easier to reason about. It now relies much more closely on d3's formatting method:

1. Instantiate a formatter function with a locale and appropriate currency symbol
2. Create a [specifier](https://observablehq.com/@ikesau/d3-format-interactive-demo)
3. Format the value with the specifier
4. Postprocess the formatted string to add percentage information and abbreviation, if necessary

Unit tests have also been added.


clientUtils/formatValue.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        ["shortNumberPrefixes true", 1000000000, "1B", { shortNumberPrefixes: true }],
+        ["shortNumberPrefixes false", 1000000000, "1,000,000,000", { shortNumberPrefixes: false }],
+        ["showPlus true", 1, "+1", { showPlus: true }],
+        ["showPlus false", 1, "1", { showPlus: false }],
+        ["showPlus false with negative number", -1, "-1", { showPlus: false }],
+        ["showPlus true with unit", 1, "+$1", { showPlus: true, unit: "$" }],
+        ["showPlus true with % and 4 decimals", 1.23456, "+1.2346%", {showPlus: true, numDecimalPlaces: 4, unit: "%"}],
+        ["showPlus false with $ and noTrailingZeroes true", 1234.5678, "$1,234.57", {showPlus: false, unit: "$", noTrailingZeroes: true}],
+        ["showPlus false with $, noTrailingZeroes false, and noSpaceUnit false", 1234.5678, "$1,234.57", {showPlus: false, unit: "$", noTrailingZeroes: false, noSpaceUnit: false}],
+    ]
```
danielgavrilov, line 36: Super nice to have these test cases 🙌 

We should try to reproduce the one that is going wrong in the svgTester? I think it might be a combination of number prefixes (due to a large number) in combination with a currency, but not sure.

clientUtils/formatValue.ts
```diff
@@ -70,10 +74,15 @@ export function formatValue(
         if (value !== 0 && Math.abs(value) < targetDigits) {
             if (value < 0) output = `>-${targetDigits}`
             else output = `<${targetDigits}`
-        } else
-            output = d3Format(`${showPlus ? "+" : ""},.${numDecimalPlaces}f`)(
+        } else if (unitIsCurrency) {
+            output = d3Format(unit)(
+                `${showPlus ? "+" : ""}$,.${numDecimalPlaces}f`
```
danielgavrilov, line 79: Nice, I wasn't aware that you can pass `$` and it gets replaced with the currency:

```ts
d3.format("($.2f")(-3.5); // localized fixed-point currency, "(£3.50)"
```

But I think this might be causing the problem with omitted units. With number prefixes, we do a recursive call overriding the unit, and the unit gets lost (probably?)

danielgavrilov (commented):
This one seems more tricky to fix than I expected because of the recursive calls. I left some comments, but didn't think of ideas for a potentially good fix...

clientUtils/formatValue.ts
ikesau: I went ahead and made a little sandbox to test the different specifiers: https://observablehq.com/@ikesau/d3-format-interactive-demo

I'm in the process of rewriting the function to rely on d3.format in a much simpler way, with some post-processing at the end for the usecases it can't handle. Hopefully I'll be out on the other side soon!

grapher/barCharts/DiscreteBarChart.test.ts
```diff
@@ -111,11 +111,11 @@ describe("barcharts with columns as the series", () => {
             },
         })
         expect(chart.formatValue(chart.series[0])).toEqual({
-            valueString: "1,002",
+            valueString: "1k",
```
danielgavrilov, line 114: I was just curiously scrolling around, can ignore if this is not yet ready, but we very likely **don't want to replace `1,002` with `1k` here**. 

And in general `k` should not be used for thousands unless we are opting for short number prefixes, which is really a proxy for "make this number take up as little space as possible" which could be a separate setting from the number prefixes, although it aligns pretty well with it so far that it may not be necessary.

danielgavrilov:
Just trying to document the output differences (WIP, feel free to edit):
- `$-` becomes `-$`

- Negative values with number prefixes are getting inconsistent decimal places? ([chart](https://ourworldindata.org/grapher/difference-in-the-value-of-goods-exported-to-and-imported-by-the-us))
Screenshot 2022-04-25 at 12 27 41

- `.5` gets rounded down instead of up? ([chart](https://ourworldindata.org/grapher/percentage-change-energy-by-source))

- Additional decimal places shown when number is below the minimum representable value for the given decimal places. e.g. for 2 decimal places, minimum value is `0.01`. If a number is lower than that, it used to be formatted as `<0.01`. Now it seems there are [cases](https://ourworldindata.org/grapher/prevalence-of-schizophrenia-by-age) where it's formatted with more decimal places than that?

   Screenshot 2022-04-25 at 12 36 41


- additional number prefixes 

   |magnitude|long|short|
   |-|-|-|
   |10¹⁵|quadrillion|P|
   |10¹⁸|quintillion|E|
   |10²¹|sextillion|Z|
   |10²⁴|septillion|Y|
   
  Screenshot 2022-04-25 at 12 40 09

danielgavrilov:
> 🔶 additional number prefixes

I would say few people know what these mean, and we shouldn't go any higher than **trillion**? Especially the short number prefixes, which I had never seen before. 

There is still the issue that in production currently we have some large values like `10 million trillion`, which make little sense formatted that way, but we can keep this issue out of scope for now. I think with the new implementation that will be formatted as `10,000,000 trillion`? Which seems better.

(The ideal fix is to switch to scientific notation for large magnitudes, but definitely out of scope here.)

> 🛑 Negative values with number prefixes are getting inconsistent decimal places?

> 🛑 Additional decimal places shown when number is below the minimum representable value for the given decimal places.

I would say these two are bugs that need fixing? 😬

clientUtils/formatValue.ts
```diff
@@ -1,92 +1,199 @@
...
     numDecimalPlaces?: number
     unit?: string
-    noTrailingZeroes?: boolean
-    noSpaceUnit?: boolean
-    numberPrefixes?: boolean
-    shortNumberPrefixes?: boolean
+    trailingZeroes?: boolean
+    spaceBeforeUnit?: boolean
     showPlus?: boolean
+    numberAbreviation?: "short" | "long" | false
```
danielgavrilov, line 10: I think it should be `Abbreviation` (double b)? Although it's a good way to ab~b~reviate it, so I hope it [catches on](https://en.wikipedia.org/wiki/HTTP_referer#Etymology).

clientUtils/formatValue.ts
```diff
@@ -1,92 +1,199 @@
...
+        long: {
+            k: "k",
+            M: " million",
+            G: " billion",
+            T: " trillion",
+            P: " quadrillion",
+            E: " quintillion",
+            Z: " sextillion",
+            Y: " septillion",
+        },
```
danielgavrilov, line 111: Let's remove all after `trillion` (`T`)? I don't think they're widely understood.

clientUtils/formatValue.ts
```diff
@@ -1,92 +1,199 @@
...
+        string: formattedString,
+        numberAbreviation,
+        spaceBeforeUnit,
+        unit,
+        value,
+        numDecimalPlaces,
+    })
+
+    return postprocessedString
+}
```
danielgavrilov, line 199: I like how neat this is!

danielgavrilov (commented):
Nice, this is much neater! I hope this task didn't take a huge hit on motivation. 😬 (especially since I have a few more comments... 👀)

There are two things I'm wondering about:

- **Should we discard short number abbreviations if the unit is `%`**? ([chart](https://ourworldindata.org/grapher/change-agricultural-inputs-outputs))

   I think we don't do it for long abbreviations, e.g. we don't output `1 million %` but `1,000,000%`, so we also probably shouldn't output `1k%`.

   Screenshot 2022-04-27 at 11 36 56

- We are getting an **additional decimal place** in some cases in the [COVID data explorer](https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&time=2020-03-01..latest&facet=none&pickerSort=desc&pickerMetric=aged_65_older&Metric=Confirmed+cases&Interval=7-day+rolling+average&Relative+to+Population=true&Color+by+test+positivity=false&country=USA~GBR~CAN~DEU~ITA~IND) sidebar:

   Screenshot 2022-04-27 at 11 45 16

   
   Is it likely related to the unit being `%` and short abbreviations enabled? Since it's potentially an explorer issue, let me know if you need help debugging!

clientUtils/formatValue.ts
ikesau: It's a bit tricky unfortunately. I agree, no one really knows these prefixes (hence remapping G to B) and there was work done in a [d3-format PR](https://github.com/d3/d3-format/pull/96) to introduce this functionality natively, but I don't think there's any out-of-the-box way to set a threshold on ab**b**reviation.

Some options we have:
- Switch to exponential notation after trillion (also handling cases where 9.5 trillion would get rounded to 1 quadrillion if precision is set to 1)
- Re-implement the logic to get it to output 1,234T (possibly quite hard to do with this flow)
- Change the prefixes from `P` to `kT` or `quad` or `thousand trillion` or whatever.

Thoughts?

danielgavrilov: On the options:

- I don't think we should be switching to exponential notation in this PR. As we discussed privately, when using exponential notation all values should be in exponential notation, not just the large ones. Also, the default formatting, e.g. `5e3` is problematic, not many people understand that it means `5 × 10^3` (we could rewrite that, but it's another potential can of worms).
- Yes the ideal fix is probably that it outputs `1,234 trillion` as it currently does.
- `thousand trillion` is worse than what we have now (which is no double prefixes until we reach `million trillion`, which in itself is a bug we wanted to fix when eventually there's an important enough use case)

This issue makes me wonder if we want to rely on a 3rd party library to this degree. It's possible there will be another feature in the future that is harder to implement because we have to circumvent `d3-format`, and maybe it would force us to re-implement a subset of `d3-format` ourselves? What do you think?

We could also maybe fork `d3-format` but then we'd have to maintain a codebase which we're not using large parts of.

clientUtils/formatValue.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        ["1 decimal with abbreviation", 1234567, "1.23 million", { numDecimalPlaces: 1, numberAbbreviation: "long" }],
+        ["1 decimal with short abbreviation", 1234, "1.23k", { numDecimalPlaces: 1, numberAbbreviation: "short" }],
+        ["2 decimal with percentage", 19.985, "19.98%", { numDecimalPlaces: 2, unit: "%" }],
+        ["with unit", 1, "$1", { unit: "$" }],
+        ["with custom unit", 1, "1pp", { unit: "pp", spaceBeforeUnit: false }],
+        ["with custom unit and space", 1, "1 pp", { unit: "pp", spaceBeforeUnit: true }],
+        ["negative with unit", -1, "-$1", { unit: "$" }],
+        ["trailingZeroes true", 1.10, "1.1", { trailingZeroes: false }], 
+        ["trailingZeroes false", 1.10, "1.10", { trailingZeroes: true }], 
+        ["$ spaceBeforeUnit false", 1.1, "$1.1", { spaceBeforeUnit: false, unit: "$" }],
```
marcelgerber, line 62: Please add a test case with `{ unit: "£" }`, because I'm currently unsure if that is handled correctly.

marcelgerber (changes requested):
Hi Ike!
This is all great work, and it's interesting how this small issue escalated into a complete rewrite.

However, I am unsure if relying on d3 so much is (still) the right way to go here. We need to do a lot of preprocessing and postprocessing to get out the result we want, and it seems that d3 isn't doing _all that much_ work.
Also, the format specifiers they use are cryptic anyway, and understanding what we need to pass to d3 to get out what is a hard ask.

For these reasons, I feel like we could self-roll our `formatValue` function instead - how hard can it be? 😆 
We got good test coverage now (thank you!), and we can also make use of the well-supported [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat), which gives us nice things like number grouping, `{ signDisplay: "exceptZero" }` and support for significant digits (if we want to go that way).

Let me know what you think, and feel free to discuss if anything's unclear!

ikesau:
Oof, oh no! I wasn't aware of the `NumberFormat` class. I probably would have gone with that if we were going from scratch, but I guess I still wasn't thinking big enough this time 😓

I've definitely come to understand the ins and outs of d3-format, but yeah obviously it's pretty oblique interface. Let me play around with NumberFormat a little bit and see how it handles before making a decision.

ikesau:
Okay, so I made a [Intl.NumberFormat demo](https://observablehq.com/@ikesau/intl-numberformat-interactive-demo)

It's basically the same thing. An object that you configure with a bunch of settings and then format with.

Some differences:
- Its prefixes only go up to trillion, and only in short form (`100K`, `1T`, `1000T`, etc)
- Its rounding, decimal place, and SF logic is all a bit more complex, but can still be configured for our use-case.

So, we _could_ rewrite it to ditch d3-format (though it will still be in our bundle because `d3-scale` requires it) but I don't think the `formatValue` function will become any more readable. If we use `NumberFormat`, we'll still have a function that generates a config and postprocesses the output.

danielgavrilov:
I am also worried about increasing reliance on `d3-format`. It makes us less able to make improvements to the implementation – the fact that we can't find an easy way around not outputting "sextillion" is a clear example of this, and I think there will probably be a few more cases in the future since we are very particular about value formatting.

If fixing the "sextillion" issue means having to re-implement `d3-format` functionality ourselves, it makes "rolling our own" not such a bad option? Especially since the old 90-line function already did this, relying very little on `d3-format`, it can't be _that_ complex? 😅

danielgavrilov (approved):
Sorry, we have a thread of a discussion here around architecture that I started and isn't too productive. I think it would be best to move that discussion elsewhere.

**This PR fixes a bug, improves readability significantly, adds comprehensive tests**, and in case we find our requirements change in the future in such a way that our custom implementation is better, we can always bring back some of the old implementation.

For now, I think it would be best to **merge and deploy** as it is. 🚀

marcelgerber (approved):
I agree, let's merge this monster of a PR. It's a big improvement, and as gavrilov says we can still revisit this in the future.

@ikesau The big green Merge button is all yours :shipit:

PR #1393, opened 2022-04-15
VS code Devcontainer fixes

danyx23:
There were a few issues with using the VS Code remote container extensions to run grapher. I have fixed a few things and ran a trial on a fresh machine.

On this machine I realized that git still has this weird default behaviour on Windows where files are changed transparently from LF line endings to CRLF. This then causes issues when e.g. shell scripts running in the devcontainer that is linux based or in a WSL environment which is also linux based.

I tried to fix this with .gitattribute settings but it doesn't seem to be enough, so I added some docs explaining the issue.


docker-compose.devcontainer.yml
```diff
@@ -51,7 +51,8 @@ services:
         ports:
             # Exposing via the port specified for Grapher
```
ikesau, line 52: We should remove these comments

docs/before-you-start-on-windows.md
```diff
@@ -0,0 +1,20 @@
+## Set up Git on Windows
+
+If you are on Windows, make sure that you configure Git to use linux line ending (LF) instead of windows line endings (CRLR) **before** checking out this repository on your local machine. If you fail to do so then the line endings will be converted to the usual windows default for all files including shell scripts and when those are then executed inside the Linux based devcontainer the scripts to load the database etc will fail.
+
+To verify the line endings, open a file (e.g. the README.md file) and have a look at the VS Code status bar in the lower rifht corner. If the settings are correct it should show "LF". If your line endings were switched by git to the windows ones then it will read "CRLF"
```
ikesau, line 5: lower right*

docs/devcontainer-setup.md
```diff
@@ -2,20 +2,50 @@
...
+make up.devcontainer
+```
+
+This will run [tmux](https://github.com/tmux/tmux/wiki/Getting-Started) and create 3 panels that you can switch between with `<C-b>, n` (i.e. press `CTLR` (on PC)/`CMD` (on Mac) and `b`, then release and press `n`). Mouse support is enabled so you should be able to scroll through the panels and click them in the bottom row. A short welcome message is printed on the initial pane - if you scroll up here you will see a quick cheatsheet with various commands.
+
+Now navigate to http://localhost:3030/admin/charts in a browser and have a look at the admin interface. The default user account is `admin@example.com` with a password of `admin`
+
+To stop the admin servers, press `<C-b>, Q` in the terminal to kill the window and end all 3 processes when you are done. Close VS Code to shut down all docker containers and free up the resources of running the MySQL docker container.
+
+## Accessing Mysql
```
ikesau, line 23: Anal-retentive style comment: can we try to always write "MySQL" when in docs? 😊

docs/devcontainer-setup.md
```diff
@@ -2,20 +2,50 @@
...
+
+If you want to access mysql you have two options. ⚠ Note that depending on which one you choose, you will have to use different ports!
+
+1. From the terminal
+
+    In the VS Code terminal that executes inside the devcontainer, run the mysql command line client:
+
+    ```bash
+    mysql -h db -u grapher -p grapher
```
ikesau, line 32:
```suggestion
    mysql -h db -u grapher -pgrapher
```

docs/devcontainer-setup.md
```diff
@@ -2,20 +2,50 @@
...
+    ```
+
+    This will ask you for the password (enter `grapher`) and then show a prompt. See the https://dev.mysql.com/doc/refman/8.0/en/getting-information.html for how to query the database in this interface.
```
ikesau, line 35: with the above, the password prompt should go away

PR #1401, opened 2022-04-19
Bulk-fastt-enable-latest-earliest

danyx23:
This PR is part of #1224 - it adds a new operator so that on time fields like maxTime you can query for entries "latest" or "earliest"

### DIFF
--- a/adminSiteClient/GrapherConfigGridEditorTypesAndUtils.tsx
+++ b/adminSiteClient/GrapherConfigGridEditorTypesAndUtils.tsx
@@ -342,6 +342,38 @@ export function SExpressionToJsonLogic(
     })
 }
 
+/** JsonLogic is the easiest format that the React Awesome Query Library can round
+    trip (i.e. deserialize from). Building the internal structure of the query library
+    would be tedious so we convert our SExpressions to JsonLogic. When React Awesome
+    Query Library parses this, we need to convert some custom things like the is_latest
+    query operator that we added to RAQL that in our SExpressions are represented as
+    field = "latest"
+ */
+export function postProcessJsonLogicTree(filterTree: JsonTree | JsonItem) {
+    if (filterTree.type === "group" && filterTree.children1) {
+        if (isArray(filterTree.children1))
+            for (const child of filterTree.children1)
+                postProcessJsonLogicTree(child)
+        else if (isPlainObject(filterTree.children1))
+            for (const child of Object.values(filterTree.children1))
+                postProcessJsonLogicTree(child)
+    } else if (filterTree.type === "rule") {
+        const isLatest =
+            filterTree.properties.value.length === 1 &&
+            filterTree.properties.value[0] === "latest"
+        const isEarliest =
+            filterTree.properties.value.length === 1 &&
+            filterTree.properties.value[0] === "earliest"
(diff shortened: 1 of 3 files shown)


danielgavrilov (approved):
Looks good!

I am doing just a basic sense-check review since I don't have a great handle of the codebase here, let me know if it needs more than that.

adminSiteClient/GrapherConfigGridEditorTypesAndUtils.tsx
```diff
@@ -339,6 +340,38 @@ export function SExpressionToJsonLogic(
...
+        const isEarliest =
+            filterTree.properties.value.length === 1 &&
+            filterTree.properties.value[0] === "earliest"
+        const isEqual =
+            filterTree.properties.operator === "equal" ||
+            filterTree.properties.operator === "select_equals"
+        if (isLatest && isEqual) {
+            filterTree.properties.operator = "is_latest"
+        } else if (isEarliest && isEqual) {
+            filterTree.properties.operator = "is_earliest"
```
ikesau, line 371: I'd deconstruct to DRY this up a bit.

```
const { properties: { value }, operator }  = filterTree

if (value.length !== 1) return

const isLatest = value[0] === "latest"

const isEarliest = value[0] === "earliest"

const isEqual = operator === "equal" || operator === "select_equals"
```
Up to you! Still getting a sense for the preferred idioms on this project.

ikesau (approved):
Same as gavrilov! Assuming the logic's good, just a few style suggestions.

adminSiteClient/GrapherConfigGridEditorTypesAndUtils.tsx
```diff
@@ -488,12 +534,17 @@ export function fieldDescriptionToFilterPanelFieldConfig(
...
         .with(EditorOption.mappingEditor, () => undefined)
         .with(EditorOption.numeric, () => "number")
+        .with(EditorOption.numericWithLatestEarliest, () => "number")
         .with(EditorOption.primitiveListEditor, () => undefined)
         .with(EditorOption.textarea, () => "text")
         .with(EditorOption.textfield, () => "text")
         .exhaustive()
 
-    if (widget !== undefined)
+    if (widget !== undefined) {
```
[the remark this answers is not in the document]
danyx23: I think the cognitive overhead of falsy checks is not great - you have to consider the type to know the behaviour. This is why I prefer to check explicitly for undefined (or lodash.isNil when either null or undefined).

adminSiteClient/GrapherConfigGridEditorTypesAndUtils.tsx
danyx23: Yeah true, that reads nicer. My brain always goes into legacy mode with these things in TS as in an ML language could write it like this:
```fsharp
match (filterTree.properties) with
| {value = ["latest"], operator = "equal"}
| {value = ["latest"], operator = "select_equal"} -> "is_latest"
| {value = ["earliest"], operator = "equal"}
| {value = ["earliest"], operator = "select_equal"} -> "is_earliest"
```

danyx23:
Thanks @ikesau and @danielgavrilov for your reviews!

PR #1406, opened 2022-04-20
improve(export-graphs.ts): add flag to support specific graph

ikesau:
I wanted to run the svg diff tester locally, but not to redundantly generate hundreds of graphs I wasn't interested in.

This option seems simpler than temporarily deleting config directories in the diff tester repo 🙃 

Example usage:

```bash
node ./itsJustJavascript/devTools/svgTester/export-graphs.js \
        -i ../owid-grapher-svgs/configs \
        -o ../owid-grapher-svgs/svgs \
        -c 1000
```

Outputs:
```bash
# in ~/owid-grapher-svgs/svgs
ls
capture-fishery-production_v14_850x600.svg results.csv
```

### DIFF
--- a/devTools/svgTester/export-graphs.ts
+++ b/devTools/svgTester/export-graphs.ts
@@ -10,7 +10,7 @@ async function main(parsedArgs: parseArgs.ParsedArgs) {
     try {
         const inDir = parsedArgs["i"] ?? "grapherData"
         const outDir = parsedArgs["o"] ?? "grapherSvgs"
-        const numPartitions = parsedArgs["n"] ?? 1
+        const targetConfig = parsedArgs["c"]
         if (!fs.existsSync(inDir))
             throw `Input directory does not exist ${inDir}`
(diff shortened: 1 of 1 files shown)


devTools/svgTester/export-graphs.ts
```diff
@@ -19,7 +19,11 @@ async function main(parsedArgs: parseArgs.ParsedArgs) {
         const directories = []
         for await (const entry of dir) {
             if (entry.isDirectory()) {
-                directories.push(entry.name)
+                if (String(targetConfig) === entry.name) {
```
danyx23, line 22: Just curious: what does the String constructor do here?

ikesau: Oop! A leftover from an earlier attempt which used fs to check if targetConfig existed, and threw an error because `parsedArgs` was exporting it as a number.

Switching to a loose equality check 👍

Issue #1346, opened 2022-03-30
Prevent bar label overlaps by decreasing font size


danielgavrilov:
## Problem

[When many entities are selected](https://ourworldindata.org/grapher/economic-decline-in-the-second-quarter-of-2020), the bar labels start to overlap:

![economic-decline-in-the-second-quarter-of-2020](https://user-images.githubusercontent.com/1308115/160827876-562acf5c-906f-4b93-873e-847e73741904.png)

## Proposed solution

We should decrease the font size based on the number of entities selected and the available height.

PR #1407, opened 2022-04-21
Fix bar chart font size

MahmoudHamdy02:
I'm still trying to fully understand the code but I had an initial idea for this fix. (Issue #1346)

I first thought of editing the `legendLabelStyle` function, but I couldn't figure out how to access `barHeight`, so I moved `fontSize` and `fontWeight` inside the `<text>` tags.

For now, this only takes into consideration `barHeight` with some approximate values for the heights and font sizes.

Please let me know if this approach is alright and I will do the same for `valueLabelStyle` as well.

### DIFF
--- a/grapher/barCharts/DiscreteBarChart.tsx
+++ b/grapher/barCharts/DiscreteBarChart.tsx
@@ -165,8 +165,9 @@ export class DiscreteBarChart
         fontSize: number
         fontWeight: number
     } {
+        const availableHeight = this.boundsWithoutColorLegend.height / this.barCount
         return {
-            fontSize: 0.75 * this.fontSize,
+            fontSize: Math.min(0.75 * this.fontSize, 1.1 * availableHeight),
             fontWeight: 700,
         }
     }
@@ -175,8 +176,9 @@ export class DiscreteBarChart
         fontSize: number
         fontWeight: number
     } {
+        const availableHeight = this.boundsWithoutColorLegend.height / this.barCount
         return {
-            fontSize: 0.75 * this.fontSize,
+            fontSize: Math.min(0.75 * this.fontSize, 1.1 * availableHeight),
             fontWeight: 400,
         }
     }
@@ -418,9 +420,7 @@ export class DiscreteBarChart
                                 fill="#555"
                                 dominantBaseline="middle"
                                 textAnchor="end"
-                                fontSize={this.barHeight < 6 ? 10
-                                : this.barHeight < 8 ? 11 : 0.75 * this.fontSize}
-                                fontWeight={700}
+                                {...this.legendLabelStyle}
                             >
                                 {series.seriesName}
                             </text>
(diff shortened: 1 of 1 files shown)


danielgavrilov:
Hi @MahmoudHamdy02! You are on the right track! And I thought this would be an easy one – just access `barHeight` in `legendLabelStyle` – but I can see now that it causes a cycle. It's very common to run into a cycle like this in our codebase.

Since the label font size has effects on other layout parameters, for example the allocated horizontal space for labels, I think we should try to define it early on, in `legendLabelStyle` as you suggested, and that will then take care of any downstream effects.

I haven't looked at it in depth, but I'll think out loud for a bit as an example of one way I might tackle the issue (and it maybe isn't the best way, feel free to experiment):

The height of the plot (just the bars, excluding title, axes, etc) seems to be defined by `innerBounds`: https://github.com/owid/owid-grapher/blob/9c64d1f6c4aeae97e78b523f413d7c60a369a47c/grapher/barCharts/DiscreteBarChart.tsx#L290-L292

And the definition of `innerBounds`:
https://github.com/owid/owid-grapher/blob/9c64d1f6c4aeae97e78b523f413d7c60a369a47c/grapher/barCharts/DiscreteBarChart.tsx#L274-L279

So the plot height could technically be derived from only `this.boundsWithoutColorLegend.height - this.yAxis.height`. It might even be the case that we don't really care about too much accuracy here for the font size, so we can get away with just using `this.boundsWithoutColorLegend.height` as the plot height. We can divide that by the bar count to get a rough estimate of the available height for the bar label.

So maybe `legendLabelStyle` could be derived this way:

```ts
@computed private get legendLabelStyle(): {
    fontSize: number
    fontWeight: number
} {
    const availableHeight = this.boundsWithoutColorLegend.height / this.barCount
    return {
        fontSize: Math.min(0.75 * this.fontSize, availableHeight),
        fontWeight: 700,
    }
}
```

But then I'm not sure that the height is a good estimate for the font size, maybe there should be some scaling factor there (like `1.2 * availableHeight`). I'd try it out and see.

And also, this logic can potentially lead to some very small and unreadable font sizes, so I'd maybe cap the minimum font size to `6px`. But this is arbitrary and maybe unnecessary, since a bar chart with that many bars will be unreadable regardless what we do.

I hope some of that helps! Let me know if you have questions.

MahmoudHamdy02:
This will take a lot of experimenting indeed, I will have another go at it and update you on any progress.

MahmoudHamdy02:
Hi @danielgavrilov! I have implemented the proposed solution, I've also added it to `valueLabelStyle`.

> maybe there should be some scaling factor there

In my opinion, 1.1 works well, the font is readable with minimal overlapping.

> maybe cap the minimum font size to 6px.

With all countries selected, the font size is close to 7.5, so I don't think this would be necessary.

This type of issue is a bit subjective so feel free to suggest more changes!

danielgavrilov:
Thanks @MahmoudHamdy02! This seems to work well! I think you can unmark as draft 🙂

One change that might be a good idea: since both the `legendLabel` and `valueLabel` have the same font size (but different font weight), maybe `labelFontSize` could be extracted as a `@computed` property, and used in both implementations?

danielgavrilov (approved):
Awesome, thanks for looking into this one @MahmoudHamdy02! 🙌  Will merge and deploy 🚀

Issue #1399, opened 2022-04-19
Escape key should exit the entity selector for charts


larsyencken:
## Core problem

In a chart like this one (https://ourworldindata.org/grapher/animal-protein-vs-gdp?country=~CHN) you can use the entity selector. It's pretty keyboard friendly, but you cannot escape it easily with a single key.

<img width="1118" alt="Screen Shot 2022-04-19 at 13 21 43" src="https://user-images.githubusercontent.com/36012/163992778-ba79a9fc-18ce-4b56-9a4f-7a5ddfddfa7e.png">

## Proposal

Have the escape key close the selection dialog.

danielgavrilov:
Fixed in https://github.com/owid/owid-grapher/pull/1409

PR #1409, opened 2022-04-21
Escape key to exit entity selector

MahmoudHamdy02:
Initial fix for issue #1399.

I moved the condition to check for "Escape" key to a separate function, and added it to `searchResults` div.

### DIFF
--- a/grapher/controls/EntitySelectorModal.tsx
+++ b/grapher/controls/EntitySelectorModal.tsx
@@ -112,7 +112,7 @@ export class EntitySelectorModal extends React.Component<{
         document.removeEventListener("click", this.onDocumentClick)
     }
 
-    @action.bound onEscapeKeyDown(
+    @action.bound onOverlayKeyDown(
         e: React.KeyboardEvent<HTMLInputElement>
     ): void {
         if (e.key === "Escape") this.props.onDismiss()
@@ -178,7 +178,7 @@ export class EntitySelectorModal extends React.Component<{
         const { searchResults, searchInput } = this
 
         return (
-            <div className="entitySelectorOverlay">
+            <div className="entitySelectorOverlay" onKeyDown={this.onOverlayKeyDown}>
                 <div
                     ref={this.base}
                     className={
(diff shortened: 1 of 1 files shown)


grapher/controls/EntitySelectorModal.tsx
```diff
@@ -112,13 +112,19 @@ export class EntitySelectorModal extends React.Component<{
         document.removeEventListener("click", this.onDocumentClick)
     }
 
+    @action.bound onEscapeKeyDown(
```
danielgavrilov, line 115: I am nitpicking, but should we maybe name this differently? 

In `onSearchKeyDown`, I think the "search" refers to the search field, which is the element we are attaching the event handler to. Here we are attaching the event handler to the overlay so maybe `onOverlayKeyDown`?

Because we have an if statement to check if the key is an escape, it doesn't seem like it's exactly `onEscapeKeyDown`.

grapher/controls/EntitySelectorModal.tsx
```diff
@@ -190,7 +196,7 @@ export class EntitySelectorModal extends React.Component<{
                         </h2>
                     </header>
                     <div className="entities wrapper">
-                        <div className="searchResults">
+                        <div className="searchResults" onKeyDown={this.onEscapeKeyDown}>
```
danielgavrilov, line 199: We could maybe attach this event handler somewhere higher in the tree to capture events even if the focus element is outside the search results? Maybe the top-level `<div className="entitySelectorOverlay">`

The ideal might be that we even attach it to the `window`... but that's messier, and probably not necessary for this context.

danielgavrilov (changes requested):
Awesome! 🙌  I just have some nitpicky comments, but the approach is good. 🙂 Thanks a lot for contributing!

grapher/controls/EntitySelectorModal.tsx
MahmoudHamdy02: Of course! I just put in the name as a placeholder but the new name definitely makes more sense.

grapher/controls/EntitySelectorModal.tsx
MahmoudHamdy02: I thought about this but didn't know which div would be the best to attach to, thanks for clearing that up! The changes should be pushed now.

danielgavrilov: I think this change might not have made a difference, because it seems `keydown` events are only captured when the focus is on a checkbox or the search input field – sorry about that!

danielgavrilov (approved):
Thank you @MahmoudHamdy02! Merging and deploying this now.

Issue #1296, opened 2022-03-21
Change the default color scheme for DiscreteBar to 'OWID Distinct'


danielgavrilov:
A request from Max ([Slack discussion](https://owid.slack.com/archives/CLDM2AWCD/p1646406804766299)).

We would like `OWID Distinct` to become the default color scheme for DiscreteBar. Currently all bars are blue by default ([chart](https://ourworldindata.org/grapher/global-warming-potential-of-greenhouse-gases-over-100-year-timescale-gwp?country=~OWID_WRL)): 

<img src="https://user-images.githubusercontent.com/1308115/159249327-4220341e-c69c-4811-ab76-9e12c34f950d.png" width=600 />

With `OWID Distinct`:

<img src="https://user-images.githubusercontent.com/1308115/159249527-2e3871dd-3e6e-43c4-b751-ff8240dd9439.png" width=600 />

danielgavrilov:
Currently we have some logic to check if a DiscreteBar is a LineChart-turned-DiscreteBar using the timeline, perhaps that logic becomes part of DiscreteBar now?

danielgavrilov:
There are some implementation details in a (now closed) PR: https://github.com/owid/owid-grapher/issues/1296

danyx23:
Closing this; as part of issue #1118 it was decided that distinct bar charts should be single color by default instead and it was implemented like this

PR #1421, opened 2022-04-30
feat(DiscreteBar): Changes default color scheme for DiscreteBar to OWID Distinct

coder-whale:
Closes issue #1296 

All discrete bar charts now use owid-distinct color scheme by default

### DIFF
--- a/grapher/barCharts/DiscreteBarChart.tsx
+++ b/grapher/barCharts/DiscreteBarChart.tsx
@@ -607,11 +607,7 @@ export class DiscreteBarChart
     }
 
     @computed private get colorScheme(): ColorScheme | undefined {
-        // If this DiscreteBarChart stems from a LineChart, we want to match its (default) color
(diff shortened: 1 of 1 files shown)


coder-whale:
Thanks for the detailed suggestions. I have attempted to implement those changes, however the colour distribution is a lot more uniform while invoking the `getColors` function as compared to `categoricalColorAssigner`. The comparison below indicates the same ([link to chart](http://localhost:3030/grapher/difference-in-the-value-of-goods-exported-to-and-imported-by-the-us)):
colorassigner
 I would no longer be able to work on this task due to other constraints. I apologize for the delayed response.

danielgavrilov:
> I have attempted to implement those changes, however the colour distribution is a lot more uniform while invoking the `getColors` function as compared to `categoricalColorAssigner`.

Ah, interesting! Interpolated/gradient color schemes might need some special handling.

> I would no longer be able to work on this task due to other constraints. I apologize for the delayed response.

No worries at all! I will close the PR, but it's still very helpful as extra documentation around the issue.

Issue #1084, opened 2022-01-24
Tooltip on hover for faceted charts


larsyencken:
## Overview

When you mouse-over a normal chart, you get a small infobox giving you a summary of the thing you're hovering on. For example, for line charts, you get the values of all series at the time point you're hovering on. But faceted charts are missing this interactivity, making it much harder to read off an exact value from a faceted chart.

## Context

When implementing faceted charts, we did not have the opportunity to complete this part. But it's the most glaring missing piece right now.

marcopiani:
Hi!

I see this is now implemented! Great job!

One idea/request: would it be possible to have the option (to be selected with a checkbox, perhaps) to have the tooltip shown in parallel on all subplots?

See example at [https://projects.fivethirtyeight.com/biden-approval-rating/](https://projects.fivethirtyeight.com/biden-approval-rating/)
![image](https://user-images.githubusercontent.com/18381804/169087204-85130d7f-9548-4406-9c1b-6badb498c6dd.png)

danyx23:
@marcopiani thanks for your continued interest! Yes we are aware that this would be nice to have but in our case we have to find enough screen real estate to show all those tooltips and make sure they don't overlap each other on smaller screens etc. We will likely continue work on this but we thought it is important enough to ship this first step and tackle this other improvement later.

marcopiani:
Hi @danyx23 many thanks for the prompt reply.

I understand the need to give priority to other features.

Thank you for considering the option, and, as always, for the work of the whole OWID team.

PR #1235, opened 2022-03-07
Scatter size legend #2: Feature work

danielgavrilov:
Fixes #1089 | [Deployed on playfair](https://playfair.owid.cloud/admin/test/embeds?type=ScatterPlot) | [Slack #design discussion](https://owid.slack.com/archives/C5BDCB2R3/p1645630994255989)

This PR adds a ScatterPlot size legend:

https://user-images.githubusercontent.com/1308115/157084760-c7fb71ac-7b03-4707-9620-d1c9ad26bfd7.mp4

## Scale changes

- The point radius scale has been changed to be more perceptually accurate – the area did not always correspond to magnitude in the past, [for example](https://playfair.owid.cloud/admin/test/embeds?ids=308&comparisonUrl=https%3A%2F%2Fourworldindata.org) (before/after):

   Screenshot 2022-03-09 at 17 14 16

   (Luxembourg **~$106**, Poland **~$26**. Perceptually the scale was wrong in some cases in the past.)

- It was possible [to have zero-radius, invisible points](https://playfair.owid.cloud/admin/test/embeds?ids=5277&comparisonUrl=https%3A%2F%2Fourworldindata.org) before:

   Screenshot 2022-03-09 at 17 42 58

PR #1425, opened 2022-05-03
improve(FacetChart.tsx): tooltips

ikesau:
Partially resolves https://github.com/owid/owid-grapher/issues/1084

## Changes

**The Tooltip module now stores tooltips in an `ObservableMap` instead of referencing a single instance.**
_This will allow for simultaneous synchronised tooltips on facet charts in the future_

**Connects  `FacetChart`'s managers to the parent `Grapher` `tooltips` instance**
_This allows the tooltip to be added to the tooltips map correctly and be rendered by Grapher._

## Example

facets

### DIFF
--- a/grapher/lineCharts/LineChart.tsx
+++ b/grapher/lineCharts/LineChart.tsx
@@ -259,7 +259,7 @@ class Lines extends React.Component<LinesProps> {
     private container?: SVGElement
     componentDidMount(): void {
         const base = this.base.current as SVGGElement
-        const container = base.closest("svg") as SVGElement
+        const container = base.closest("g.LineChart") as SVGElement
         container.addEventListener("mousemove", this.onCursorMove)
         container.addEventListener("mouseleave", this.onCursorLeave)
         container.addEventListener("touchstart", this.onCursorMove)
@@ -364,14 +364,10 @@ export class LineChart
         return table
     }
 
-    // todo: rename mouseHoverX -> hoverX and hoverX -> activeX
-    @observable mouseHoverX?: number = undefined
-    @action.bound onHover(hoverX: number | undefined): void {
-        this.mouseHoverX = hoverX
-    }
+    @observable hoverX?: number = undefined
 
-    @computed get hoverX(): number | undefined {
-        return this.mouseHoverX ?? this.props.manager.annotation?.year
+    @action.bound onHover(hoverX: number | undefined): void {
+        this.hoverX = hoverX ?? this.props.manager.annotation?.year
     }
 
     @computed private get manager(): LineChartManager {
@@ -707,7 +703,7 @@ export class LineChart
                 />
             )
 
-        const { manager, tooltip, dualAxis, hoverX, clipPath } = this
+        const { hoverX, manager, tooltip, dualAxis, clipPath } = this
         const { horizontalAxis, verticalAxis } = dualAxis
 
         const comparisonLines = manager.comparisonLines || []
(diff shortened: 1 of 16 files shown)


grapher/lineCharts/LineChart.tsx
```diff
@@ -259,7 +259,7 @@ class Lines extends React.Component<LinesProps> {
     private container?: SVGElement
     componentDidMount(): void {
         const base = this.base.current as SVGGElement
-        const container = base.closest("svg") as SVGElement
+        const container = base.closest("g.LineChart") as SVGElement
```
danielgavrilov, line 262: Since this element is in the render tree of the component:

https://github.com/owid/owid-grapher/blob/f0fb2f0352853da969392375e7ce86aaac84604c/grapher/lineCharts/LineChart.tsx#L718

...it might be better to attach event listeners through React, instead of the DOM?

danielgavrilov:
I'm not sure if this PR is the cause or the code it's based on, but the map tooltips don't highlight the selected value:

https://user-images.githubusercontent.com/1308115/168818575-1af3db52-96a1-4d79-9046-e0d942901e1c.mp4

Are properties potentially getting dropped somehow?

danielgavrilov:
The tooltip positioning logic needs some tweaking, but can be out of scope for this PR, since having tooltips at all is such a worthwhile feature.

https://user-images.githubusercontent.com/1308115/168819569-8f6ccada-754f-4159-8028-190137783ce9.mp4

danielgavrilov (changes requested):
Nice way to support multiple tooltips and also easily pass it onto faceted charts! 

I think the only blocker is that map tooltips seem to be missing a highlight bubble, the rest is nitpicking, can be addressed later.

ikesau:
@danielgavrilov Fixes and changes implemented! 

See [this diff](https://github.com/owid/owid-grapher/pull/1425/files/b6f5a1ac7058e19be6890285922689bb06c61f24..3513cf8f86195b0acd59c2e75c1ca61ceb0bdbb5) to exclude a find-and-replace commit I made. If we don't want to degrade git blame by boyscouting in this way, or disagree with the "check" verb pattern, let me know. 🙂

danielgavrilov:
I think boyscouting like this is fine. If there is a big boyscouting diff, in the past I've split up PRs, e.g.:

- https://github.com/owid/owid-grapher/pull/1234 and https://github.com/owid/owid-grapher/pull/1235
- https://github.com/owid/owid-grapher/pull/1159 and https://github.com/owid/owid-grapher/pull/1160
- https://github.com/owid/owid-grapher/pull/1061 and https://github.com/owid/owid-grapher/pull/1057

That takes some time (and potentially rebasing pain) but makes reviews somewhat easier.

---

On the "check" verb pattern specifically, I think it's good to quickly discuss with the team whether we want to adopt this pattern, and maybe even make one sweep at renaming a lot of things rather than renaming them one by one. The risk I see is that we end up even more inconsistent if we do it very slowly, some parts of the codebase rarely get touched. And it's better if everyone does the renaming rather than just by yourself.

danielgavrilov:
I think something might still going wrong with `hoverX` – the bubble is not shown in some cases? I can help with debugging if you want, after reading through for a few minutes I couldn't find anything obviously wrong. 

The chart I'm looking at is at http://localhost:3030/admin/test/embeds?ids=4703&tab=map

https://user-images.githubusercontent.com/1308115/168991429-df7e6bb5-18f5-4f5e-837f-81b50d0e4e74.mp4

danielgavrilov (approved):
Awesome! 🙌  Let's give people tooltips! 💬

PR #1442, opened 2022-05-17
fix: consistent grapher schema URLs

### LINKED ISSUE #1439: Use consistent Grapher schema URL
We are currently linking to files in `nyc3.digitaloceanspaces.com` which is harder to redirect from if we moved ever the files in the future. Instead it's better to link to `files.ourworldindata.org` which we have complete control over. We could also use a different domain for these things, but this one is already (halfway) set up.

There was an old PR for this issue (#1152) that got halted because we need to first set up a Cloudflare Worker in front of `files.ourworldindata.org` in order to support CORS (https://github.com/owid/cloudflare-workers/pull/1). Currently, the redirect for `files.ourworldindata.org` to the DigitalOcean bucket is set up through the Cloudflare UI, and it drops CORS headers.

danielgavrilov:
Fixes #1439, supersedes #1152.

Now we have a Cloudflare Worker in place for `files.ourworldindata.org`, so we can change all references.

### DIFF
--- a/adminSiteClient/GrapherConfigGridEditor.tsx
+++ b/adminSiteClient/GrapherConfigGridEditor.tsx
@@ -1086,8 +1086,6 @@ export class GrapherConfigGridEditor extends React.Component<GrapherConfigGridEd
     }
 
     async getFieldDefinitions() {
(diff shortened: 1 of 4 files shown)


danyx23 (approved):
Looks great - thanks for your work on this!

PR #1443, opened 2022-05-17
fix(facets): do not share Y axis on faceted DiscreteBar

danielgavrilov:
`DiscreteBar` used to try to collapse the axis when "uniform Y axis" is checked, this fixes that ([chart](https://ourworldindata.org/grapher/share-who-currently-had-depression-by-age-and-gender)):

https://user-images.githubusercontent.com/1308115/168875112-277ddf4a-ac95-450f-bba8-a79c62d20f67.mp4

### DIFF
--- a/grapher/controls/Controls.tsx
+++ b/grapher/controls/Controls.tsx
@@ -122,7 +122,7 @@ export class FacetYDomainToggle extends React.Component<{
                     onChange={this.onToggle}
                     data-track-note="chart-facet-ydomain-toggle"
                 />{" "}
-                &nbsp;Uniform y-axis
+                &nbsp;Align axis scales
             </label>
         )
     }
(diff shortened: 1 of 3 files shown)


danielgavrilov:
The "Uniform y-axis" label is also not accurate for bar charts. I wonder what a good name is:

- Uniform axis
- Share axis between all charts
- Same axis for all charts

Any more suggestions? I think I'll collect some and post it in #design.

danielgavrilov:
> a bit confusing because _uniform_ and _shared_ axes are usually, but not always, the same thing.

Yes I agree, we should rename the config setting, but we would need to migrate Grapher & Explorer configs 😭 

For now I left it as a `TODO`: https://github.com/owid/owid-grapher/blob/48e0a48408c8eb19fc9f6cae0708b6b550aaa1de/grapher/core/GrapherConstants.ts#L50-L53

Issue #1338, opened 2022-03-30
Delete or archive old datasets that are not used in any chart


larsyencken:
## Proposal

We could enumerate, manually verify, then delete all datasets >6m old which are not used in any chart.

## Scope

- [x] Enumerate datasets to delete (>6 month old never used non-bulk OR >3y old never used bulk)
- [x] Manually verify
- [x] ~~Delete in a migration~~ Archive in a migration
- [ ] Hide archived data from the chart builder admin

## Rationale

We would like to move the "source of truth" of our data from MySQL to the ETL over time, which involves backporting old datasets for which we have no recipe from MySQL to the ETL, to be republished to the general public. We do not want to ship data to the public that we're not happy with, and we do not want to QA data that by definition wasn't important enough to use. So we should delete it instead.

## Related issues

- https://github.com/owid/etl/issues/24
- #1337

larsyencken:
@edomt Thoughts on this?

edomt:
This was generally implemented in Tenerife already, but (following @maxroser's advice) I only deleted unused datasets up to the YC period (2019).

@HannahRitchie @maxroser Any one thoughts on whether we should extend this to more recent datasets?

HannahRitchie:
Yes, we should do a search of _all_ datasets not being used and filter them all.

Happy to help with filtering because I probably have a lot of legacy knowledge of any we shouldn't delete.

danielgavrilov:
I think anything under the `owid` namespace could be deleted like this, but potentially we should be more conservative with bulk imported datasets? Especially if they are the "current" version.

larsyencken:
A draft SQL query that finds variables in BigQuery finds quite a lot:

```sql
with var_used_in_charts as (
    select chart_id, CAST(variable_id AS INT64) variable_id
    from `owid-analytics.prod_owid.charts`
    cross join unnest(variable_ids) as variable_id
),

dataset_used_in_charts AS (
    SELECT
        d.id dataset_id,
        any_value(d.name) dataset_name,
        any_value(d.createdAt) dataset_created_at,
        count(distinct uc.chart_id) n_charts
    FROM `prod_mysql.variables` v
    inner join prod_mysql.datasets d on (v.datasetId = d.id)
    left join var_used_in_charts uc on (v.id = variable_id)
    group by dataset_id
),

unused_datasets as (
    select * 
    from dataset_used_in_charts
    where n_charts = 0
)

select
    dataset_id,
    dataset_name,
[...]

PR #1463, opened 2022-05-24
chore(migration): remove unused datasets

Marigold:
Migration for ~~deleting~~ archiving unused datasets from https://github.com/owid/owid-grapher/issues/1338. The original query and datasets themselves are in [this spreadsheet](https://docs.google.com/spreadsheets/d/10pJoPIIFKeEKmvmk2rQ6Uwl6pUTIFoL9_stSy1Weo2M/edit?usp=sharing).

This migration just adds additional column `isArchived` to `datasets` table and labels some datasets as archived. Archived datasets will be excluded from backporting and hidden from catalog. Any idea if adding extra column could cause problems anywhere?

Next steps as per [@larsyencken comment](https://github.com/owid/owid-grapher/issues/1338#issuecomment-1137498976)

> But an archived dataset should not show up in search, and should not be possible to publish a grapher chart
from. I.e. it should be totally hidden, and only restorable with some
manual SQL.

I might need some help with that ;).


db/migration/1653646292637-ArchiveDatasets.ts
```diff
@@ -0,0 +1,190 @@
...
+    revert: boolean = false
+) => {
+    // label datasets as archived in DB
+    const placeholders = datasetIds.map(() => "?")
+
+    await queryRunner.query(
+        `UPDATE datasets SET isArchived = ${revert ? 0 : 1}
+        WHERE id in (${placeholders})`,
+        datasetIds
+    )
```
danielgavrilov, line 145: nit: I think there is no need for placeholders, you can pass an array as a parameter, like here: https://github.com/owid/owid-grapher/blob/9a94b86554e6782c94e624352553b5e29008ebfb/baker/siteRenderers.tsx#L557-L560

I think this should work:

```suggestion
    const placeholders = datasetIds.map(() => "?")

    await queryRunner.query(
        `UPDATE datasets SET isArchived = ${revert ? 0 : 1}
        WHERE id in (?)`,
        [datasetIds]
    )
```

danielgavrilov (approved):

danielgavrilov:
I am a bit out of the loop: Are we not deleting them because authors might need them, or because of something else? Just wondering if we need to allow authors to unarchive a dataset through the Admin UI.

larsyencken:
> I am a bit out of the loop: Are we not deleting them because authors might need them, or because of something else? Just wondering if we need to allow authors to unarchive a dataset through the Admin UI.

Well, we want to be a bit aggressive here, but the thinking was to give ourselves an out by "soft deleting" them instead, in case we find there was a mistake later. We don't want to allow unarchiving by authors, since that would introduce a gap in data coverage between the ETL and grapher.

PR #1468, opened 2022-05-26
feat(details on demand): add details table to grapher DB

### LINKED ISSUE #1467: Add a `details` table to the Grapher DB
The table should be as follows, with a unique constraint on category + term

name | type
-- | --
category | varchar
term | varchar
title | varchar
content | varchar

ikesau:
Resolves #1467

~Realized there was no point in having an `AUTO_INCREMENT` id if we can always just use the category and term to index.~

Update: added an `AUTO_INCREMENT` id 😊

### DIFF
--- a/db/migration/1653584089031-DetailsOnDemand.ts
+++ b/db/migration/1653584089031-DetailsOnDemand.ts
@@ -1,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class DetailsOnDemand1653584089031 implements MigrationInterface {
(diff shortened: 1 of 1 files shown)


danyx23:
I think we should have an auto-incrementing id. Yes technically the category and term can be used as primary key but I think having an integer primary key is a lot safer in the long term. A unique constraint on category + term might still be a good idea though

danielgavrilov (approved):
Looks good!

(I think the auto-incrementing ID is helpful for foreign keys, which I think we might eventually be using to replicate the content graph in MySQL.)

PR #1490, opened 2022-06-09
Migration to add sourceChecksum column for datasets

wcox13:
Adds a yet-unused `sourceChecksum` column to `datasets` table. See https://github.com/owid/etl/pull/208 for the logic to actually write this column.


wcox13:
I'm not sure how to fix this `prettier` check. When I run 
```yarn prettier db/migration/1654553380446-AddSourceChecksum.ts```
It completes successfully, but doesn't change the file at all. Then when I run 
```yarn prettier --check db/migration/1654553380446-AddSourceChecksum.ts```
It still fails

Marigold (approved):
LGTM!

marcelgerber (approved):
I just pushed a commit to replace tabs with spaces. Otherwise this all looks good!

wcox13:
👍 Thanks for the review guys!

PR #1491, opened 2022-06-09
Details on demand editor implementation

### LINKED ISSUE #1482: Include DoD data in bake

ikesau:
Resolves #1482 

Adds support for the adding Details to our grapher configs.

How it works:
1. fetch all the Details from the DB
2. Watch the `subtitle` and `note` fields for changes, specifically for the inclusion of any text that matches the `hover::category::term` syntax
3. If there's a match, check to see if it corresponds with a Detail defined in the DB
4. Pass all corresponding Details through to Grapher as an object with the shape `Record<string, Record<string, Detail>>`
5. Track the Detail references that _don't_ match, and warn the user

https://user-images.githubusercontent.com/11844404/172952619-a3c59e2e-2fdf-4fc7-9701-2214d72d1b42.mov

### DIFF
--- a/adminSiteClient/ChartEditorPage.tsx
+++ b/adminSiteClient/ChartEditorPage.tsx
@@ -272,28 +272,32 @@ export class ChartEditorPage
         this.fetchTopics()
     }
 
-    dispose!: IReactionDisposer
-    disposeDetailHandler!: IReactionDisposer
+    disposers: IReactionDisposer[] = []
+
     componentDidMount(): void {
         this.refresh()
 
-        this.dispose = reaction(
-            () => this.editor && this.editor.previewMode,
-            () => {
-                if (this.editor) {
-                    localStorage.setItem(
-                        "editorPreviewMode",
-                        this.editor.previewMode
-                    )
+        this.disposers.push(
+            reaction(
+                () => this.editor && this.editor.previewMode,
+                () => {
+                    if (this.editor) {
+                        localStorage.setItem(
+                            "editorPreviewMode",
+                            this.editor.previewMode
+                        )
+                    }
                 }
-            }
+            )
         )
 
-        this.disposeDetailHandler = reaction(
-            () => this.currentlyReferencedDetails,
-            (currentlyReferencedDetails = {}) => {
-                this.grapher.details = currentlyReferencedDetails
-            }
+        this.disposers.push(
+            reaction(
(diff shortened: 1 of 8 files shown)


adminSiteClient/ChartEditorPage.tsx
```diff
@@ -80,6 +82,15 @@ class TabBinder extends React.Component<{ editor: ChartEditor }> {
     }
 }
 
+function extractDetailsFromSyntax(str: string): [string, string][] {
+    const pattern = /\(hover::(\w+)::(\w+)\)/g
```
danyx23, line 86: I think we should elevate this regex and potential siblings to a shared place where it can be used by different pieces (e.g. in the admin we will want to check that the individual terms adhere to the same requirements - it would be bad if you can add numbers in terms in the admin but then when extracting them we don't find them because the regex excludes numbers).

This will come in handy I think when we add (include:: ...) as an alternative to hover::

I think we also want to be more permissive than \w, at least for the term. `int-$` would currently not work but might be more along what authors want to write than `international_dollars`. We should exclude : and () so that we don't get into parsing issues but I think adding a few more characters would be nice. In theory we could even allow whitespace but maybe that leads to more subtle issues than what it is worth. So maybe for now something like [A-Za-z0-9_\-$%\.+]? I'm not particularly fixed on any particular character and I think it makes sense to be more strict now and then extend but stuff like the $ sign is a pretty obvious candidate to be used IMHO.

danyx23 (changes requested):
Nice work! I have one request to change the regex - once this is done it can be merged (I'll be away now so don't wait for a second ok from me :) )

adminSiteClient/ChartEditorPage.tsx
```diff
@@ -206,6 +273,7 @@ export class ChartEditorPage
     }
 
     dispose!: IReactionDisposer
+    disposeDetailHandler!: IReactionDisposer
```
danielgavrilov, line 276: One mobx pain is to have to write these disposers... a slight improvement might be to add them to an array, we use this pattern in other places: https://github.com/owid/owid-grapher/blob/f7cd315cc8ac99c2c3dbafae6d2450f9e673dbb6/grapher/core/Grapher.tsx#L917-L948

danielgavrilov (approved):
All good! I just have the same comment as B: to make the regex more permissive and to make it usable in other places like the admin UI.

PR #1495, opened 2022-06-13
Rebake graphers when updating details on demand

### LINKED ISSUE #1494: Rebake affected Graphers when a DoD is updated

ikesau:
Resolves #1494 

Flow:
1. User updates a detail
2. Update that detail in the DB
3. Find all Graphers that reference that detail
4. Update their syntax to use the new one (e.g. `hover::animals::dogs` -> `hover::mammals::dogs`)
5. Update the details data structure in each Grapher's config (e.g. `{ animals: dogs: { } }` -> `{ mammals: dogs: { } }`)
6. Save each Grapher with the `saveGrapher` function, which triggers rebuilds if the server is configured to do so.


clientUtils/owidTypes.ts
```diff
@@ -265,6 +265,14 @@ export interface FullPost {
     glossary: boolean
 }
 
+export interface Detail {
```
ikesau, line 268: This interface has been defined in other branches already. Once everything's merged I'll prune down the duplication.

ikesau:
Hey @danyx23

I’m not sure how a suggested chart revision could ever lead to invalid details,  because any time a DoD is updated, its associated charts also get updated which would invalidate the revision.

As for the variable table grapherConfig check, I’m not sure I know what exact structure that information would take, it seems like it’d be easier to implement the check once we’ve implemented that feature?

ikesau:
Though I suppose we could add validation on the backend when updating a grapher to make sure its DoD references are still good. Right now it would be possible to use the bulk editor to add a subtitle like `[this](hover::fake::fake)`. 🤔

danielgavrilov:
I am wondering if it might be better to inject the `details` into the config at the bake step, in order to avoid any of the staleness issues.

An unintended overwrite I can think of:
- Author has two tabs open: chart and details
- Author updates details (triggering an update to the chart)
- Author switches to chart tab and saves the chart, overwriting the updated details

I think this would lead to a stale DoD?

What I'm not sure is how fiddly injecting it at the bake step would be...

danielgavrilov:
Ah one more: you can revert to previous chart config from `chart_revisions`.

PR #1503, opened 2022-06-17
Details on demand desktop grapher implementation

### LINKED ISSUE #1497: Desktop tooltip component for DoD

ikesau:
Resolves #1497 

https://user-images.githubusercontent.com/11844404/174406038-93bfdfaf-2dca-43e9-b8a7-ed368444b61a.mov


danielgavrilov (changes requested):
I mentioned in a comment – can check if `tippy` is a good fit for the tooltip.

I'm also hesitant about parsing DoD in the `TextWrap` component like this, feels like the parsing logic should maybe be outside the text-wrapping procedure? Are there other options you considered? I'm thinking about how we might extend this to support bold, italic in subtitles eventually. 

Could we maybe pass tokenised input to `TextWrap`? Or even `TextWrap` itself could be implemented in two steps – parse markdown to produce tokens, then try to wrap those tokens. 

Also related: do Markdown parsing libraries handle the `[](hover::...)` syntax just fine? We could use them and avoid parsing DoD syntax ourselves?

What I'm suggesting might be tricky and take a lot of time, so going ahead with the existing implementation is definitely an option.

clientUtils/Util.ts
```diff
@@ -848,6 +848,16 @@ export function scrollIntoViewIfNeeded(
     }
 }
 
+export function checkIsChildOfClass(
```
ikesau, line 851: This can be removed

db/migration/1653584089031-DetailsOnDemand.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class DetailsOnDemand1653584089031 implements MigrationInterface {
```
ikesau, line 3: Not sure why this is in here. I think git's confused.

grapher/detailsOnDemand/detailsOnDemand.tsx
```diff
@@ -0,0 +1,86 @@
...
+                return acc
+            },
+            new ObservableMap()
+        )
+        this.details = this.details.merge(observableDetails)
+    }
+}
+
+// An object that can be shared throughout the codebase to make details globally accessible
+export const globalDetailsOnDemand = new DetailsOnDemand()
```
ikesau, line 35: Not sure about this pattern, was just trying to set something up for if/when we start having DoDs in articles.

It also makes it much easier for the MarkdownTextWrap to hook into

grapher/downloadTab/DownloadTab.tsx
```diff
@@ -283,6 +301,30 @@ export class DownloadTab extends React.Component<DownloadTabProps> {
...
                         </button>
                     </div>
+                    {!isEmpty(this.manager.detailRenderers) && (
+                        <div className="static-exports-options">
+                            <input
+                                type="checkbox"
+                                id="shouldIncludeDetailsInStaticExport"
+                                name="shouldIncludeDetailsInStaticExport"
+                                onChange={(): void => {
+                                    this.isReady = false
```
ikesau, line 311: this line makes the grapher loading spinner re-appear for 500ms while the images re-render. 

If we don't have it, then instead you see the static export thumbnails flash and change their aspect ratio. I'm not sure which is better.

grapher/text/MarkdownTextWrap.stories.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { createRef } from "react"
```
ikesau, line 1: and more. can be ignored

PR #1512, opened 2022-06-24
Markdown parser

ikesau:
Bringing in [Parsimmon](https://github.com/jneen/parsimmon) to help us make a markdown AST that we can easily render text in Grapher with.

The general idea is that you use `mdParser.value.parse` with some markdown input and get an AST back along these lines:

```
{
// root
children: [
  {
    type: 'bold',
    children: [
      {
        type: "text",
        value "hello",
      },
      { type: "whitespace" },
      {
        type: "text",
        value "world",
      },
    ]
  },
]
}
```

If you want a scratchpad to explore Parsimmon with, I made one [here](https://observablehq.com/@ikesau/parsimmon-markdown-parser).

Note that this isn't a complete implementation. Many (most) markdown features currently aren't supported.

Nor is it even a partial [MDAST](https://github.com/syntax-tree/mdast) implementation, as this parser tokenizes based on words, and keeps track of newlines.

I haven't yet done anything around error-handling, it'll be easier to know what pattern to use once I actually implement this in Grapher.


danielgavrilov:
Is it the case that whitespace is implicit, and there are no whitespace tokens? I think it's better to have them explicit, otherwise you can't have a token without triggering whitespace, e.g. **bolded**-text would become **bolded** -text. Right?

ikesau:
T'was a very good point @danielgavrilov - thanks for your help on this! I've updated the parser to use `many` and it now treats whitespace as a token (squashing multiple spaces into one).

grapher/text/parser.ts
```diff
@@ -0,0 +1,115 @@
...
+// Deviates because we want to track individual words, whitespace, and newlines to use with TextWrap and our SVG exporter
+
+// The root literal that should be the leaf of every branch that isn't a newline or whitespace
+interface Text {
+    type: "text"
+    value: string
+}
+
+export const textParser = (): P.Parser<Text> =>
+    P.regex(/[A-Za-z'\-":!?,.]+/).map((value) => ({ type: "text", value }))
```
danielgavrilov, line 13: Below are all the characters used in subtitles today, and we probably need to support all. Might be better to create a regex that consumes everything apart from some subset of characters?

```
 
!
"
#
$
%
&
'
(
)
+
,
-
.
/
0
1
2
3
4
5
6
7
8
9
:
;
<
=
>
?
A
B
C
D
E
F
G
H
I
J
K
L
M
N
O
P
Q
R
S
T
U
V
W
X
Y
Z
[
\
]
a
b
c
d
e
f
g
h
i
j
k
l
m
n
o
p
q
r
s
t
u
v
w
x
y
z
 
£
­
°
±
²
³
µ
¹
é
ü
 
‑
–
—
‘
’
“
”
•
⁵
₀
₁
₂
₃
₄
₅
−
≥
ﬁ
```

danielgavrilov: This seems like a very annoying problem to solve, since it leads to ambiguity in the parser?

ikesau: Oh boy.

It may be possible to get around this with something else from the toolbox. Some sort of lookahead that chains accordingly. (I'm more on the "just saying words" side of the solution at this point 😅)

Worst case, we can add a more inclusive character regex that still omits characters that will break the parser (`[]*_` etc), and implement error handling that returns the plain string for those errors (because it's safe to assume that any graphers that currently have brackets in them won't have markdown syntax or DoDs currently.)

PR #1527, opened 2022-06-30
enhance(map): automatic legend width

danielgavrilov:
I implemented [automatic legend sizing](https://github.com/owid/owid-grapher/commit/6129ac3c26d8a0c6d716f1acbd0ed678daf1d747) several months ago, but didn't enable it on maps to avoid a large svgTester diff.

By enabling it, denser legends are now wider, and sparse legends narrower. Relatedly, [Max recently suggested legends on maps should be bigger](https://owid.slack.com/archives/C46U9LXRR/p1656510679657209), so this seems like a good time to make the change.

#### Before / after:

Screenshot 2022-07-01 at 00 51 42

Screenshot 2022-07-01 at 00 39 48

Screenshot 2022-07-01 at 00 42 28

### DIFF
--- a/grapher/horizontalColorLegend/HorizontalColorLegends.tsx
+++ b/grapher/horizontalColorLegend/HorizontalColorLegends.tsx
@@ -236,7 +236,8 @@ export class HorizontalNumericColorLegend extends HorizontalColorLegend {
 
     private getNumericLabelMinWidth(bin: NumericBin): number {
         if (bin.text) {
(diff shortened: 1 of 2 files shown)


danielgavrilov:
In some cases it leads to text overflow, but mostly it's an improvement.

Screenshot 2022-07-01 at 01 07 13

If we want to deal with the overflow issue, we have to make the legends a bit smarter to take into account the widths of the first/last labels.

marcelgerber (approved):

PR #1531, opened 2022-07-05
enhance(axis): do not show solid 0 line for date columns

### LINKED ISSUE #1172: ScatterPlot with a time X dimension should not show vertical solid line for 21 Jan 2020
## Description

ScatterPlots with a date column in X have a solid vertical line on 21 Jan 2020:

([link to chart](https://ourworldindata.org/grapher/ai-training-computation))

image (3)

## Context

If you create a ScatterPlot with the X column blank in Admin, it uses the time column by default (experimentally implemented in #1065).

In ScatterPlots, we show the Y=0 and X=0 as solid lines. For `Year` columns this seems sort of ok, but for `Day` columns this doesn't work well, because no matter what the variable epoch date is, all variables get transformed relative to our default epoch which is 21 Jan 2020:

https://github.com/owid/owid-grapher/blob/bfc4280b7c0ab6ed087ea27fa5121185d34ff637/clientUtils/owidTypes.ts#L2

(There is nothing special about this date, except that the WHO COVID data used to start on that date, and we had our first charts with `0` day being that date.)

## Expected behaviour

No solid vertical lines at all when a Date column is used as an X dimension on ScatterPlot.

## Additional context

Initially [reported by Charlie](https://owid.slack.com/archives/C0146HTMV2R/p1644935513096149?thread_ts=1644931416.138679&cid=C0146HTMV2R) and later again by [Max](https://owid.slack.com/archives/C46U9LXRR/p1644959735012219), for the [same chart](https://ourworldindata.org/grapher/ai-training-computation).

danielgavrilov:
Fixes #1172.

No longer shows solid line for value=0 when the axis `formatColumn` is a date column.

Screenshot 2022-07-05 at 11 40 49

### DIFF
--- a/grapher/axis/Axis.ts
+++ b/grapher/axis/Axis.ts
@@ -12,7 +12,7 @@ import {
 import { Bounds, DEFAULT_BOUNDS } from "../../clientUtils/Bounds.js"
 import { TextWrap } from "../text/TextWrap.js"
 import { AxisConfig } from "./AxisConfig.js"
-import { CoreColumn } from "../../coreTable/CoreTableColumns.js"
+import { ColumnTypeMap, CoreColumn } from "../../coreTable/CoreTableColumns.js"
 import { ValueRange } from "../../coreTable/CoreTableConstants.js"
 import {
     AxisAlign,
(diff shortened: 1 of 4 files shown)


marcelgerber (approved):
Looks good!

PR #1535, opened 2022-07-05
Parsimmon-based markdown parser and AST renderer

ikesau:
Incorporates the contributions made by @danielgavrilov and @danyx23 to the Parsimmon parser.


grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,527 @@
...
+                {lines.map((line, i) => (
+                    <div key={i}>
+                        {line.length ? (
+                            line.map((token, i) => token.toHTML(i))
+                        ) : (
+                            <br />
+                        )}
+                    </div>
+                ))}
+            </span>
```
danielgavrilov, line 524: In `TextWrap`, we use`<span>`: https://github.com/owid/owid-grapher/blob/31fc73b21799963e97f908b72b601a85f39979aa/grapher/text/TextWrap.tsx#L166-L190

I'm not sure if it makes a difference, but if we match the output identically, we can check the diffs in `owid-static` for any regressions.

I made them `<div>`s for no good reason, it was just an easy way to implement lines.

danielgavrilov: On second thought maybe not: no charts are rendered as HTML in the baked files, but might still be a good idea to match output as closely as possible – up to you!

ikesau: Yup I'll get these details right once I've got the high level work in the right place, this was just a temporary commit to show off the overall strategy 👍

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,527 @@
...
+                <text
+                    fontSize={fontSize.toFixed(2)}
+                    fontWeight={fontWeight}
+                    x={props.x.toFixed(1)}
+                    y={yOffset.toFixed(1)}
+                    {...props.svgOptions}
+                >
+                    {lines.map((line) =>
+                        line.map((token, i) => token.toSVG(i))
+                    )}
```
danielgavrilov, line 509: We have to wrap each line in a `<tspan>` like `TextWrap` does it. 

We could either do that here directly, or we could also create an `IRLine` token that wraps in `<tspan>`, but it would need to receive the offsets as well... probably option 1 is better.

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    // @computed get height(): number {
+    //     if (this.lines.length === 0) return 0
+
+    //     return (
+    //         this.lines.reduce((total, line) => total + line.height, 0) +
+    //         this.lineHeight * (this.lines.length - 1)
+    //     )
+    // }
```
danielgavrilov, line 475: I think the `height()` method on the original class might be incorrect/non-sensical: https://github.com/owid/owid-grapher/blob/31fc73b21799963e97f908b72b601a85f39979aa/grapher/text/TextWrap.tsx#L132-L139

The `lineHeight` is just used as padding between lines? Or am I missing something? 

Everywhere else `lineHeight` is interpreted as a factor to multiply fontSize by.

ikesau: Yeah seems like a mis-implementation of the line height formula (`font-size * line-height`)

We only need it for SVG so I've just put it in-line.

ikesau:
I had to revert the "React-first" changes I made once I realized that the height calculations were important for SVG rendering 😞 

I've partially kept the JSX changes though so that we can use the MarkdownTextWrap in JSX if we want.

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,549 @@
...
+    getClone(children: IRToken[]): IRBold {
+        return new IRBold(children, this.fontParams)
+    }
+    toHTML(key: number | string): JSX.Element {
+        return (
+            <strong key={key}>
+                {this.children.map((child, i) => child.toHTML(i))}
+            </strong>
+        )
+    }
```
danielgavrilov, line 161: nit: could we make `key` optional? (also the `React.Key` type might be slightly better to use, although it just maps to `string | number` so technically the same)

---

Also we could maybe drop the `key` param by wrapping the output of `toHTML` and `toSVG` in `<React.Fragment>`. E.g. there could be a small utility function:

```js
function renderChildren(children, renderFunc) {
    return children.map((child, i) => (
        <React.Fragment key={i}>
            {renderFunc(child)}
        </React.Fragment>
    ))
}
```

and then this becomes

```suggestion
    toHTML(key: number | string): JSX.Element {
        return (
            <strong>
                {renderChildren(children, (child) => child.toHTML())}
            </strong>
        )
    }
```

but not clear that it's better, just makes calling `toHTML()` cleaner.

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,549 @@
...
+    maxWidth?: number
+    style?: CSSProperties
+}
+
+export class MarkdownTextWrap extends React.Component {
+    props: MarkdownTextWrapProps
+    constructor(props: MarkdownTextWrapProps) {
+        super(props)
+        this.props = props
+    }
```
danielgavrilov, line 442: nit: I think this also works, don't need a `constructor`?

```suggestion
export class MarkdownTextWrap extends React.Component<MarkdownTextWrapProps> {
```

danielgavrilov (changes requested):
Reviewed everything except `parser.ts` and `parser.test.ts` – in case there are changes to those, might be better for @danyx23 to review since he has more context.

Otherwise I have some small suggestions, and a bugfix: https://github.com/owid/owid-grapher/pull/1541

Feel free to ignore suggestions you think are not necessary in v1, in order to get this shipped and have a break before next cycle.

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,549 @@
...
+export class IRBold extends IRElement {
+    @imemo get width(): number {
+        const plaintext = this.children
+            .map((token) => token.toPlaintext())
+            .join("")
+        return Bounds.forText(plaintext, {
+            ...this.fontParams,
+            fontWeight: 700,
+        }).width
+    }
```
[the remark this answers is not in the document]
danielgavrilov: Some context:

I think the weird thing with my implementation is that the correct `FontParams` need to be passed into `IRBold` when instantiating it, it doesn't by itself propagate the bold `FontParams`. 

This is because I decided that the element itself (whether text, bold or whatever) should have access to its own width. If we assume that, then the information for its context (font size, weight, etc) needs to be available for the element, and you can't simply instantiate it independently of its context.

So there is the:
- Rendering logic: just wrap the output in `<strong>` and the DOM will take care of propagating the styles to children. This is all that `IRBold` does by itself.
- Text width logic: `IRBold` doesn't propagate this by itself, we need to make sure we propagate `FontParams` down to the `IRText`, as done here: https://github.com/owid/owid-grapher/blob/facc9e60e794405563c7e7308609d4844f21ec14/grapher/text/MarkdownTextWrap.tsx#L391-L396

This is confusing, because the same information lives in two places basically – both outside and inside `IRBold`. I wonder if there is some way to implement it that makes it simpler.

grapher/text/MarkdownTextWrap.tsx
ikesau: Thanks for the review @danielgavrilov !

Changes:
- `key` is optional now, and typed with `React.Key`
  - I didn't implement `renderChildren`, as it seems like a bit of a leaky abstraction. These classes are probably going to be called via `MarkdownTextWrap` 99% of the time so I don't think it's a huge concern?
- Fixed italic and bold width calculation, including propagating `isItalic` to `pixelWidth`
- Removed `style` - we can wait till we need it
- `whitespace: pre-wrap` fixes the collapsing whitespace problem
  - I stuck with the block span rendering method as it makes calculating line heights easier (text content and br elements behave weirdly with line height when inline, whereas a block span is always just `line-height * font-size`)

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,510 @@
...
+    return lines
+}
+
+export function splitLineAtBreakpoint(
+    tokens: IRToken[],
+    breakWidth: number
+): { before: IRToken[]; after: IRToken[] } {
+    let i = 0
+    let offset = 0
+    // finding the token where the split should be
```
danyx23, line 255: Should be what?

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -0,0 +1,510 @@
...
+            )
+        } else if (node.type === "detailOnDemand") {
+            // TODO: create a IR for this
+            return new IRLink(
+                node.term,
+                parsimmonToTextTokens(node.children, fontParams)
+            )
+        } else {
+            throw new Error(`Unknown node type: ${(node as any).type}`)
+        }
```
danyx23, line 401: Fine like this but just as an FYI - the ts-pattern library that does pattern matching with exhaustiveness checking can be pretty neat for such cases (would need a union type that merges all possible node types but that might be a good idea anyhow). Nothing to here, just wanted to mention this.

danyx23 (commented):
I think this looks good! I couldn't really test the text wrap implementation as is without the rendering/admin features so I'll do that in another PR. The code for the text wrap looks fine but I only read it, so I'll trust you and Daniel G that it makes sense for now :)

grapher/text/MarkdownTextWrap.tsx
danielgavrilov: made

danielgavrilov:
@danyx23 To test the text wrapping, you can use the storybook:

```sh
yarn startStorybookServer
```

And then navigate to MarkdownTextWrap. This is the better way to test it, now that Ike added both SVG and HTML rendering there.

danielgavrilov:
@ikesau Everything looks (& sounds) good to me!

Just one more change needed that I mentioned on Slack: **some existing subtitles are not parsing**, because they contain special characters: [U+2008](https://www.compart.com/en/unicode/U+2008) (punctuation space) and [U+00A0](https://www.compart.com/en/unicode/U+00A0) (non-breakable space).

Can see the failing subtitles in the Observable notebook: https://observablehq.com/@danielgavrilov/parsimmon-owid-markdown

PR #1553, opened 2022-07-18
Details on Demand: TextWrap output comparison

danielgavrilov:
Adds a storybook to compare the outputs of our existing `TextWrap` and our new `MarkdownTextWrap`:
Screenshot 2022-07-18 at 10 42 09

### DIFF
--- a/grapher/text/parser.test.ts
+++ b/grapher/text/parser.test.ts
@@ -801,4 +801,69 @@ how **are** you?`)
             },
         })
     })
+    it("Parses whitespace preceding a newline", () => {
+        const input =
+            "this-line-ends-with-a-space" +
+            " " +
+            "\n" +
+            "but-the-newline-should-be-tracked-separately"
+        expect(mdParser.markdown.parse(input)).toEqual({
+            status: true,
+            value: {
+                type: "MarkdownRoot",
+                children: [
+                    {
+                        type: "text",
+                        value: "this-line-ends-with-a-space",
+                    },
+                    {
+                        type: "whitespace",
+                    },
+                    {
+                        type: "newline",
+                    },
+                    {
+                        type: "text",
+                        value: "but-the-newline-should-be-tracked-separately",
+                    },
+                ],
+            },
+        })
+    })
+    it("Parses newlines surrounded by whitespace", () => {
+        const input =
+            "this-line-ends-with-a-space" +
+            " " +
+            "\n\n" +
+            " " +
+            "but-the-newline-should-be-tracked-separately"
+        expect(mdParser.markdown.parse(input)).toEqual({
+            status: true,
+            value: {
+                type: "MarkdownRoot",
+                children: [
+                    {
+                        type: "text",
+                        value: "this-line-ends-with-a-space",
+                    },
+                    {
+                        type: "whitespace",
+                    },
+                    {
+                        type: "newline",
+                    },
+                    {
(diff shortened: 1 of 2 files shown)


danielgavrilov:
Added a fix:
Screenshot 2022-07-18 at 12 05 33

You can see the red dashed boxes are a bit shorter than they should be, this is because `TextWrap` incorrectly calculates height. I created a PR to fix that: https://github.com/owid/owid-grapher/pull/1554

danielgavrilov:
The svgTester will find differences, but I have been testing the PR locally, and it seems to match the `TextWrap` positioning exactly now.

There is the difference with how we specify the `font-size` that leads to differences in the SVGs (but no differences in the image).

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -556,13 +555,23 @@ export class MarkdownTextWrap extends React.Component<MarkdownTextWrapProps> {
...
     ): JSX.Element | null {
-        const { lines } = this
+        const { lines, fontSize, lineHeight } = this
         if (lines.length === 0) return null
 
+        // Magic number set through experimentation.
+        // The HTML and SVG renderers need to position lines identically.
+        // This number was tweaked until the overlaid HTML and SVG outputs
+        // overlap.
+        const HEIGHT_CORRECTION_FACTOR = 0.74
```
danielgavrilov, line 565: Copied this from `TextWrap`:

https://github.com/owid/owid-grapher/blob/fba2aa94ecd7d2d06bd9a0161bc6ea410ef68326/grapher/text/TextWrap.tsx#L201-L205

Maybe the magic number should be shared... (in order to do that, the base branch (`details-on-demand`) needs to get the latest file from master)

danielgavrilov:
With some local adjustments to avoid non-meaningful diffs, I ran the svgTester manually: https://github.com/owid/owid-grapher-svgs/commit/1572ceef1dc99264d4bd315ad60958334d00c2c0

danielgavrilov:
I ran the **svgTester manually: https://github.com/owid/owid-grapher-svgs/commit/292a062902dbb7830fd116d60a441e9235245d2d**

It **omits** some inconsequential differences:
- The old `TextWrap` for some reason doesn't fill the available space, it wraps to a new line earlier than `MarkdownTextWrap`. I didn't try to figure out why and don't think it's important.

Some inconsequential differences are **still included**:
- Multiple consecutive spaces get detected as a difference, even though it leads to zero changes in the image (svg ignores whitespace, like HTML)

The only other differences I found:
- ✅ URLs are automatically wrapped in `<a>`
- ⚠️ The automatic links don't handle dashes in hashes ([chart](https://owid.cloud/admin/charts/2703/edit)): 
   Screenshot_2022-07-18_at_21_24_21

grapher/text/MarkdownTextWrap.tsx
```diff
@@ -571,8 +580,8 @@ export class MarkdownTextWrap extends React.Component<MarkdownTextWrapProps> {
                         key={i}
                         x={x}
                         y={(
-                            this.lineHeight * this.props.fontSize * (i + 1) +
-                            y
+                            yOffset +
+                            lineHeight * this.props.fontSize * i
```
danielgavrilov, line 584: hm, why `this.props.fontSize` instead of `fontSize`?...

PR #1552, opened 2022-07-18
Details on Demand

ikesau:
🎉  It's finally here!

- [x] New DB Table
- [x] Admin API
- [x] Admin Detail editor UI
- [x] Admin Chart editor UI
- [x] MarkdownTextWrap integration
- [x] Grapher implementation 
- [x] Static exports
- [x] Keyboard navigation

Something very strange happened with the feature branch amidst all this where several commits were dropped and had to be re-added on top of the branch. I'll research a better rebase workflow to ensure that doesn't happen again.

I'll start the final review on the 18th to squeeze in any last minute fixes and polish deemed necessary.

https://user-images.githubusercontent.com/11844404/179438661-f6901de4-d359-4a46-beb4-ec6e5e609825.mov

https://user-images.githubusercontent.com/11844404/179438754-afd6d66a-fa54-4d44-8b4b-fb4655defae1.mov

PR #1554, opened 2022-07-18
fix(TextWrap): match SVG and HTML rendering

danielgavrilov:
In Details on Demand (#1552) we noticed that the height `TextWrap` returns is not correct – `lineHeight` is not multiplied by the `fontSize`.

### DIFF
--- a/grapher/captionedChart/CaptionedChart.tsx
+++ b/grapher/captionedChart/CaptionedChart.tsx
@@ -79,7 +79,7 @@ interface CaptionedChartProps {
 }
 
 const OUTSIDE_PADDING = 15
-const PADDING_BELOW_HEADER = 18
+const PADDING_BELOW_HEADER = 16
 const CONTROLS_ROW_HEIGHT = 36
 const PADDING_ABOVE_FOOTER = 25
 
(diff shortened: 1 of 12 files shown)


grapher/captionedChart/CaptionedChart.tsx
```diff
@@ -79,7 +79,7 @@ interface CaptionedChartProps {
 }
 
 const OUTSIDE_PADDING = 15
-const PADDING_BELOW_HEADER = 18
+const PADDING_BELOW_HEADER = 16
```
danielgavrilov, line 82: Decreased the padding to compensate a bit for the subtitle `height` being inaccurate – it used to not take into account `lineHeight` correctly, it was an underestimate.

ikesau (approved):
Checked it out, definitely within acceptable tolerances. 🚀

danielgavrilov:
Screenshot 2022-07-18 at 17 45 35

I just did the calculation on paper and the spacing between the title & subtitle is correctly calculated, but when I take screenshots of HTML & SVG and try to compare in Figma, I always get a pixel difference (less spacing on HTML).

danielgavrilov:
Fixes a longstanding issue with footer text overflowing:

Screenshot 2022-07-18 at 18 11 55

PR #1559, opened 2022-07-21
Perf: adds a new implementation for full joining multiple tables.

danyx23:
This should allevieate performance issues with many variables as recently reported. In one test the time to do the full join went down from 4 seconds to 600 ms and the memory use from 250MB to 85MB


danyx23:
Superseded by #1565

PR #1560, opened 2022-07-21
DoD export style touchups

ikesau:
Implements Marwa's design feedback for static exports.

Business logic outline:
1. Scan through the subtitle and note text, find all DoD references and create a unique ordered list (because in the case where the same DoD is referenced twice, we don't want to increment the citation number)
2. Traverse the Markdown AST and whenever we find a DoD, append a IRSuperscript token into its children that uses the corresponding index from the unique ordered list

Rendering superscript in SVG has some quirks but it seems to work fairly well.

TODO:
- [x] link restyling
- [x] fix right-side text clipping

renewable-share-energy(2)

### DIFF
--- a/grapher/detailsOnDemand/detailsOnDemand.tsx
+++ b/grapher/detailsOnDemand/detailsOnDemand.tsx
@@ -4,7 +4,6 @@ import { MarkdownTextWrap } from "../text/MarkdownTextWrap.js"
 import { computed, observable, ObservableMap } from "mobx"
 import { observer } from "mobx-react"
 import { Detail } from "../core/GrapherConstants.js"
-import { throttle } from "lodash"
 
 class DetailsOnDemand {
     @observable details = new ObservableMap<
@@ -42,15 +41,9 @@ interface DoDWrapperProps {
 }
 
 @observer
-export class DoDWrapper extends React.Component<
-    DoDWrapperProps,
-    { innerHeight: number }
-> {
+export class DoDWrapper extends React.Component<DoDWrapperProps> {
     constructor(props: DoDWrapperProps) {
         super(props)
-        this.state = {
-            innerHeight: window.innerHeight,
-        }
     }
     @computed get category() {
         return this.props.category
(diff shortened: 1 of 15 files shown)


grapher/detailsOnDemand/detailsOnDemand.tsx
```diff
@@ -63,25 +91,35 @@ export class DoDWrapper extends React.Component<{
...
-                <span
-                    aria-label={`A definition of the term ${this.title}`}
-                    tabIndex={0}
-                    className="dod-term"
+            <span>
+                <Tippy
+                    // Tippy gets it wrong when viewport height is very low
+                    // because it doesn't realize that the .GrapherComponent
+                    // container is overflow: hidden
+                    placement={this.state.innerHeight < 300 ? "right" : "auto"}
```
danielgavrilov, line 99: Isn't it enough to just make it `placement="auto"` always, and not worry about resize? I think the problem was that by default it's `placement="top"`.

Tippy does in fact know that `.GrapherComponent` is `overflow: hidden`, it has different behaviour when it's `visible`.

danielgavrilov (changes requested):
😥 I didn't realise numbering would add this much complexity.

Hopefully these are the last few tweaks!

grapher/detailsOnDemand/detailsOnDemand.tsx
```diff
@@ -63,25 +91,35 @@ export class DoDWrapper extends React.Component<{
...
-                </span>
-            </Tippy>
+                    <span
+                        aria-label={`A definition of the term ${this.title}`}
+                        tabIndex={0}
+                        className="dod-term"
+                    >
+                        {this.props.children}
+                    </span>
+                </Tippy>
```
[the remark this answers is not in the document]
ikesau: Fixed! I wrapped the DoDWrapper in a span to stop Tippy from complaining about interactivity contexts, but the new span wrapper wasn't display: inline-block by default. Firefox was fine but Chrome didn't like that.

danielgavrilov (approved):

grapher/detailsOnDemand/detailsOnDemand.tsx
ikesau: This one was strange. I tried just auto initially and it was still glitching out at low heights, but now it's working 🤷

ikesau:
SVG diffs are all due to underlines on links. Should be fine to merge.

PR #1563, opened 2022-07-26
Disable DoD footnotes on bake

ikesau:
I ran the local baker with configs that included DoDs before and after these changes: it's fixed!


danielgavrilov (approved):
🙌

PR #1565, opened 2022-07-28
Perf: adds a new implementation for full joining multiple tables

danyx23:
This should allevieate performance issues with many variables as recently reported. In one test the time to do the full join went down from 4 seconds to 600 ms and the memory use from 250MB to 85MB

Slack discussion: https://owid.slack.com/archives/CQQUA2C2U/p1659005364871929

> The original motivation was performance improvements. Those were quite straightforward but then I kept running into breaking charts with the svg tester that led me to add several iterations of fixes (felt so close every time ). The end result is several improvements over the status quo and one open question:
> 
> **Benefits**
> 
> -   Performance - the chart initialization time for some heavy multi-variable charts improved from ~4000 ms join cpu time to ~600ms and a memory footprint reduction by about half. (e.g. this chart <https://ourworldindata.org/grapher/share-of-deaths-by-cause> that Pablo mentioned recently and Fiona complained about perf for similar charts)
> -   Day and Year variables are now joined properly. This means that you can now do scatter and marimekko charts that use day variables (e.g. covid vaccinations) and year variables (e.g. GDP/Population) and days are matched to year values of the corresponding year. One caveat: the year variable has to have values for the years the days are in (i.e. you need values for population or gdp for 2022 for this to show up correctly). For population we have a shared historical + predicted population variable that has values for 2022 but for GDP we might need such a unified variable for day based joins to work correctly. The problem when the year variable does not have values for the correct year is that the join picks an arbitrary year (usually the first in the time series). Properly fixing this would require some bigger refactorings. For now this means that the charts that use population with day variables should switch to the population version that has projections.
> -   Multi-variable joins are now more correct in some edge cases. E.g. population used to be joined after some other joins and for scatter plots could then often implicitly use the year for population that was picked in the join before. Now all years are lined up at once and special scatter logic only applies for variables that have a target year set. This also recovers a few countries that were previously dropped as they fell out of the tolerance ranges because of the old join logic - e.g. In <https://owid.cloud/admin/charts/867/edit> Yemen and Afghanistan have high values but are missing in the old join code; or Nigeria is missing currently in <https://owid.cloud/admin/charts/3041/edit> )
> 
> **Unwanted changes**
> 
> -   Continent colors are shifted in scatter plots that join day and year variables (i.e. covid scatter plots). This has to do with how Antarctica is treated that doesn't appear in the daily data series and is discarded in the new algorithm but was still present in the inputTable in the old algorithm. I am inclined to ignore this for now, especially as Marwa and Matt are working on improving color anyhow and then we'll probably want to implement stable mappings for continents anyhow. Thoughts?
> 
> You can see all differences in charts in the svg tester images for the branch: <https://github.com/owid/owid-grapher-svgs/commit/292fa892f0791bc8335b395adb8b717ca05d6512> (use swipe to easily see a larger version of each chart in the before/after view). For a few charts I will do some minor fixes after this is merged so that e.g. population will be used correctly for day variables.

### DIFF
--- a/grapher/core/LegacyToOwidTable.ts
+++ b/grapher/core/LegacyToOwidTable.ts
@@ -219,10 +219,6 @@ export const legacyToOwidTableAndDimensions = (
         else variableTablesToJoinByYear.push(variableTable)
     })
 
-    const variablesJoinedByYear = fullJoinTables(variableTablesToJoinByYear, [
-        OwidTableSlugs.year,
-        OwidTableSlugs.entityId,
-    ])
     const variablesJoinedByDay = fullJoinTables(variableTablesToJoinByDay, [
         OwidTableSlugs.day,
         OwidTableSlugs.entityId,
@@ -258,14 +254,17 @@ export const legacyToOwidTableAndDimensions = (
             variablesJoinedByDay.appendColumns([newYearColumn])
 
         joinedVariablesTable = fullJoinTables(
-            [variablesJoinedByDayWithYearFilled, variablesJoinedByYear],
+            [variablesJoinedByDayWithYearFilled, ...variableTablesToJoinByYear],
             [OwidTableSlugs.day, OwidTableSlugs.entityId],
             // [OwidTableSlugs.year, OwidTableSlugs.entityId]
             [OwidTableSlugs.entityId] // Id' rather join with year here but the legacy behaviour
             // is to join with entityId only and filter year based variables to single years beforehand
         )
     } else if (variableTablesToJoinByYear.length > 0)
-        joinedVariablesTable = variablesJoinedByYear
+        joinedVariablesTable = fullJoinTables(variableTablesToJoinByYear, [
+            OwidTableSlugs.year,
+            OwidTableSlugs.entityId,
+        ])
     else if (variableTablesToJoinByDay.length > 0) {
         mainIndexColumns = [OwidTableSlugs.day, OwidTableSlugs.entityId]
         joinedVariablesTable = variablesJoinedByDay
--- a/grapher/core/LegacyToOwidTable.ts
+++ b/grapher/core/LegacyToOwidTable.ts
@@ -236,7 +236,7 @@ export const legacyToOwidTableAndDimensions = (
     // has values for all countries but only for the year 2015. If we join that with covid era days
     // we still want to retain the continents so we have the fallback to entity only (this was also
     // the only behaviour prior to July 2022). Remember that tolerance will only be applied much later -
-    // here we are only concerned with merging mutliple variables into an inputTable that retains information.
+    // here we are only concerned with merging multiple variables into an inputTable that retains information.
     // Another approach would be to convert years into days when we have days - then we could simplify the fallback
     // join key logic described above.
     // Another caveat is that by switching to day+entity as the primary index that we use to join we can drop some entities.
@@ -281,7 +281,7 @@ export const legacyToOwidTableAndDimensions = (
         const variablesJoinedByDayWithYearFilled =
             variablesJoinedByDay.appendColumns([newYearColumn])
 
-        // Now join the already merged days table with all the years. It is imporant
+        // Now join the already merged days table with all the years. It is important
         // to not join the years together into one table already before so that each
         // table lookup for fallback values is looked at individually.
         // See the longer comment above for the idea behind the fallback cascade here of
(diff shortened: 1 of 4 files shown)


danyx23:
Charts to edit the config for when this is merged into prod: 
* https://owid.cloud/admin/charts/335/edit
* https://owid.cloud/admin/charts/5134/edit
* https://owid.cloud/admin/charts/379/edit

danielgavrilov:
> Charts to edit the config for when this is merged into prod:
> 
> * https://owid.cloud/admin/charts/335/edit
> * https://owid.cloud/admin/charts/5134/edit
> * https://owid.cloud/admin/charts/379/edit

What needs to be edited for [335](https://owid.cloud/admin/charts/335/edit) to to work correctly? The before/after for context:

Screenshot 2022-08-03 at 10 03 41

danielgavrilov (approved):
Awesome to have these joining logic fixes, in addition to performance fixes 🎉 

I'm only just wondering what needs to be changed in the config to get those 2 charts to work.

grapher/core/LegacyToOwidTable.ts
```diff
@@ -202,16 +212,134 @@ export const legacyToOwidTableAndDimensions = (
...
+                    daysColumn.values[i] as number
+                )
+            )
+        }
+        const newYearColumn = {
+            ...daysColumn,
+            slug: OwidTableSlugs.year,
+            name: OwidTableSlugs.year,
+            values: yearsForDaysValues,
+        } as unknown as OwidColumnDef
```
marcelgerber, line 280: I think this should rather be:
```suggestion
        const newYearColumn = {
            ...daysColumn.def,
            slug: OwidTableSlugs.year,
            name: OwidTableSlugs.year,
            values: yearsForDaysValues,
        } as OwidColumnDef
```

grapher/core/LegacyToOwidTable.ts
```diff
@@ -232,8 +360,189 @@ export const legacyToOwidTableAndDimensions = (
...
+    if (tables.length === 0) return new OwidTable()
+    else if (tables.length === 1) return tables[0]
+
+    // Get all the index values per table and then figure out the full set of all stringified index values
+    const indexValuesPerTable = tables.map((table) =>
+        // When we get a mergeFallbackLookupColumn then it can happen that a table does not have all the
+        // columns of the main index. In this case, just return an empty map because that will lead all
+        // lookups by main index to fail and we'll try the fallback index
+        !mergeFallbackLookupColumns ||
+        difference(indexColumnNames, table.columnSlugs).length === 0
```
marcelgerber, line 397: I had trouble understanding this condition, and negating it to `mergeFallbackLookupColumns && difference(indexColumnNames, table.columnSlugs).length > 0` in my head helped me a lot to better understand it. I campaign for turning the ternary around :)

marcelgerber:
> The problem when the year variable does not have values for the correct year is that the join picks an arbitrary year (usually the first in the time series).

This sounds a bit problematic maybe? It is certainly something that we need to communicate well. And as you say, it's problematic for GDP for example, where we probably don't even have projection data available.
Also, it's probably pretty common for a daily variable to have data well into the current year, and for a yearly variable to lag one or two years behind the current year.

marcelgerber (commented):
Well, what can I say - this is great work and I totally get how this took a lot of your time 😄 

I won't even pretend to understand most of it, in reading through it I was maybe grasping 40% or so, but at least these didn't look all too bad :)

In `legacyToOwidTableAndDimensions`, I don't like how we now have three different code paths for "has day variables", "has year variables", and "has both of them". Is it at all possible to combine code paths?

Also, it would be great if you could add tests for joining days and years.

Overall, what to do next with this PR depends a bit on how comfortable you are with your code. If you're comfortable, go ahead and merge it and we'll see what happens :)
If you're not quite comfortable yet, I would be up for a call where we talk through the different cases, you talk through the code and what it does, and we can brainstorm together whether it all makes sense and if there's anything we can do differently.

grapher/core/LegacyToOwidTable.ts
```diff
@@ -393,8 +393,8 @@ const fullJoinTables = (
         // When we get a mergeFallbackLookupColumn then it can happen that a table does not have all the
         // columns of the main index. In this case, just return an empty map because that will lead all
         // lookups by main index to fail and we'll try the fallback index
-        !mergeFallbackLookupColumns ||
-        difference(indexColumnNames, table.columnSlugs).length === 0
+        mergeFallbackLookupColumns &&
+        difference(indexColumnNames, table.columnSlugs).length > 0
             ? table.rowIndex(indexColumnNames)
             : new Map()
```
marcelgerber, line 399: You will also have to switch these two up.

danyx23: 🤦of course

PR #1575, opened 2022-08-04
DoD static exports superscript fix

ikesau:
Thanks @marcelgerber for ID-ing this one. Constructor names were getting minified, so string matching was breaking. Switched to the better technique and validated on [staging](https://ptolemy-owid.netlify.app/grapher/annual-inflation-of-consumer-prices?country=Landlocked+developing+countries+%28LLDCs%29~Least+developed+countries+%28LDCs%29~Small+Island+Developing+States+%28SIDS%29) 👍

### DIFF
--- a/grapher/text/MarkdownTextWrap.tsx
+++ b/grapher/text/MarkdownTextWrap.tsx
@@ -310,17 +310,17 @@ export class IRDetailOnDemand extends IRElement {
 }
 
 function checkIsIRElement(token: IRToken): token is IRElement {
-    return [
-        "IRBold",
-        "IRSpan",
-        "IRItalic",
-        "IRLink",
-        "IRDetailOnDemand",
-    ].includes(token.constructor.name)
+    return (
+        token instanceof IRDetailOnDemand ||
+        token instanceof IRBold ||
+        token instanceof IRSpan ||
+        token instanceof IRItalic ||
+        token instanceof IRLink
+    )
 }
 
 function checkIsIRDetailOnDemand(token: IRToken): token is IRDetailOnDemand {
-    return "IRDetailOnDemand" === token.constructor.name
+    return token instanceof IRDetailOnDemand
 }
 
 function splitAllOnNewline(tokens: IRToken[]): IRToken[][] {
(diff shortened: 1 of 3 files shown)


grapher/text/MarkdownTextWrap.tsx
```diff
@@ -310,17 +310,17 @@ export class IRDetailOnDemand extends IRElement {
...
-        "IRLink",
-        "IRDetailOnDemand",
-    ].includes(token.constructor.name)
+    return (
+        token instanceof IRDetailOnDemand ||
+        token instanceof IRBold ||
+        token instanceof IRSpan ||
+        token instanceof IRItalic ||
+        token instanceof IRLink
+    )
```
danielgavrilov, line 319: Could this just be:

```suggestion
    return token instanceof IRElement
```

ikesau: 🙊

danielgavrilov:
Could they be `instanceof` expressions instead of function calls? We already have these:

https://github.com/owid/owid-grapher/blob/eebacbda402a27c5ee85b8e260b2cb7eb50dffcd/grapher/text/MarkdownTextWrap.tsx#L333

https://github.com/owid/owid-grapher/blob/eebacbda402a27c5ee85b8e260b2cb7eb50dffcd/grapher/text/MarkdownTextWrap.tsx#L364

https://github.com/owid/owid-grapher/blob/eebacbda402a27c5ee85b8e260b2cb7eb50dffcd/grapher/text/MarkdownTextWrap.tsx#L340

https://github.com/owid/owid-grapher/blob/eebacbda402a27c5ee85b8e260b2cb7eb50dffcd/grapher/text/MarkdownTextWrap.tsx#L377

marcelgerber (approved):

marcelgerber:
This was not introduced in this PR, but still flagging it:
The [generated SVG](https://ptolemy-owid.netlify.app/grapher/annual-inflation-of-consumer-prices) contains a [U+2028 line separator](https://unicode-table.com/en/2028/) right after the superscript, which shows up for me as a weird sign:  
Maybe this is a Windows thing, but still something we should address and maybe change the sign used there.
This is what it looks like on Windows:
image

PR #1582, opened 2022-08-09
fix: avoid redundant recomputation in choroplethMap on hover

danyx23:
@coder-whale discovered that choropleth maps recompute too much when hovering over countries. This PR fixes it by getting rid of the indirection for the choroplethData property that is passed down through props. When caching this with a computed property it was causing frequent re-evaluations.

### DIFF
--- a/clientUtils/ComponentWithReactiveProps.ts
+++ b/clientUtils/ComponentWithReactiveProps.ts
@@ -1,0 +1,42 @@
+import { autorun, IReactionDisposer, observable } from "mobx"
+
+import React from "react"
+/** This is a helper class for React components that don't use the Manager pattern
+    we often use but just want to use normal props in an efficient way.
+
+    The problem with MobX and React props is that creating props at the usage site
+    creates a new prop object and thus all properties in a class using computed
+    properties to derive values from props will be regenerated.
+
+    This class solves this by adding a new observable called reactiveProps that
+    should be used instead of this.props when computing derived values in a React
+    component. The reactiveProps are updated from props automatically but because
+    this is copied to a separate observable object the usual mobx magic works again
+    and only properties that change trigger updates in derived classes.
+
+    When using the manager pattern this is not necessary because we usually pass
+    a reference to the parent object anyhow that does not change. It is also unnecessary
+    if all props are themselves observable objects (since then mobx tracking works again).
+*/
+export class ComponentWithReactiveProps<T> extends React.Component<T> {
+    constructor(props: T) {
+        super(props)
+        const reactiveProps = {}
+        Object.assign(reactiveProps, this.props)
+        this.reactiveProps = reactiveProps as T
+        this.reactivePropsCleanup = autorun(() =>
+            Object.assign(this.reactiveProps, this.props)
+        )
+    }
+
+    override componentWillUnmount(): void {
+        if (super.componentWillUnmount) super.componentWillUnmount()
+        this.reactivePropsCleanup()
+    }
+
+    reactivePropsCleanup: IReactionDisposer
+
+    /** A copy of the React props that tracks updates more effeciently to minimize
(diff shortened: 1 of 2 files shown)


danielgavrilov:
We might have misunderstood each other on the issue. How do you verify that this fixes it?

To emulate what is happening in Karthik's code, I created a `derived` property which I call in the `render()` method:

```ts
@observer
class ChoroplethMap extends React.Component<ChoroplethMapProps> {

   // create a new prop...

    @computed get derived(): boolean {
        trace()
        if (this.props.choroplethData) {
            return true
        }
        return false
    }

    // retrieve it in the render method...

    render() {
        // ...
        const derived = this.derived
        // ...
    }
}
```

My expectation was that this `derived` property should not get recalculated on every `render` call, since `choroplethData` stays the same. But it ends up being recalculated both before and after this PR:

```
[mobx.trace] 'ChoroplethMap@1160.derived' is invalidated due to a change in: 'reactive props'
```

The issue that Karthik has is there is some computation-heavy transformation we need to cache that is derived from `choroplethData`.

danielgavrilov:
I haven't dived into it, I might've misunderstood how mobx works – it might be that the props themselves observable, so changing any prop will trigger a change to all?

danielgavrilov:
I think this could be solved with the Manager pattern – instead of passing individual props, make sure `MapChart` implements a `ChoroplethMapManager` and then pass `this` as the only parameter to `<ChoroplethMap manager={this}>`. Mobx seems to encourage this kind of thing, in order to have more fine-grained dependency relationships. 

Alternatively, we need some other way to make sure that the dependency is made between specific properties, e.g. `MapChart#choroplethData` and properties on `ChoroplethMap`. This could be done by passing a getter function to it, like:

```ts
<ChoroplethMap getData={() => this.choroplethData} />
```

And then anything using that getter within `MapChart` will create a dependency between the individual computeds.

danyx23:
... yeah it looks like I shouldn't be doing fixes shortly before midnight. I think I accidentally removed the trace statement too early and thought I had solved the issue.

The manager pattern is one way we could do this, yes.

danielgavrilov:
This new change kind of makes sense to me, but I'm not exactly sure about how mobx does its magic, and wonder if there are cases where this breaks. 

E.g. what if we conditionally pass one computed or another, but the computeds themselves are constants?


```ts
@observer
class Example extends React.Component {
    @computed get a() { return { count: 1 } }
    @computed get b() { return { count: 2 } }
    render() {
        return <Test prop={somethingThatToggles ? this.a : this.b} />
    }
}

@observer
class Test extends ComponentWithReactiveProps {
    @computed get c() {
        return this.reactiveProps.prop.count + 1
    }
    render() {
        // Does mobx know that `c` should be recalculated, given that `c` is observing either `a` or `b`, which haven't changed?
        // Does the `props` itself need to be an observable object in order to catch updates like this?
        return <span>{this.c}</span>
    }
}
```

danielgavrilov:
I'm not comfortable getting this to production without testing a bunch of these cases (what if we add a property? what if we remove a property? how are shallow & deep observables handled?) and making sure this is robust. These kind of bugs can take up a lot of debugging time later on.

I found potentially related issues/PRs, but haven't read them:
- https://github.com/mobxjs/mobx-react/pull/436
- https://github.com/mobxjs/mobx-react/pull/496
- https://github.com/mobxjs/mobx-react/issues/380

Unfortunately this robustifying would also take quite some time probably, and maybe we don't have the appetite for that, but we still want to fix the issue. I'm happy if this gets merged in case someone else on the team is more comfortable with how mobx works and can approve this.

danyx23:
Yeah good call - I did some more tests and it's hard to make this work in a predictable way for various edge cases. Abandoning this PR.

PR #1600, opened 2022-08-16
chore: add new makefile commands for updating data

larsyencken:
Newly added commands that work from within a running session:

- `make refresh`: grapher-only
- `make refresh.wp`: wordpress-only
- `make refresh.full`: both grapher and wordpress


larsyencken:
See: https://github.com/owid/owid-grapher/issues/1381

PR #1619, opened 2022-08-24
Webpack5 esm

danyx23:
Hopefully the final push to merge Marcel's heroic effort to migrate our codebase to webpack 5 and es modules.

List of issues to resolve before this can be merged:
- [x] Fix Github action step that most of our check use that runs set-node-yarn-deps that currently fails (weird 404 error on restoring tar 🤨 )
- [ ] Make all tests run
- [x] Fix bundlewatch not running (webpack error)
- [ ] Fix github code scanning issues
- [ ] Local bake fails with Scroll menu being a commonJS module (import from react-horizontal-scrolling-menu)


danyx23:
Jest errors when running locally - there seem to be more like these but only one is reported at the very bottom. Maybe an artifact of running them in parallel?
```
FAIL   jsdom  itsJustJavascript/grapher/stories.jsdom.test.js
  ● Test suite failed to run

    Jest encountered an unexpected token
```
There are a ton of `ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From itsJustJavascript/grapher/stories.jsdom.test.js.`

larsyencken:
Seems from our discussion that the future of ESM is not in this branch, but that we might take lessons here and apply them gradually to some of the sub-packages that we're setting up with Lerna.

PR #1620, opened 2022-08-25
Remove underline from Grapher tabs in .article-content

ikesau:
Changes to link styling mistakenly affected the tabs on graphers in articles. This overrides those changes.


mlbrgl (approved):
LGTM!

PR #1621, opened 2022-08-25
Research & Writing block

ikesau:
[Staging demo](https://ptolemy.owid.cloud/research-and-writing-demo)

Screen Shot 2022-08-25 at 15 25 05

image


mlbrgl (changes requested):
Hey! I came across an unfortunate side-effect on entry H2s (more below). The rest looks good!

Current h2s would break, since [this default](https://github.com/owid/owid-grapher/blob/be23c3871e95757cc02535a767757203cd76d0d6/baker/formatWordpressPost.tsx#L593) doesn't make it into LongFormPage. The formatting options passed to LongFormPage are (rather confusingly) the ones [extracted from the HTML comment](https://github.com/owid/owid-grapher/blob/be23c3871e95757cc02535a767757203cd76d0d6/baker/siteRenderers.tsx#L176).

H2s from summary blocks currently also need [to be excluded from that extra section ToC processing](https://github.com/owid/owid-grapher/blob/be23c3871e95757cc02535a767757203cd76d0d6/site/formatting.tsx#L377). It is less declarative than the formatting option but I think this is actually desirable here, since otherwise we require authors to set that option on all topic pages, when it should really be a default.

Somewhat less important, setting a default at the page level makes topic pages block less reusable in other contexts, since it forces all other H2s on the page to be toc-less.

mlbrgl (approved):
LGTM 👍

Issue #1626, opened 2022-08-29
Smoother workflow for editing suggestions in suggested chart revisions


larsyencken:
## Problem

When you go to edit a suggested chart revision, the behaviour is currently:

- If you save the edit, then it updates both the chart config and the chart data to your edited suggestion
- The suggestion revision is left in-place (listed as `PENDING`), but can no longer be approved
    - You have to dismiss the suggestion, even though it's actually been applied already

## Proposed solution

- Minimal fix: same behaviour, but do not leave the suggestion with status `PENDING`, mark it as `APPROVED`
- Better fix: allow editing the suggestion with live preview, but without approving it yet
- Docs: update the README at the top of the page to explain this scenario better

danyx23:
@danielgavrilov had a bunch of good thoughts on this topic that should be revisited before working on this tool: https://owid.slack.com/archives/CQQUA2C2U/p1661263683983659

@larsyencken and I chatted about this - it makes sense to look at the tool as a whole and make sure that it is in a good place for us to maintain

PR #1627, opened 2022-08-29
doc: improve help text for updating chart suggestions

larsyencken:
The current behaviour is not ideal, but as a minimal first step, document it more clearly.

See: #1626


larsyencken:
Confirmed this behaviour in the wild with @lucasrodes

danielgavrilov (approved):
Definitely simpler to understand!

I think we maybe need to allocate some attention to any potential changes to this chart-updating process, I'm noticing more people running into issues. I was [monologuing about it on Slack](https://owid.slack.com/archives/CQQUA2C2U/p1661263683983659).

PR #1654, opened 2022-09-13
fix(Url.ts): update types for url-parse

### LINKED ISSUE #1644: Can't recreate yarn.lock file
## Description

There is an issue that the current `yarn.lock` file can't be reproduced from scratch. 

## Expected behaviour

Removing and generating a new `yarn.lock` file should leave the project in a working condition but doesn't. 

## Steps to reproduce

1. Clone the repo
2. Run `rm yarn.lock`
3. Run `yarn`
4. Run `make up.full`
5. See error:

```
clientUtils/urls/Url.ts:9:9 - error TS2314: Generic type 'URLParse<Query>' requires 1 type argument(s).

9 ): Omit<urlParseLib, "query"> & { query: string } => {
          ~~~~~~~~~~~
``` 

## Environment

For desktop, provide:

-   OS: MacOS
-   Node Version: this occurs with both Node v16 and v18

## Additional context

Not an urgent problem, but we should fix it as it makes e.g. solving merge conflicts related to dependencies more difficult to resolve.

ikesau:
Fixes #1644

It seems url-parse fixed its typing in a minor version.


marcelgerber (changes requested):
Did you upgrade the whole lockfile?
That's... probably not a good idea. I have tried this before, and there was always something that was broken afterwards.

It's safer to just update single packages.

ikesau:
❗ 

As evidenced by the failing tests! I didn't realize how much everything would bump, one moment.

marcelgerber (approved):

PR #1659, opened 2022-09-15
Replace old CSS colour variables with new ones

ikesau:
The grep equivalent of

image

As the entire diff is impossible to review, here's the approximate order of operations I followed:

Replace `$primary-color` with `$oxford-blue`

Replace ` color: $primary-color` with ` color: $blue-100` (to try and nudge it towards mostly being used for text)

Went down the list of `$primary-color-X00` and replaced them with the `$blue-X0` equivalent (some rounding required, e.g. `$primary-color-700` went to `$blue-90` — this was the smallest possible [Delta E](http://colormine.org/delta-e-calculator/))

Remove variables that weren't being used

Replace `$body-bg` with `$white`

Replace `$secondary-colour` with `$vermillion`

Remove tertiary colours that weren't being used

Reorder the variables in the file.

`$grey-text-color` and some of the tertiary colours (shades of amber) aren't stored in Figma currently. When Marwa gets back we can do a second pass to make sure each colour is accounted for.


larsyencken:
@ikesau Sounds like a great case for shipping the work to a staging server, so that it can be visually checked out in a pinch.

ikesau:
@larsyencken I agree! The only reason it already isn't is because Ptolemy has topic page stuff on it. Now that we're awake I can ask who's got a server available 🙂

mlbrgl (approved):
Went through a few pages and and components, and nothing seems off at first glance color-wise 👍

It might be helpful to borrow Marwa's eyes on this one, to get a third pass at this - and we should be good.

PR #1677, opened 2022-09-22
Table of Contents improvements

ikesau:
A couple of ToC glitches that I wanted to fix.

[Staging demo](https://ptolemy.owid.cloud/admin/posts/preview/32527)

I have tested the changes on several COVID pages and articles (with and without y-overflow in the sidebar), and the All Charts page.

### Changes:
- Background is now `blue-10` all the time
- Remove the sticky tracking logic
- Fix `.section-heading` positioning on widescreens
- Fix top and bottom margin between the header and footer

### Before:

https://user-images.githubusercontent.com/11844404/191842712-4cc13db3-3ae3-482d-8bc1-da17ac33bbc5.mov

### After:

https://user-images.githubusercontent.com/11844404/191842931-936ab76e-c19b-4d06-9880-e2cef549d525.mov

### DIFF
--- a/site/owid.scss
+++ b/site/owid.scss
@@ -631,11 +631,12 @@ div.owid-presentations li:nth-of-type(1) a:hover {
 .with-sidebar .section-heading {
     @include xxlg-up {
         position: relative;
-        width: calc(100vw - #{$sidebar-content-width + 2 * $padding-x-md});
-        margin-left: calc(
-            50% - 50vw - #{($sidebar-content-width + 2 * $padding-x-md) * 0.5}
+        width: calc(100vw - $sidebar-width);
+        margin-left: 0;
+        left: min(
+            calc((100vw - 100% - $sidebar-width) / -2),
+            calc(-1 * $padding-x-md)
         );
-        left: $sidebar-content-width + 2 * $padding-x-md;
     }
 }
 
--- a/grapher/controls/globalEntitySelector/GlobalEntitySelector.scss
+++ b/grapher/controls/globalEntitySelector/GlobalEntitySelector.scss
@@ -1,8 +1,11 @@
 .global-entity-control-container {
     position: sticky;
     top: 5px;
-    margin-left: 7rem; // HACK, making horizontal space for "Contents" button
     z-index: $zindex-global-entity-select;
+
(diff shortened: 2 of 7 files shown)


site/owid.scss
```diff
@@ -631,11 +631,12 @@ div.owid-presentations li:nth-of-type(1) a:hover {
...
     @include xxlg-up {
         position: relative;
-        width: calc(100vw - #{$sidebar-content-width + 2 * $padding-x-md});
-        margin-left: calc(
-            50% - 50vw - #{($sidebar-content-width + 2 * $padding-x-md) * 0.5}
+        width: calc(100vw - $sidebar-width);
+        margin-left: 0;
+        left: min(
+            calc((100vw - 100% - $sidebar-width) / -2),
+            calc(-1 * $padding-x-md)
```
ikesau, line 638: This is pretty heinous but here's what's going on.

magic margin

`100vw - 100% - $sidebar-width` gives us a width equivalent to the two `remainder` rectangles in the illustration. 

We divide that by two so that we just have the remainder on the left, and make it negative.

The `min` function is needed because there is a small window where this formula doesn't work. 🙂

mlbrgl (commented):
I'm noticing some overflow around the 1600px mark:

Screenshot 2022-09-26 at 15 30 23

Any chance to restore some of the bottom margin below the toggle? This looks a bit squished (before / after):

image

We are now entering the realm of personal preference, but I find the top of the page quieter and somewhat less overwhelming when the ToC and content share the same background color, while the title banner is visible. It also makes the difference in vertical and horizontal padding around the content less jarring (before / after):

image (1)

Maybe something to run by Marwa before pulling the trigger.

ikesau:
Thanks for your comments @mlbrgl !

Changes:

- sidebar background-color -> `$white`
- sidebar button hover/focus background-color -> `$blue-10`
- fixed the spacing at the top of articles
- fixed the container sizing at large resolutions
- repositioned the toggle button (with a little animation)


https://user-images.githubusercontent.com/11844404/192379462-0f452e64-a5c6-4f27-98a9-05740516194e.mov

mlbrgl (approved):

PR #1684, opened 2022-09-28
Prominent links with Grid

ikesau:
A rewrite of the Prominent Link CSS to get it to play nice with our new columns.

[Staging demo](https://ptolemy.owid.cloud/admin/posts/preview/50492)

~@mlbrgl could you check these out for usecases I may have missed? I'm not sure what the lightbox interaction is, for example.~ Done.

Screenshot 2022-09-29 at 12-32-50 Prominent links demo

### DIFF
--- a/devTools/cypress/integration/kitchenSink.js
+++ b/devTools/cypress/integration/kitchenSink.js
@@ -160,11 +160,11 @@ describe("Kitchen sink", function () {
             src: DEFAULT_THUMBNAIL_URL,
         })
         // N/A
-        // testProminentLink({
-        //     testTitle:
-        //         "[NOT RECOMMENDED] Grapher link redirected to grapher (in wordpress)",
-        //     title: "Adjusted net savings per capita",
-        //     src: "http://localhost:3030/grapher/exports/adjusted-net-savings-per-person.svg",
-        // })
+        testProminentLink({
+            testTitle:
+                "[NOT RECOMMENDED] Grapher link redirected to grapher (in wordpress)",
+            title: "Adjusted net savings per capita",
+            src: "http://localhost:3030/grapher/exports/adjusted-net-savings-per-person.svg",
+        })
     })
 })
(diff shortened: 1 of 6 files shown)


ikesau:
I think I'll slightly modify this to introduce a grid mixin that makes it easier to define subgrids. It's going to be annoying to constantly have to write out the template each time.

Also need to handle other column layouts, the current rules won't align if someone puts a prominent link in the right-hand side of a sticky-left column, for example.

mlbrgl:
Sounds good! 

Just so you know, there is a suite of cypress tests for prominent links. You can run it with `yarn testCypress` and then select "kitchenSink.js"

Screenshot 2022-09-29 at 13 32 36

There shouldn't be any lightbox popping up when clicking a prominent link.

ikesau:
Cool, seems like everything works then. 👍 

After a discussion with Marwa we decided to change the links back to having padding instead of an outline,  because otherwise the boxes were entirely flush with sticky graphers in the other column.

So, a whole refactor for what could have basically just been a one-line change in the padding. Alas! 😅

devTools/cypress/integration/kitchenSink.js
```diff
@@ -159,11 +159,12 @@ describe("Kitchen sink", function () {
...
-            title: "Adjusted net savings per capita",
-            src: "http://localhost:3030/grapher/exports/adjusted-net-savings-per-person.svg",
-        })
+        // N/A
+        // testProminentLink({
+        //     testTitle:
+        //         "[NOT RECOMMENDED] Grapher link redirected to grapher (in wordpress)",
+        //     title: "Adjusted net savings per capita",
+        //     src: "http://localhost:3030/grapher/exports/adjusted-net-savings-per-person.svg",
+        // })
```
mlbrgl, line 168: does this test need updating?

ikesau: Seems so, yeah. I no longer have `adjusted-net-savings-per-person` in my exports after running a local build. We should probably try to not have these tests rely on the state of someone's local development DB.

mlbrgl: Yes could you please revert this so I'll take a look next time I run this and update the test?

mlbrgl (changes requested):
https://github.com/owid/owid-grapher/pull/1684/files#r987658657

mlbrgl (approved):

PR #1688, opened 2022-10-03
chore: add "make refresh.meta" to only refresh grapher metadata

larsyencken:
For some development workflows, it's not important to have the data values from grapher, only up-to-date metadata.

This PR adds `make refresh.meta` which does this, downloading and fetching only the grapher metadata.


mlbrgl (approved):
Very nice addition, thanks @larsyencken!

PR #1725, opened 2022-10-26
fix: SKIP_CHARTDATA use collided with set -o nounset

danyx23:
This commit creates a temp variable that is initialized with a default value instead


ikesau (approved):
👍 

thanks! verified this fixes the issue on mac by running `make refresh` 

master fails with `./devTools/docker/refresh-grapher-data.sh: line 33: SKIP_CHARTDATA: unbound variable`

this branch runs correctly.

larsyencken (approved):
Looks good 👍 

If it comes up again we can probably just add the `:-` to the specific check that fails (e.g. `${SKIP_CHARTDATA:-}`)

PR #1730, opened 2022-10-27
fix(duckdb): move require(duckdb) from top to inline to prevent segfaults

Marigold:
We're still getting segfaults in production even though we're not using duckdb at all. It doesn't play nicely with `workerpool` from deploys and can result in non-deterministic segfaults.

I've been testing moving `require` inline (which never gets called) and it has been working on staging. It's non-deterministic though, so one can never be sure :(. If we still get errors after this we should just remove duckdb altogether.


danyx23 (approved):
Sounds good, worth a shot. Thanks for trying all these things!

PR #1732, opened 2022-10-27
additional-information h3 CSS fix

ikesau:
Fixes some minor CSS scoping glitches with the Additional Information block

I tested how this would impact other h3's if they were in the additional information block and found that any child h3 gets lifted up to the "main heading" of the block, so presumably multiple h3's is a case we don't have to worry about.

The `&.grapherPreview img` change addresses the `img` rule which was also affecting all `img`'s within a Grapher inside `article-content`. I suspect it was written to only apply to the preview `img` before the embed is hydrated, so I made the rule only apply when `grapherPreview` is active.

## Before
image

## After
image

### DIFF
--- a/site/blocks/AdditionalInformation.tsx
+++ b/site/blocks/AdditionalInformation.tsx
@@ -106,6 +106,7 @@ const AdditionalInformation = ({
             className={classes.join(" ")}
         >
             <h3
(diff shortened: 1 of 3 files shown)


mlbrgl (approved):
makes sense to me!

PR #1743, opened 2022-11-02
fix: use -l in cron so that nvm env variables are there

larsyencken:
This is to address a problem on `live` where after upgrade to 20.04 calls to `node` no longer worked from `cron`.


larsyencken:
Merging this one without review just to make sure it reflects what we have in prod.

PR #1744, opened 2022-11-02
Fix missing URLs in Explorer grapher configs

ikesau:
Two things broke when we stopped baking the site URLs into Grapher's code directly:
- Explorers weren't injecting the URLs correctly into their Graphers
- These values weren't being exported by Graphers' `toObject` method

This fixes that.

[Staging preview](https://ptolemy-owid.netlify.app/explorers/co2)

### DIFF
--- a/baker/siteRenderers.tsx
+++ b/baker/siteRenderers.tsx
@@ -30,6 +30,8 @@ import {
     BLOG_POSTS_PER_PAGE,
 } from "../settings/serverSettings.js"
 import {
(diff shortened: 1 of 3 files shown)


danyx23 (approved):

danyx23:
@marcelgerber maybe you can have your eyes over this after we merged it - you are the most across the various ways of embedding different pieces and it would be good just to have your sanity check as well :)

PR #1755, opened 2022-11-08
chore: improve "make deploy" and "make stage" with nx reset

larsyencken:
This change aims to avoid accidental carry-over state from Lerna if you switched branch before deploy, by doing `yarn nx reset` before deploy or stage.

It also enhances the `make stage` command. Previously, you could only target the `staging` server, but now you can run `STAGING=myserver make stage` to target a different one.

## Testing

- [ ] Works on a local built environment
- [ ] Works on a fresh clone of the repo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,6 +238,7 @@ deploy:
 	yarn buildAndDeploySite live
 
 stage:
+	@echo '==> Preparing to deploy to $(STAGING)'
 	@echo '==> Starting from a clean slate...'
 	rm -rf itsJustJavascript
 	
@@ -246,7 +247,7 @@ stage:
 	yarn nx reset
 	yarn run tsc -b
 	
-	@echo '==> Deploying to staging...'
+	@echo '==> Deploying to $(STAGING)...'
 	yarn buildAndDeploySite $(STAGING)
 
 test: 
--- a/Makefile
+++ b/Makefile
@@ -232,6 +232,8 @@ deploy:
 	@echo '==> Building...'
 	yarn
 	yarn nx reset
+	yarn lerna bootstrap
+	yarn lerna run build
 	yarn run tsc -b
 	
 	@echo '==> Deploying...'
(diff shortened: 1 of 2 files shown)


danyx23 (approved):

larsyencken:
With the new `nx` commands in the `Makefile`, it can kick off the deploy, but the deploy always fails. It fails here:

```
88% sealing runtime requirements
91% sealing chunk assets processing
92% sealing asset processing TerserPlugin
Killed
error Command failed with exit code 137.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Connection to 165.22.127.239 closed.
Baking and deploying to halley [===============----------] 15/25 510.7s 📡⛔️ failed running /home/owid/tmp/halley-lars/tempDeployScript.webpack.sh [exit code: 137]
```

Any ideas @ikesau ? 

A comment on this issue: https://stackoverflow.com/questions/49184787/92-chunk-asset-optimization-webpack suggested `yarn cache clean` might help, I'll try that out.

Makefile
```diff
@@ -228,21 +231,28 @@ deploy:
 	
 	@echo '==> Building...'
 	yarn
+	yarn nx reset
+	yarn lerna bootstrap
+	yarn lerna run build
```
ikesau, line 236: After some digging, I've now realized that this isn't the right place for these commands.

In this Makefile, we only run `yarn run tsc -b` to transpile `buildAndDeploySite.ts`, which is node script that `yarn buildAndDeploySite` calls.

That script ssh's to the remote server, rsyncs the directory over, and then runs [these scripts](https://github.com/owid/owid-grapher/blob/master/baker/Deployer.ts#L217) in sequence. 

So if anywhere's missing some extra lerna commands, it's there. The reason it had been working till this point is because we [were rsync'ing over locally built copies of the lerna libraries](https://github.com/owid/owid-grapher/blob/master/baker/Deployer.ts#L69) so it didn't need to rebuild them (though this would 100% introduce errors if you changed a lerna package on a different branch, built the packages, then switched back to master without rebuilding them.)

Posting the rest of my memory allocation notes in slack.

larsyencken:
Merged it now without checking again that it works from a fresh clone. Deploying takes too long 😓

PR #1799, opened 2022-12-07
copyedits on homepage

mathisonian:
Based on a discussion with Max, these updates align with our policy to use the terminology of "articles" and "topic pages" instead of "posts" and "entries". 

Note that I don't use "topic pages" explicitly here, but favor using clear language that is still in line with those terms. 

I've also changed the phrasing of "all our X" to "all of our X" because the "of" is needed for the phrase to be grammatical (because ['our' is a pronoun](https://proofed.com/writing-tips/word-choice-all-of/#:~:text=When%20deciding%20whether%20to%20write,that%20begins%20with%20a%20determiner.), weird rule.)

### DIFF
--- a/site/FrontPage.tsx
+++ b/site/FrontPage.tsx
@@ -112,7 +112,7 @@ export const FrontPage = (props: {
                             data-smooth-scroll
                             data-track-note="homepage-scroll"
                         >
-                            Scroll to all our articles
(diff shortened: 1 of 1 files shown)


ikesau (approved):
Seems good to all of me

PR #1824, opened 2022-12-13
Fix leading/trailing whitespace links

danyx23:
In google docs it often happens that links or other styling spans include whitespaces (especially at the beginning of a word in the SDG pages). We want to avoid that. 

To do so, this code goes through the tree of spans recursively and scans while the content is only whitespace. All initial whitespace only span-simple-text spans and the whitespace start of the first non-whitespace containing span are then collected, deleted from the original nested spans and the collected whitespace inserted as one single span-simple-text span at the beginning (and the whole procedure vice versa for the trailing WS)


ikesau:
Closed, as we decided to solve this problem another way in #1832

PR #1826, opened 2022-12-15
Fix explorer crash on iOS

ikesau:
This change in resize logic did not play safely with the way explorers use graphers.

The explorer render path doesn't leverage `renderGrapherIntoContainer` , but `componentDidMount` still calls and something about the config was not persisting correctly causing the view to break when changing tabs.

This reverts the change.


marcelgerber (approved):
Tested it on ptolemy on desktop and mobile, all looking very good! thank you!

PR #1832, opened 2022-12-20
feat: fix the leading whitespace underlined issue in links in gdoc posts

danyx23:
This solution supersedes the approach in #1824. After a discussion today we decided that simplicity trumps possible correctness at the extreme - i.e. that using regexp to solve the immediate problem in two lines is better than adding a lot of recursive tree walking logic to the same ultimate effect.


ikesau (approved):
🥳 🎉 

Tested on some very simple input, works great!

PR #1842, opened 2023-01-04
Add edit button to Google Docs index page

ikesau:
Small quality of life improvement, I kept wanting to edit articles directly from the index page and assume that other people are the same.

image


mlbrgl (approved):
👍 I would just put it in another column to avoid making the list longer

PR #1843, opened 2023-01-04
Change "grey-section" to "gray-section"

ikesau:
Per Hannah, our authors aim to write with American English as it is the variety most common among our readership.

So we should ensure authors' Google Docs spellcheck is set to the NA locale, and that it doesn't auto-correct hardcoded strings.

Hence this PR.

image
image


mlbrgl (approved):
LGTM. We just need to replace it in the current gdocs too. I only found an occurence of `grey-section` in https://docs.google.com/document/d/1G3W3V8GytW6S_z78ZaEIHRcOKmtwayLWkBXz0PjEfO8/edit

PR #1867, opened 2023-01-19
feature(baking): get variable data from S3 instead of data_values

Marigold:
Add new column `variables.dataPath` with link to variable data in JSON format saved in S3 (e.g. https://catalog.ourworldindata.org/baked-variables/live_grapher/8872.json.gz). With that we can skip `data_values` table and load data from there and also avoid baking data for those variables and serve them from S3 instead. This is fairly similar to `variables.catalogPath` which is an address to parquet file in the catalog.

There's a script running in ETL that scans new datasets and creates JSON files on the fly. It's almost real-time (it takes less than a minute to detect a change and update S3) and it updates `variables.dataPath` column.

The purpose of this change is twofold:

1. Make our DB smaller by removing (part of) `data_values`
2. Serve data JSON files from S3 instead of Netlify to save bandwidth

### Handling staging/dev environment

Copying live DB still keeps `dataPath`s pointing to datasets from live. If we add any new data (e.g. running ETL locally that upserts data to `data_values`), they would have null `dataPath` and so grapher would use `data_values` instead.

### The future of `data_values`

`data_values` is now used more as a temporary storage for data until we sync it to S3 (or for dev/staging). We could easily replace it with "local catalog" of JSON data and use it as `dataPath` (i.e. making `dataPath` non-null). For that we'd need to make sure nothing is inserting to `data_values` any more (ETL, CSV importer).

Removing `data_values` would probably simplify codebase, though I'd rather do it as a next step after making sure storing data as JSON files works for us. Pruning `data_values` should unblock dynamic staging environments for us.

A lot of our tools uses `data_values`. If we wanted to avoid rewriting them all, we could just prune the largest datasets from `data_values` and keep the rest.

### Questions

1. Should I save it as `[variable_id].json.gz` or `[variable_id].json` (both same gzipped files, is adding `.gz` suffix a good idea?)
3. Not sure what are the implications for `Grapher`?
4. This would replace fetching data from parquet and from API. Perhaps it's a good time to remove them both from our codebase?

### TODO

* [ ] find all uses of `data_values`

### DIFF
--- a/baker/GrapherBaker.tsx
+++ b/baker/GrapherBaker.tsx
@@ -45,7 +45,10 @@ import {
 import workerpool from "workerpool"
 import ProgressBar from "progress"
 import fetch from "node-fetch"
-import { getVariableData } from "../db/model/Variable.js"
+import {
+    getVariableData,
+    getOwidVariableDataPath,
+} from "../db/model/Variable.js"
 
 const grapherConfigToHtmlPage = async (grapher: GrapherInterface) => {
     const postSlug = urlToSlug(grapher.originUrl || "")
@@ -194,6 +197,13 @@ async function getDataMetadataFromMysql(
     ]
 }
 
+async function assertFileExistsInS3(dataPath: string): Promise<void> {
+    const resp = await fetch(dataPath, { method: "HEAD" })
+    if (resp.status !== 200) {
+        throw new Error("URL not found on S3: " + dataPath)
+    }
+}
+
 export const bakeVariableData = async (
     bakeArgs: BakeVariableDataArguments
 ): Promise<BakeVariableDataArguments> => {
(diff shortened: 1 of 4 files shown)


Marigold:
Here are some places where `data_values` is used. None of them looks like a major blocker, though updating them to use `dataPath` could be a fair amount of work (mostly validating that they still work).

### ETL
- [backporting](https://github.com/owid/etl/blob/master/backport/backport.py#L209)
- [chart revision suggester](https://github.com/owid/etl/blob/master/etl/chart_revision_suggester.py#L421) (could be replaced by variable statistics)
- [compare](https://github.com/owid/etl/blob/master/etl/compare.py#L269)

### Grapher
- CSV importer
- deleting variable
- country profiles
- exportChartData
- exportDataset

#### analytics
- sync `data_values` to bigquery

PR #1873, opened 2023-01-20
Images in Archie

ikesau:
[Staging preview](https://ptolemy-owid.netlify.app/ikes-test-document)

### Changes
- Adds a flow for hosting and baking images with Google Docs, Drive, and S3
- Adds an `images` table
- Adds a `posts_gdocs_x_images` table to track when a Google Docs article uses an image
- Adds a INT id column to the `posts_gdocs` table for easier foreign keying
- ~Adds `lazysizes` to automate the `sizes` attribute calculation for responsive images~

Here's a [flowchart of how the preview flow works](https://flowchart.fun/c#NoJwpgDgNghgxmAugAgKoQCYwC5mdgCzwAMBRAOzhAEs4iMBJAWxgHMwAhKAezgGtiyAO7VCyJmGwwsU5ADMQ3JsgAiNAG5gAUMl3IASpICuIcshjII4ddTBDk3OfiLIMvIxPLYtwJtRCKICgAYpJ0znjULOzIAEY8sQDO8orKatSaOnoAsv6BEXEJydjcyADKAMxZenqGcNwgGAVRbHgSUjIW1GaEeCoc1TVDABTg0PBgAJRahgCORmCJ2K7unstWYDZ2KUrIiWAgmiCDKtyLBW5wHmBeyAQwmsgt7IkA-IM1AJ6LAFzIodhwhYoNQlg4nM82pJpDgumZ0pkhkjVKVek9oucwAAPUHYd7IgnIb6JP4ACQeeCEeBgUHA0k+4jyDTATTRkOKpUq+MJPOJfwAguACn4AsymvUILZkkYIABaEqymRgbk81VE37IUaQWAIaZqtXkbh-YYiwJ6-VIw3G00Nc0Wq3IUiikAAQg+yAdhmwJjMFg2W3sjguqxu3i0AF5I5GtABvAC+Eaj4aAA)

**Note**: any time an image is previewed, its metadata will be updated if necessary. This will be reflected in the next bake even if a new version of the article wasn't published. 

e.g. a document referencing `chart.png` v1 is published, then  `chart.png` v2 is uploaded using Drive's "Manage Versions" feature. If the article is previewed, we'll update the copy in S3 to v2.

### The Baking process
When an article is published, any images it references are added to the `posts_gdocs_x_images` table. **Only published articles have these references tracked.** 

If a bake is triggered, we find all unique image ID's in `posts_gdocs_x_images` and fetch them from S3. For non-SVG images, we generate a set of smaller webp files. Both PNG and Jpeg are converted to lossless webp (because it took the same amount of time as saving them as their original filetype, and the compression is better)

These compressed versions (or the SVG) are copied to the `bakedSite/images` directory.

The `Image` component that we use to render these images will either use the Digital Ocean CDN (if we're in a preview) or load from Netlify. The component is aware of the image's original max width, which it uses to calculate which resized sources will be available.

~We use the lazysizes library (3.4kB gzipped+minified) to listen to all images with the class `lazyload`. Their [sizes](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/sizes) attribute will be updated any time the `img` element changes size, allowing the browser to always select the most appropriate image from its src-set.~

I still need to robustly QA the baking/lightning-update process, I suspect it's not working quite right at the moment.


danyx23:
Nice work! The high level description is super useful, thanks for that. I didn't have time yet to do a detailed pass over the code, but I have a few high level comments that I thought might be useful to already surface:

The previews get a bit slower now because they have to wait for the image checking, right? Glancing at it this is already done in parallel but in theory it could still make the previews slow if one of the requests takes a long time, right? Should we maybe log if that takes longer than a second or so?

On the unique names - as we briefly discussed in the P&D meeting I think it would be good if errors are loud and clear in preview but in baking I would try to warn on slack but be lenient. Some cases like copying an image into a different folder so that all of a sudden it exists twice could happen easily and would brake bakes for other users. Probably the best would be to revert to using whatever image we already have in S3 for a given filename if there are duplications - what do you think?

As also mentioned in the P&D meeting, with such large user numbers the long tail of users is substantial so we should try to have a fallback for browsers that don't support wepb - probably the original png/jpeg is fine. This will mean switching to the picture element if I understood responsive images right but I think is worth it in this case as this impacts the "basic static site" viewing experience, not just the interactive one where we can be more adventurous with breaking experiences for some small percentage of users.

Finally, I wonder if the overhead of the sizes library is worth it. As you said it has to load early and blocking to do it's magic. As far as I understand responsive images, images with only a srcset would lead the browser to pick the image that is best to fill the full width of the browser viewport. This is substantially too wide for desktop in many scenarios with our new, narrow layout, but for mobile where it counts the most it would be just fine. Alternatively I think it wouldn't be too hard to write the media queries that give a reasonable approximation for the largest size that an image would actually be visible at on desktop. My hunch is that either would be a better solution than having to eagerly load a library before other content.

mlbrgl (commented):
Very thorough effort 👏 And yes the description and the flowchart really help!

This is only a half first pass review, I'm still trying to get an image to go through the system in the happy path.

> Note: any time an image is previewed, its metadata will be updated if necessary. This will be reflected in the next bake even if a new version of the article wasn't published.

In the same vein, I'm a bit uneasy about `get` routes having an effect on the production S3 store.

db/model/Image.ts
```diff
@@ -0,0 +1,255 @@
...
+                    originalWidth: google.imageMediaMetadata?.width || 100,
+                }))
+
+            const duplicateFilenames = findDuplicates(
+                images.map((image) => image.filename)
+            )
+
+            if (duplicateFilenames.length) {
+                throw new Error(
+                    `Multiple images are named ${duplicateFilenames.join(", ")}`
```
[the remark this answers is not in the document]
ikesau: Ah yes, this is due to shared drives keeping files around even after they've been deleted so we need to filter out images where `image.trashed === true`

I've fixed it and will push up my changes soon. Sorry about that, didn't expect you to be reviewing today.

adminSiteClient/GdocsPreviewPage.tsx
```diff
@@ -31,61 +32,114 @@ import { GdocsEditLink } from "./GdocsEditLink.js"
...
     const [errors, setErrors] = React.useState<ErrorMessage[]>()
-
     const { admin } = useContext(AdminAppContext)
     const store = useGdocsStore()
 
-    const hasChanges = useGdocsChanged(originalGdoc, gdoc)
-    const syncingError = useUpdatePreviewContent(id, gdoc, setGdoc, admin)
-    const isLightningUpdate = useLightningUpdate(originalGdoc, gdoc, hasChanges)
-    useAutoSaveDraft(gdoc, setOriginalGdoc, hasChanges)
+    const fetchGdoc = useCallback(
```
ikesau, line 60: @mlbrgl I made these changes because race conditions were emerging on the intial load. Tracking state and responsibilities between this component and the `useUpdatePreviewContent` hook was getting tricky to debug so I lifted everything back into this component. Hope it makes sense to you!

adminSiteClient/gdocsValidation.ts
```diff
@@ -106,6 +111,36 @@ export class ExcerptHandler extends AbstractHandler {
...
+                        property: "body",
+                        type: parseError.isWarning
+                            ? ErrorMessageType.Warning
+                            : ErrorMessageType.Error,
+                        message: parseError.message,
+                    })
+                })
+                // For now, only images can have data errors
+                if (block.type === "image") {
+                    block.dataErrors.forEach((dataError) => {
```
ikesau, line 129: I wanted a separate interface to store errors that aren't to do with parsing the archie syntax - in this case, if something's wrong with the image file that's being referenced.

adminSiteClient/GdocsStore.tsx
```diff
@@ -23,36 +27,42 @@ export class GdocsStore {
...
+                gdoc,
+                "PUT"
+            )
+            .then(getArticleFromJSON)
     }
 
-    async publish(gdoc: OwidArticleType) {
-        const publishedGdoc = {
+    async publish(gdoc: OwidArticleType): Promise<OwidArticleType> {
+        const publishedGdoc = await this.update({
```
ikesau, line 41: Using the response object as it has the correct `updatedAt` value from SQL's `CURRENT_TIMESTAMP`

site/gdocs/ArticleBlock.tsx
```diff
@@ -158,6 +156,17 @@ export default function ArticleBlock({
             />
         )
     }
+    if (block.type === "image" && block.dataErrors.length > 0) {
```
ikesau, line 159: Eventually we'll probably want a more robust abstraction for these errors but I didn't want to prematurely optimise

mlbrgl (commented):
I think I ran into the issue you were trying to fix with #1926. The migration  `GdocPostsXImages1674238961678` fails.

It seems the mismatch issue between `gdocId` and `posts_gdoc.id` comes from the collation. Removing the foreign key constraint, the migration runs and shows the following data structure :

```sql
CREATE TABLE `posts_gdocs_x_images` (
  `id` int NOT NULL AUTO_INCREMENT,
  `gdocId` varchar(255) NOT NULL,
  `imageId` int NOT NULL,
  PRIMARY KEY (`id`),
  KEY `imageId` (`imageId`),
  CONSTRAINT `posts_gdocs_x_images_ibfk_1` FOREIGN KEY (`imageId`) REFERENCES `images` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```

whereas `posts_gdocs` uses the case sensitive variation :
```sql
  `id` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
```
Changing one or the other lets the FK constraint run.

ikesau:
Thank you for the help there @mlbrgl ! I've made the change - hopefully you'll now be able to test it locally 🙂

ikesau:
Notes from call with Matthieu:

- gray-section images not working
- see if i can lift out the archieToEnriched traversal code
- nullable OG width
- duplicate filename error not showing on FE

mlbrgl:
Another partially explored thought is why use the `filename` and not the `fileurl` given the latter is likely easier to grab and less prone to typos especially with longer filenames that might not be as clean as `cartogram.png`?

Screenshot 2023-02-21 at 18 44 17

db/model/Gdoc/Gdoc.ts
```diff
@@ -68,32 +83,91 @@ export class Gdoc extends BaseEntity implements OwidArticleType {
...
+        }
+        return
+    }
+
+    get filenames(): string[] {
+        return Gdoc.extractImagesFilenames(this.content)
+    }
+
+    static extractImagesFilenames(enriched: OwidArticleContent): string[] {
+        const filenames: string[] = []
```
mlbrgl, line 131: nit: could also make this a `Set` since we need unique filenames

packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -1291,6 +1292,25 @@ export const getArticleFromJSON = (
     }
 }
 
+export function recursivelyMapArticleBlock(
+    block: OwidEnrichedArticleBlock,
+    callback: (block: OwidEnrichedArticleBlock) => OwidEnrichedArticleBlock
+): OwidEnrichedArticleBlock {
+    if (block.type === "gray-section") {
+        block.items.map((block) => recursivelyMapArticleBlock(block, callback))
```
mlbrgl, line 1300: `forEach` since we're not using the result of the `map`? (another instance below) might not be relevant if refactoring into the other traversal function as mentioned above.

packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -1291,6 +1292,25 @@ export const getArticleFromJSON = (
...
+    }
+    if (
```
mlbrgl, line 1302: nit: `else if`?

ikesau:
Oh, interesting. It looks like Docs has had a little revamp.

A month ago I'd experimented with the _Insert Image From Drive_ feature to see if we could use it. It opened a modal window and would insert the image inline. Back then, I didn't see any properties in the API response that I could use to get the original image with, just the `contentUri` which was resized and re-generated for every instance image. 

This new functionality _would_ be very useful, except it still doesn't support inserting SVG's, which is the way we want to go for our static charts. 😞 

As for filenames versus google IDs, I prefer the legibility of the filenames. It makes them easier to sanity check whereas IDs always have to be investigated.

mlbrgl (commented):
"Committing" these comments but still WIP

mlbrgl:
> This new functionality would be very useful, except it still doesn't support inserting SVG's, which is the way we want to go for our static charts. 😞

Ah, right...

> For now, authors can insert images inline if they want to remember what the image is (like they do with graphers) though ideally we'd automate this with the extension so we never have to worry about inline images getting out of sync if they upload a new revision of the image via Drive.

👍

> As for filenames versus google IDs, I prefer the legibility of the filenames. It makes them easier to sanity check whereas IDs always have to be investigated.

Makes sense. I might be overestimating the typing issue, and this makes having cryptic references a higher price to pay.

mlbrgl (approved):
Again, impressive effort!

I ran this locally and tried to break it on various fronts, but couldn't find any substantial use cases that would blow up (you already noticed the WSOD on duplicate filenames).

The use of filename vs fileid is probably the main thing  that I'm ambivalent about, where the ratio benefit / complexity seems off.

As a side note, there are quite a few moving parts here, and this made grasping the architecture and its various choices somewhat challenging to me. For instance, I could still benefit from confirming my intuition about the reasons for using an additional file storage. Maybe that's worth documenting in our architecture documents at some point ... and / or through tests, although I'm well aware of the complexity here.

I haven't addressed the admin refactor. As noted above, it'd be easier for me to look at a possible side-effect in isolation (and would also avoid blocking this PR longer).

Pending resolution on some of bigger open topics, I think we're good to merge!

mlbrgl:
I think @mallikasnyder used a bunch of fixed-graphic blocks already for SDG. So we should provide a (manual) way to resolve the soon-to-be faulty blocks.

site/gdocs/Image.tsx
```diff
@@ -1,14 +1,90 @@
...
+} from "../../settings/clientSettings.js"
+import { ArticleContext } from "./OwidArticle.js"
+import { Container } from "./ArticleBlock.js"
+
+// generates rules that tell the browser:
+// below the medium breakpoint, the image will be 95vw wide
+// above that breakpoint, the image will be (at maximum) some fraction of 1280px
+const generateResponsiveSizes = (numberOfColumns: number): string =>
+    `(max-width: 960px) 95vw, (min-width: 960px) ${Math.floor(
+        1280 * (numberOfColumns / 12)
```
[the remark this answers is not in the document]
ikesau: [1280px as the maximum width of any content in our columns](https://github.com/owid/owid-grapher/blob/82b76b1279285e96c3abb02f23bd38d1222c7d40/site/css/grid.scss#L63)

image

baker/SiteBaker.tsx
```diff
@@ -336,11 +344,66 @@ export class SiteBaker {
...
+    private async bakeDriveImages() {
+        const images: Image[] = await db
+            .queryMysql(
+                `SELECT * FROM images WHERE id IN (SELECT DISTINCT imageId FROM posts_gdocs_x_images)`
+            )
+            .then((results: ImageMetadata[]) =>
+                results.map((result) => Image.create<Image>(result))
+            )
+
+        this.ensureDir("images")
```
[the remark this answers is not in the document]
ikesau: Hmm yeah we may as well avoid this footgun. 

I'll put it all inside `images/published`

db/model/Gdoc/htmlToEnriched.ts
```diff
@@ -650,9 +650,11 @@ function cheerioToArchieML(
                         content: [
                             {
                                 type: "image",
-                                src: image?.attribs.src ?? "",
-                                caption: figcaptionElement?.value ?? [],
+                                // TODO ike
+                                filename: "",
+                                alt: "",
                                 parseErrors: [],
+                                dataErrors: [],
```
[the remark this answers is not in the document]
ikesau: Yeah I think I'll do it later when I have a better idea of how to handle the WP backporting

ikesau:
Hey @mlbrgl 

What's the Saving issue you've been having? I haven't been able to reproduce it and I'd prefer not to have to extract it into another PR.

The added complexity of using filenames doesn't seem that significant to me.

1. Right click on the file -> Click "Rename" -> Copy -> Paste into doc
2. Right click on file -> Click "Get link" -> Click "Copy link" -> Paste into a doc

Except when it's in the doc there's no way of telling what the google link points to, whereas a filename should be self-explanatory.

And it's already in our interests not to have filename collisions, so that our charts can all be unique and nicely named (otherwise it'd be possible to have two charts with the same name and we'd have to append google IDs to them or something to discriminate them)

Storing the images in S3 is... less necessary in hindsight given that we ended up deciding that we're okay with all articles automatically updating their version of an image if the Drive version is updated.

Some upsides: it makes us robust to Drive outages, and makes it so authors can't accidentally break articles by deleting things in Drive, which is nice. It will also allow us to cache the resizing (if we do end up finding that resizing everything each bake is a bottleneck) and could eventually transition to being our actual image hosting solution if we move away from Netlify.

mlbrgl:
> The added complexity of using filenames doesn't seem that significant to me.
> Right click on the file -> Click "Rename" -> Copy -> Paste into doc
> Right click on file -> Click "Get link" -> Click "Copy link" -> Paste into a doc

Fair enough on generating the link / filename. 
It was more on the handling duplicates side of things, but good point on the required slug unicity later on when we make static charts first class citizens.

> What's the Saving issue you've been having? I haven't been able to reproduce it and I'd prefer not to have to extract it into another PR.

If I remember correctly, I was making a change in the gdoc and the "Saving ..." component in the top right wouldn't change back to "Saved" when done. I can't quite recall if it was changing non-image or image content.

PR #1882, opened 2023-01-27
Add ability to run tests with a testing instance of MySQL

danyx23:
This PR adds the infrastructure to run tests written in JS using jest that can read/write to a database with the same schema as our production database but without any data.

The idea is for this to be used for tests that want to assert certain properties of the database - e.g. that triggers or views work in the expected way.

We only started doing migrations at some point in 2020 or so when there was already a database with a certain schema. This PR therefore includes a SQL script that recreates the state the schema had back then. 

On top of this base schema all migrations are run. Currently this happens in one go but if testing migrations is interesting then the script could also run migrations 1-5, run a test, run migration 6, run some tests and then run the rest of the migrations and run the rest of the tests.

To test this add the GRAPHER_TEST_DB_NAME etc env variables into your .env file (see the updated .env.example-grapher) and run `make dbtest`

The tests are also set up to run on github actions (including bootstrapping the database via migrations)


db/migration/pre-migrations-schema.sql
```diff
@@ -0,0 +1,447 @@
...
+  `createdAt` datetime NOT NULL,
+  `updatedAt` datetime NOT NULL,
+  `namespace` varchar(255) COLLATE utf8mb4_0900_as_cs NOT NULL,
+  `isPrivate` tinyint(1) NOT NULL DEFAULT '0',
+  `createdByUserId` int NOT NULL,
+  `metadataEditedAt` datetime NOT NULL,
+  `metadataEditedByUserId` int NOT NULL,
+  `dataEditedAt` datetime NOT NULL,
+  `dataEditedByUserId` int NOT NULL,
+  -- `nonRedistributable` tinyint(1) NOT NULL DEFAULT '0',
```
ikesau, line 150: Oof this file must have taken a while 😓 

What's up with the commented out statements throughout?

db/tests/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+It can be useful to fetch segments from the prod or a staging database for testing. The nicest way to do this is to write a where clause that selects the right rows from the table you are interested in and then run a mysqldump command to create insert statements for these rows, like so:
+
+```bash
+mysqldump -h $HOST -u $USER -p --no-create-info --no-tablespaces --single-transaction live_grapher charts --where "id=3963" > grapher3963.sql
+```
+
+This SQL can then be specified in the code that sets up the database.
+
+It is good practice to be deliberate about how your tests leave the database behind for the next test file. By default the teardown of a file should remove any rows that were created, but it can also make sense to build up a set of state that is shared across tests. Communicating either clearly is key.
```
ikesau, line 15: 💯 

👌

db/tests/run-db-tests.sh
```diff
@@ -0,0 +1,36 @@
...
+: "${GRAPHER_TEST_DB_PASS:?Need to set GRAPHER_TEST_DB_PASS non-empty}"
+: "${GRAPHER_TEST_DB_NAME:?Need to set GRAPHER_TEST_DB_NAME non-empty}"
+
+echo '==> Starting & initializing Mysql instance for testing'
+
+docker-compose -f docker-compose.dbtests.yml up -d
+
+./devTools/docker/wait-for-tests-mysql.sh
+
+sleep 3s # This is to give the db init script time to create the user, database and tables
```
ikesau, line 20: Not sure if this is a WSL thing, but `sleep 3s` doesn't work on Mac. `sleep 3` does

jest.db.config.js
```diff
@@ -0,0 +1,16 @@
...
+            testEnvironment: "node",
+            testMatch: ["<rootDir>/itsJustJavascript/db/tests/**/*.test.js"],
+            modulePathIgnorePatterns: ["<rootDir>/wordpress/"],
+        },
+    ],
+}
+
+// special envs for tests
+process.env = Object.assign(process.env, {
+    CATALOG_PATH: "https://owid-catalog.nyc3.digitaloceanspaces.com",
```
ikesau, line 15: Which tests is this used in?

ikesau (approved):
Very cool! 

There's one little glitch with a sleep command, but everything worked fine after making the change. I also wrote some broken tests and confirmed that they broke 👍

db/tests/run-db-tests.sh
larsyencken:
```suggestion
sleep 3 # This is to give the db init script time to create the user, database and tables
```

Yes, this also broke for me. On Linux, the `s` suffix is optional, it defaults to seconds anyway. On MacOS it does not accept the suffix and only accepts seconds.

larsyencken: Actually, even after getting it working, I was getting tons of spurious failures due to this sleep. Added a PR on top of this branch with a suggested fix: https://github.com/owid/owid-grapher/pull/1890

devTools/docker/create-test-db.sh
```diff
@@ -0,0 +1,42 @@
...
+_mysql() {
+    mysql -uroot -p$DB_ROOT_PASS -h$DB_ROOT_HOST --batch --skip-column-names "$@"
+}
+
+createTestDb() {
+    echo "Waiting for DB to be online"
+    mysqladmin ping -h$DB_ROOT_HOST -uroot --password=$DB_ROOT_PASS --wait=30
+
+    # DB_EXISTS=$(_mysql -e "SHOW DATABASES LIKE '"${GRAPHER_TEST_DB_NAME}"';" | grep "${GRAPHER_TEST_DB_NAME}" > /dev/null; echo "$?")
+
```
larsyencken, line 25:
```suggestion
```

larsyencken (approved):
Nice work! All the pieces are here, though it still feels like black magic that you brought them together 🐈‍⬛ 

You should ship it! :shipit: 

Note a PR I layered on top of this to fix a race condition I was having https://github.com/owid/owid-grapher/pull/1890

jest.db.config.js
danyx23: Leftover from copying this form the other config - removing it

PR #1887, opened 2023-01-31
refactor(baking): replace `data_values` table by fetching data JSON files

Marigold:
Rip out `data_values` table and fetch data from data JSON files instead. To make this work I had to refactor a few things:

1. Start using [nodejs-polars](https://github.com/pola-rs/nodejs-polars) to manipulate with dataframes
2. Only bake new variables for `country_latest_data` (country profiles) - this should be fine given that variables data never changes (we only do revisions), also we'll be replacing it by new country profiles soon
3. Simplify Makefile by removing SQL dump with data values
4. Raise an error when someone tries to use "Import CSV" tool (got replaced by Fast-track)
5. Remove DuckDB and option for reading Parquet
6. Remove option for fetching data from Data API

## Polars

[nodejs-polars](https://github.com/pola-rs/nodejs-polars) is a cool Rust alternative to Pandas. It has a lot of momentum in python world, unfortunately its Node version is not as mature (it is being actively developed, but has only 74 stars and one dev). Even though I had to debug a few cryptic Rust errors and its Node documentation is practically non-existent, I found it more natural to work with than manipulating objects (e.g. joins, writes to csv, etc.). In the long-term we're probably going to move all data stuff to Data API, but until then Polars could be a nice addition to grapher.

## Questions
* Should I split it into multiple PRs?
  * refactoring with polars
  * removing DuckDB and Data API
  * remove `data_values`
* I removed an option for fetching data from Data API... I'm not sure how far are we from using it. Should I keep it in code or bury it in git history?
* Should I add some unit tests (there are basically none at the moment)

## TODO
* [ ] compare all baked JSON files from S3 to those baked from data_values

### DIFF
--- a/db/model/Variable.ts
+++ b/db/model/Variable.ts
@@ -382,14 +382,12 @@ export const _dataAsDFfromS3 = async (
 
     const dfs = await Promise.all(
         variableIds.map(async (variableId) => {
-            return pl
-                .DataFrame(await readValuesFromS3(variableId))
-                .select(
-                    pl.col("entityId").cast(pl.Int32),
-                    pl.col("year").cast(pl.Int32),
-                    pl.col("value").cast(pl.Utf8),
-                    pl.lit(variableId).cast(pl.Int32).alias("variableId")
-                )
+            return createDataFrame(await readValuesFromS3(variableId)).select(
+                pl.col("entityId").cast(pl.Int32),
+                pl.col("year").cast(pl.Int32),
+                pl.col("value").cast(pl.Utf8),
+                pl.lit(variableId).cast(pl.Int32).alias("variableId")
+            )
         })
     )
 
@@ -482,7 +480,7 @@ const httpsAgent = new https.Agent({
     maxSockets: 50,
 })
 
-const fetchS3Values = async (
+export const fetchS3Values = async (
     variableId: OwidVariableId
 ): Promise<OwidVariableMixedData> => {
     const dataPath = await getOwidVariableDataPath(variableId)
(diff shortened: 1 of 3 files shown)


danyx23:
Hi Mojmir! It looks like you created another PR that does the Polars stuff. Can you re-submit this one with only the other work? I'll gladly take a look at it then.

Removing the data api and leaving it to git history sounds fine with me for now.

Marigold:
@danyx23 rebased and moved `data_values` removal from scripts and docs to other PR so that this one is easier to review. I'm working on another one that uses metadata JSON files from S3. Thanks!

danyx23:
I might be missing something but if we don't bake variables that have a data path to S3, isn't that going to cause issues on prod when baking runs? The normal grapher requests would still hit netlify data.json files and those would not exist once a file has been created in S3 - what am I missing?

Marigold:
> I might be missing something but if we don't bake variables that have a data path to S3, isn't that going to cause issues on prod when baking runs? The normal grapher requests would still hit netlify data.json files and those would not exist once a file has been created in S3 - what am I missing?

Sorry, I was too ambitious and tried to load data from S3 to avoid storing baked files on Netlify. It turned out it's probably not as trivial and would require changes to Grapher (I couldn't easily change [data prefetching](https://github.com/owid/owid-grapher/blob/master/site/GrapherPage.tsx#L88)... I guess I'd have to somehow point `bakedGrapherURL` to S3, but that'd be a mess since metadata still lives on Netlify).

Rolling it back and adding tests... ⌛

Marigold:
@danyx23 looks better now, it's ready for review. I'm testing it out on staging. I don't really like the way unit tests are written with all those mocks for MySQL. I think this might be a nice use case for your new DB tests. With them I could create sample data in MySQL and run all tests against it. Not sure how hard that could be. It'd be certainly nicer if we had proper TypeORM definition for `Variable` (and potentially others like `entities` or `data_values`) and then migrate it to DB tests.

Marigold:
@danyx23 side question about grapher data & metadata links - do you know what would it take to switch from Netlify links to S3 links in grapher? From code it looks that `bakedGrapherURL` passed to grapher is then used to construct links like https://ourworldindata.org/grapher/data/variables/data/525715.json from variable id.

If we are to switch to links like https://catalog.ourworldindata.org/baked-variables/staging_grapher/data/525715.json, we'd have to pass the link to `Grapher` instance somehow. For instance we could create new parameters `variableDataURLs`, `variableMetadataURLs` and pass
```
variableDataURLs={
  525715: https://catalog.ourworldindata.org/baked-variables/staging_grapher/data/525715.json,
}
variableMetadataURLs={
  525715: https://catalog.ourworldindata.org/baked-variables/staging_grapher/metadata/525715.json,
}
```
Would it make sense or not? I bet there must be a better way how to do it.

Marigold:
@danyx23 could you take a look, please? Just to recap - at baking time we look at whether `dataPath` is null or not and based on that [read values either from S3 or from `data_values`](https://github.com/owid/owid-grapher/pull/1887/files#diff-3e4306be594cb95ff9fe006c3c020b190c208882db96d4b8dbb9f38c501c8199R468). That's all that this PR does, besides adding new unit tests (and minor refactorings). I've compared all baked data JSON files from MySQL and from S3 and they're identical. Loading data from S3 doesn't affect performance much compared to loading from MySQL.

Next step is removing the part that reads from MySQL to only read from S3. I'll just have to figure out how local development should work (and get `datasync` service merged and deployed).

db/model/Variable.ts
```diff
@@ -366,24 +434,75 @@ export const dataAsDF = async (
...
+    if (!dataPath) {
+        throw new Error(`Missing dataPath for variable ${variableId}`)
+    }
+    return fetchS3ValuesByPath(dataPath)
+}
+
+export const fetchS3ValuesByPath = async (
+    dataPath: string
+): Promise<OwidVariableMixedData> => {
+    return (await (
```
danyx23, line 503: You don't need the outer await here (but then you need to change the type to a promise like in the return type). In this case I think that is a bit nicer to read

danyx23 (approved):
Looks good! Had a minor style nit but feel free to disregard

PR #1889, opened 2023-02-01
Migrate to mkdocs

### LINKED ISSUE #1883: Redo site docs in mkdocs + material theme
## Overview

As we prepare to onboard new people, it becomes more important to have a great overview of our systems for them.

Today, we have getting started docs for _codebases_, but nothing that starts at the top and walks you through. Furthermore, the `owid-grapher` name is a misnomer; it's actually the monorepo for the Our World In Data site.

We would like to redo the docs for Grapher in a similar way that we did for the ETL, using `mkdocs` with a material theme.

## Related

- ETL docs: https://docs.owid.io/projects/etl/en/latest/
- Architecture overview: https://gist.github.com/larsyencken/28fab0e49da145ab0477a002670d3a75

swaptr:
This PR will port the existing docs to mkdocs with material theme as discussed in #1883 
Fixes #1883


swaptr:
Hey @larsyencken Could you please take a look at this PR and let me know if you have any feedback?

larsyencken:
Hey @swaptr, thanks for your efforts to dive in here. I'd actually already started on this issue locally, as you can see it's already been assigned. 

Since there's a lot of editing work to do as I go, I won't try to integrate what we've both done, and will instead close your PR here.

Before I do, as high-level feedback: your basic conversion direction was good. We would want to drive this with a `Makefile`, and since it's a `node` project, probably to hide the Python dependencies in the `docs/` directory to keep the top-level folder clean.

PR #1890, opened 2023-02-01
fix: get rid of race condition on "make dbtests"

larsyencken:
When running "make dbtest", a new MySQL testing database is intialised with a base schema and users. It must be ready before tests can be run.

Currently, the command we use to wait may return before the database is in fact ready, which we workaround with "sleep 3", although on some machines 3s is not enough.

This change replaces this strategy with by adding a "tombstone" table called `_test_db_ready` which is added at the end of db init. The wait script waits for this table to be created, meaning there is no longer a race condition on first run.

To avoid a race condition on second run, we delete the tombstone after every test run, meaning that on subsequent runs we will wait until the db is re-initialised again.

### DIFF
--- a/db/tests/run-db-tests.sh
+++ b/db/tests/run-db-tests.sh
@@ -17,8 +17,6 @@ docker-compose -f docker-compose.dbtests.yml up -d
 
 ./devTools/docker/wait-for-tests-mysql.sh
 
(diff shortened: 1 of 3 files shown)


danyx23:
Very nice, thanks a lot!

PR #1891, opened 2023-02-01
enhance(mysql): exclude pageviews from export

Marigold:
Part of https://github.com/owid/analytics/pull/70

### DIFF
--- a/db/exportMetadata.ts
+++ b/db/exportMetadata.ts
@@ -22,7 +22,7 @@ const filePath =
         ? "/tmp/owid_metadata.sql"
         : "/tmp/owid_metadata_with_passwords.sql")
 
-const excludeTables = ["sessions", "dataset_files", "data_values"]
+const excludeTables = ["sessions", "dataset_files", "data_values", "pageviews"]
 
 async function dataExport(): Promise<void> {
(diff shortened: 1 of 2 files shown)


marcelgerber (commented):
So, I know that https://github.com/owid/analytics/pull/70 creates/updates this table every day.
However, that only happens on `live` and not on any other server or local environment. And if the table doesn't exist and this script is run, then `mysqldump` will fail with an error.

So, in my mind we should also create a proper migration so that our database schema is consistent no matter in what environment we came up with it. In particular, this also means that the schema in the new db-tests will be the same as the one we get from importing the live database.

Curious to know what @danyx23 thinks.

Marigold:
@marcelgerber good points, I'll add the migration that creates the table. I guess we don't need TypeORM definition yet.

danyx23:
Agreed - adding a migration is the way to go. Otherwise this looks good and I'm looking forward to having this!

Marigold:
Added that migration and tested that `node itsJustJavascript/db/exportMetadata.js` still works. @marcelgerber is this good to go?

marcelgerber (approved):

PR #1892, opened 2023-02-01
refactor(Variable): use Polars for table manipulations & remove DataAPI & remove DuckDB

Marigold:
* Refactor data-oriented code with nodejs-polars
* Remove DuckDB support
* Remove DataAPI support
* Only bake new variables for table `country_latest_data` (country profiles) - this should be fine given that variables data never changes (we only do revisions after creating new variables), also we'll be replacing it by new country profiles soon anyway

The question is whether we should remove DataAPI from code and bury it in git history or keep maintaining it. it's hard to say as it's unclear yet how (and whether) will Data API interact with baking. It's also possible we move to static site builder like Next.js that might cache things on its own without needing special code for it.

Should I write unit tests for untested functions like `writeCSV` that I'm changing? (There are currently none and I think it'd be nice to have at least some test coverage)

## Polars

[nodejs-polars](https://github.com/pola-rs/nodejs-polars) is a cool Rust alternative to Pandas. It has a lot of momentum in python world, unfortunately its Node version is not as mature (it is being actively developed, but has only 74 stars and one dev). Even though I had to debug a few cryptic Rust errors and its Node documentation is practically non-existent, I found it more natural to work with than manipulating objects (e.g. joins, writes to csv, etc.). In the long-term we're probably going to move all data stuff to Data API, but until then Polars could be a nice addition to grapher.


adminSiteServer/adminRouter.tsx
```diff
@@ -130,7 +131,14 @@ adminRouter.get("/datasets/:datasetId.csv", async (req, res) => {
     ).name
     res.attachment(filenamify(datasetName) + ".csv")
 
-    return Dataset.writeCSV(datasetId, res)
+    const writeStream = new Writable({
```
Marigold, line 134: Perhaps there's a better way how to work with streams that emit chunks (and need calling `.toString()`)? I've tried a few things and nothing worked (could be polars issue).

danyx23 (approved):
This looks very promising - thanks, Mojmir! All of it seems reasonable to me. I hope that the Polars library won't cause any surprising issues.

If I understand the code correctly then the main thing that changes now is the CSV baking and some intermediate storage of the data when baking variables - is that right? For the tests I think it would be great to add a few somewhat more odd examples if they can come up - e.g. can there be null values or number.NaN? Would be good to have test cases that document how those are serialized.

Other than the minor extension of the tests I think this is good to merge.

Marigold:
> I think it would be good to make sure that polars produces semantically identically results as before - around number.NaN handling or so I'm not entirely confident that there might not be some weird things

I've checked both data/*.json and metadata/*.json and they're identical (except of ordering which has changed, but only for JSON files, not for CSVs).

PR #1900, opened 2023-02-03
Extract Textwrap, MarkdownTextwrap, and DetailsOnDemand from grapher

ikesau:
This was a bit more annoying than I would have liked it to be.

`@ourworldindata/utils` doesn't already export SCSS so I took the lazy route and simply pasted the ~20 lines into grapher.scss for the meantime.

I lifted imemo out of core-table and a few other type definitions to fix tests that were breaking due to circular dependencies (found with [dpdm](https://github.com/acrazing/dpdm)).

[Staging preview](https://ptolemy-owid.netlify.app/grapher/primary-energy-cons)


marcelgerber (changes requested):
I only had a brief look over the code at this point, but just to jot this down somewhere already:
For completeness, we will want to put `parsimmon` and `@tippyjs/react` into utils' `package.json`. Maybe there are more of these, not sure.

marcelgerber (approved):
I fixed the package.json issue now.
Looks good now! It's a bummer that doing this is so so hard though.

ikesau:
Thanks @marcelgerber !

I've tested explorers and graphers (plus static exports) with and without DoDs on Safari and Firefox. I think we're good to go.

danyx23 (approved):
LGTM

PR #1909, opened 2023-02-07
Importing wpdb should not automatically create a connection

danyx23:
It used to be that just importing wpdb.ts would create a connection but this causes issues when you want to only connect if the settings are available. This commit changes it so that a function call is required to initialize the connection

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -167,7 +167,7 @@ const getReferencesByChartId = async (
     // careful not to erroneously match those, which is why we switched to a
     // REGEXP.
     try {
-        posts = await wpdb.singleton.query(
+        posts = await wpdb.getSingleton().query(
             `
                 SELECT ID, post_title, post_name
                 FROM wp_posts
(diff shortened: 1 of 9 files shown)


mlbrgl (approved):
> when you want to only connect if the settings are available

`wpdb.singleton.connect()` is never called without checking for `isWordpressAPIEnabled` (connect() is either called directly in `connectToDatabases()`, `wpdb.singleton.query()` or `wpdb.singleton.get()`, all of which are guarded by a `isWordpressAPIEnabled` check at some point in the call chain).

So it seems like the settings (or the approximation of it that we settled for) are always available when we want to connect?

What am I misunderstanding or overlooking here? Would you mind going through the use case you have in mind?

I'm just trying to clarify my understanding here, but all things being equal, this approach is more predictable.

It also makes me wonder now why this `isWordpressAPIEnabled` check is not done in `connect()` rather than relying on manually checking in every caller?

While we're here: are we still using Knex anywhere? It seems like this function (along with a few isolated declarations) are never called.
https://github.com/owid/owid-grapher/blob/53f955178b143319abb749defb81c37b28344cd3/db/wpdb.ts#L55-L74

danyx23:
... right. I realized now that I didn't read the code carefully enough. I assumed that the WPDB class instance could automatically make a connection but as you say that is actually guarded against. I think I'll retract this work now and do something more along the lines of what you describe. Thanks for raising this!

danyx23:
Closing this PR now as I don't have time to redo this and it's mostly code aesthetics anyway

PR #1913, opened 2023-02-08
nit: properly type Chart config in typeorm Chart definition

danyx23:
Just a small PR to get rid of another any type


adminSiteServer/testPageRouter.tsx
```diff
@@ -257,7 +257,7 @@ async function propsFromQueryParams(
 
     const charts: ChartItem[] = (await query.getMany()).map((c) => ({
         id: c.id,
-        slug: c.config.slug,
+        slug: c.config.slug!,
```
ikesau, line 260: Without knowing how we actually use these objects throughout the codebase, shouldn't this be `c.config.slug ?? ""` instead? Otherwise we might accidentally call a method on `undefined`.

PR #1918, opened 2023-02-09
enhance: drop obsolete tables: settings and knex_migrations

danyx23:


Marigold (approved):
I've never used these tables, but they don't seem to be used anymore in code. Better get a confirmation from @marcelgerber

marcelgerber (approved):
Yep, can confirm these are no longer in use.
For further reference:
- `knex_migrations` got introduced in https://github.com/owid/owid-grapher/commit/023e58fcffad2e0b73acde1e36fdfbec1b058212 (2019) and removed in https://github.com/owid/owid-grapher/commit/e89e597a443b96b6eabfe9ef9b23307d4498d2fd (2020)
- The last use of `settings` was removed in c42fa7c4c47839f1a8162d9b9ddecd24492a8249 (2018) - back then it was in `grapher_admin/views.py`.

PR #1920, opened 2023-02-09
Complete backport publishing workflow

danyx23:
This PR completes the workflow of converting selected wordpress posts into gdocs and then allowing publishing those. Before we did create gdocs documents but they were not registered in the posts_gdocs table and there was no way to publish them from there. This PR now adds that capability.

To test it, go to the posts table. You can view compare for all posts and then also hit "Create GDoc" (this needs the .env setting entry GDOCS_BACKPORTING_TARGET_FOLDER to be set). Once the GDoc is created, a link brings you to the preview page from where you can also go the the gdoc and edit it.

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -2291,19 +2291,27 @@ apiRouter.post("/posts/:postId/createGdoc", async (req: Request) => {
         .where({ id: postId })
         .select("*")
         .first()) as PostRow | undefined
+
     if (!post) throw new JsonError(`No post found for id ${postId}`, 404)
     if (post.gdocSuccessorId)
         throw new JsonError("A gdoc already exists for this post", 400)
     const archieMl = JSON.parse(post.archieml) as OwidArticleType
     const gdocId = await createGdocAndInsertOwidArticleContent(archieMl.content)
     post.gdocSuccessorId = gdocId
+    // This is not ideal - we are using knex for on thing and typeorm for another
+    // which means that we can't wrap this in a transaction. We should probably
+    // move posts to use typeorm as well or at least have a typeorm alternative for it
     await db
         .knexTable(postsTable)
         .where({ id: postId })
         .update("gdocSuccessorId", gdocId)
-    // TODO: fill the rest of content required for an entry in the gdoc table,
-    // mark the post in the posts table as no-longer editable/publish-able,
-    // publish through gdocs?
+    const gdoc = new Gdoc(gdocId)
+    gdoc.slug = post.slug
+    gdoc.published = false
+    gdoc.createdAt = new Date()
+    gdoc.publishedAt = post.published_at
+    await dataSource.getRepository(Gdoc).insert(gdoc)
+
     return { googleDocsId: gdocId }
 })
 
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -2286,6 +2286,7 @@ apiRouter.get("/posts/:postId.json", async (req: Request, res: Response) => {
 
 apiRouter.post("/posts/:postId/createGdoc", async (req: Request) => {
     const postId = expectInt(req.params.postId)
+    const allowRecreate = !!req.body.allowRecreate
     const post = (await db
         .knexTable(postsTable)
         .where({ id: postId })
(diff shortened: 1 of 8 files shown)


adminSiteServer/apiRouter.ts
```diff
@@ -2271,24 +2271,40 @@ apiRouter.get("/posts/:postId.json", async (req: Request, res: Response) => {
...
-        .where({ id: postId })
-        .update("gdocSuccessorId", gdocId)
-    // TODO: fill the rest of content required for an entry in the gdoc table,
-    // mark the post in the posts table as no-longer editable/publish-able,
-    // publish through gdocs?
+    const gdocId = await createGdocAndInsertOwidArticleContent(
+        archieMl.content,
+        post.gdocSuccessorId
+    )
+    if (!existingGdocId) {
```
ikesau, line 2290: What happens if there _is_ an `existingGdocId` ? Shouldn't we still be updating the DB?

db/migration/1676022628012-AddPostsWithGdocPublishStatusView.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class AddPostsWithGdocPublishStatusView1676022628012
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`-- sql
```
ikesau, line 7: ooc, what's up with the `-- sql` comments ?

db/model/Gdoc/archieToGdoc.ts
```diff
@@ -195,7 +195,8 @@ function articleToBatchUpdates(
 }
 
 export async function createGdocAndInsertOwidArticleContent(
-    content: OwidArticleContent
+    content: OwidArticleContent,
+    existingGdoc: string | undefined
```
ikesau, line 199:
```suggestion
    existingGdocId: string | undefined
```

db/model/Gdoc/archieToGdoc.ts
```diff
@@ -212,19 +213,51 @@ export async function createGdocAndInsertOwidArticleContent(
...
-            name: content.title,
-        },
-        media: {
-            mimeType: docsMimeType,
-            body: "",
-        },
-    })
-    const documentId = createResp.data.id!
+    let documentId = existingGdoc
+    if (existingGdoc) {
```
ikesau, line 217: This section had me going up and down a lot understand what was going on. It'd be nice if you could lift the branches up into description functions 🙂 

e.g.
```js
  await deleteGdocContents()
} else {
  await createNewGdoc()
}
await setGdocContents()
return documentId
```

db/wpdb.ts
```diff
@@ -110,7 +110,12 @@ class WPDB {
     }
 }
 
-export const singleton = new WPDB()
+let singleton: WPDB | undefined = undefined
+
+export function getSingleton(): WPDB {
+    if (!singleton) singleton = new WPDB()
```
ikesau, line 116: What does this change do?

ikesau (commented):
Overall looks good (though I don't envy your imminent merge conflicts!)

Will test locally soon 👍

db/migration/1676022628012-AddPostsWithGdocPublishStatusView.ts
danyx23: With [this extension](https://marketplace.visualstudio.com/items?itemName=qufiwefefwoyn.inline-sql-syntax) you get SQL syntax highlighting if you prefix strings in programming languages with `-- sql`

db/wpdb.ts
danyx23: I'll take this out again - I thought the old code initialized a db connection as a side effect of being imported but [Matthieu made me realize that this is not the case](https://github.com/owid/owid-grapher/pull/1909).

adminSiteServer/apiRouter.ts
danyx23: In this case the posts table already contains a link to this gdoc document and we only needed to regenerate the gdoc content. I've added a comment to make this more clear

ikesau:
Running locally and finding that `appendDotEndIfMultiline` can be passed `null` which causes an error when trying to read `line.includes`

Seems simple enough to fix but I'm not sure if those nulls are an important case to handle.

ikesau:
An example of one of those null cases is `1588`

image

PR #1923, opened 2023-02-10
Remove `data_values` table

Marigold:
Completely remove `data_values` from grapher.

Services still using `data_values`:

* [x] covid19
* [x] importers (code is still there, but not being used for updates as far as I know)
* [x] chart revision suggester
* [x] backporting


danyx23:
I still want this some time soon :)

marcelgerber (approved):
Real nice work, burn it 🔥 🧯

PR #1956, opened 2023-02-27
docs(covid): change source for cases and deaths

lucasrodes:
This is part of https://github.com/owid/owid-issues/issues/957.

Change references to JHU in our codebase so that the [deaths](https://ourworldindata.org/covid-deaths#global-comparison-where-are-confirmed-deaths-increasing-most-rapidly) and [cases](https://ourworldindata.org/covid-cases#global-comparison-where-are-confirmed-cases-increasing-most-rapidly) table are updated.

This should _only_ be merged once we finalize the transition JHU → WHO (i.e. finish https://github.com/owid/owid-issues/issues/957).

### DIFF
--- a/site/covid/CovidConstants.ts
+++ b/site/covid/CovidConstants.ts
@@ -2,8 +2,8 @@ import { SortOrder } from "@ourworldindata/core-table"
 import { NounKey, NounGenerator } from "./CovidTypes.js"
 import { createNoun } from "./CovidUtils.js"
 
-export const JHU_DATA_URL =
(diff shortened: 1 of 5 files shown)


lucasrodes:
We might end up removing this table from the site: https://owid.slack.com/archives/CV5RY8F1B/p1677576744567419

lucasrodes:
This table has been removed from our confirmed COVID-19 [Cases](https://ourworldindata.org/covid-cases) and [Deaths](https://ourworldindata.org/covid-deaths) articles.

Therefore, there is no need to move forward with this PR

PR #1977, opened 2023-03-06
Disallow splitting by product for charts with more than one unit

### LINKED ISSUE #1928: On Discrete Bar charts that are split by metric or product, the units are wrong when splitting by product
## Description

On charts that have the option to split by metric or by product (e.g. [this chart](https://owid.cloud/admin/charts/5525/edit)), when splitting by product, the units are wrong.

## Expected behaviour

When splitting by product, each row shown corresponds to a different metric, and hence should have a different unit.

## Steps to reproduce

1. Go to [this chart](https://owid.cloud/admin/charts/5525/edit).
2. Click on 'split by product'.
3. See that all metrics have the same units.

## Screenshots

environmental-footprint-milks

## Environment

Desktop:
-   OS: macOS Big Sur 11.6.4
-   Browser: chrome
-   Version: 109.0.5414.119 (Official Build) (arm64)

sophiamersmann:
fixes #1928

[As suggested by Marcel](https://owid.slack.com/archives/C5BDCB2R3/p1677880789986719?thread_ts=1677861957.213109&cid=C5BDCB2R3), we'll disallow splitting by product if that means we would be comparing different units along the same axis.

This PR does both:
- Fixes the particular issue in #1928 (if splitting by product were allowed)
- Disallows splitting by product when appropriate (such that the fix from above shouldn't be necessary)

### DIFF
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
@@ -78,7 +78,7 @@ export interface Label {
 }
 
 interface DiscreteBarItem {
-    yColumnSlug: string
+    yColumn: CoreColumn
     seriesName: string
     value: number
     time: number
@@ -417,15 +417,11 @@ export class DiscreteBarChart
                 {this.hasColorLegend && (
                     <HorizontalNumericColorLegend manager={this} />
                 )}
-                {/* only show the axis if all columns share a unit */}
-                {new Set(this.yColumns.map((column) => column.unit)).size ===
-                    1 && (
-                    <HorizontalAxisComponent
-                        bounds={boundsWithoutColorLegend}
-                        axis={yAxis}
-                        preferredAxisPosition={innerBounds.bottom}
-                    />
-                )}
+                <HorizontalAxisComponent
+                    bounds={boundsWithoutColorLegend}
+                    axis={yAxis}
+                    preferredAxisPosition={innerBounds.bottom}
+                />
                 <HorizontalAxisGridLines
                     horizontalAxis={yAxis}
                     bounds={innerBounds}
@@ -524,23 +520,19 @@ export class DiscreteBarChart
     }
 
     formatValue(series: DiscreteBarSeries): Label {
-        let column = this.yColumns.find(
-            (yCol) => yCol.slug === series.yColumnSlug
-        )
-        if (column == undefined) column = this.yColumns[0]
-
         const { transformedTable } = this
+        const { yColumn } = series
 
         const showYearLabels =
             this.manager.showYearLabels || series.time !== this.targetTime
-        const displayValue = column.formatValueShort(series.value)
+        const displayValue = yColumn.formatValueShort(series.value)
         const { timeColumn } = transformedTable
         const preposition = OwidTable.getPreposition(timeColumn)
 
         return {
             valueString: displayValue,
             timeString: showYearLabels
-                ? ` ${preposition} ${column.formatTime(series.time)}`
+                ? ` ${preposition} ${yColumn.formatTime(series.time)}`
                 : "",
         }
     }
@@ -598,7 +590,7 @@ export class DiscreteBarChart
                       entityNames[index] as string
                   )
             return {
-                yColumnSlug: col.slug,
+                yColumn: col,
                 seriesName,
                 value: values[index] as number,
                 time: originalTimes[index] as number,
@@ -771,10 +763,10 @@ export class DiscreteBarChart
 
     @computed get series(): DiscreteBarSeries[] {
         const series = this.sortedRawSeries.map((rawSeries) => {
-            const { value, time, colorValue, seriesName, color, yColumnSlug } =
+            const { value, time, colorValue, seriesName, color, yColumn } =
                 rawSeries
             const series: DiscreteBarSeries = {
-                yColumnSlug,
+                yColumn,
                 value,
                 time,
                 colorValue,
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
@@ -352,9 +352,20 @@ export class DiscreteBarChart
         if (
             this.yColumns.length > 1 &&
             this.selectionArray.numSelectedEntities > 1
-        )
+        ) {
+            // if we have more than one unit, then faceting by entity is not allowed
+            // as comparing multiple units in a single chart isn't meaningful
+            const uniqueUnits = new Set(
+                this.yColumns.map((column) => column.unit)
+            )
+            if (uniqueUnits.size > 1) {
+                return [FacetStrategy.metric]
+            }
+
             return [FacetStrategy.entity, FacetStrategy.metric]
-        else return [FacetStrategy.none]
+        }
+
+        return [FacetStrategy.none]
     }
 
     componentDidMount(): void {
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
@@ -356,7 +356,7 @@ export class DiscreteBarChart
             // if we have more than one unit, then faceting by entity is not allowed
             // as comparing multiple units in a single chart isn't meaningful
             const uniqueUnits = new Set(
-                this.yColumns.map((column) => column.unit)
+                this.yColumns.map((column) => column.shortUnit)
             )
             if (uniqueUnits.size > 1) {
                 return [FacetStrategy.metric]
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
@@ -78,6 +78,7 @@ export interface Label {
 }
 
 interface DiscreteBarItem {
+    yColumnSlug: string
     seriesName: string
     value: number
     time: number
@@ -405,11 +406,15 @@ export class DiscreteBarChart
                 {this.hasColorLegend && (
                     <HorizontalNumericColorLegend manager={this} />
                 )}
-                <HorizontalAxisComponent
-                    bounds={boundsWithoutColorLegend}
-                    axis={yAxis}
-                    preferredAxisPosition={innerBounds.bottom}
-                />
+                {/* only show the axis if all columns share a unit */}
+                {new Set(this.yColumns.map((column) => column.unit)).size ===
+                    1 && (
+                    <HorizontalAxisComponent
+                        bounds={boundsWithoutColorLegend}
+                        axis={yAxis}
+                        preferredAxisPosition={innerBounds.bottom}
+                    />
+                )}
                 <HorizontalAxisGridLines
                     horizontalAxis={yAxis}
                     bounds={innerBounds}
@@ -508,7 +513,11 @@ export class DiscreteBarChart
     }
 
     formatValue(series: DiscreteBarSeries): Label {
-        const column = this.yColumns[0] // todo: do we need to use the right column here?
+        let column = this.yColumns.find(
+            (yCol) => yCol.slug === series.yColumnSlug
+        )
+        if (column == undefined) column = this.yColumns[0]
+
         const { transformedTable } = this
 
         const showYearLabels =
@@ -578,6 +587,7 @@ export class DiscreteBarChart
                       entityNames[index] as string
                   )
             return {
+                yColumnSlug: col.slug,
                 seriesName,
                 value: values[index] as number,
                 time: originalTimes[index] as number,
@@ -750,8 +760,10 @@ export class DiscreteBarChart
 
     @computed get series(): DiscreteBarSeries[] {
         const series = this.sortedRawSeries.map((rawSeries) => {
-            const { value, time, colorValue, seriesName, color } = rawSeries
+            const { value, time, colorValue, seriesName, color, yColumnSlug } =
+                rawSeries
             const series: DiscreteBarSeries = {
+                yColumnSlug,
                 value,
                 time,
                 colorValue,
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
@@ -1,9 +1,9 @@
 import { ChartManager } from "../chart/ChartManager"
-import { CoreValueType, Time } from "@ourworldindata/core-table"
+import { CoreColumn, CoreValueType, Time } from "@ourworldindata/core-table"
 import { ChartSeries } from "../chart/ChartInterface"
 
 export interface DiscreteBarSeries extends ChartSeries {
-    yColumnSlug: string
+    yColumn: CoreColumn
     value: number
     time: Time
     colorValue?: CoreValueType
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
@@ -3,6 +3,7 @@ import { CoreValueType, Time } from "@ourworldindata/core-table"
 import { ChartSeries } from "../chart/ChartInterface"
 
 export interface DiscreteBarSeries extends ChartSeries {
+    yColumnSlug: string
     value: number
     time: Time
     colorValue?: CoreValueType
--- a/packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
+++ b/packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
@@ -310,7 +310,9 @@ export class AbstractStackedChart
         const hasMultipleEntities =
             this.selectionArray.selectedEntityNames.length > 1
         const hasMultipleYColumns = this.yColumns.length > 1
-        const uniqueUnits = new Set(this.yColumns.map((column) => column.unit))
+        const uniqueUnits = new Set(
+            this.yColumns.map((column) => column.shortUnit)
+        )
         const hasMultipleUnits = uniqueUnits.size > 1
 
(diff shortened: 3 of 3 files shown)


marcelgerber:
Good work!

This is a case where a perfectly great solution hits our imperfect chart configurations: We have lots and lots of charts where the units are inconsistently set, or only the first column is fully given, etc.
You can see this already in the diff produced by the svg tester (https://github.com/owid/owid-grapher-svgs/commit/a3a595205a0962feeff2465c537896f75d61620d), which automatically runs any change against all our charts to see what changed: many charts lose their axis even though the units are perfectly comparable, but the chart hasn't been configured entirely correctly.

[I even composed an absurdly complex query in Datasette](https://owid-datasette-y43qr.ondigitalocean.app/owid?sql=with+dims+as+%28%0D%0A++select%0D%0A++++c.id%2C%0D%0A++++slug%2C%0D%0A++++ifnull%28json_extract%28config%2C+%22%24.type%22%29%2C+%22LineChart%22%29+as+type%2C%0D%0A++++value+as+dim%0D%0A++from%0D%0A++++charts+c%2C%0D%0A++++json_each%28c.config%2C+%22%24.dimensions%22%29%0D%0A++where%0D%0A++++json_array_length%28config%2C+%22%24.dimensions%22%29+%3E+1+--+not+absolutely+necessary%2C+but+speeds+up+the+query+by+a+lot%21%0D%0A%29%0D%0Aselect%0D%0A++id%2C%0D%0A++slug%2C%0D%0A++type%2C%0D%0A++count%28id%29+as+numDimensions%2C%0D%0A++json_group_array%28%0D%0A++++distinct+ifnull%28json_extract%28dim%2C+%22%24.display.unit%22%29%2C+%22NULL%22%29%0D%0A++%29+as+units%2C%0D%0A++json_group_array%28%0D%0A++++distinct+ifnull%28json_extract%28dim%2C+%22%24.display.shortUnit%22%29%2C+%22NULL%22%29%0D%0A++%29+as+shortUnits%0D%0Afrom%0D%0A++dims%0D%0Awhere%0D%0A++type+%21%3D+%22ScatterPlot%22%0D%0Agroup+by%0D%0A++id%0D%0Ahaving%0D%0A++count%28%0D%0A++++distinct+ifnull%28json_extract%28dim%2C+%22%24.display.unit%22%29%2C+%22NULL%22%29%0D%0A++%29+%3E+1)[^1] that surfaces all the charts that have more than one unique unit, and it's quite a large number. (as expected, most of them are only multi-unit because of `null` or `""`, though)

Given this situation, the PR is not ready to be merged as-is, since it's effects would be net-negative at the moment.
In my mind we have the following ways forward:
1. Remove the axis-hiding code. This is in my mind the worst offender, since a chart without axis is way less useful already (and, currently, we _never_ hide the axis).
  Since the `availableFacetStrategies` logic change gets rid of almost all scenarios where we would show a wrong axis anyway, this should be fine. I would rather have a very small number of wrong axes than a somewhat large number of charts missing their axis.
2. Clean up the chart configs. This is probably what we should do, but it's also daunting and will take a long time.
3. Add some heuristics. And by heuristics, I basically mean "ignoring `null` and `""`". This should go a long way already, if you want to keep the axis-hiding code in.

I prefer (1), but feel free to discuss if you think differently. I'm also curious to hear what @danyx23 thinks.


[^1]: The query doesn't take into account inherited units from variables, so take it with a grain of salt.

packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
```diff
@@ -405,11 +417,15 @@ export class DiscreteBarChart
...
                     <HorizontalNumericColorLegend manager={this} />
                 )}
-                <HorizontalAxisComponent
-                    bounds={boundsWithoutColorLegend}
-                    axis={yAxis}
-                    preferredAxisPosition={innerBounds.bottom}
-                />
+                {/* only show the axis if all columns share a unit */}
+                {new Set(this.yColumns.map((column) => column.unit)).size ===
+                    1 && (
```
marcelgerber, line 422: I feel like this is too complex to be inline; you could instead extract it to `uniqueColumnUnits` or a similar `@computed` property.

packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
```diff
@@ -405,11 +417,15 @@ export class DiscreteBarChart
                 {this.hasColorLegend && (
                     <HorizontalNumericColorLegend manager={this} />
                 )}
-                <HorizontalAxisComponent
-                    bounds={boundsWithoutColorLegend}
-                    axis={yAxis}
-                    preferredAxisPosition={innerBounds.bottom}
-                />
+                {/* only show the axis if all columns share a unit */}
+                {new Set(this.yColumns.map((column) => column.unit)).size ===
```
marcelgerber, line 421: In addition, I think we should use `shortUnit` here: It's what we display on the axis.

```suggestion
                {new Set(this.yColumns.map((column) => column.shortUnit)).size ===
```

### Reasoning
I can imagine cases where long units are verbose like `% of males` and `% of females`, and the corresponding short unit is `%` for both.
In this case, they are of course comparable.

We have a similar situation in [this chart](https://ourworldindata.org/grapher/births-and-deaths-projected-to-2100?time=earliest), where the long units are `births` and `deaths`, but the short unit is empty. This chart definitely should show an axis.

marcelgerber (changes requested):
Great work overall!
Please take these review comments with a grain of salt: As I suggested in my above comment, the behavior should be reworked in some way, and depending on which way you choose not all review comments will apply.

sophiamersmann:
Thanks for the thorough review, Marcel! This uncovered lots of unknown ground for me, so that's great! (Also, I am a huge fan of Datasette, so fun that you are maintaining an OWID instance (and thank you for coming up with that query 😄))

I agree with your thinking. If we disallow splitting by group on the abstract bar chart, then this would get rid of many of the offending charts anyway and I agree that having few charts with a sub-optimal axis is better than many without any!

If Daniel agrees, I would then:
- Move the logic that disallows splitting by group for mulit-unit charts to the abstract class
- Remove the clause that hides the axis
- Keep the code that chooses the correct `yColumn` bus pass the actual column around, not the slug

Not strictly necessary to to work on this PR, but I was wondering:
I am not sure what option (2) clean up chart configs would look like. I can see that the chart configs have a field `dimensions` with a list of variables and each variable has a field `display` with the variable's name and unit, and this information is later written to the metadata file. If we were to clean up the chart configs, what would be the appropriate place to update this information? Would we attempt to update the unit/shortUnit column in the `variables` table or overwrite this information elsewhere?

danyx23:
Nice review, @marcelgerber, thank you! (Nice query also :) ). I agree with what you outline - I think removing the axis hiding code makes sense and then let's see what the svg tester tells us. We should then manually check a few of the facetted charts - the SVGTester only tests the default view of standalone charts - i.e. it doesn't check what charts look like after interaction and it doesn't take into account grapher views that only come up in explorers.

The plan you outline looks good, @sophiamersmann. To your question about cleaning up the chart configs - I'd have to check this in code but I'm pretty sure that most of the display settings come from the variables (although you can override them in the admin). Fixing this looks like a very manual process as you'd have to look at the chart, change the config (often ideally in the ETL) and then see if the on-hover labels or whatever else uses the short unit still look good.

I think what we should do is bubble this up to the data team that we are using short units now to check if variables can be co-plotted - this might help them decide how to phrase short units.

danyx23:
Oh and the string "null" values are really not nice - maybe we should do a migration to remove those. Thoughts, @marcelgerber?

marcelgerber:
> Oh and the string "null" values are really not nice - maybe we should do a migration to remove those. Thoughts, @marcelgerber?

Ah, that's a red herring. They are really proper `null` values, but I included an `ifnull(unit, "NULL")` clause in the query, because otherwise sqlite excludes NULL values from `json_group_array`, and we care about them in this case.

marcelgerber:
> but I didn't dare include that complication in the query ;)

okay, I did dare just now!
This is now the more accurate query, which takes variable-level inheritance into account: [Super complicated query](https://owid-datasette-y43qr.ondigitalocean.app/owid?sql=with+dims+as+%28%0D%0A++select%0D%0A++++c.id%2C%0D%0A++++slug%2C%0D%0A++++ifnull%28json_extract%28config%2C+%22%24.type%22%29%2C+%22LineChart%22%29+as+type%2C%0D%0A++++value+as+dim%2C%0D%0A++++coalesce%28%0D%0A++++++json_extract%28value%2C+%22%24.display.unit%22%29%2C%0D%0A++++++json_extract%28display%2C+%22%24.unit%22%29%0D%0A++++%29+as+unit%2C%0D%0A++++coalesce%28%0D%0A++++++json_extract%28value%2C+%22%24.display.shortUnit%22%29%2C%0D%0A++++++json_extract%28display%2C+%22%24.shortUnit%22%29%0D%0A++++%29+as+shortUnit%0D%0A++from%0D%0A++++charts+c%2C%0D%0A++++json_each%28c.config%2C+%22%24.dimensions%22%29%0D%0A++++join+variables+v+on+json_extract%28value%2C+%22%24.variableId%22%29+%3D+v.id%0D%0A++where%0D%0A++++json_array_length%28config%2C+%22%24.dimensions%22%29+%3E+1+--+not+absolutely+necessary%2C+but+speeds+up+the+query+by+a+lot%21%0D%0A%29%0D%0Aselect%0D%0A++id%2C%0D%0A++slug%2C%0D%0A++type%2C%0D%0A++count%28id%29+as+numDimensions%2C%0D%0A++json_group_array%28distinct+ifnull%28unit%2C+%22NULL%22%29%29+as+units%2C%0D%0A++json_group_array%28distinct+ifnull%28shortUnit%2C+%22NULL%22%29%29+as+shortUnits%0D%0Afrom%0D%0A++dims%0D%0Awhere%0D%0A++type+%21%3D+%22ScatterPlot%22%0D%0Agroup+by%0D%0A++id%0D%0Ahaving%0D%0A++count%28distinct+ifnull%28unit%2C+%22NULL%22%29%29+%3E+1)

packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
```diff
@@ -351,9 +352,20 @@ export class DiscreteBarChart
         if (
             this.yColumns.length > 1 &&
             this.selectionArray.numSelectedEntities > 1
-        )
+        ) {
+            // if we have more than one unit, then faceting by entity is not allowed
+            // as comparing multiple units in a single chart isn't meaningful
+            const uniqueUnits = new Set(
+                this.yColumns.map((column) => column.unit)
```
marcelgerber, line 359: I'm not entirely sure about it, but I think we should also use `shortUnit` here (and in AbstractStackedChart).
The reason being similar to what I said elsewhere: [A chart like this](https://ourworldindata.org/grapher/births-and-deaths-projected-to-2100?time=earliest&facet=entity&uniformYAxis=0&country=OWID_WRL~DEU~NAM~BOL) still makes sense, but the long units are "births" and "deaths".

sophiamersmann: Makes perfect sense!

sophiamersmann:
I just checked each failing chart (we have 4) to see what's going on and I think it's mostly small issues around how units are specified (or not specified) that we can fix upstream. That being said, I'm not really sure what that would entail – so second (and third) opinions are welcome!

I'll go through them one by one:

**Failing chart 1 (diff):** https://github.com/owid/owid-grapher-svgs/commit/0aae50f18f0036005ecdde9fc04bce577f4cc43b#diff-7aaad35dfa37b9d119c108d19731d4e2643ed55d506808e161d62d397ea26d5a

- The `display` field in the metadata json is empty for the second variable labelled "Terrorism"

**Failing chart 2 (diff):** https://github.com/owid/owid-grapher-svgs/commit/0aae50f18f0036005ecdde9fc04bce577f4cc43b#diff-513c7721bb57620b9dddc8f73e46edf17af710fa0d0846b05f4d1dc04e3d9b6f

- The variable labelled "Outdoor particulate matter polution" shares a unit wir all other charts ("DALYs") but additionally has a short unit ("LDL") that the other variables don't have (I'm not sure what LDL means in this context)
- Also note: The label on "Zinc deficiency" is rounded differently (I don't think that's a problem but I also don't understand why that would happen)

**Failing chart 3 (diff):** https://github.com/owid/owid-grapher-svgs/commit/0aae50f18f0036005ecdde9fc04bce577f4cc43b#diff-3a5c9026e9387aa9ef6856422b7dd9c4b5132a161a478778c14091a65a4f9b22

This one I don't really get:
- All three variables have the same `unit`, `shortUnit` and `conversionFactor` in their `display` field in the metadata, so not sure why one of them ("Forest") would be displayed differently
- In the config file though, the variable "Forest" is listed without unit and short unit, so that might have something to do with this?

**Failing chart 4 (diff):** https://github.com/owid/owid-grapher-svgs/commit/0aae50f18f0036005ecdde9fc04bce577f4cc43b#diff-62cd829ae7956aaf8e815803cc70804949e50f95a81582f2ad5c13d66baa2cbd

- Another small rounding issue that I don't think is a big deal, but I also don't understand why that would happen.

marcelgerber:
Good work figuring all of these out!

> * Also note: The label on "Zinc deficiency" is rounded differently (I don't think that's a problem but I also don't understand why that would happen)

This one I can explain: "Number of decimal points" is a per-dimension setting, and because we switched from always using the first dimension ( = column) for number formatting to using the correct one, this changed.

image



> * In the config file though, the variable "Forest" is listed without unit and short unit, so that might have something to do with this?

Yeah exactly - the two other dimensions override the variable-level config, but "Forests" doesn't have a unit set and inherits the `shortUnit` of `1000 ha` from the variable.

marcelgerber (approved):
Overall this is now good code-wise!

Before we can merge this, though, we will have to fix a few charts. (Sorry that this is taking so many steps before we can ship it - that's not something we realized beforehand!)

We will definitely want to fix the 4 charts that the SVG tester brought up, directly in the Admin. If you want I can show you through this tomorrow afternoon.

Then, we have a bunch more charts that are also partially affected: You will maybe know already that all our line charts can be "collapsed" to a discrete bar chart, and there are many of these which will now look a bit weird in two ways when collapsed:
- They will not be able to be faceted by entity, which is not much of a problem, since we don't often expose faceting anyhow.
- They will show weird and inconsistent units, which is not a super-big problem but we should still fix it at some point.

In particular, [this slightly modified query](https://owid-datasette-y43qr.ondigitalocean.app/owid?sql=with+dims+as+%28%0D%0A++select%0D%0A++++c.id%2C%0D%0A++++slug%2C%0D%0A++++ifnull%28json_extract%28config%2C+%22%24.type%22%29%2C+%22LineChart%22%29+as+type%2C%0D%0A++++value+as+dim%2C%0D%0A++++coalesce%28%0D%0A++++++json_extract%28value%2C+%22%24.display.unit%22%29%2C%0D%0A++++++json_extract%28display%2C+%22%24.unit%22%29%0D%0A++++%29+as+unit%2C%0D%0A++++coalesce%28%0D%0A++++++json_extract%28value%2C+%22%24.display.shortUnit%22%29%2C%0D%0A++++++json_extract%28display%2C+%22%24.shortUnit%22%29%0D%0A++++%29+as+shortUnit%0D%0A++from%0D%0A++++charts+c%2C%0D%0A++++json_each%28c.config%2C+%22%24.dimensions%22%29%0D%0A++++join+variables+v+on+json_extract%28value%2C+%22%24.variableId%22%29+%3D+v.id%0D%0A++where%0D%0A++++json_array_length%28config%2C+%22%24.dimensions%22%29+%3E+1+--+not+absolutely+necessary%2C+but+speeds+up+the+query+by+a+lot%21%0D%0A%29%0D%0Aselect%0D%0A++id%2C%0D%0A++slug%2C%0D%0A++type%2C%0D%0A++count%28id%29+as+numDimensions%2C%0D%0A++json_group_array%28distinct+ifnull%28unit%2C+%22NULL%22%29%29+as+units%2C%0D%0A++json_group_array%28distinct+ifnull%28shortUnit%2C+%22NULL%22%29%29+as+shortUnits%0D%0Afrom%0D%0A++dims%0D%0Awhere%0D%0A++type+in+%28%22LineChart%22%2C+%22DiscreteBar%22%29%0D%0Agroup+by%0D%0A++id%0D%0Ahaving%0D%0A++count%28distinct+ifnull%28shortUnit%2C+%22NULL%22%29%29+%3E+1) surfaces the line/bar charts that are affected by this change.

So, as next steps I suggest:
1. Fixing the 4 svg tester charts in the admin.
2. Looking at the above query and finding the 5-or-so most popular charts of these (by pageviews; I can hopefully do that tomorrow) and fix them for good measure.
3. File a cleanup issue in https://github.com/owid/owid-issues to get the rest of them looked at and cleaned up.

marcelgerber:
> 2\. Looking at the above query and finding the 5-or-so most popular charts of these (by pageviews; I can hopefully do that tomorrow) and fix them for good measure.

Okay did that now ;)

These are the few charts that have >400 views per 14 days:
- [ ] https://ourworldindata.org/grapher/primary-sub-energy-source
- [ ] https://ourworldindata.org/grapher/deaths-globally-by-age
- [ ] https://ourworldindata.org/grapher/co2-emissions-and-gdp
- [ ] https://ourworldindata.org/grapher/environmental-footprint-milks
- [ ] https://ourworldindata.org/grapher/voter-turnout-rate-by-age-usa
- [ ] https://ourworldindata.org/grapher/share-of-deaths-by-cause
- [ ] https://ourworldindata.org/grapher/index-of-cereal-production-yield-and-land-use
- [ ] https://ourworldindata.org/grapher/life-expectation-at-birth-by-sex

sophiamersmann:
Sounds good! I think I figured out how to update units in the admin but I'd be more comfortable if you could show me around before I start updating these charts. I'll drop you a message in Slack :)

Kind of related questions: Which charts are tested automatically – all charts in their default configuration? Or a subset of charts? Or... something else?

sophiamersmann:
I've checked the four failing charts and the charts you listed above. Some required me to update the short unit, others were fine as is. I have also deployed this branch to [kelley](https://kelley.owid.cloud/).

marcelgerber:
Great work! In that case, everything's ready for merge! (and sorry again that this was so much work)

PR #1982, opened 2023-03-08
feat: add an unlink button - this keeps the gdoc but severs the link to posts/gdocs

danyx23:
This was requested by Joe.  It happens that authors are working on converting an entry that is not supposed to be converted as one entity. In those cases it is useful if, after creating a gdoc and copying some fragments, the link to the doc can be cut again so as not to confuse
yourself or others in the future.

### DIFF
--- a/adminSiteClient/PostsIndexPage.tsx
+++ b/adminSiteClient/PostsIndexPage.tsx
@@ -140,7 +140,7 @@ class PostRow extends React.Component<PostRowProps> {
                         </>
                     </Link>
                     <button
-                        onClick={async () => await this.onRecreateGdoc()}
+                        onClick={this.onRecreateGdoc}
                         className="btn btn-primary alert-danger"
                     >
                         <FontAwesomeIcon
@@ -151,7 +151,7 @@ class PostRow extends React.Component<PostRowProps> {
                         />
                     </button>
                     <button
-                        onClick={async () => await this.onUnlinkGdoc()}
+                        onClick={this.onUnlinkGdoc}
                         className="btn btn-primary alert-danger"
                     >
                         <FontAwesomeIcon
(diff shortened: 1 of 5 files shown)


ikesau:
Just making sure I understand: 
1. An author is writing an article in Archie 
2. There's a WordPress article with some useful content in it
3. They convert that WP article to Archie so that they can copy the section they need
4. They unlink the converted copy because they know a) isn't going to be used anymore, or b) will be properly converted later

danyx23:
@ikesau yes pretty much. The flow can also be: 
* The author converts 10 posts of a topic at once
* They start working on some
* They realize only 3 of them are really making progress
* They want to signal to themselves or others that the other 7 have not been worked on
* They delete the 7 gdocs without changes for now

Stuff like this. I discussed a couple of alternatives with Joe but this one seemed like the most flexible and least confusing

danyx23:
@ikesau I just realized that there was a merge conflict with base, force pushed an update now to resovle that (just FYI)

adminSiteClient/PostsIndexPage.tsx
```diff
@@ -112,13 +135,29 @@ class PostRow extends React.Component<PostRowProps> {
...
-                        Recreate
+                        <FontAwesomeIcon
+                            icon={faRecycle}
+                            title={
+                                "Recreate the Gdoc, replacing existing content"
+                            }
+                        />
+                    </button>
+                    <button
+                        onClick={async () => await this.onUnlinkGdoc()}
```
ikesau, line 154: `onClick={this.onUnlinkGdoc}` is fine here. TS only complains if the method takes an argument 

I don't mind either way, just pointing it out in case you didn't know 🙂

ikesau (approved):
tested locally, LGTM!

I'll have to factor this into the linking PR once it's merged, shouldn't be too hard.

I'm very much looking forward to the eventual removal of all this stuff again once we've finished the migration 😂

adminSiteClient/PostsIndexPage.tsx
danyx23: Yeah it reads a bit nicer :)

PR #1999, opened 2023-03-14
feat(admin): split "Hide automatic time/entity" toggle into multiple controls

### LINKED ISSUE #1094: Split "Hide automatic time/entity" toggle into multiple controls
Parent Notion issue: https://www.notion.so/owid/Split-Hide-automatic-time-entity-toggle-into-multiple-ones-a3056259ca2248ad865a01a55af12807

## Core problem

We have a single admin control for a chart which controls how the chart title gets modified in multiple ways. This is not clear enough for authors, and not enough control for situations where you want some changes but not others.

automatic-time

This tiny toggle here - "Hide automatic time/entity" - [really controls three different aspects](https://github.com/owid/owid-grapher/blob/5a4400afae21f9342a8a14f93fc52cde9ab668f9/grapher/core/Grapher.tsx#L1120-L1148) of the chart title:

- Automatically append the shown entity name - e.g. "Greenhouse gas emissions, France"
- Automatically append the shown time (range) - e.g. "Greenhouse gas emissions, 1990 to 2021"
- Automatically prepend "Change in" for line charts in relative mode - e.g. "Change in greenhouse gas emissions"

## Proposed solution

Having the ability to disable the automatic annotations can be helpful when our heuristics fail or when *someone* is abusing the grapher's capabilities.

However, that's all-or-nothing at the moment.

Instead, we should split it up into *at least*:

- Hide automatic time
    - see also ~related issue: [Title of svg/png export doesn't include date annotation, but interactive chart does](https://www.notion.so/Title-of-svg-png-export-doesn-t-include-date-annotation-but-interactive-chart-does-89bce5655f3a4ae5a66083e9aae88a6f)
    - A chart where we would definitely want the time annotation disabled: [https://ourworldindata.org/grapher/coal-phase-out-timeline](https://ourworldindata.org/grapher/coal-phase-out-timeline)
- Hide automatic entity
    - There are definitely some cases where our current code would show a nonsensical entity annotation, e.g. here: [https://ourworldindata.org/grapher/exports-from-western-europe-to-other-regions](https://ourworldindata.org/grapher/exports-from-western-europe-to-other-regions) (would be: "Western European exports by region of destination, Western Europe")

## Alternatives

If we want to take this further, we could also consider making these three-way options. Basically:

- Show automatic time: Always / Automatic / Never
- Show automatic entity: Always / Automatic / Never

sophiamersmann:
fixes #1094 

Splits the "Hide automatic time/entity" toggle into multiple controls, including:
- "Hide automatic entity (where possible)"
- "Hide automatic time (where possible)"
- "Don't prepend 'Change in' automatically (where possible)"

To do this, I have replaced `hideTitleAnnotation` with an object that holds three boolean properties, `entity`, `time` and `change`. I haven't changed the logic at all though. I've looked at some example charts and I couldn't find one where I thought updating the logic might be beneficial (doesn't mean these cases don't exist though).

Questionable:
- Wording
- Down migration (at the moment, if _any_ flag in the `hideTitleAnnotation` object is set to true, `hideTitleAnnotation` will be set to true)
- I created a new config schema and wrote a migration, but it's very well possible that I missed something here...

Open questions:
- The three toggles are pretty close to the URL field, which is a bit weird as they really belong to the Title text field. I didn't want to mess around with global css.. maybe we could simply move the toggles above the Title field? Or something else could be done? I am also happy to let this go if you don't think it matters.
- ~Do explorers need to be updated accordingly?~ I have updated the explorers in such a way that if someone sets `hideTitleAnnotation` to `true` in an explorer, all three fields (entity/time/change) will be set to true.

### DIFF
--- a/adminSiteClient/EditorTextTab.tsx
+++ b/adminSiteClient/EditorTextTab.tsx
@@ -57,20 +57,20 @@ export class EditorTextTab extends React.Component<{ editor: ChartEditor }> {
 
     @action.bound onToggleTitleAnnotationEntity(value: boolean) {
         const { grapher } = this.props.editor
-        grapher.hideTitleAnnotations = grapher.hideTitleAnnotations || {}
-        grapher.hideTitleAnnotations.entity = value || undefined
+        grapher.hideTitleAnnotation = grapher.hideTitleAnnotation || {}
+        grapher.hideTitleAnnotation.entity = value || undefined
     }
 
     @action.bound onToggleTitleAnnotationTime(value: boolean) {
         const { grapher } = this.props.editor
-        grapher.hideTitleAnnotations = grapher.hideTitleAnnotations || {}
-        grapher.hideTitleAnnotations.time = value || undefined
+        grapher.hideTitleAnnotation = grapher.hideTitleAnnotation || {}
+        grapher.hideTitleAnnotation.time = value || undefined
     }
 
     @action.bound onToggleTitleAnnotationChange(value: boolean) {
         const { grapher } = this.props.editor
-        grapher.hideTitleAnnotations = grapher.hideTitleAnnotations || {}
-        grapher.hideTitleAnnotations.change = value || undefined
+        grapher.hideTitleAnnotation = grapher.hideTitleAnnotation || {}
+        grapher.hideTitleAnnotation.change = value || undefined
     }
 
     @computed get errorMessages() {
@@ -107,17 +107,17 @@ export class EditorTextTab extends React.Component<{ editor: ChartEditor }> {
                     />
                     <Toggle
                         label="Hide automatic entity (where possible)"
-                        value={!!grapher.hideTitleAnnotations?.entity}
+                        value={!!grapher.hideTitleAnnotation?.entity}
                         onValue={this.onToggleTitleAnnotationEntity}
                     />
                     <Toggle
                         label="Hide automatic time (where possible)"
-                        value={!!grapher.hideTitleAnnotations?.time}
+                        value={!!grapher.hideTitleAnnotation?.time}
                         onValue={this.onToggleTitleAnnotationTime}
                     />
                     <Toggle
                         label="Don't prepend 'Change in' automatically (where possible)"
-                        value={!!grapher.hideTitleAnnotations?.change}
+                        value={!!grapher.hideTitleAnnotation?.change}
                         onValue={this.onToggleTitleAnnotationChange}
                     />
                     <AutoTextField
--- a/adminSiteClient/EditorTextTab.tsx
+++ b/adminSiteClient/EditorTextTab.tsx
@@ -55,6 +55,24 @@ export class EditorTextTab extends React.Component<{ editor: ChartEditor }> {
         grapher.relatedQuestions.splice(idx, 1)
     }
 
+    @action.bound onToggleTitleAnnotationEntity(value: boolean) {
+        const { grapher } = this.props.editor
+        grapher.hideTitleAnnotations = grapher.hideTitleAnnotations || {}
+        grapher.hideTitleAnnotations.entity = value || undefined
+    }
+
+    @action.bound onToggleTitleAnnotationTime(value: boolean) {
+        const { grapher } = this.props.editor
+        grapher.hideTitleAnnotations = grapher.hideTitleAnnotations || {}
+        grapher.hideTitleAnnotations.time = value || undefined
+    }
+
+    @action.bound onToggleTitleAnnotationChange(value: boolean) {
+        const { grapher } = this.props.editor
+        grapher.hideTitleAnnotations = grapher.hideTitleAnnotations || {}
+        grapher.hideTitleAnnotations.change = value || undefined
+    }
+
     @computed get errorMessages() {
         const { invalidDetailReferences } = this.props.editor.manager
         const keys = getIndexableKeys(invalidDetailReferences)
@@ -88,13 +106,19 @@ export class EditorTextTab extends React.Component<{ editor: ChartEditor }> {
                         softCharacterLimit={100}
                     />
                     <Toggle
-                        label="Hide automatic time/entity"
-                        value={!!grapher.hideTitleAnnotation}
-                        onValue={action(
-                            (value: boolean) =>
-                                (grapher.hideTitleAnnotation =
-                                    value || undefined)
-                        )}
+                        label="Hide automatic entity (where possible)"
+                        value={!!grapher.hideTitleAnnotations?.entity}
+                        onValue={this.onToggleTitleAnnotationEntity}
+                    />
+                    <Toggle
+                        label="Hide automatic time (where possible)"
+                        value={!!grapher.hideTitleAnnotations?.time}
+                        onValue={this.onToggleTitleAnnotationTime}
+                    />
+                    <Toggle
+                        label="Don't prepend 'Change in' automatically (where possible)"
+                        value={!!grapher.hideTitleAnnotations?.change}
+                        onValue={this.onToggleTitleAnnotationChange}
                     />
                     <AutoTextField
                         label="/grapher"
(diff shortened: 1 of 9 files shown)


sophiamersmann:
The automatic SVG tester failed _all_ charts. I expected those charts to fail that use the `hideTitleAnnotation` field as that option has been removed in the code but I'm a bit surprised that all other charts fail as as well. I had a look at the diffs and it appears that the font and background colour has been updated. Isn't that something you recently touched, @marcelgerber?

sophiamersmann:
The SVG tester fails for 900+ charts because the config schema has been updated in this branch. Not sure what best to do in this situation but for the moment I updated the configs in [owid/owid-grapher-svgs](https://github.com/owid/owid-grapher-svgs) locally to match the new schema and then ran the SVG testing on my machine – none of the charts failed :)

db/migration/1678783599815-SplitHideTitleAnnotationIntoMultipleFlags.ts
```diff
@@ -0,0 +1,25 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class SplitHideTitleAnnotationIntoMultipleFlags1678783599815
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        queryRunner.query(`
+            UPDATE charts
+            SET config = JSON_REPLACE(config, "$.hideTitleAnnotation", JSON_OBJECT("entity", TRUE, "time", TRUE, "change", TRUE))
+            WHERE config->"$.hideTitleAnnotation" IS true
```
marcelgerber, line 10: sadly, `json_extract(...) IS TRUE` doesn't work as expected in mysql 8; it also pull up `false` values (see https://github.com/owid/owid-grapher/pull/1268 and the mysql issue linked there)

the less intuitive but more correct way is:
```suggestion
            WHERE config->>"$.hideTitleAnnotation" = "true"
```

please also make sure to change it below ⬇️

adminSiteClient/EditorTextTab.tsx
```diff
@@ -55,6 +55,24 @@ export class EditorTextTab extends React.Component<{ editor: ChartEditor }> {
         grapher.relatedQuestions.splice(idx, 1)
     }
 
+    @action.bound onToggleTitleAnnotationEntity(value: boolean) {
+        const { grapher } = this.props.editor
+        grapher.hideTitleAnnotation = grapher.hideTitleAnnotation || {}
```
marcelgerber, line 60: we can use modern JS features here:
```suggestion
        grapher.hideTitleAnnotation ??= {}
```

or, alternatively, if you like it more:
```ts
        grapher.hideTitleAnnotation = { ...grapher.hideTitleAnnotation, entity: value || undefined }
```

adminSiteClient/EditorTextTab.tsx
```diff
@@ -88,13 +106,19 @@ export class EditorTextTab extends React.Component<{ editor: ChartEditor }> {
...
+                        value={!!grapher.hideTitleAnnotation?.entity}
+                        onValue={this.onToggleTitleAnnotationEntity}
+                    />
+                    <Toggle
+                        label="Hide automatic time (where possible)"
+                        value={!!grapher.hideTitleAnnotation?.time}
+                        onValue={this.onToggleTitleAnnotationTime}
+                    />
+                    <Toggle
+                        label="Don't prepend 'Change in' automatically (where possible)"
```
marcelgerber, line 119: Since this is the most obscure one (it's only relevant for a small subset of charts), I would try to make its name clearer:
`Don't prepend 'Change in' in relative line charts`.

Also, we sometimes hide controls that are not relevant for the current chart.
In this case, whether it's relevant depends on `grapher.isLineChart && (grapher.isRelativeMode || grapher.canToggleRelativeMode)`.

marcelgerber:
> The three toggles are pretty close to the URL field, which is a bit weird as they really belong to the Title text field. I didn't want to mess around with global css.. maybe we could simply move the toggles above the Title field? Or something else could be done? I am also happy to let this go if you don't think it matters.

I quickly tried this out, and adding a simple `<hr>` below the toggles works well.

> Down migration (at the moment, if any flag in the hideTitleAnnotation object is set to true, hideTitleAnnotation will be set to true)

Yeah makes sense to me. This won't change any existing charts (despite of what the svg tester says), and we can then look into ones that we would like changed.

marcelgerber (commented):
Nice work! I had a bunch of comments, but overall this is looking great!

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -299,7 +299,11 @@ export class Grapher
     @observable.ref subtitle = ""
     @observable.ref sourceDesc?: string = undefined
     @observable.ref note = ""
-    @observable.ref hideTitleAnnotation?: boolean = undefined
+    @observable hideTitleAnnotation?: {
```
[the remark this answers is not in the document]
sophiamersmann: For my understanding only: Wouldn't `{ entity: true }` and `{ entity: false }` be structurally equal?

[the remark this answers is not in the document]
marcelgerber: Oh yeah, you're right (in a way).
`@observable.struct` is definitely not the thing we want to use, it is not deep and only observes the top-level property.

packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
```diff
@@ -41,7 +41,11 @@ export interface GrapherInterface extends SortConfig {
     subtitle?: string
     sourceDesc?: string
     note?: string
-    hideTitleAnnotation?: boolean
+    hideTitleAnnotation?: {
```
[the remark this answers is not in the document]
marcelgerber: Yeah, I can see that it's a PITA with explorers.
I think we should rename it in explorers too. It should be a simple find-and-replace.

Just let me know when you're planning to deploy this change, and then I can coordinate the deploy with the explorer update.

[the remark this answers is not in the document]
sophiamersmann: Ok cool, will do, thanks!

marcelgerber (approved):
Looks good, thank you! We can coordinate on when to deploy this.

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -1210,28 +1212,32 @@ export class Grapher
     @computed get currentTitle(): string {
         let text = this.displayTitle
         const selectedEntityNames = this.selection.selectedEntityNames
-        const showTitleAnnotation = !this.hideTitleAnnotation
+        const showEntityAnnotation = !this.hideAnnotationFieldsInTitle?.entity
+        const showTimeAnnotation = !this.hideAnnotationFieldsInTitle?.time
+        const showChangeInPrefix =
+            !this.hideAnnotationFieldsInTitle?.changeInPrefix
+
+        const isFacetingEnabled =
```
marcelgerber, line 1220:
```suggestion
        const isFacetingActive =
```

`isFacetingActive` is more expressive here.

sophiamersmann: I thought `this.facetStrategy` refers to the facet strategy that is currently active? For example, if faceting is possible but currently only a single entity is selected, then `this.facetStrategy` would be `none` (currently not active) but `this.selectedFacetStrategy` would be set?

sophiamersmann:
I think this is okay to merge now, @marcelgerber? Once these changes are deployed, I'm happy to update these charts:

> The following two charts need fixing though:
>
> * [ ]  https://ourworldindata.org/grapher/cereal-yield-vs-extreme-poverty
> * [ ]  https://ourworldindata.org/grapher/soy-production-yield-area

Do you mind updating the explorers, Marcel? (`hideTitleAnnotation` needs to be renamed to `hideAnnotationFieldsInTitle` across all explorers.) If you want me to, I can also do it, but I'm scared haha. Non of this is urgent of course, so just let me know when you have some time :)

packages/@ourworldindata/grapher/src/core/Grapher.tsx
marcelgerber: Ah, right, I see now. Makes sense!

Issue #1995, opened 2023-03-13
Project data pages: Extend Grapher embedding to work with variable level partial grapher configs


danyx23:
## Overview

Enable Grapher to be embedded in a data page from given a variable id alone, and any partial Grapher config that variable has in its metadata.

## Config inheritance model 

```mermaid
graph TB

defaults["Default grapher settings"] --> variable-meta["Partial config from variable"] --> archieml-config["Partial config in ArchieML"] --> grapher["Rendered grapher"]
```

## Background

To clarify what this issue is about, let's start with a look at how Grapher configs and embedding works today:

Our website is static HTML and our grapher is a React based Javascript widget that needs a bit of json as its configuration. How exactly we run Grapher is a bit different depending on where it is used - the main uses are:
* On our main website. The baking process produces a `<figure>` element containing a static thumbnail. The grapher config is persisted into the HTML of the standalone /grapher/ pages [in a script tag](https://github.com/owid/owid-grapher/blob/4a9c8bf7c32f73c85f4301adde84425744995630/site/GrapherPage.tsx#L65C3-L70) that stores the config in a global JS variable. When a visitor comes to our site the siteFooter scripts set up an IntersectionObserver instance that trigger when one of these figures comes into view (on desktop) and then replaces the thumbnail with a new Grapher instance. This instance is saved as a React ref so that the heavy, stateful Grapher is not reconstructed all the time (this is not very React idiomatic of course and something we might want to change). The grapher config is extracted either from the same site if it is a standalone /grapher/ page or by fetching the html of that standalone grapher page.
* Google docs articles are hydrated as a whole. Currently they rely on the same method as above to get instantiated but there is a [useEmbedChart hook](https://github.com/owid/owid-grapher/blob/179f2f7cb1e165993039bfe194238d141c69708b/site/hooks.ts#L59) that we'll probably use instead soon (see #1974). 
* The grapher admin and some other admin components like the bulk chart editor have an `@observable.ref grapher = new Grapher()` and an `@observable.ref grapherElement?: JSX.Element` that they keep around - i.e. this is an instance of Grapher that is kept around and which is refreshed using functions like  and then the element is included in the render function.

[...]

PR #2000, opened 2023-03-14
Archieml component variable-chart

danyx23:
This PR implements a solution half-way between #1948 and #1995. It adds a function to stack grapher configs but it does not use this function yet or fetch the variable defaults. What it does add though is an ArchieML component `variable-chart` that shows a chart just from a variable id with the option to change the chart type and the ability to hide the tabs. 

This is an example for how to use the new component:
```archieml
{.variable-chart}
variable: 539022
chartType: DiscreteBar
hideTabs: true
{}
```

This implementation side-steps the InteractionObserver that we usually use via the MultiEmbedder and instead it just creates a Grapher immediately - so this doesn't scale to many Graphers on a page but it should be enough for the initial Data Pages project work.


larsyencken:
Closing this for now, since Data Pages did a pivot. We can re-discuss again once you're back!

PR #2017, opened 2023-03-17
Script to sync images from prod S3

ikesau:
Our refresh scripts were failing because the Grapher DB's `images` table is a record of files that have been uploaded to its S3 `owid-image-upload` bucket.

Each environment has its own folder in the bucket, so if we are resetting an environment's DB to be the same as prod's, we need to do the same for its images folder.

### DIFF
--- a/devTools/docker/sync-s3-images.sh
+++ b/devTools/docker/sync-s3-images.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env  bash
+
+# This script assumes that you have ssh access to the OWID production db server at live-db.owid.io It will
+# fail if you don't have these credentials
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+s3cmd sync s3://owid-image-upload/production/ s3://$IMAGE_HOSTING_BUCKET_PATH/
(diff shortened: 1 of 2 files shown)


devTools/docker/sync-s3-images.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env  bash
+
+# This script assumes that you have ssh access to the OWID production db server at live-db.owid.io It will
+# fail if you don't have these credentials
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
```
larsyencken, line 9: It doesn't error out here as expected due to `-o nounset` because it gets set to empty in the default `.env` file. It would be nice to add a usability message like the one below.

```
if [[ -n "$IMAGE_HOSTING_BUCKET_PATH" ]]; then
  echo 'Please set IMAGE_HOSTING_BUCKET_PATH in .env'
  exit 1
fi
```

larsyencken (approved):
Suggested one change for usability, then feel free to ship it :shipit: 

In a separate discussion, it would be awesome to be able to use a local folder for this, recalling that we want to move to a world where we have one staging environment per pull request.

Issue #1341, opened 2022-03-30
StackedArea allows selecting entities for which there is no data


danielgavrilov:
## Description

Some StackedArea charts show entities in select dialog that we seem to have no data for:

- Seems like only regions and the World have data, but many countries are listed: https://ourworldindata.org/grapher/vegetable-oil-production?country=~IDN
- https://ourworldindata.org/grapher/per-capita-consumption-of-cereals-by-commodity-type-daily-kilocalories?country=~AGO

## Expected behaviour

Entities without (plottable) data should not be listed in select dialog for StackedArea.

larsyencken:
Guesstimating at 1d, but it could be more complex than that.

larsyencken:
Reviewed again with @danyx23, we still think this is important to fix

sophiamersmann:
For reference, here is another example where this is an issue:
- https://ourworldindata.org/grapher/population-breakdown-by-highest-level-of-education-achieved-for-those-aged-15-in?country=~ABW

(The first example given above actually looks fine to me, the second one is a broken link)

sophiamersmann:
I did an investigation into this one but can't actually fix it. I'll drop my findings here.

The gist is: The select dialog and the chart are based on two different tables – the chart is based on the transformed table, but the select dialog is based on the input table. Using the input table rather than the transformed table is necessary for almost all charts, as the transformed table usually only contains a single entity (the one that is plotted).

The select dialog thus displays all entities that exist in the input table. But: In the case of stacked area charts, the `transformTable` function sometimes deletes all data that is available for a single entity – and no chart can be rendered. In such cases, an entity would show up in the select dialog that has in fact no _plottable_ data.

In the examples I looked at, this often seemed to be the offending line in the `transformTable` function:

https://github.com/owid/owid-grapher/blob/28812826c8576d6ca02d7502c92cd14b9774f5b7/packages/%40ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx#L71

It drops rows when _any_ column is missing. If a stacked area charts displays, for example, four different variables and one happens to be missing for an entity, than all data would be discarded and no chart plotted. This is what happens here:

https://ourworldindata.org/grapher/population-breakdown-by-highest-level-of-education-achieved-for-those-aged-15-in?country=~ABW

Aruba comes with data for the variable "Population aged 0-14" only, which means that no stacked area chart can be plotted. (But Aruba shows up in the select dialog as it is does exist in the input table!)

[...]

PR #2027, opened 2023-03-21
feat(grapher): allow authors to specify a preferred missing data strategy

### LINKED ISSUE #1177: Option to only show countries with data for all variables in selection dialog (hide countries with partial data)
## Core problem

In some charts where we plot multiple variables we would like to hide countries which are missing data for some variables, because it leads to weird views that have no resemblance to the title, or are in some cases wrong. 

For example in [this chart of GDP and CO2 emissions](https://ourworldindata.org/grapher/co2-emissions-and-gdp-per-capita?time=1750..2019&country=~GBR) there are three variables which should be shown. But for some countries or regions, only GDP is available (see for example [Africa Eastern and Southern](https://ourworldindata.org/grapher/co2-emissions-and-gdp-per-capita?time=1750..2019&country=~Africa+Eastern+and+Southern)). Therefore when you select that country, you only get one variable plotted which is out-of-sync with the title and the point of the chart. 

co2-emissions-and-gdp-per-capita

**Other examples**:

- https://ourworldindata.org/grapher/production-vs-consumption-co2-emissions?country=~MKD
- https://ourworldindata.org/grapher/daily-new-estimated-covid-19-infections-icl-model?country=~Asia

## Proposed solution

An option in the Admin to hide countries from the "change country" dialog if they don't have data for all variables.

sophiamersmann:
fixes #1177 

Grapher currently implements two different strategies to deal with entities that only have partial data (i.e. entities that are missing data for at least one variable). Here are two charts that exemplify the two strategies that Grapher currently employs:

- [LineChart](https://ourworldindata.org/grapher/production-vs-consumption-co2-emissions?country=~MKD): Consumption-based CO2 emissions are missing for North Macedonia → chart is rendered with missing data (which is usually fine, but in this case problematic as described in #1177)
- [StackedArea](https://ourworldindata.org/grapher/population-breakdown-by-highest-level-of-education-achieved-for-those-aged-15-in?country=~ABW): Aruba does have data for one of the variables, but most variables are missing → No chart is rendered (which is usually a good default to avoid showing misleading charts, but can be undesired in some cases, see comments on Slack [here](https://owid.slack.com/archives/C04PZLMLVFY/p1679351095481309?thread_ts=1678986529.574209&cid=C04PZLMLVFY) and [here](https://owid.slack.com/archives/C04PZLMLVFY/p1679351095481309?thread_ts=1678986529.574209&cid=C04PZLMLVFY))

This PR adds an additional “Missing data” section in the “Data” tab that allows authors to specify a preferred “Missing data strategy”. Three options are available:

- “Automatic” (default): picks a reasonable strategy based on the chart type (this makes the current behaviour explicit)
- “Hide entities with missing data”: Only charts with complete data are rendered. Charts with missing data are not shown – instead, the chart canvas says “No matching data” and it is suggested to change the country/entity
- “Show entities with missing data”: All charts are rendered, including those with missing data.

Caveat: Selecting “Hide charts with missing data” makes #1341 worse as charts for countries/entities with missing data will not be rendered but show up in the select dialog. ~(I made an attempt to fix this issue in this PR.)~

Chart types that hide charts with missing data by default (to do: implement the opposite behaviour):

- [x]  StackedArea
- [x]  StackedBar

Chart types that render charts with missing data by default (to do: implement the opposite behaviour):

- [x]  StackedDiscreteBar
- [x]  DiscreteBar
- [x]  LineChart

Chart types that solve the missing data problem visually (nothing to be done here):

- Marimekko

Doesn’t apply (nothing to be done here):

- ScatterPlot
- TimeScatter
- SlopeChart

Other TODOs:

- [x]  Apply schema changes to schema version 3 once available
- [x]  Hide “Missing Data” section when not applicable?
- [x] Only enable this option for multi-entity charts?

### DIFF
--- a/adminSiteClient/EditorCustomizeTab.tsx
+++ b/adminSiteClient/EditorCustomizeTab.tsx
@@ -718,7 +718,9 @@ export class EditorCustomizeTab extends React.Component<{
                         </FieldsRow>
                     </Section>
                 )}
-                <MissingDataSection editor={this.props.editor} />
+                {features.canSpecifyMissingDataStrategy && (
+                    <MissingDataSection editor={this.props.editor} />
+                )}
                 {features.comparisonLine && (
                     <ComparisonLineSection editor={this.props.editor} />
                 )}
--- a/adminSiteClient/EditorCustomizeTab.tsx
+++ b/adminSiteClient/EditorCustomizeTab.tsx
@@ -8,6 +8,7 @@ import {
     ColorSchemeName,
     FacetAxisDomain,
     FacetStrategy,
+    MissingDataStrategy,
 } from "@ourworldindata/grapher"
 import {
     NumberField,
@@ -37,6 +38,7 @@ import {
 } from "./ColorSchemeDropdown.js"
 import { EditorColorScaleSection } from "./EditorColorScaleSection.js"
 import Select from "react-select"
+
 @observer
 export class ColorSchemeSelector extends React.Component<{ grapher: Grapher }> {
     @action.bound onChange(selected: ColorSchemeOption) {
@@ -386,6 +388,68 @@ class TimelineSection extends React.Component<{ editor: ChartEditor }> {
 }
 
 @observer
+class MissingDataSection extends React.Component<{ editor: ChartEditor }> {
+    missingDataStrategyLabels = {
+        [MissingDataStrategy.auto]: "Automatic",
+        [MissingDataStrategy.hide]: "Hide charts with missing data",
+        [MissingDataStrategy.show]: "Render charts with missing data",
+    }
+
+    @computed get grapher() {
+        return this.props.editor.grapher
+    }
+
+    @computed get missingDataStrategyOptions(): {
+        value: MissingDataStrategy
+        label: string
+    }[] {
+        return Object.values(MissingDataStrategy).map((strategy) => {
+            return {
+                value: strategy,
+                label: this.missingDataStrategyLabels[strategy],
+            }
+        })
+    }
+
+    @computed get selectedMissingDataStrategyOption(): {
+        value: MissingDataStrategy
+        label: string
+    } {
+        const { grapher } = this
+
+        const selectedOption = this.missingDataStrategyOptions.find(
+            (opt) => opt.value === grapher.missingDataStrategy
+        )
+
+        if (selectedOption) return selectedOption
+
+        return {
+            value: MissingDataStrategy.auto,
+            label: this.missingDataStrategyLabels[MissingDataStrategy.auto],
+        }
+    }
+
+    render() {
+        const { grapher } = this
+
+        return (
+            <Section name="Missing data">
+                <div className="form-group">
+                    Missing data strategy
+                    <Select
+                        options={this.missingDataStrategyOptions}
+                        value={this.selectedMissingDataStrategyOption}
+                        onChange={(option) =>
+                            (grapher.missingDataStrategy = option?.value)
+                        }
+                    />
+                </div>
+            </Section>
+        )
+    }
+}
+
+@observer
 class ComparisonLineSection extends React.Component<{ editor: ChartEditor }> {
     @observable comparisonLines: ComparisonLineConfig[] = []
 
@@ -654,6 +718,7 @@ export class EditorCustomizeTab extends React.Component<{
                         </FieldsRow>
                     </Section>
                 )}
+                <MissingDataSection editor={this.props.editor} />
                 {features.comparisonLine && (
                     <ComparisonLineSection editor={this.props.editor} />
                 )}
--- a/adminSiteClient/EditorCustomizeTab.tsx
+++ b/adminSiteClient/EditorCustomizeTab.tsx
@@ -19,6 +19,7 @@ import {
     TextField,
     Button,
     RadioGroup,
+    SelectField,
 } from "./Forms.js"
 import {
     debounce,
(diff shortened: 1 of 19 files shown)


adminSiteClient/EditorCustomizeTab.tsx
```diff
@@ -376,6 +379,50 @@ class TimelineSection extends React.Component<{ editor: ChartEditor }> {
     }
 }
 
+@observer
+class MissingDataSection extends React.Component<{ editor: ChartEditor }> {
```
marcelgerber, line 383: I would move all this to `EditorDataTab`, actually.

adminSiteClient/EditorCustomizeTab.tsx
```diff
@@ -376,6 +379,50 @@ class TimelineSection extends React.Component<{ editor: ChartEditor }> {
...
+        this.grapher.missingDataStrategy = value as MissingDataStrategy
+    }
+
+    render() {
+        const { grapher } = this
+
+        return (
+            <Section name="Missing data">
+                <SelectField
+                    label="Missing data strategy"
```
marcelgerber, line 416: May clarify this a bit more:
```suggestion
                    label="Missing data strategy (for when some variables are missing data)"
```

adminSiteClient/EditorCustomizeTab.tsx
```diff
@@ -376,6 +379,50 @@ class TimelineSection extends React.Component<{ editor: ChartEditor }> {
...
+        return this.props.editor.grapher
+    }
+
+    get missingDataStrategyOptions(): {
+        value: MissingDataStrategy
+        label: string
+    }[] {
+        const missingDataStrategyLabels = {
+            [MissingDataStrategy.auto]: "Automatic",
+            [MissingDataStrategy.hide]: "Hide charts with missing data",
```
marcelgerber, line 394: We have this slight problem that in our lingo, "chart" usually refers to the whole chart chrome (i.e. grapher as a whole), which is a bit confusing here.
I'm trying to come up with alternative words, but it's tough: plot, entities, visualisation, ...

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -614,6 +616,22 @@ export class Grapher
...
+                (this.isLineChart ||
+                    this.isDiscreteBar ||
+                    this.isStackedDiscreteBar)) ||
+            (this.missingDataStrategy !== MissingDataStrategy.show &&
+                (this.isStackedArea || this.isStackedBar))
+        ) {
+            return this.inputTable.dropRowsWithErrorValuesForAnyColumn(
+                this.yColumnSlugs
+            )
+        }
```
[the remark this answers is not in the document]
marcelgerber: Could make this into an exported `dropRowsAccordingToMissingDataStrategy(table: OwidTable)` instead, so it can be used by both `tableForSelection` as well as its own `transformTable` method.

marcelgerber (commented):
Cool, this works well and is a sound approach 👏🏻

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -614,6 +616,22 @@ export class Grapher
         // for other chart types, the `transformTable()` call would sometimes remove too many
         // entities, and we want to use the inputTable instead (which should have exactly the
         // entities where data is available)
+
+        // however, sometimes we choose not to render a chart for an entity in case of missing
+        // data. In that case, we want to remove the entity from the selection table as well.
```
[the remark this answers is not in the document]
sophiamersmann: Ahh, weird. I'll look into it but I'm also still struggling with MobX, so will likely get back to you for this one...

packages/@ourworldindata/grapher/src/core/Grapher.tsx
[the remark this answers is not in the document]
sophiamersmann: Yeah, agreed that this is very brittle!

adminSiteClient/EditorFeatures.tsx
```diff
@@ -101,4 +101,15 @@ export class EditorFeatures {
     @computed get canHideTotalValueLabel() {
         return this.grapher.isStackedDiscreteBar
     }
+
+    @computed get canSpecifyMissingDataStrategy() {
+        return (
+            this.grapher.canChangeEntity &&
```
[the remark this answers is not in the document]
sophiamersmann: mhh.. yes, I think you're right!

adminSiteClient/EditorCustomizeTab.tsx
sophiamersmann: Ahh, okay. I think I'm leaning towards entities then? But it's a difficult one...

sophiamersmann:
Thanks for the thorough review, Marcel, it's appreciated! :)))

adminSiteClient/EditorCustomizeTab.tsx
sophiamersmann: I said "entities" for now but we can sleep on it :)

sophiamersmann:
Lars brought to my attention that this approach sometimes trim the data series. An example is the chart with id 3622.

Default strategy (no hiding)
Screenshot 2023-03-31 at 13 40 58

Hiding entities when missing data has this undesired side effect:
Screenshot 2023-03-31 at 13 41 08

This happens because we drop every year for which at least one column is missing.

marcelgerber:
> Lars brought to my attention that this approach sometimes trims the data series. An example is the chart with id 3622.

Ah, that's a bummer. Not sure how we would want to handle these instead.

sophiamersmann:
**Regarding undesired data trimming (see [my comment](https://github.com/owid/owid-grapher/pull/2027#issuecomment-1491802638)):** 
I think it's fine for StackedArea and StackedBar charts to continue using `dropRowsWithErrorValuesForAnyColumn` as we really do want to make sure that we never render a row where values are missing for these chart types (because that could be misleading). That is not the case for LineChart, DiscreteBar and StackedDiscreteBar charts however. Here, showing entities with missing data is usually ok, but sometimes undesired. In these cases, instead of dropping rows individually, we check if one of the columns is missing for an entity and if so, drop all data.

**Regarding the input dialog and `tableForSelection` (see https://github.com/owid/owid-grapher/issues/1341):**
I'm not sure how to address this properly, and the data trimming issue complicated the problem further. I also realised that checking for valid values has to run after interpolation. So, I really can't see a solution that doesn't involve re-running basically all transforms for all available entities... which feels wrong. I think we should decouple this issue from this PR for now, is that's okay with you, Marcel? (If you have an approach in mind, I'd be happy to try to implement that, of course)

marcelgerber:
Makes sense.
Anyhow, I would like to wait on this until after Daniel is back, since Daniel has previously thought about missing data a lot.

danyx23:
Great work, @sophiamersmann! I think we should discuss this in the group in the next data viz meeting. Sorry that my input was blocking this for so long :)

sophiamersmann:
Sounds good, thanks for checking it out! Looking forward to chatting about this on Wednesday :)

sophiamersmann:
Hey @danyx23, I just rebased this onto master, reviewed and tested the code again and updated the PR description – I think this is ready for your review (not urgent of course).

sophiamersmann:
Thanks for the review! I wanted to merge this today but it slipped by mind 🤦‍♀️ Will do on Thursday!

PR #2037, opened 2023-03-23
enhance: Hide the Covid-19 alert banner on every page

larsyencken:
This change simply removes the alert banner from the site, but leaves the `AlertBanner` class around for us to reuse later if we want to.


larsyencken:
I also had the first instinct to cut deeper, but I scaled myself back. We've wanted to use this space for hiring as well in the past, we may still have a use for being able to put something out quick.

PR #1306, opened 2022-03-22
enhance(scatter): set "Other" as default label for points without color

danielgavrilov:
Fixes #1167. [Slack thread](https://owid.slack.com/archives/C5BDCB2R3/p1647944889676269). We decided that `Other` is better than `No data`. 

### To-do

- [ ] Add tests
- [ ] Check/think how case where data contains `"Other"` should be handled
- [ ] Define no data color outside `customCategoryLabels`? Could be 2 new fields for `ColorScale`: `noDataLabel` and `noDataColor`

PR #2058, opened 2023-03-28
fix: hide No Data item in scatterplot legends

samizdatco:
This PR offers a partial solution to issue #1167 by hiding the No Data label that appears in the legend for scatterplots when aggregates are part of the data set. It's unresolved whether anything should be done to hide the actual bubbles for regions (especially for the World dot which will typically dwarf all other country sizes).

The "No Data" item is being filtered out based on the `label` value of the legend's ColorScaleBin meaning this will break if it's ever renamed (e.g. to "Other" as proposed in #1306).

### DIFF
--- a/packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx
+++ b/packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx
@@ -54,7 +54,7 @@ import {
 import { DualAxisComponent } from "../axis/AxisViews"
 import { DualAxis, HorizontalAxis, VerticalAxis } from "../axis/Axis"
 
-import { ColorScale, ColorScaleManager } from "../color/ColorScale"
+import { ColorScale, ColorScaleManager, NO_DATA_LABEL } from "../color/ColorScale"
 import { AxisConfig } from "../axis/AxisConfig"
 import { ChartInterface } from "../chart/ChartInterface"
 import {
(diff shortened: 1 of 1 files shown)


samizdatco:
Counterexample (where the 'other' case is called "Not categorized"): https://ourworldindata.org/grapher/aid-vs-gdp-per-capita?yScale=linear

Seems like we need a more universal predicate...

packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx
```diff
@@ -564,8 +564,10 @@ export class ScatterPlotChart
     }
 
     @computed get legendItems(): ColorScaleBin[] {
-        return this.colorScale.legendBins.filter((bin) =>
-            this.colorsInUse.includes(bin.color)
+        return this.colorScale.legendBins.filter(
+            (bin) =>
+                this.colorsInUse.includes(bin.color) &&
+                !bin.label?.match(/no data/i)
```
larsyencken, line 570: It probably makes sense to explicitly reference `NO_DATA_LABEL` from `ColorScale.ts` here, which also makes it easier for someone to follow the link back to where that bin is created.

larsyencken (changes requested):
Tiny suggestion, but if you fix that up it's good to merge.

marcelgerber (approved):

packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx
samizdatco: Ah, that's much better. It was definitely feeling wrong to compare against a copy the phrasing we happen to be using for the moment…

PR #2067, opened 2023-03-31
List recent charts first on test pages

sophiamersmann:
Reverses the order of test charts to show recent charts on the first page


larsyencken (approved):
Looks good, works well :shipit:

PR #2077, opened 2023-04-04
WIP: Changes needed by automated staging servers

Marigold:
Small changes required for the ops work [on staging servers](https://github.com/owid/ops/pull/53)

* use insecure cookie if using HTTP protocol and `ENV=production` (for staging server behind Tailscale)

### DIFF
--- a/adminSiteServer/adminRouter.tsx
+++ b/adminSiteServer/adminRouter.tsx
@@ -104,10 +104,15 @@ adminRouter.post(
                 req.body.username,
                 req.body.password
             )
+            // secure cookie if in production and using https
+            // our staging servers use http and `production`, so passing insecure
+            // cookie wouldn't work
+            const secure = ENV === "production" && req.protocol !== "http"
+
             res.cookie("sessionid", session.id, {
                 httpOnly: true,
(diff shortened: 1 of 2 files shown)


larsyencken (commented):
It's not clear to me whether or not a cookie could leak this way, in prod, over HTTP by accident.

How hard would it be to have an ENV=staging and just disable it for staging?

marcelgerber (approved):
This is interesting.
In terms of attack risks, a stolen `sessionid` to our live site is a pretty serious incident, since they never expire and give full access to the Admin. Hence, we should protect them as best as possible.

On the other hand, we thankfully have [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) enabled on owid.cloud and subdomains, meaning that the browser will not connect to it via http and thus downgrade attacks are harder.

Also, I just checked that `req.protocol` is in fact `https` on current staging servers, _and it is_.
One part of the traffic is encrypted owid.cloud <-> Cloudflare (using LetsEncrypt certificates, managed ourselves using certbot), then the other part is encrypted using Cloudflare.

Issue #1876, opened 2023-01-25
Filter out aggregates in tables and only show countries


danyx23:
## Overview

The table tab in Grapher shows all entities which is often not just countries but also aggregates like World or continents. It would be useful to easily filter those out because they tend to dominate tables if you sort by value. 

## Proposal

Add a small control to the table view that lets you hide aggregates.

## Technical notes

- For this it would be useful to have #1849 in place as it contains the necessary information.

larsyencken:
Suggested faster approach is to simply look at `EntityCodes.ts` in grapher, take note of the very few aggregates that are in that list, and make a second list to allow exclusion of them.

Probably needs a small design change to the table view to allow it.

samizdatco:
One option I've been playing with is keeping the aggregates around but grouping them together either at the top or the bottom of the table with a horizontal line separating them from the country-by-country rows.

To make this work, the sorting would have to reorder the two groups separately as well (i.e., the aggregate rows would always be on one end of the table, but within the group the rows would be ordered by the current sort column).

<img width="1003" alt="image" src="https://user-images.githubusercontent.com/469523/228270728-39256d2b-ac21-45aa-916e-5da77762bbaf.png">

PR #2096, opened 2023-04-11
Display aggregates separately in tables

samizdatco:
Implements the proposal sketched in #1876 to group countries and aggregates into separate regions within the table. Countries are always listed first and their names are styled in bold. Aggregates are moved to the end of the table and their names are italicized. When the table is sorted by a column, the two groupings within the table are sorted independently.

## Additional styling tweaks

To improve the readability of individual rows, the formatter now takes an option to use a no-break space to join the value and its unit, preventing awkward cases with unstable vertical baselines such as:
image

Vertically aligning cells to the bottom of the cell improves the chances that numerals will all be on the same line regardless of the presence of a data-not-available annotation:
image

A new formatting option that applies to [formatValue](https://github.com/owid/owid-grapher/blob/master/packages/%40ourworldindata/utils/src/formatValue.ts) and friends is the optional boolean flag `useNoBreakSpace`. It is `false` by default and only applies in cases where `spaceBeforeUnit` is `true`. If a no-break space is requested, a `\00a0` space will be used instead of the standard `\u0020`.

## Design questions

- Should the divider between the countries & aggregates sections have a text label along the lines of "Other groupings"?
- Should the header for the entities column retitle itself from "Countries" to something else ("Aggregates"? "Other regions"?) when the user scrolls into the aggregate section?

## Preview

The above screenshots come from:
http://staging-site-restyle-aggregates-in-tables/grapher/kaya-identity-co2?tab=table

Also note that for tables whose `entityType` doesn't [contain the string `country`](https://github.com/owid/owid-grapher/issues/2173), the country-vs-aggregate sorting and styling isn't applied:
http://staging-site-restyle-aggregates-in-tables/grapher/artificial-intelligence-training-computation?tab=table

### DIFF
--- a/packages/@ourworldindata/grapher/src/dataTable/DataTable.jsdom.test.tsx
+++ b/packages/@ourworldindata/grapher/src/dataTable/DataTable.jsdom.test.tsx
@@ -93,7 +93,7 @@ describe("when you select a range of years", () => {
 
     it("renders absolute change values", () => {
         const cell = view.find("tbody .dimension-delta").first()
-        expect(cell.text()).toBe("-0.89 pp")
+        expect(cell.text()).toBe("-0.89\u00a0pp") // uses no-break space separator
     })
 
     it("renders relative change values", () => {
--- a/packages/@ourworldindata/grapher/src/dataTable/DataTable.scss
+++ b/packages/@ourworldindata/grapher/src/dataTable/DataTable.scss
@@ -41,7 +41,7 @@ $cell-border: 1px solid rgba(0, 0, 0, 0.08);
 
 .data-table {
     font-size: 0.9375rem;
-    line-height: 16px;
+    line-height: 1.333em;
     border: none;
     background: white;
     width: 100%;
@@ -101,6 +101,10 @@ $cell-border: 1px solid rgba(0, 0, 0, 0.08);
     td {
         padding-left: 15px;
         padding-right: 15px;
+
+        &:not(.entity) {
+            vertical-align: bottom;
+        }
     }
 
     th.dimension {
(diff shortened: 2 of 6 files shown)


sophiamersmann:
> * Should the header for the entities column retitle itself from "Countries" to something else ("Aggregates"? "Other regions"?) when the user scrolls into the aggregate section?

The new `entityType` default is likely going to be "country or region" (#2175), so I think not changing it on scroll is fine :)

danyx23 (approved):
Looks good! I like the design choices here. There was [one case in the fish explorer](http://staging-site-restyle-aggregates-in-tables/explorers/fish-stocks?tab=table&facet=none&country=Albacore+tuna+North+Pacific+Ocean~Albacore+tuna+South+Atlantic~Atlantic+bluefin+tuna+Western+Atlantic&Metric=Fish+stocks+%28biomass%29&Total+or+Relative=Relative+to+Max+Sustainable+Yield) that I saw that looks a tiny bit funky - the different-year-note has a trailing space that leads to the right side of the text no longer being flush. If you have a quick idea how to remedy that then that might be nice but otherwise don't bother - this is clearly better than what we have atm for almost all cases.

samizdatco:
> the different-year-note has a trailing space that leads to the right side of the text no longer being flush. 

Good catch—that bit of padding was also rendering differently in chrome vs safari for some reason. Turning it into a margin seems to give the intended effect (providing space when it's on a single line, but being flush right when it's the line-break point).

That explorer you linked to also shows some weird behavior with the header rows having a transparent horizontal slit that lets the table contents peek through once you start scrolling. I reworked the sticky-positioning setup to fix that and to make sure the header borders don't scroll off the screen anymore.

PR #2106, opened 2023-04-14
Consolidate competing methods to specify gdoc cover images

ikesau:
A month ago, I added `cover` as a gdoc property without realizing that we already had a `cover-image` property.

This removes that redundant property and also makes sure we're using `cover-image` and `featured-image` correctly throughout the codebase.

`cover-image` is an optional image that will overlay the yellow banner. At the moment it's very primitive and only supports one file, which means that we have to make the image square so that it works at all aspect ratios. If we want to use these more seriously, we'll have to revisit this so that multiple images can be specified.

`featured-image` is the image that we use for the open graph social media previews and prominent link thumbnails. If it's unset, we'll fallback to `default-thumbnail.jpg` for social media shares, and blank for prominent links.


mlbrgl:
- [ ] is the `copilot:summary` accurate?
- [ ] set up vscode remote host: staging-site-PR-NAME
- [ ] what is the happy path?

mlbrgl:
**blackbox test (short image)**

Screenshot 2023-05-18 at 12 34 57

mlbrgl:
Simply noticing the error a wrong cover-image displays: not totally visible, but good enough.

Screenshot 2023-05-18 at 14 49 01

site/gdocs/OwidGdocPage.tsx
```diff
@@ -29,6 +32,14 @@ export default function OwidGdocPage({
 }) {
     const { content, slug, createdAt, updatedAt } = gdoc
 
+    const featuredImageUrl =
+        content["featured-image"] &&
+        path.join(
+            baseUrl,
+            "images",
+            "published",
```
mlbrgl, line 40: `images/published` is being used in a few places by now, it might be worth saving that path into a variable

site/gdocs/centered-article.scss
```diff
@@ -18,6 +18,19 @@ $banner-height: 200px;
     z-index: -1;
 }
 
+// Until we implement a way to specifiy src-sets in archie, this image will need to have a 1:1 aspect ratio
+// to fit on both mobile and desktop resolutions
```
mlbrgl, line 22: I don't quite get this 1:1 ratio constraint. Whether on mobile or desktop, the image is never quite square or meant to be, is it?

I also don't get why we would want multiple cover images.

mlbrgl (approved):
Makes sense!

My main question is around the ambition for `cover-image` (cf. comment below).

site/gdocs/centered-article.scss
ikesau: True, 1:1 was just an approximation of the smallest view that the site reasonably supports:

image

And so eventually I imagine we'd want to be able to design one cover image to use when the aspect ratio is like that, and a separate one to use on wider screens that would be more rectangular.

Of course, we could stick to using the same asset in both cases and just center and scale it - it depends on what kind of designs we want to use for assets here. I'm kicking that can with this PR until we actually have articles that we want to cover.

Issue #2146, opened 2023-04-26
Faceting drop-down uses "country" even for non-country entities


larsyencken:
## Overview

We have charts with non-country entities, but we still show "Country" in the faceting dropdown.

## Example

<img width="818" alt="Screenshot 2023-04-26 at 10 24 15" src="https://user-images.githubusercontent.com/36012/234532395-95549c76-9e8f-42ce-a204-d6879d706978.png">

https://ourworldindata.org/grapher/low-earth-orbits-objects

Expected behaviour would be to use a user-defined word for this set in the admin, as we do in the country selector.

larsyencken:
This specific case turned out to be an issue because the admin did not let you edit the entity type. Submitted a patch to allow entityType editing in the admin now.

larsyencken:
@sophiamersmann is doing an audit of the existing charts for more weird cases like this.

sophiamersmann:
Should be done! This was a bit of a manual process, so can't promise I got them all but I went through the chart configs twice – so I'm pretty confident I got most of them.

sophiamersmann:
Closing this as completed but please reopen if I missed some :)

sophiamersmann:
Re-opening because I only checked StackedArea, StackedDiscreteBar and StackedBar chart, not LineCharts

sophiamersmann:
Should be done 🙏🏻

PR #2147, opened 2023-04-26
fix: show entityType editor in admin in more cases

larsyencken:
Currently, we only show the `entityType` editor in the admin when entity selection is enabled. But now that we enable faceting for more cases, those cases also need an `entityType`.

This change makes `entityType` editable in more cases, even if entity selection is not enabled for a chart.

See: #2146


sophiamersmann (approved):
Looks good to me. Thank you, Lars!

PR #2149, opened 2023-04-26
refactor(baker): remove silent failures

Marigold:
If baker fails on a chart, it just logs the error with `console.error(err)` and continues. See what charts are failing like this and either keep them as errors or at least send them to Slack.


marcelgerber (approved):
We should definitely watch out if any deploys fail because of this in the future, but at least we'll now get notified about them ;)
LGTM!

PR #2162, opened 2023-04-27
fix(baker): await country profiles

Marigold:
Country profiles were never awaited. This was making debugging with `--steps countryProfiles` hard.


ikesau (approved):
nice catch

marcelgerber (approved):

PR #2182, opened 2023-05-03
feat(admin): support custom "Split by metric" label

### LINKED ISSUE #2163: Allow to customise the word "metric" in "Split by metric"
## Core problem

For some charts (not many), "Split by metric" is poor wording in the context of the chart.

## Proposed solution

Add a field to the admin that allows to overwrite the word "metric".

## Alternatives

Leave Grapher as is. (Not many charts are impacted – for most charts, saying "Split by metric" might not be optimal but kind of works)

## Context

Here, the "metrics" are actually countries!

Screenshot 2023-04-27 at 18 52 23

"Split by product" would be better here:

Screenshot 2023-04-27 at 18 53 09

"Split by category" would be better here:

Screenshot 2023-04-27 at 18 54 20

sophiamersmann:
fixes #2163 

### Summary

This pull request adds a new field to the admin that allows to customise the name of the y-variable type when multiple y-variables are in use. The default is "metric". This is used to customise the "Split by metric" label for faceting.

Screenshot 2023-06-05 at 16 21 40

The config field is called `facettingLabelByYVariables`

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
+++ b/packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
@@ -149,6 +149,7 @@ export const grapherKeysToSerialize = [
     "hideRelativeToggle",
     "entityType",
     "entityTypePlural",
+    "yVariableType",
     "hideTimeline",
     "zoomToSelection",
     "showYearLabels",
--- a/packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
+++ b/packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
@@ -149,7 +149,7 @@ export const grapherKeysToSerialize = [
     "hideRelativeToggle",
     "entityType",
     "entityTypePlural",
-    "yVariableType",
+    "facettingLabelByYVariables",
     "hideTimeline",
     "zoomToSelection",
     "showYearLabels",
--- a/adminSiteClient/EditorCustomizeTab.tsx
+++ b/adminSiteClient/EditorCustomizeTab.tsx
@@ -634,7 +634,7 @@ export class EditorCustomizeTab extends React.Component<{
                                         Split by <s>metric</s>
                                     </>
                                 }
-                                field="yVariableType"
+                                field="facettingLabelByYVariables"
                                 store={grapher}
                                 helpText={
                                     "When facetting is active, one option is to split " +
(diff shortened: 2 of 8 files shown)


packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
```diff
@@ -149,6 +149,7 @@ export const grapherKeysToSerialize = [
     "hideRelativeToggle",
     "entityType",
     "entityTypePlural",
+    "yVariableType",
```
danyx23, line 152: I'm not a fan of `yVariableType `- this makes me think of a data type or so but this is actually a label. What about `facettingLabelByYVariables` or similar?

danyx23 (commented):
I know that the related fields are called entityType etc and that you wanted to keep the name of the new field similar but I really think we should find a better name that makes it clear that this is just used as a label. 

Instead of `yVariableType` I suggested `facettingLabelByYVariables` but maybe we can find something better still.

In the admin UI I think the same thing applies, we should make it clearer what this is. One way to do it is to add a small help text below the text field like this: 'When facetting is active, one option is to split by entity/country, the other is by metric. This option lets you override "metric" with a custom word like "products" or "species"'.

Another option, a bit unconventional, would be to label it with a strikethrough of metric: 
Override facetting label: "Split by ~metric~"

packages/@ourworldindata/grapher/src/core/GrapherInterface.ts
sophiamersmann: Yeah.. I'm not a fan of `yVariableType` either... I struggled to come up with something and then landed on the one that was closest to `entityType`.

I tried to choose a name that describes what it is rather than what it is used for in the UI (in case we wanted to use it for something else in the future). But I don't have a particular second use case in mind, so we might as well give it a use case-specific name.

I like `facettingLabelByYVariables`! How about `facetControlLabelForSplitByVariable`?

sophiamersmann:
> In the admin UI I think the same thing applies, we should make it clearer what this is. One way to do it is to add a small help text below the text field like this: 'When facetting is active, one option is to split by entity/country, the other is by metric. This option lets you override "metric" with a custom word like "products" or "species"'.
> 
> Another option, a bit unconventional, would be to label it with a strikethrough of metric: Override facetting label: "Split by ~metric~"

I like the strikethrough label, very smart! If we label the field "Split by ~metric~", then I think we might not need the additional info text.. right?

danyx23 (approved):
Looks good!

sophiamersmann:
Thanks for the review!

PR #2187, opened 2023-05-04
migration: rename countries, especially Timor to East Timor

Marigold:
Migration that renames countries according to
```
const renameMap = {
    Timor: "East Timor",
    "Saint Barthélemy": "Saint Barthelemy",
    "Åland Islands": "Aland Islands",
}
```


samizdatco (commented):
This is looking good but I'd be interested in testing whether anything unexpected happens after updating both the db and the etl-derived regions.json file. Is there a staging url for the [ETL PR](https://github.com/owid/etl/pull/1077) that I can try pointing the `regionsUpdater` at? [And, if I'm not misunderstanding the config, setting `CATALOG_PATH` to the staging url should also get grapher to do its S3 variable fetches there as well?]

Marigold:
Great idea! I'm building the entire catalog on staging now, new regions file [is here](https://catalog-staging.ourworldindata.org/grapher/regions/latest/regions/regions.csv)

> I'm not misunderstanding the config, setting CATALOG_PATH to the staging url should also get grapher to do its S3 variable fetches there as well?]

CATALOG_PATH is kinda deprecated. There's no global ENV that controls source, but there are fields `variables.dataPath` and `variables.metadataPath` that point to source of variables data. 

I had to connect to MySQL running on staging server with
```
mysql -h staging-site-rename-countries -u owid --port 3306 -D owid
```
and then ran
```
UPDATE variables
SET metadataPath = REPLACE(metadataPath, 'live_grapher', 'staging_grapher'),
        dataPath = REPLACE(dataPath, 'live_grapher', 'staging_grapher');
```
to point them to staging.

Marigold:
@samizdatco I realised it won't work just with redirecting to staging (variable ids are different on staging and on staging-site-rename-countries), but I at least made it work [with population](http://staging-site-rename-countries/grapher/population?tab=map&time=latest) and it's showing east timor! Would you mind reviewing & approving?

samizdatco (approved):
The updated `regions.json` data looks solid to me and I can confirm that the page you got working on the staging server is working as expected (the failure-mode previously was for East Timor to appear in the tooltip but for it to be displayed as having No Data since the values were still associated with "Timor").

I added one item to the TODO list above: the entities whose slugs have changed need to have redirects added via wordpress from `/country/old-slug` to `/country/new-slug`

PR #2192, opened 2023-05-05
fix: Fix wordpress posts missing in the sitemap by making the posts_with_gdoc_publish_status easier to use

danyx23:
The view tried to join on the posts_gdocs table and then gave as the isGdocPublished field a nullable boolean. This is easy to use incorrectly so this fix adds COALESCE(false).

This fixes wordpress posts not showing up in the sitemap anymore.


ikesau (approved):
Rebased to fix the failing tests.

Can confirm this fixes the sitemap on a local bake 🙂

danyx23:
Thanks @ikesau!

PR #2210, opened 2023-05-12
Channel server errors through Bugsnag

danyx23:
_published by Daniel for convenience, please direct your feedback to @mlbrgl_

### <samp>🤖 Generated by Copilot at 6a8a2c1</samp>

This pull request replaces Slack with Bugsnag for error reporting and monitoring across the project. It adds a new `errorLog.ts` module that exports functions for logging errors and warnings, and updates the imports and function calls in various files to use them. It also installs the Bugsnag plugin for Express and removes the Slack packages from `package.json`. It deletes the `slackLog.ts` module and modifies the environment variables in `serverSettings.ts` to include the Bugsnag API key.

- [x] add BUGSNAG_NODE_API_KEY to  live .env
- [ ] remove slack key from live .env
- [ ] close #content-errors channel


mlbrgl:
Test integration for new errors and repeat errors (100 in <1h): https://owid.slack.com/archives/C5JJW19PS/p1683905832627659

mlbrgl:
Concept reviewed and approved yesterday during P&D meeting by Daniel, Ike, Joe and Matthieu

marcelgerber:
Didn't have time to look at the code, but just some things to watch out for:
- Maybe it makes sense to trial-run the switch by doing both things; i.e. sending the error to Slack using our current integration and also sending it to Bugsnag (which we may set up to post to a different channel, to avoid some confusion) and keep track that looking at the Bugsnag feed gives a good and timely overview.
- I'm still wondering about whether the Bugsnag notification settings are a good fit for our server errors. There is no trigger "an error occurs for the first time within 24h" sadly, which would be great for us. I'm a bit worried that we might miss some events?
image

Would be great if you can also have a look at this, @danyx23. Sadly, I cannot request a review from you since you've opened the PR 👀

danyx23:
I'm about to go over the code in detail but as a general comment on the conversation so far: I think it would be worth running bugsnag and direct slack notifications to a different channel in parallel for a week or so as Marcel suggested. 

For configuring bugsnag I would say the best thing is probably to set a very low frequency threshold (e.g. 2 per 15 min or per 1 hour). This would mean that we should hear about errors that occur more than once relatively timely. In bugsnag we could then still choose to silence an error that we know is benign. We could then tweak the settings from there. What do you think?

adminSiteServer/apiRouter.ts
```diff
@@ -1962,7 +1962,7 @@ apiRouter.put("/datasets/:datasetId", async (req: Request, res: Response) => {
             commitEmail: res.locals.user.email,
         })
     } catch (err) {
-        logErrorAndMaybeSendToSlack(err)
+        logErrorAndMaybeSendToBugsnag(err)
```
danyx23, line 1965: Looking at [the docs](https://docs.bugsnag.com/platforms/javascript/express/#reporting-handled-errors) it seems to me that we should include the request whenever possible to get the added context. I think to have the full flexbility we should add the request as a second parameter and in logErrorAndMaybeSendToBugsnag call `req.bugsnag.notify()`

baker/SiteBaker.tsx
```diff
@@ -715,7 +715,7 @@ export class SiteBaker {
             return await execWrapper(cmd)
         } catch (error) {
             // Log error to Slack, but do not throw error
```
danyx23, line 717:
```suggestion
            // Log error to Bugsnag, but do not throw error
```

danyx23 (commented):
Looks good - I think it would be good to include the request data if we have it, otherwise I think we could go ahead with this once bugsnag is configured the way we like it and prod has the new env variables

danyx23:
@mlbrgl pinging you this way because officially this is my PR and I don't know which notifications you get :)

adminSiteServer/apiRouter.ts
mlbrgl: Make sense, thanks!

For reference, routes used for testing in app.tsx:
```js
        // routes to generate errors for testing
        app.get("/error", (req, res) => {
            throw new Error("This is a test error")
        })

        app.get("/log-error-with-req", (req, res) => {
            const error = new Error("This is a test error with req")
            logErrorAndMaybeSendToBugsnag(error, req)
            res.send("logged error")
        })

        app.get("/log-error-without-req", (req, res) => {
            const error = new Error("This is a test error without req")
            logErrorAndMaybeSendToBugsnag(error, req)
            res.send("logged error")
        })
```

mlbrgl: Logging error with `req`:

Screenshot 2023-05-26 at 10 45 38

PR #2216, opened 2023-05-16
:sparkles: add default for `minTime` to chart config schema

lucasrodes:
Add missing default values for `minTime` to chart config schema.

I also added linebreaks, which should make diffs easier to read in the future. Unsure if the previous styling was on purpose.


lucasrodes:
@danyx23 I am now unsure if we actually want this. For charts `DiscreteBar` or `StackedDiscreteBar`, it seems like the field `minTime` is not present. And this makes sense since we can not have a range selected in the time slider for this kind of charts ([example](http://localhost:3030/admin/charts/1567/edit)).

danyx23 (approved):
Agreed - merge it!

PR #2235, opened 2023-05-19
🔨 add @types/node-fetch as dev dependency

sophiamersmann:
Fixes https://owid.slack.com/archives/CQQUA2C2U/p1684485967941479


danyx23 (approved):
Good catch!

PR #2262, opened 2023-05-31
Research and writing block

ikesau:
Oof! That's a lot of links! ⛓️ 

[Staging preview](https://ptolemy.owid.cloud/admin/gdocs/1MfJkgmcH8-wM0Rd4AcksSrGaw6Hh4Uyrx2aouGge3gM/preview)

image

Overall a pretty simple component, just a lot of boilerplate.

For a while I was building it out such that if one tile/link was invalid, we'd show the error there instead of blocking out the entire component.

But I realized it was going to make validation convoluted for very little gain (seeing as we want authors to fix all these errors before they can publish anyway, so it's good enough to just treat it as a top level parse error)

@danyx23 if you agree with that verdict then I'll properly strip out the code that put parseErrors at the link level.


danyx23 (approved):
Looks good! I agree that we don't need individual errors - but make sure that it is understandable in the error message which link is broken. There also seem to be some conflicts vs master now but I assume those will be trivial. Can be merged from my side whenever you think it's ready!

PR #2266, opened 2023-05-31
fix: respect given height in gdoc charts

sophiamersmann:
I think `d.height` is accidentally ignored at the moment.


ikesau (approved):
aha, nice spotting!

PR #2278, opened 2023-06-01
Include featured image in wpPost -> ArchieML migration

ikesau:
We didn't have all the other plumbing for featured images set up when this migration script was written. I'm trying to add it in now.

### DIFF
--- a/baker/postUpdatedHook.ts
+++ b/baker/postUpdatedHook.ts
@@ -53,18 +53,36 @@ const syncPostToGrapher = async (
             left join wp_term_taxonomy t on t.term_taxonomy_id = r.term_taxonomy_id
             WHERE t.taxonomy = 'author' and p.ID = ?
             group by p.ID
-        )
+        ),
+        post_featured_image AS (
+        SELECT
+            p.ID,
+            (
+            SELECT
+                meta_value
+            FROM
+                wp_postmeta
+            WHERE
+                post_id = p.ID
+                AND meta_key = '_thumbnail_id') AS featured_image_id
+        FROM
+            wp_posts p
+        WHERE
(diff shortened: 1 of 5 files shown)


danyx23 (approved):
Looks good - I assume you have tried running the migrate script and it looks good? I haven't done that but if you have then this can be merged. 

After you merge this don't forget to run the script on prod so we get the thumbnails migrated :)

danyx23:
@ikesau was this work merged some other way or was this lost?

ikesau:
Looks like it didn't make it! 🤔 🙈 

Let me test it out again and merge if all's well

ikesau:
I've tested this and it's not working correctly. Will try to fix it soon :)

danyx23:
@ikesau let me know if you need help on this or want to push this over to my side - I will need this soon for the work on the automated "related research and writing" block for data pages - see #2736

danyx23:
@ikesau how should we proceed with this one? Could you have a look some time soon or should I try to take this over?

ikesau:
@danyx23 working! 👍 

To get these images in archie as part of the `migrateWpPostsToArchieMl` job, it requires the `posts` table `featured_image` column to be populated, which this PR does via the `syncPostToGrapher` function.

There's no mechanism to run this job on every entry in the `posts` table, as I'm not aware of any system we have to run a wpdb-dependent migration. 

Thus I think some small manual script to re-run `syncPostToGrapher` on each post might be the way to go? What do you think?

ikesau:
@danyx23 Code added to syncPostsToGrapher. Maybe just test it on your local too and then we can merge 🙂

danyx23 (approved):

PR #2282, opened 2023-06-02
:sparkles: Allow all page types on the homepage

ikesau:
The `featuredWork` feature only lets us specify posts that are in the blog roll.

Because we also want to feature topic pages we're changing this to allow all post types.

To test:
1. Ensure the [dev front page config document](https://docs.google.com/document/d/1LpZ5LFDTA6buEb_uL-IOWQC1YLAEbpj7odup-zgg1II/edit) is registered in your environment
2. Publish it
3. `yarn buildLocalBake --steps specialPages`
4. Check `localBake/index.html` and verify that the slugs specified in your published version of the config doc are present and in the correct order


danyx23 (approved):
Looks good - the staging server creation failed but I think that is unrelated to the changes. If it works for you locally then I'd say go ahead and merge :)

PR #2291, opened 2023-06-06
fix(facet): preserve relative mode after faceting

### LINKED ISSUE #2158: StackedArea: Can't go back to relative mode after faceting when relative toggle is hidden
## Description

For StackedAreaCharts with relative mode enabled, clicking on "Split by metric" shows facets in absolute mode (which makes sense), but switching back to the default view doesn't bring back relative mode. If the chart does not have a relative mode toggle, then the reader has no way to go back to relative mode.

## Expected behaviour

Re-apply relative mode when going back to the default view

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/parents-who-agree-with-spanking-to-discipline-a-child
2. Notice that the chart is in relative mode
3. Click on 'Split by metric'
4. Notice that facets are in absolute mode (which makes sense)
5. Click on 'All together'
6. Notice that the chart is in absolute mode and there is no way of going back to relative mode since the toggle is hidden

## Screenshots

Default view (relative mode):

Screenshot 2023-04-27 at 12 00 06

Faceted (absolute mode):

Screenshot 2023-04-27 at 12 00 12

End view (absolute mode):

Screenshot 2023-04-27 at 12 00 17

sophiamersmann:
fixes #2158

### Summary

- We don't allow relative mode in some cases, in particular when faceting
- Grapher didn't restore relative mode correctly when switching from a faceted view (where absolute mode was enforced) to the normal view
- Fixed by removing hacky code that manipulated `stackMode` in the facetStrategy setter

### Caveat

`hasSingleMetricInFacets`, `hasSingleEntityInFacets` and `isStackedChartSplitByMetric` should ideally be using `facetControl` (not `selectedFacetControl`) but we can't because that would result in a cyclic dependency.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -1555,6 +1555,13 @@ export class Grapher
     // NB: The timeline scatterplot in relative mode calculates changes relative
     // to the lower bound year rather than creating an arrow chart
     @computed get isRelativeMode(): boolean {
+        // don't allow relative mode in some cases
+        if (
+            this.hasSingleMetricInFacets ||
+            this.hasSingleEntityInFacets ||
+            this.isStackedChartSplitByMetric
+        )
+            return false
         return this.stackMode === StackMode.relative
     }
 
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
Yeah makes sense!

sophiamersmann:
Thank you for the review!

PR #2293, opened 2023-06-06
🎉 Rich text in gdoc footnotes

### LINKED ISSUE #2299: Gdoc lists can't have multi-line footnotes
In trying to reproduce the Poverty topic page in Gdocs, I've encountered an issue with the way we parse refs and lists.

To reproduce:
1. Create a list in Gdocs
2. Add `{ref}{/ref}` to the end of one of the list items and fill the tags with multiple paragraphs

The new paragraphs will be formatted by Gdocs as a list and render like so:

image

And if you try to break the ref onto a newline, the superscript number won't be placed correctly and any subsequent list items won't be parsed properly either.

## Possible solution
Create a `refs` top level archie block e.g.
```
[refs]
id: example_id
[.+content]
I am a 

multi-line ref
[]

id: another
[.+content]
Blah blah

Blah
[]
[]
```
As JSON:
```json
{
  "refs": [
    {
      "id": "example_id",
      "content": [
        {
          "type": "text",
          "value": "I am a"
        },
        {
          "type": "text",
          "value": "multi-line ref"
        }
      ]
    },
    {
      "id": "another",
      "content": [
        {
          "type": "text",
          "value": "Blah blah"
        },
        {
          "type": "text",
          "value": "Blah"
        }
      ]
    }
  ]
}
```

Then to use them:

`Due to limitations in survey methods{ref}example_id{/ref}, the following must be taken with a grain of salt`

This would also allow us to still use the inline ref syntax, and also let us cite the same source throughout an article without having to duplicate the content.

ikesau:
Resolves #2299

Changes refs to use a ID reference system instead of defining them inline.

Upsides:
- Multi-line refs in lists without messing up parsing. 
- Define a ref once and use it multiple times throughout an article.

We have three articles that are published with the old ref style. **Instead of writing a complex migration for them, I've opted to simply strip the refs out of the article during baking.** Once this PR is merged we can update the 3 source docs and republish them.

# Archie example
```
[.refs]
[.refs]
id: example_id
[.+content]
I am a ref
[]

id: multiline_id
[.+content]
I am a ref with [links](https://en.wikipedia.org/wiki/Wiki)

And multiple lines

And formatting

And
A list
whoa

[]
[]

[+.body]

If this works it’ll be great{ref}Some more inline stuff{/ref}

blah{ref}multiline_id{/ref}

blah blah{ref}example_id{/ref}

And an old style ref{ref}The [blah](https://en.wikipedia.org/wiki/Main_Page) of the blah blah has been known to blah.

I go multiple lines too{/ref}

And one more {ref}Some more inline stuff{/ref}

[]
```

# Preview example
image

### DIFF
--- a/db/model/Gdoc/archieToEnriched.ts
+++ b/db/model/Gdoc/archieToEnriched.ts
@@ -153,13 +153,13 @@ function formatCitation(
 export function extractRefs(text: string): {
     extractedText: string
     refsByFirstAppearance: Set<string>
-    parsedInlineRefs: RefDictionary
+    rawInlineRefs: unknown[]
 } {
     let extractedText = text
     const RefRegExp = "{ref}(.*?){/ref}"
 
     const refsByFirstAppearance = new Set<string>()
-    let parsedInlineRefs: RefDictionary = {}
+    const rawInlineRefs: unknown[] = []
     const rawRefStrings = text.match(new RegExp(RefRegExp, "gims")) || []
 
     for (const rawRef of rawRefStrings) {
@@ -195,28 +195,21 @@ export function extractRefs(text: string): {
         )
 
         if (isInlineRef) {
-            const { refs } = load(`
-                [.refs]
+            const rawInlineRef = load(`
                 id: ${id}
                 [.+content]
                 ${contentOrId}
                 []
-                []
             `)
-            const parsedInlineRef = parseRefs({
-                refs,
-                refsByFirstAppearance,
-                isInline: true,
-            })
-            parsedInlineRefs = merge(parsedInlineRefs, parsedInlineRef)
+            rawInlineRefs.push(rawInlineRef)
         }
     }
 
-    return { extractedText, refsByFirstAppearance, parsedInlineRefs }
+    return { extractedText, refsByFirstAppearance, rawInlineRefs }
 }
 
 export const archieToEnriched = (text: string): OwidGdocContent => {
-    const { extractedText, refsByFirstAppearance, parsedInlineRefs } =
+    const { extractedText, refsByFirstAppearance, rawInlineRefs } =
         extractRefs(text)
     text = extractedText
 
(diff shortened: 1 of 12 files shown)


ikesau:
todo: this will require a migration as well.

db/model/Gdoc/archieToEnriched.ts
```diff
@@ -177,7 +178,14 @@ export const archieToEnriched = (text: string): OwidGdocContent => {
...
 
-    parsed.refs = refs.map(htmlToEnrichedTextBlock)
+    parsed.refs = refs.map(({ blocks }) =>
+        blocks.map(parseRawBlocksToEnrichedBlocks)
+    )
+
+    console.log(
+        "JSON.stringify(parsed.refs, null, 2)",
+        JSON.stringify(parsed.refs, null, 2)
+    )
```
danyx23, line 188: Superfluous comment

danyx23 (commented):
This makes footnotes almost too powerful but I think in practice authors will not go totally crazy (... 🤞)

One log statement made it through and as you said the migration is still missing. Do you want to ping me again once that is added?

danyx23:
@ikesau I think it might be better to deal with two kinds of refs - old style refs (everything between the ref tags becomes the footnote) and single word refs which are treated as ids of items in the ref block. I think this would make for a nicer experience for existing content (there might be quite a few posts that use the old style that are not yet publisehd) and also it would be useful to keep the migration working. Do you see a big issue in doing this?

ikesau:
@danyx23 done!

I hash the contents of each inline ref so that we can potentially re-use footnote numbers if the contents is identical. This also allows us to mostly reuse the same logic that I've written for the ID-based ref system.

It's quite a big function though so let me know if you want a quick call to talk it through. 👍

danyx23 (changes requested):
Nice! This is getting very close. I found two cases of unexpected behaviour: 
* Inline refs lead to a document error entry about the hash not being defined (it works but the error is shown in the document sidepanel)
* When you define refs but then misspell the reference id at the point of use then you get a document error entry but also the footnote indicator renders, does not give you an on-hover footnote but there is an endnote entry that seems to reference the first ref (or the ref at the corresponding index?). See this example doc: https://docs.google.com/document/d/1ImYJfRy_cXvsRTsI1lYgXmh1vG1mMtvnXWrGuDtrGLw/edit

ikesau:
Oops! Very silly oversight on that inline ref error message. I've refactored slightly to enrich all the refs at the same time, which fixes the issue (and improves the readability of the code, imo)

I don't think it's worth adding the necessary complexity to improve how we render broken refs. 

If you misspell the ID of the ref, it's going to break and we won't let you publish the document. That seems fine to me.

danyx23:
Fair enough about the misspelling. I haven't checked it after your last changes but if you feel confident then feel free to go ahead and merge and tell authors about it :)

PR #2295, opened 2023-06-07
enhance(admin): add short name, version and number of charts to datasets table

sophiamersmann:
### Summary

Improves admin datasets table:
- renames "Dataspace" column to "Namespace"
- adds three columns:
    - dataset short name
    - version
    - number of charts

### Screenshot

Screenshot 2023-06-07 at 10 45 50

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -1764,6 +1764,14 @@ apiRouter.delete("/variables/:variableId", async (req: Request) => {
 
 apiRouter.get("/datasets.json", async (req) => {
     const datasets = await db.queryMysql(`
+        WITH variable_counts AS (
+            SELECT
+                v.datasetId,
+                COUNT(DISTINCT cd.chartId) as numCharts
+            FROM chart_dimensions cd
+            JOIN variables v ON cd.variableId = v.id
+            GROUP BY v.datasetId
+        )
         SELECT
             ad.id,
             ad.namespace,
@@ -1776,8 +1784,10 @@ apiRouter.get("/datasets.json", async (req) => {
             mu.fullName AS metadataEditedByUserName,
             ad.isPrivate,
             ad.nonRedistributable,
-            d.version
+            d.version,
+            vc.numCharts
         FROM active_datasets ad
+        JOIN variable_counts vc ON ad.id = vc.datasetId
         JOIN users du ON du.id=ad.dataEditedByUserId
         JOIN users mu ON mu.id=ad.metadataEditedByUserId
         JOIN datasets d ON d.id=ad.id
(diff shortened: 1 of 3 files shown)


adminSiteServer/apiRouter.ts
```diff
@@ -1765,34 +1765,60 @@ apiRouter.delete("/variables/:variableId", async (req: Request) => {
...
-        JOIN users mu ON mu.id=d.metadataEditedByUserId
-        ORDER BY d.dataEditedAt DESC
+            ad.isPrivate,
+            ad.nonRedistributable,
+            d.version
+        FROM active_datasets ad
+        JOIN users du ON du.id=ad.dataEditedByUserId
+        JOIN users mu ON mu.id=ad.metadataEditedByUserId
+        JOIN datasets d ON d.id=ad.id
+        ORDER BY ad.dataEditedAt DESC
```
danyx23, line 1784: Instead of getting the counts separately below and having to merge them locally it's more efficient and IMHO nicer to let SQL do this:
(suggestions don't work here for some reason)
```sql
with variable_counts as
(
SELECT
	v.datasetId as datasetId,
	COUNT(DISTINCT cd.chartId) as numCharts
FROM
	chart_dimensions cd
JOIN variables v ON
	cd.variableId = v.id
GROUP BY
	v.datasetId)
SELECT
	ad.id,
	ad.namespace,
	ad.name,
	d.shortName,
	ad.description,
	ad.dataEditedAt,
	du.fullName AS dataEditedByUserName,
	ad.metadataEditedAt,
	mu.fullName AS metadataEditedByUserName,
	ad.isPrivate,
	ad.nonRedistributable,
	d.version,
	vc.numCharts
FROM
	active_datasets ad
join variable_counts vc on
	ad.id = vc.datasetId
JOIN users du ON
	du.id = ad.dataEditedByUserId
JOIN users mu ON
	mu.id = ad.metadataEditedByUserId
JOIN datasets d ON
	d.id = ad.id
ORDER BY
	ad.dataEditedAt DESC
```

This uses a CTE (Common Table Expression) to basically define a temporary table that we then join with to get the count for each active dataset. The temporary table with the chart counts per dataset is constructed from the chart_dimensions table which is in a way a materialized "cache" of charts.config '$->>dimensions'.

danyx23 (changes requested):
Looks good! I have one request to change the SQL query - if you want to understand more how this works in detail feel free to grab me!

adminSiteServer/apiRouter.ts
sophiamersmann: I didn't know about chart_dimensions, that's brilliant!
Thank you for taking the time to write this down 🙏🏻  (I'm very keen to learn more sql stuff)

PR #2304, opened 2023-06-08
:honeybee: make the new gdocs env variable optional

larsyencken:
A new env var was introduced, causing all dev environments to fail validation if they did not have it configured. But in fact the variable is optional.

This PR introduces the ability to comment that a variable is optional, which then will mean that validation does not check for it.


ikesau (approved):
Tested locally: it does what it says on the tin :shipit:

danyx23 (approved):

PR #2316, opened 2023-06-13
:sparkles: Simpler schema publishing

danyx23:
This PR removes the old schema files and the json file of the current schema. This leaves only grapher-schema.003.yaml as the current schema file to edit. 

The github action that publishes the schema files is modified to create the json file from the yaml and to create a .latest.yml and .latest.json copy. These 4 files are then uploaded to S3 so they are available.

Going forward, if we do breaking schema changes we will rename the schema.yaml file. This will lead to the new version number being added, the old existing ones remain on S3 unchanged and latest will always be a copy of the most current file.

When fetching the .latest.json schema file, inspecting the $id field will always tell you the current version. Grahpher config files that are persisted should always include a $schema key that points to the authorative url of the json schema they adhere to (i.e. not .latest.json but .003.json or whatever)


marcelgerber (approved):
Cool, thank you!

I haven't tried out the nushell scripts to see that they do the right thing, but I trust that they do :)

PR #2323, opened 2023-06-15
Improve chart title

### LINKED ISSUE #2320: Chart title improvements
Discussion on [Slack](https://owid.slack.com/archives/C46U9LXRR/p1686593864029519)

We often get questions about the "Hide automatic entity" and "Hide automatic time" toggles because their behaviour can be difficult to understand.

Current state:
- Entity and time are automatically added to chart titles
- Authors can choose to hide these fields using the "Hide automatic entity" and "Hide automatic time" toggles
- In some cases, Grapher prevents hiding the time or entity if doing so would cause crucial information to be missing from the chart title (this is especially important for svg exports that don't have a timeline or controls)

Improvements:
- Fix bug: Prevent hiding the time field for discrete bar charts (see [Slack](https://owid.slack.com/archives/C46U9LXRR/p1686593864029519))
- Make the admin UI more intuitive:
    - Consider adding help text or an info icon
    - Consider hiding the toggles if they are not applicable
- Consider forcing the addition of entity/time when exporting the chart to svg (but allow time/entity to be hidden for interactive charts)

sophiamersmann:
resolves #2320 (not urgent)

Collection of small improvements around the annotation fields in chart titles:

- Add DiscreteBar, StackedDiscreteBar, and Marimekko charts to the list of chart types where hiding time is not allowed
    - These chart types only refer to the current time in the timeline (for SVG exports without controls, the current time would be missing if not mentioned in the title)
- Hide the "Hide automatic time" and "Hide automatic entity" toggles in cases where Grapher ignores their state
- Add help text to better explain the behaviour

----

The SVG tester fails four charts that I think should in fact all mention the time in the title.

### DIFF
--- a/adminSiteClient/EditorFeatures.tsx
+++ b/adminSiteClient/EditorFeatures.tsx
@@ -113,4 +113,29 @@ export class EditorFeatures {
                 this.grapher.isLineChart)
         )
     }
+
+    @computed get showChangeInPrefixToggle() {
+        return (
+            this.grapher.isLineChart &&
+            (this.grapher.isRelativeMode || this.grapher.canToggleRelativeMode)
+        )
+    }
+
+    @computed get showEntityAnnotationInTitleToggle() {
+        return (
+            !this.grapher.canChangeEntity &&
+            !this.grapher.canSelectMultipleEntities
+        )
+    }
+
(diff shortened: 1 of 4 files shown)


marcelgerber (approved):
Makes sense, cool!

PR #2333, opened 2023-06-19
feat(explorer): allow indicator metadata to be overwritten

sophiamersmann:
relates to #2313

### Summary

Allows indicator metadata in explorers to be overwritten. (Technically, column definitions are overwritten.) New columns that are derived from indicators can be added using the `transform` keyword.

#### Overwriting metadata

- if a `columns` block is not linked to a table (i.e. the cell next to the `columns` keyword is empty), then the manually provided column definitions map to the indicators referenced in the `graphers` block to create charts
- mapping is done via a new column called `indicatorId` that is added to the `columns` grammar
- the manually provided column definition for an indicator is merged with the original column definition of that indicator

#### Adding derived columns

- add a new column that is derived from indicators by adding an entry to the `columns` block that has a `slug` and a `transform`
- the `transform` string can refer to indicators by id, e.g. `multiplyBy 539022 2` multiples the indicator with id `539022` by 2
- to create a chart from a derived column, refer to it by slug in the `ySlugs` column

### Screenshots

Screenshot 2023-06-26 at 15 49 43

### To do

- [x] Chart creation mode: Be more explicit about what is supported and what is not (explain that multi-content explorers are meant to be dropped in the future #2374)
- [x] Rename indicator to variable

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -4,7 +4,7 @@ import {
     BlankOwidTable,
     ColumnTypeNames,
     CoreColumnDef,
-    extractDataSlugsFromTransform,
+    extractPotentialDataSlugsFromTransform,
     isNotErrorValue,
     OwidColumnDef,
     OwidTable,
@@ -419,7 +419,8 @@ export class Explorer
     ): string[] {
         const def = this.columnDefsNotLinkedToTableByIdOrSlug[slug] ?? {}
         if (!def.transform) return baseColumns
-        const dataSlugs = extractDataSlugsFromTransform(def.transform)
+        const dataSlugs =
+            extractPotentialDataSlugsFromTransform(def.transform) ?? []
         for (let i = 0; i < dataSlugs.length; i++) {
             this.getBaseColumnsForColumnWithTransform(dataSlugs[i], baseColumns)
             baseColumns.push(dataSlugs[i]) // deepest dependency at the start of the array
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -347,13 +347,13 @@ export class Explorer
     @computed private get chartCreationMode(): ExplorerChartCreationMode {
         const { decisionMatrix, grapherConfig } = this.explorerProgram
         const { grapherId } = grapherConfig
-        const yIndicatorIdsColumn = decisionMatrix.table.get(
-            GrapherGrammar.yIndicatorIds.keyword
+        const yVariableIdsColumn = decisionMatrix.table.get(
+            GrapherGrammar.yVariableIds.keyword
         )
-        // referring to an indicator in a single row triggers
-        // ExplorerChartCreationMode.FromIndicatorIds for all rows
-        if (yIndicatorIdsColumn.numValues)
-            return ExplorerChartCreationMode.FromIndicatorIds
+        // referring to a variable in a single row triggers
+        // ExplorerChartCreationMode.FromVariableIds for all rows
+        if (yVariableIdsColumn.numValues)
+            return ExplorerChartCreationMode.FromVariableIds
         if (grapherId && isNotErrorValue(grapherId))
             return ExplorerChartCreationMode.FromGrapherId
         return ExplorerChartCreationMode.FromExplorerTableColumnSlugs
@@ -364,8 +364,8 @@ export class Explorer
             case ExplorerChartCreationMode.FromGrapherId:
                 this.updateGrapherFromExplorerUsingGrapherId()
                 break
-            case ExplorerChartCreationMode.FromIndicatorIds:
-                this.updateGrapherFromExplorerUsingIndicatorIds()
+            case ExplorerChartCreationMode.FromVariableIds:
+                this.updateGrapherFromExplorerUsingVariableIds()
                 break
             case ExplorerChartCreationMode.FromExplorerTableColumnSlugs:
                 this.updateGrapherFromExplorerUsingColumnSlugs()
@@ -429,22 +429,22 @@ export class Explorer
         ])
     }
 
-    // gets the IDs of all indicators that a transformed column depends on;
+    // gets the IDs of all variables that a transformed column depends on;
     // for example, if a there are two columns, 'slug' and 'other_slug', that
     // are defined by the transforms 'divideBy 170775 other_slug' and 'multiplyBy 539022 2',
-    // respectively, then getBaseIndicatorIdsForColumnWithTransform('slug')
-    // returns ['539022', '170775'] as these are the IDs of the two indicators
+    // respectively, then getBaseVariableIdsForColumnWithTransform('slug')
+    // returns ['539022', '170775'] as these are the IDs of the two variables
     // that the 'slug' column depends on
-    private getBaseIndicatorIdsForColumnWithTransform(slug: string): string[] {
+    private getBaseVariableIdsForColumnWithTransform(slug: string): string[] {
         const { columnDefsWithoutTableSlug } = this.explorerProgram
-        const baseIndicatorIdsAndColumnSlugs =
+        const baseVariableIdsAndColumnSlugs =
             this.getBaseColumnsForColumnWithTransform(slug)
         const slugsInColumnBlock: string[] = columnDefsWithoutTableSlug
             .filter((def) => !def.owidVariableId)
             .map((def) => def.slug)
-        return baseIndicatorIdsAndColumnSlugs.filter(
-            (indicatorIdOrColumnSlug) =>
-                !slugsInColumnBlock.includes(indicatorIdOrColumnSlug)
+        return baseVariableIdsAndColumnSlugs.filter(
+            (variableIdOrColumnSlug) =>
+                !slugsInColumnBlock.includes(variableIdOrColumnSlug)
         )
     }
 
@@ -469,14 +469,14 @@ export class Explorer
         grapher.downloadData()
     }
 
-    @action.bound private async updateGrapherFromExplorerUsingIndicatorIds() {
+    @action.bound private async updateGrapherFromExplorerUsingVariableIds() {
         const grapher = this.grapher
         if (!grapher) return
         const {
-            yIndicatorIds = "",
-            xIndicatorId,
-            colorIndicatorId,
-            sizeIndicatorId,
+            yVariableIds = "",
+            xVariableId,
+            colorVariableId,
+            sizeVariableId,
             ySlugs = "",
             xSlug,
             colorSlug,
@@ -489,47 +489,47 @@ export class Explorer
             hideEntityControls: this.showExplorerControls,
         }
 
-        // set given indicator IDs as dimensions to make Grapher
-        // download the data and metadata for these indicators
+        // set given variable IDs as dimensions to make Grapher
+        // download the data and metadata for these variables
         const dimensions = config.dimensions ?? []
-        if (yIndicatorIds) {
-            const yIndicatorIdsList = yIndicatorIds
+        if (yVariableIds) {
+            const yVariableIdsList = yVariableIds
                 .split(" ")
                 .map((item) => parseInt(item, 10))
                 .filter((item) => !isNaN(item))
-            yIndicatorIdsList.forEach((yIndicatorId) => {
+            yVariableIdsList.forEach((yVariableId) => {
                 dimensions.push({
-                    variableId: yIndicatorId,
+                    variableId: yVariableId,
                     property: DimensionProperty.y,
                 })
             })
         }
-        if (xIndicatorId) {
+        if (xVariableId) {
             dimensions.push({
-                variableId: xIndicatorId,
+                variableId: xVariableId,
                 property: DimensionProperty.x,
             })
         }
-        if (colorIndicatorId) {
+        if (colorVariableId) {
             dimensions.push({
-                variableId: colorIndicatorId,
+                variableId: colorVariableId,
                 property: DimensionProperty.color,
             })
         }
-        if (sizeIndicatorId) {
+        if (sizeVariableId) {
             dimensions.push({
-                variableId: sizeIndicatorId,
+                variableId: sizeVariableId,
                 property: DimensionProperty.size,
             })
         }
 
-        // Slugs that are used to create a chart refer to columns derived from indicators
+        // Slugs that are used to create a chart refer to columns derived from variables
         // by a transform string (e.g. 'multiplyBy 539022 2'). To render such a chart, we
-        // need to download the data for all indicators the transformed columns depend on
+        // need to download the data for all variables the transformed columns depend on
         // and construct an appropriate Grapher table. This is done in three steps:
-        //   1. find all indicators that the transformed columns depend on and add them to
+        //   1. find all variables that the transformed columns depend on and add them to
         //      the config's dimensions array
-        //   2. download data and metadata of the indicators
+        //   2. download data and metadata of the variables
         //   3. append the transformed columns to the Grapher table (note that this includes
         //      intermediate columns that are defined for multi-step transforms but are not
         //      referred to in any Grapher row)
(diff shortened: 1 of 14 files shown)


sophiamersmann:
* master
  * **PR #2332** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2332" target="_blank">Graphite</a> 
    * **PR #2333** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2333" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2333?utm_source=stack-comment).

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -311,23 +399,35 @@ const availableTransforms: any = {
...
-            (word) => availableTransforms[word] !== undefined
-        )
-        if (!transformName) {
-            console.warn(`Warning: transform '${transformName}' not found`)
-            return
-        }
-        const params = words.filter((word) => word !== transformName)
-        const fn = availableTransforms[transformName]
+    for (let i = 0; i < defs.length; i++) {
+        const def = defs[i]
```
marcelgerber, line 422: I like to use indexing for-loops only when the index is actually relevant in the loop; otherwise a for-of loop is more straightforward in my mind.
```suggestion
    for (const def of defs) {
```

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -340,6 +440,29 @@ export const applyTransforms = (
...
+    }
     return columnStore
 }
+
+export const extractDataSlugsFromTransform = (
+    transform: string
+): ColumnSlug[] => {
+    const { transformName, params = [] } =
+        extractTransformNameAndParams(transform) ?? {}
+    if (!transformName) return []
```
marcelgerber, line 452: We may want to return `undefined` in this case?

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -217,88 +271,122 @@ enum WhereOperators {
...
-
-    return values.map((value, index) =>
-        passes(conditionValues[index]) ? value : ErrorValueTypes.FilteredValue
-    )
+const where: Transform = {
+    params: [
+        { type: TransformParamType.DataSlug },
+        {
+            type: TransformParamType.ColumnSlug,
+        },
```
marcelgerber, line 279: Open question to me right now: What's the difference between the two?

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -340,6 +440,29 @@ export const applyTransforms = (
...
+    const dataParams = availableTransforms[transformName].params
+        .map((param, index) => ({ ...param, index }))
+        .filter(
+            (param) =>
+                param.type === TransformParamType.DataSlug ||
+                param.type === TransformParamType.ColumnSlug // might be a data slug
+        )
+    if (dataParams.length === 0) return []
+    const dataParamInds = dataParams.map((param) => param.index)
+    const dataSlugs = params.filter((_, index) => dataParamInds.includes(index))
```
marcelgerber, line 462: You can use `zip(availableTransforms[transformName].params, params)` maybe to make that all easier (lodash's `zip` works just like in Python), and to not have to deal with keeping track of indices.

packages/@ourworldindata/core-table/src/Transforms.ts
marcelgerber, line 1: Nice job specifying the input types to a transform, I think that'll also help us in transitioning to a more sane syntax, just like we talked about last week. 🏅

explorer/ExplorerProgram.ts
```diff
@@ -254,6 +254,10 @@ export class ExplorerProgram extends GridProgram {
         return columnDefs
     }
 
+    get columnDefsNotLinkedToTable(): OwidColumnDef[] {
+        return this.columnDefsByTableSlug.get(undefined) ?? []
+    }
```
marcelgerber, line 259: I think you know this, but:
If an explorer only has one table, then that table can be unnamed, and in that case the `tableSlug` column is also empty or nonexistent.
I guess in that case, `columnDefsByTableSlug.get(undefined)` would just return every column def??

---

If it's easier, we can also totally migrate that weirdness away, and just change all existing explorers to have named tables only, and remove all code paths that handle the special case.

explorer/ExplorerProgram.ts
```diff
@@ -409,3 +423,41 @@ export const trimAndParseObject = (config: any, grammar: Grammar) => {
...
+        ])
+        .renameColumn("indicatorId", "owidVariableId")
+        .combineColumns(
+            ["slug", "owidVariableId"],
+            {
+                slug: "slugOrIndicatorId",
+                type: ColumnTypeNames.String,
+                name: "slugOrIndicatorId",
+            },
+            (values) => values.slug ?? values.owidVariableId.toString()
```
marcelgerber, line 445:
```suggestion
            (values) => values.slug ?? values.owidVariableId?.toString()
```

explorer/Explorer.tsx
```diff
@@ -341,130 +343,270 @@ export class Explorer
...
+        slug: string,
+        baseColumns: string[] = []
+    ): string[] {
+        const def = this.columnDefsNotLinkedToTableByIdOrSlug[slug] ?? {}
+        if (!def.transform) return baseColumns
+        const dataSlugs = extractDataSlugsFromTransform(def.transform)
+        for (let i = 0; i < dataSlugs.length; i++) {
+            this.getBaseColumnsForColumnWithTransform(dataSlugs[i], baseColumns)
+            baseColumns.push(dataSlugs[i]) // deepest dependency at the start of the array
+        }
```
marcelgerber, line 423:
```suggestion
        for (const dataSlug of dataSlugs) {
            this.getBaseColumnsForColumnWithTransform(dataSlug, baseColumns)
            baseColumns.push(dataSlug) // deepest dependency at the start of the array
        }
```

explorer/Explorer.tsx
```diff
@@ -341,130 +343,270 @@ export class Explorer
...
+        if (!grapher.id) grapher.id = 0
+    }
+
+    @computed private get columnDefsNotLinkedToTableByIdOrSlug(): Record<
+        number | string,
+        OwidColumnDef
+    > {
+        const { columnDefsNotLinkedToTable } = this.explorerProgram
+        return keyBy(
+            columnDefsNotLinkedToTable as OwidColumnDef[],
```
marcelgerber, line 408: The typecast shouldn't be necessary?
```suggestion
            columnDefsNotLinkedToTable,
```

explorer/Explorer.tsx
```diff
@@ -341,130 +343,270 @@ export class Explorer
...
-        const grapherConfig =
-            creationMode === ExplorerChartCreationMode.FromGrapherId
-                ? this.grapherConfigs.get(grapherId!) ?? {}
-                : {}
+    @action.bound private getBaseColumnsForColumnWithTransform(
+        slug: string,
+        baseColumns: string[] = []
+    ): string[] {
+        const def = this.columnDefsNotLinkedToTableByIdOrSlug[slug] ?? {}
+        if (!def.transform) return baseColumns
```
marcelgerber, line 418:
```suggestion
        const def = this.columnDefsNotLinkedToTableByIdOrSlug[slug]
        if (!def?.transform) return baseColumns
```

marcelgerber:
## Review progress

I reviewed all but `Explorer.tsx` at this point, which is arguably where the meat of this is.

Review TODOs:
- [ ] Take another good look at `Explorer.tsx`
- [ ] Run through `Explorer.tsx` in a debugger to see what happens, or alternatively in a call with Sophia
- [ ] Hunt for edge cases: Unnamed tables? Any amount of other weird things?
- [ ] In the above screenshot, how does it work for the `life-expectancy-doubled` column? The relevant explorer view is in `ExplorerChartCreationMode.FromExplorerTableColumnSlugs` mode, and if I'm reading it right then the kind of dependency-tracking doesn't apply in this case?

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -741,7 +741,7 @@ export class Grapher
     }
 
     @action.bound
-    private async downloadLegacyDataFromOwidVariableIds(): Promise<void> {
+    async downloadLegacyDataFromOwidVariableIds(): Promise<void> {
```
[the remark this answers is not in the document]
sophiamersmann: Agreed! Maybe we should write a clean-up issue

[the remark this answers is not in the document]
sophiamersmann: (I got used to it but it was very confusing for me when I started)

packages/@ourworldindata/core-table/src/CoreColumnDef.ts
```diff
@@ -54,6 +54,7 @@ export interface CoreColumnDef extends ColumnColorScale {
 
     // Computational
     transform?: string // Code that maps to a CoreTable transform
+    transformHasRun?: boolean // If true, the transform has been applied
```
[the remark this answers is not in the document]
sophiamersmann: Yeah, you're right. I'll move it!

packages/@ourworldindata/core-table/src/Transforms.ts
sophiamersmann: `ColumnSlug` is a slug for a column with entities, years or data values. `DataSlug` only refers to columns with data values. So, every `DataSlug` is a `ColumnSlug` but not every `ColumnSlug` is a `DataSlug`

packages/@ourworldindata/core-table/src/Transforms.ts
sophiamersmann: :)

explorer/ExplorerProgram.ts
sophiamersmann: Ahh, yes, you're right. I guess it doesn't matter that much because indicator-based chart creation and csv data-based chart creation can't be mixed at the moment. But the name is def not fitting then

sophiamersmann:
I will let you finish your review and then do revisions in one go. seems less confusing to me :) Thank you!

packages/@ourworldindata/grapher/src/core/Grapher.tsx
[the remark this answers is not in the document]
sophiamersmann: I added an issue here, so that we might get to it at some point: #2378

explorer/Explorer.tsx
```diff
@@ -341,130 +343,275 @@ export class Explorer
...
+        if (uniqueSlugsInGrapherRow.length) {
+            const allColumnSlugs = uniq(
+                uniqueSlugsInGrapherRow.flatMap((slug) => [
+                    ...this.getBaseColumnsForColumnWithTransform(slug),
+                    slug,
+                ])
             )
-            this.futureGrapherTable.set(this.tableLoader.get(tableSlug))
+            const existentColumnSlugs = grapherTable.columnSlugs
+            const requiredColumnSlugs = allColumnSlugs.filter(
```
marcelgerber, line 579: Maybe using the word `outstanding` makes it clearer what's going on here
```suggestion
            const outstandingColumnSlugs = allColumnSlugs.filter(
```

explorer/Explorer.tsx
```diff
@@ -341,130 +343,275 @@ export class Explorer
...
+        // add derived columns to grapher table
+        if (uniqueSlugsInGrapherRow.length) {
+            const allColumnSlugs = uniq(
+                uniqueSlugsInGrapherRow.flatMap((slug) => [
+                    ...this.getBaseColumnsForColumnWithTransform(slug),
+                    slug,
+                ])
             )
-            this.futureGrapherTable.set(this.tableLoader.get(tableSlug))
+            const existentColumnSlugs = grapherTable.columnSlugs
```
marcelgerber, line 578:
```suggestion
            const existingColumnSlugs = grapherTable.columnSlugs
```

explorer/Explorer.tsx
```diff
@@ -341,130 +343,275 @@ export class Explorer
...
-        if (
-            creationMode ===
-            ExplorerChartCreationMode.FromExplorerTableColumnSlugs
-        ) {
-            // Clear any error messages, they are likely to be related to dataset loading.
-            this.grapher?.clearErrors()
-            // Set a table immediately. A BlankTable shows a loading animation.
-            this.setGrapherTable(
-                BlankOwidTable(tableSlug, `Loading table '${tableSlug}'`)
+        // add derived columns to grapher table
```
marcelgerber, line 570: It honestly took me way longer than I'd like to admit until I realized why we need to this. This is either a sign of my dwindling attention, or a sign that there should be more of a comment on why we do this, and why we need all this sophisticated dependency tracking.

I hope I'm getting this right:
We need all of that because the transform expects its "source" data columns to all be there in the `OwidTable`, but that's not a given when we construct the table based on indicator ids and the `dimensions` array.
Whereas with csv inputs we would get an `OwidTable` with all columns that are present in the csv file, here we instead get one that is as slim as possible, only containing the variables we directly specify.

However, columns that are source columns for transforms are only indirectly specified (in the transform), and we do all this work to make sure that we detect the dependency on them and also include them in the `dimensions` array, so they are then also part of the table we construct and the transform can run without issues.

marcelgerber (changes requested):
What a beast of a PR!
I kind of had a hard time making sense of all the dependency tracking, and only at the very end had the epiphany of why we need it all.

Still, it would maybe be good to add comments here and there, or maybe explain what a method does based on an example transform. Tests would also be great for documenting this, but I think in many cases they would be quite hard to create.

explorer/Explorer.tsx
sophiamersmann: yes, that's a perfect summary

if spelled out, these are the basic steps:
- given are a bunch of transforms that reference indicators by id or other intermediate columns
- first, figure out which data needs to be downloaded and put those in the `dimensions` array
- after running `downloadLegacyDataFromOwidVariableIds`, the table contains the indicators needed for the transforms (but not the transformed and intermediate columns themselves)
- as a last step, append the transformed and intermediate columns to the input table (this is when the transforms actually run)

that dependency tracking is necessary, like you said, because the price of simply downloading all indicators in the explorers when we encounter a transform would be too high

explorer/Explorer.tsx
```diff
@@ -341,130 +343,275 @@ export class Explorer
...
-            creationMode = ExplorerChartCreationMode.FromIndicatorIds
-        else
-            creationMode =
-                ExplorerChartCreationMode.FromExplorerTableColumnSlugs
-
-        const grapherConfig =
-            creationMode === ExplorerChartCreationMode.FromGrapherId
-                ? this.grapherConfigs.get(grapherId!) ?? {}
-                : {}
+    @action.bound private getBaseColumnsForColumnWithTransform(
```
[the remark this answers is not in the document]
sophiamersmann: 💡

explorer/Explorer.tsx
```diff
@@ -341,130 +343,275 @@ export class Explorer
...
+        slug: string,
+        baseColumns: string[] = []
+    ): string[] {
```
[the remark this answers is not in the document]
marcelgerber: Yes exactly!

explorer/Explorer.tsx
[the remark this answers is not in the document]
sophiamersmann: Removing `@action.bound` from `getBaseColumnsForColumnWithTransform` worked but removing it from `getBaseIndicatorIdsForColumnWithTransform` broke the code 🤷🏻‍♀️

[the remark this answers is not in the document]
marcelgerber: Ah, interesting. I think if you change the call to an array function like this:
```ts
                uniqueSlugsInGrapherRow.flatMap(
                    (slug) => this.getBaseIndicatorIdsForColumnWithTransform(slug)
                )
```

that should work? `this` behaves differently in arrow functions, and I think that's what would be breaking things here then.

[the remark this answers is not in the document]
sophiamersmann: excellent! that worked 👌

sophiamersmann:
Revisions are done ✨

Moving `transformHasRun` from `CoreColumnDef` to `CoreColum` doesn't really work because `CoreTable.transform` creates a new table and only passes the data values and column definitions. I tried accessing the `parent` table to check if a transform has already been applied (though that might be an anti-pattern) and I tried passing a new advanced option to retain information regarding transforms – but both didn't work (not sure why).

sophiamersmann:
Based on @danyx23 feedback, I added an alert for cases where authors try to use indicators as well as grapher Ids or tabular data to create charts (which we don't support). That looks like this:

Screenshot 2023-07-05 at 14 42 33

sophiamersmann:
@sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2333).

sophiamersmann:
@sophiamersmann merged this pull request with [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2333).

PR #2335, opened 2023-06-20
fix(explorers): respect entity type when explorers are embedded

### LINKED ISSUE #1008: Embedded explorers do not respect entity type
Hi, I'm not totally sure if this is the right repo to raise this bug, but noticed this on https://ourworldindata.org/environmental-impacts-of-food
image

sophiamersmann:
fixes #1008 

Makes embedded explorers respect the entity type.

Instead of passing a selection array to explorers, we only pass the selected entity names, then construct the selection array in the explorer component with the correct entity type.


sophiamersmann:
I think this breaks the global entity selector. Closed in favour of #2337

PR #2342, opened 2023-06-20
fix(explorer): respect entity type when explorer is embedded

### LINKED ISSUE #1008: Embedded explorers do not respect entity type

sophiamersmann:
fixes #1008 

My third attempt to make embedded explorers respect the entity type.


sophiamersmann:
haha I'm actually glad you're saying that – I don't particularly like the solution I have come up with. I'll look into it!

sophiamersmann:
😭😭😭  

Thanks for your help!

PR #2345, opened 2023-06-21
fix(explorer): reset time when single-year view is not meaningful

### LINKED ISSUE #1693: Transition from map to faceted line chart should discard time selection
## Steps to reproduce

1. Navigate to https://ourworldindata.org/explorers/poverty-explorer
2. Ensure you are viewing the map version of a chart
3. Change the "Poverty line" dropdown to "Multiple lines"

## Observed behaviour

You get a bar chart of of just one of the multiple poverty variables, without indicating which one.

## Expected behaviour

- (ideal) Time selection is discarded and you get faceted line charts with full time-span
- (less ideal, but valid) Faceted bar charts for the specific time point in the map

## Visual example

https://user-images.githubusercontent.com/36012/193861844-d7b767f6-40c4-4035-b364-827c612d480b.mp4

sophiamersmann:
fixes #1693

Explorers persist the time when switching charts. This makes sense for most cases, but it can sometimes lead to sub-optimal chart views, such as collapsed line charts or stacked bar charts with only a single year. This came up in #1693, specifically for the case where switching from a map (i.e. single-year view) to a chart led to a collapsed line chart.

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -18,6 +18,7 @@ import {
     GrapherManager,
     GrapherProgrammaticInterface,
     GrapherQueryParams,
+    GrapherTabOption,
     SelectionArray,
     setSelectedEntityNamesParam,
     SlideShowController,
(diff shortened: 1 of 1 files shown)


sophiamersmann:
I'm closing this pr – I'm not comfortable with the level of micromanagement

PR #2355, opened 2023-06-21
🎉 update scrollY when collapsing expandable-parapraph

ikesau:
Another little QOL thing asked for by Marwa

https://github.com/owid/owid-grapher/assets/11844404/62bb6f39-8827-43e6-ad7d-0941610cc7a8

### DIFF
--- a/site/blocks/ExpandableParagraph.tsx
+++ b/site/blocks/ExpandableParagraph.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { useLayoutEffect, useRef, useState } from "react"
 import cx from "classnames"
 import ReactDOM from "react-dom"
 import ReactDOMServer from "react-dom/server.js"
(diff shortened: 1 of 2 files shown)


mlbrgl (approved):
I can't quite pinpoint it but I'm somehow not convinced by the asymmetry of the open / close behaviour. It's less confusing than before though, especially with longer texts.

ikesau:
@mlbrgl I think you're right - refactored it to animate instead.

Tested on Safari/Chrome/Firefox

https://github.com/owid/owid-grapher/assets/11844404/04a35b1a-9d58-47b4-b295-99fe3c1ccbf9

mlbrgl:
@ikesau nice! I think that's clearer now 👍

I just noticed that the margin with the text got reduced [here](http://staging-site-expandable-paragraph-location/poverty) and feels a bit too close (right is live): 
Screenshot 2023-06-22 at 18 29 47

ikesau:
Hmm yeah that's not up to date. Is the CSS cached somehow?

PR #2361, opened 2023-06-22
🐛 pre-fetch linkedCharts before baking

ikesau:
Found a pretty glaring oversight in the Gdoc baking process - we weren't including `linkedCharts`, which are needed to resolve any URLs / metadata to charts/explorers within a Gdoc.

It's a slightly more involved process because we need to resolve redirects first.

e.g. if a gdoc has a prominent link:

```
{.prominent-link}
url: https://ourworldindata.org/grapher/some-slug
{}
```
We use that slug to get the title and thumbnail for the prominent link.

If we update that slug, a redirect is created and we have to then resolve that redirect first before we can fetch the correct metadata.

An individual Gdoc already can already do this by calling `loadLinkedCharts`, but I didn't want to call that once per document so I fetch every single published chart & explorer and attach them instead. 

The reason we don't just do that all the time is because, for previews, we need to pass `linkedCharts` (along with the rest of the Gdoc) to the browser so passing 5000 grapher configs isn't really feasible.

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -14,6 +14,7 @@ import {
     IMAGE_HOSTING_CDN_URL,
     IMAGE_HOSTING_BUCKET_SUBFOLDER_PATH,
     GDOCS_DETAILS_ON_DEMAND_ID,
+    BAKED_GRAPHER_URL,
 } from "../settings/serverSettings.js"
 
 import {
@@ -47,6 +48,7 @@ import {
     clone,
     getFilenameAsPng,
     extractDetailsFromSyntax,
+    LinkedChart,
 } from "@ourworldindata/utils"
 import { execWrapper } from "../db/execWrapper.js"
 import { countryProfileSpecs } from "../site/countryProfileProjects.js"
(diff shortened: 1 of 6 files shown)


packages/@ourworldindata/utils/src/owidTypes.ts
```diff
@@ -1113,8 +1113,8 @@ export interface OwidGdocTag {
 
 // A minimal object containing metadata needed for rendering prominent links etc in the client
 export interface LinkedChart {
-    slug: string
-    path: string
```
ikesau, line 1117: This implementation was actually flat out broken. I don't know why I went with such ambiguous names the first time 😵

baker/SiteBaker.tsx
```diff
@@ -307,17 +314,55 @@ export class SiteBaker {
...
             publishedGdoc.imageMetadata = imageMetadataDictionary
             publishedGdoc.linkedDocuments = publishedGdocsDictionary
-            const publishedExplorersBySlug =
-                await this.explorerAdminServer.getAllPublishedExplorersBySlugCached()
+            publishedGdoc.linkedCharts = {
+                ...publishedChartsBySlug,
+                ...publishedExplorersBySlug,
+            }
+            // this is a no-op if the gdoc doesn't have an all-chart block
+            await publishedGdoc.loadRelatedCharts()
```
ikesau, line 365: This is from #2326 , where I made this commit initially but then cherry-picked over to a new branch to isolate it for a PR.

It fixes a bug so it's fine to merge it here too.

marcelgerber (changes requested):
Yeah makes sense overall!
But found a few cases where this will still go wrong atm 🤐

marcelgerber (approved):

PR #2386, opened 2023-06-29
chore(grapher): add population id

sophiamersmann:
Discussion on [Slack](https://owid.slack.com/archives/C46U9LXRR/p1670855145129159)

Adds variable 597930 to the list of variable IDs whose source shouldn't be mentioned in the footer when used in Scatter plots and Marimekko charts.

I wasn't sure whether to add 597930 to the list or replace 597929 with 597930. 597929 seems to be still in use, so I opted for the former. And should we also replace 597929 with 597930 here? https://github.com/owid/owid-grapher/blob/a47c39ea877c26818882eb6926b0f46295e493c8/adminSiteClient/EditorBasicTab.tsx#L312

I want to merge this today, so the fix goes live. Can you have a quick look at this when you're back on Monday, @marcelgerber? Thank you!

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -1389,7 +1389,8 @@ export class Grapher
         const isPopulationVariableId = (id: string): boolean =>
             id === "525709" || // "Population (historical + projections), Gapminder, HYDE & UN"
             id === "525711" || // "Population (historical estimates), Gapminder, HYDE & UN"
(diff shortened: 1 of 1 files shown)


marcelgerber (commented):
Great!

The difference between the two is that
- 597929 is historical estimates + future projections
- 597929 is only historical estimates

So, imo it's fine to be using both into the future, and also that 597929 remains the default in scatterplots.

sophiamersmann:
Ahh, I see! Thanks for looking at this so quickly (and on your free day!) 🙏🏻

PR #2391, opened 2023-07-03
:tada: add origins and origins_variables table

Marigold:
New tables needed by [origins in ETL](https://github.com/owid/etl/pull/1246).


db/migration/1687770112891-CreateOriginsTable.ts
```diff
@@ -0,0 +1,44 @@
...
+            id INT AUTO_INCREMENT PRIMARY KEY,
+            datasetTitleOwid VARCHAR(512),
+            datasetDescriptionOwid TEXT,
+            datasetDescriptionProducer TEXT,
+            producer VARCHAR(255),
+            citationProducer TEXT,
+            datasetUrlMain TEXT,
+            datasetUrlDownload TEXT,
+            dateAccessed DATE,
+            datePublished DATE
```
danyx23, line 16: I think this should maybe be a text field since it can be just a year - what do you think?

danyx23: Varchar actually

danyx23:
@larsyencken can you have a look at the DB changes and see if there is anything that you disagree with? Thanks!

db/migration/1688372371221-AddPresentationToVariables.ts
```diff
@@ -0,0 +1,91 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class AddPresentationToVariables1688372371221
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            `ALTER TABLE variables
+                ADD COLUMN schemaVersion INT NOT NULL DEFAULT 1,
+                ADD COLUMN processingLevel ENUM('minor', 'major'),
```
larsyencken, line 10: Enum fields are thought to be bad news in MySQL since they're quite annoying to change later and can have surprising properties.

Better to use a short `VARCHAR(30)` here as well.

If you want to, you can add a check constraint on it, they're new to MySQL since 8.0.16 (we're on 8.0.28):

```sql
ALTER TABLE variables
ADD CONSTRAINT checkProcessingLevel CHECK (processingLevel IN ('minor', 'major');
```

db/migration/1688372371221-AddPresentationToVariables.ts
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+        // make variables.souceId nullable
+        await queryRunner.query(
+            `ALTER TABLE variables MODIFY sourceId int NULL;`
+        )
+
+        await queryRunner.query(`
+        CREATE TABLE posts_gdocs_variables_faqs (
+            gdocId varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
```
larsyencken, line 37: Any reason for the special charset and collation? (leave a one-line comment perhaps)

db/migration/1688372371221-AddPresentationToVariables.ts
```diff
@@ -0,0 +1,91 @@
...
+                -- ADD COLUMN topicTagsLinks JSON,
+                ADD COLUMN keyInfoText JSON,
+                ADD COLUMN processingInfo TEXT,
+                ADD COLUMN licenses JSON,
+                ADD COLUMN presentationLicense JSON,
+                ADD COLUMN grapherConfigETL JSON;`
+        )
+        await queryRunner.query(
+            `ALTER TABLE datasets
+                ADD COLUMN updatePeriod INT;`
```
larsyencken, line 27: An integer duration field should really have its unit as part of the name. E.g. `updatePeriodMonths`

db/migration/1688372371221-AddPresentationToVariables.ts
```diff
@@ -0,0 +1,91 @@
...
+                ADD COLUMN producerShort TEXT,
+                ADD COLUMN citationInline TEXT,
+                ADD COLUMN descriptionShort TEXT,
+                ADD COLUMN descriptionFromProducer TEXT,
+                -- ADD COLUMN topicTagsLinks JSON,
+                ADD COLUMN keyInfoText JSON,
+                ADD COLUMN processingInfo TEXT,
+                ADD COLUMN licenses JSON,
+                ADD COLUMN presentationLicense JSON,
+                ADD COLUMN grapherConfigETL JSON;`
```
larsyencken, line 23: Are we set on having two `grapherConfig` objects on the variable? If so, could we rename the other one, to say what it's meant for?

If the ETL one is always the base, perhaps it could be `grapherConfigBase` and `grapherConfigOverlay`, or `grapherConfigETL` and `grapherConfigAdmin`?

larsyencken (changes requested):
Hey, just some small feedback and suggestions here.

I just looked at it as a whole, I didn't double check the exact fields against the schema, so I'm assuming that side is all good.

db/migration/1688372371221-AddPresentationToVariables.ts
Marigold: Good to know, changed it to `VARCHAR(30)`. (I didn't add the constraint, we validate this field in ETL anyway)

db/migration/1688372371221-AddPresentationToVariables.ts
Marigold: This is a foreign key to `posts_gdocs.id` which has this charset and collation (no idea why). I'll add a comment.

db/migration/1688372371221-AddPresentationToVariables.ts
Marigold: That makes sense, @danyx23 are you ok with this change?

db/migration/1688372371221-AddPresentationToVariables.ts
danyx23: I don't know when or how exactly this happened but our database has the default collation to do case INsensitive comparisons but then most (all?) tables define it as case sensitive which is why we have this explicitly in quite a few places.

db/migration/1688372371221-AddPresentationToVariables.ts
danyx23: yeah good idea although what we have is `updatePeriodDays`

db/migration/1688372371221-AddPresentationToVariables.ts
danyx23: Yes I think it's cleaner if we have two grapher objects for now. I think grapherConfigAdmin is a name that is easier to understand. 

My current best guess is that we'll want to use grapherConfigETL as our primary config. grapherConfigAdmin could be used as the place where we can store the copy of currently existing single variable grapher charts. This way, the grapherConfigAdmin can be filled once for a lot of variables, be written to from the Variable Annotation tool in the admin and is going to be immune from overwrites in the ETL. In the medium term we can then hopefully curate all the grapher config changes in the ETL and then evenutally drop grapherConfigAdmin.

danyx23:
Thank you @larsyencken! It's good to have a fresh set of eyes on this once more. @Marigold I added the two remaining todos to the notion table.

db/migration/1688372371221-AddPresentationToVariables.ts
```diff
@@ -0,0 +1,103 @@
...
+                ADD COLUMN schemaVersion INT NOT NULL DEFAULT 1,
+                ADD COLUMN processingLevel VARCHAR(30),
+                ADD COLUMN processingLog JSON,
+                ADD COLUMN titlePublic VARCHAR(512),
+                ADD COLUMN titleVariant VARCHAR(255),
+                ADD COLUMN producerShort VARCHAR(512),
+                ADD COLUMN citationInline TEXT,
+                ADD COLUMN descriptionShort TEXT,
+                ADD COLUMN descriptionFromProducer TEXT,
+                -- ADD COLUMN topicTagsLinks JSON,
```
ikesau, line 19: I see we're creating `tags_variables_topic_tags` instead of this. Does this mean the python model needs to change? Or should we maybe add a comment here explaining the discrepancy?

db/migration/1688372371221-AddPresentationToVariables.ts
ikesau: I first encountered this when wanting to create a FK for `posts_gdocs` - gdoc IDs _are_ case-sensitive but I don't think that would affect the operations we do with that table, either way.

@danyx23 `UpdateCharaterSet1647947484386` collated everything to `COLLATE utf8mb4_0900_as_cs` - just for my own understanding, does that concur with what you're saying about our DB's default collation?

db/migration/1688372371221-AddPresentationToVariables.ts
Marigold: I removed this outdated line. You're right there's a python ORM model, I'm trying to keep it in sync with these changes.

danyx23:
@ikesau for some reason I can't reply inline to your question about collation. So my theory is that this has to do with the wrong default collation that we had on our dev servers. Our prod server has the case sensitive version `utf8mb4_0900_as_cs` as the default, but our dev servers accidentally had `utf8mb4_0900_ai_csi` as the default. For all tables that we got from prod it didn't matter as the collation is specified in the create table statement - so on every refresh, we would sync and recreate the existing tables locally and this would set them to the correct collation.

But if you created new tables or relationships locally then those would fall back to the wrong dev default and then it became necessary to set the utf8mb4_0900_as_cs collation explicitly as in here. 

If all of the above is correct then with a dev database that is created today (e.g. via `make refresh`), or on prod, the above should work exactly the same without explicitly specifying the collation. I'm tempted to try this but have to finish a presentation rn :)

PR #2399, opened 2023-07-04
Add password hints

danyx23:
When authors use a staging server admin they are often puzzled about the user/password screen because normally we use cloudflare. This PR adds hints with the default password


sophiamersmann (approved):
Nice, I think that's helpful!
The email address is cut off for me, and I'm not sure there is a way to see the complete email? If this is just me not knowing how to operate a computer, please feel free to ignore!

Screenshot 2023-07-10 at 10 37 47

PR #2413, opened 2023-07-10
Sync explorers to Algolia

ikesau:
See #2053

Extracts all the unique descriptive text from our Explorer entries in the DB, chunks them, and uploads them to Algolia.


danyx23:
@ikesau is this still relevant? If so we should pin it

ikesau:
I'm going to close this and merge #2533 into master instead of a feature branch; review of the indexing script can be done in #2533 instead.

PR #2415, opened 2023-07-11
Fixes bug where narrative charts stopped responding to gdoc changes

sophiamersmann:
Sometimes narrative charts stopped reacting to changes made in the source gdoc, first described by Joe on [Slack](https://owid.slack.com/archives/C46U9LXRR/p1688466499021549).

### How to reproduce the problem

- Add a chart with a custom title to a gdoc
- Note that all controls are hidden
- Configure the chart to show a particular control, e.g. the `entitySelector`
- Note that the entity selector is visible
- Edit the gdoc to hide the entity selector by removing the `entitySelector` keyword
- Note that the entity selector is still visible
- Note that the console logs a hydration diff:
    - On the server: `hideEntityControls` is `false`
    - On the client: `hideEntityControls` is `true`

### Why this happened

- The `merge` function merges all objects into the first one
- `grapherInterfaceWithHiddenControlsOnly` was thus updated on each iteration, and the mutated `grapherInterfaceWithHiddenControlsOnly` lingered around
- This led to unexpected errors (like the one described)
- This only happened on the server and not on the client, hence the hydration diff

### DIFF
--- a/site/gdocs/Chart.tsx
+++ b/site/gdocs/Chart.tsx
@@ -43,10 +43,9 @@ export default function Chart({
             .filter(identity) as GrapherProgrammaticInterface[]
 
         config = merge(
-            !showAllControls
-                ? { ...grapherInterfaceWithHiddenControlsOnly }
-                : {},
-            !showAllTabs ? { ...grapherInterfaceWithHiddenTabsOnly } : {},
+            {},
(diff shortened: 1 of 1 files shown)


marcelgerber (commented):
Very good, and incredibly annoying, find!

marcelgerber (approved):

ikesau (approved):

Issue #2392, opened 2023-07-03
Add Origins to Grapher and extend variables DB classes and variable metadata


larsyencken:
After the database has gained new fields on the variables, the datasets table and new relationship tables were added in #1286 and #2481, we need to bring the the grapher DB types into alignment again.

- [x] add OwidOrigin
- [x] add new fields on OwidVariable
- [x] add/update the database classes
- [x] make Grapher use Origin in the current designs so that Origin-only indicators will still render correctly

larsyencken:
@danyx23 This one is all ticked off, should it be closed?

danyx23:
@larsyencken I'll close the issue when the PRs are merged which should happen later today, finally :)

PR #2417, opened 2023-07-11
:hammer: update types to match new metadata schema

danyx23:
This implements #2392. It adds an Origin class where a Variable can now have one Source and/or multiple Origins. It also adds several metadata fields to Variable that were added to the database.

The full list of added fields can be seen in the [proposal in Notion](https://www.notion.so/owid/2023-06-26-Proposed-metadata-structure-that-enable-data-pages-83b3602d12004796a2034fc870911137?pvs=4).


ikesau:
Hey @danyx23 - still playing around and testing things. A lot of what I've been trying has worked exactly as I'd expected, so that's good! 😅 

Most of my questions so far have been answered by TODOS so far, but I'll keep poking around to make sure I understand everything tomorrow 👍

danyx23:
@ikesau thanks a lot!

danyx23:
@ikesau thanks a lot for going through it all and exploring a very different corner of our codebase and infra :)

Yes indeed - the idea is now that only FAQ blocks are going to be written in gdocs with gdoc component support. The rest of the fields like key insights, description from source etc can only use markdown (our own markdown flavour with DoD support, hence the recent #2489). FAQs will only show up data pages for now but many of the other texts will show up in all sorts of places like the redesigned sources tab. I think it will help us keep the technical complexity at bay if all the texts are limited to a rather restricted set of markup, and only FAQs can go crazy.

ikesau:
Cool okay, that sounds good.

Do you plan to update the variable pages in the admin with this new metadata too?

ikesau (commented):
Hey @danyx23 

I spent a while tinkering on this branch to continue to piece together where everything's coming from 😅 

image

I started modifying `attribution: ACME project` in `dummy.meta.yml` and re-running the ETL, where everything worked as expected.

When I commented out the line though, the ETL would run and update the DB with a a `NULL` attribution, but in the FE, `ACME project` would still be visible... is it coming from `dummy.csv.dvc` ? 

Beyond that, nothing broke, so I suppose in some sense it's working - but if there are more things you'd like me to (or think it would be good for me to) try, let me know!

danyx23:
@ikesau what is the "FE"?

PR #2420, opened 2023-07-12
Add schema property to Grapher

sophiamersmann:
We recently introduced the `$schema` field in Grapher configs that points to the json schema that the config is checked against. This PR makes Grapher aware that the schema field exists such that:
- New charts are being assigned a schema
- Updated charts keep their schema

The `$schema` property is special because we always want it to be present, even if it is the default. (Grapher doesn't usually include default values in configs.) This means that we loose the beauty of having an empty Grapher serialize to an empty object :(

Aside: I called the new Grapher property `$schema` (with `$`) to make it clear that this maps to the `$schema` config field but I can see how that might be controversial.


danyx23 (approved):
Looks good, thanks for catching and fixing this!

PR #2426, opened 2023-07-13
Change default header background color for articles 

mathisonian:
This changes the default color from a bright yellow to a more subtle blue. We would like to use this color for the SDG pages, and think it makes sense as the default generally for all articles, although we may revisit this at some point. 

See Marwa's comment on it in slack [here](https://owid.slack.com/archives/C040429G2SD/p1689233175622379).

Screen Shot 2023-07-13 at 12 30 43 PM


ikesau:
This effectively makes #2423 redundant. We should close that PR and add `amber` as a valid value for `cover-color` here.

marcelgerber (approved):

PR #2427, opened 2023-07-13
🔨 disable `no-inferrable-types`

ikesau:
Something recently changed with eslint causing it to start [removing inferrable type annotations](https://typescript-eslint.io/rules/no-inferrable-types/) 

e.g.

```typescript

const obviouslyANumber: number = 1

// changes to

const obviouslyANumber = 1
```

But this breaks TypeORM which requires these annotations

e.g.

```typescript
@Column() slug: string = ""
```

With them removed, the Grapher admin throws an error as soon as you try to load it:

```
DataTypeNotSupportedError: Data type "Object" in "Gdoc.slug" is not supported by "mysql" database.
```

So this PR disables this rule for our project.


ikesau:
This appears to have been an issue with. my local VSCode instance.

PR #2436, opened 2023-07-18
Implement icons on the SDG homepage

ikesau:
Took a little while to find a nice way to handle the colour animations. I went with the mask+background strategy as it's [well-supported](https://caniuse.com/css-masks) and still lets use use images.

https://github.com/owid/owid-grapher/assets/11844404/0689774b-db2e-443b-97e3-8f683e6928dc

I also found that breadcrumbs with particularly long items start to wrap and overflow on mobile. I fixed this with a naive length calculation that sets a class on the header with a slightly higher margin at our small breakpoint.

An alternative would be to restructure our header so that the flow of the breadcrumbs affects the position of the white background, but that seemed like less of a sure thing when our launch is tomorrow.

https://github.com/owid/owid-grapher/assets/11844404/1094885c-e8a8-458b-befe-a2f4c919cecc


marcelgerber (approved):
Real nice work! I've just pushed up three commits that make some small changes, but nothing visible.
I also tested many times that all is working well in Chrome/Edge, Firefox, Safari - great work!

PR #2431, opened 2023-07-14
chore: add explorer variables table

sophiamersmann:
> **Warning**
> Not ready to be merged.

Adds a new table to the database that keeps track of indicators that are used in explorers.

PR #2437, opened 2023-07-18
fix: bake indicators used in explorers

sophiamersmann:
Bakes indicators used in explorers.

Gets a list of indicators that are currently used in explorers from the `explorer_variables` table that is created in https://github.com/owid/owid-grapher/pull/2431 and populated in https://github.com/owid/automation/pull/8


danyx23 (approved):
Looks good!

PR #2446, opened 2023-07-18
🔨 bake gdoc images in chunks

ikesau:
This should be optimized further to not rebake/resize every image on every bake, but because this is breaking deploys this is a temporary quick fix.


marcelgerber (approved):

danyx23:
@ikesau not urgent but can you add a comment there as well that this should be optimized further? In case we forget to follow up then there would be another little nudge in the code to tell us to do this :)

PR #2447, opened 2023-07-18
:tada: enable bugsnag performance monitoring

danyx23:
We have a free trial for bugsnag performance monitoring which looks very useful for our switch to serving data directly to the client from S3. This PR adds the required library and enables sending preformance data to Bugsnag


marcelgerber (approved):
Beautifully easy, let's go 🚀

PR #2463, opened 2023-07-19
Change url of `/blog` to `/latest`

mathisonian:
This PR updates the URL of our blogroll from `/blog` and `/blog/*` to `/latest` and `/latest/*` as requested by Max. It includes redirect rules as well in case the old URLs are linked. I believe we should also update `serverSettings.BLOG_SLUG` to reflect this. 

I'm having issues testing this locally due to a failing DB migration that I don't have time to look into atm, so would appreciate a thorough test/review on this. I've marked it as a draft PR for now until it gets properly tested.


mathisonian:
Okay, have tested this locally now. All looks good except when running locally I see this when a redirect is expected. 

Screen Shot 2023-07-19 at 6 05 58 PM

marcelgerber (approved):
Very nice!

Yes, redirects won't be applied on a local dev server, nor will they be applied on an automated staging server [they will be applied on Netlify-hosted staging servers, though]. So that's to be expected.

There was a very old empty WP page that used `/blog` as its slug; that one I just deleted so it won't interfere with this work.

Issue #2468, opened 2023-07-19
Nested lists in Gdocs


ikesau:
Currently we can only have a single level of bullet points in our Gdoc lists.

In [gdocToArchie](https://github.com/owid/owid-grapher/blob/2d388caded627e0bbc5af4aa994c0e7c2d6e8221/db/model/Gdoc/gdocToArchie.ts#L32) we convert Gdoc lists to Archie lists ("[Arrays of strings](http://archieml.org/)") which look like this:

```
[.list]
* Broccoli
* Spinach
* Apples
* Banana
[]
```

But this syntax has no support for nesting.

We'll have to strip this part of the code out and change the way we parse google doc lists to some new shape. This new shape could be quite complex if we want to support Google Docs' arbitrary levels of nesting:

<img width="351" alt="image" src="https://github.com/owid/owid-grapher/assets/11844404/7275fa3f-321a-401e-a3d2-24a42aa2917c">

As we'll be manually manipulating string data which will be parsed by the Archie parser, it's important to ensure that this new syntax is correctly placed on a newline so as to not break parsing when [following styled text.](https://github.com/owid/owid-grapher/pull/2470)

### Archie:
```
{.list}
listType: ul
[.+children]

[.+li]
I am a list item
[]

{.list}
listType: ol
[.+children]

[.+li]
I am a nested ordered list item
[]

[]
{}

[]

{}
```

### TypeScript:
```ts
type RawListType = "ul" | "ol"

interface RawListItem =  {
    type: "li"
    value: RawBlockText[]
  }

interface RawListBlock {
  type: "list"
  listType: RawListType
  children: (RawListBlock | RawListItem)[]
}
```

ikesau:
Currently, lists also break when italicizing their items. We should ensure that works correctly when implementing this.

### Gdoc
![image](https://github.com/owid/owid-grapher/assets/11844404/1f0e22bb-b8a3-4272-aaf3-61afa0bc3b13)

### Archie
```
[.list]
* <i>Apple
</i>* <i>Banana
</i>* <i>Orange</i>
[]
```

ikesau:
I personally like the idea of our articles supporting this functionality, but we've gotten this far without our authors asking for it so it's probably safe to close.

PR #2429, opened 2023-07-13
🐛 fix bold newline parsing issue

ikesau:
Fixes #2330

Input: 
https://github.com/owid/owid-grapher/assets/11844404/2d18125d-cd95-4348-ab43-16f73edadd60

Before
```
[+body]
<b>I am some bold stuff
</b>{.heading} // ERROR
text: Will this break?
level: 2
{}
[]
```

After:
```
[+body]
<b>I am some bold stuff
</b>
{.heading}
text: Will this break?
level: 2
{}
[]
```

PR #2470, opened 2023-07-20
🐛 fix lists breaking when following formatting tags

ikesau:
The same fix as https://github.com/owid/owid-grapher/pull/2429 but for the other syntax that we're manually converting atm.

We'll need to keep this in mind when implementing https://github.com/owid/owid-grapher/issues/2468 as well.


danyx23 (approved):
:shipit:

PR #2478, opened 2023-07-21
Datapage preview from metadata

danyx23:


site/DataPageV2Content.tsx
```diff
@@ -0,0 +1,973 @@
...
+                                    <RelatedCharts
+                                        charts={datapageData.allCharts}
+                                    />
+                                </div>
+                            </div>
+                        ) : null}
+                    </div>
+                    <div className="bg-gray-10">
+                        <div className="wrapper">
+                            {faqEntries?.faqs.length && (
```
ikesau, line 577: needs !!

db/model/Gdoc/Gdoc.ts
```diff
@@ -709,6 +709,21 @@ export class Gdoc extends BaseEntity implements OwidGdocInterface {
             dodDocumentErrors.push(...errors)
         }
 
+        const faqs = this.content.faqs
```
ikesau, line 712: Hmmm, the dod document doesn't do this either, but we're not tracking links in these articles. oops!

We should fix that because we're counting on the links table being reliable for our content graph / 404 tracker.

db/model/Variable.ts
```diff
@@ -127,6 +127,23 @@ export function parseVariableRows(
...
 
+export async function getMergedGrapherConfigForVariable(
+    variableId: number
+): Promise<GrapherInterface | undefined> {
+    const row = await db.mysqlFirst(
+        `SELECT grapherConfigAdmin, grapherConfigETL FROM variables WHERE id = ?`,
+        [variableId]
+    )
+    if (!row) return
+    const grapherConfigAdmin = row.grapherConfigAdmin
```
ikesau, line 138: Could also have just used `&&`'s here 🙂

packages/@ourworldindata/utils/src/GdocsConstants.ts
```diff
@@ -18,6 +18,7 @@ export const IMAGES_DIRECTORY = "/images/published/"
 
 export const gdocUrlRegex =
     /https:\/\/docs\.google\.com\/.+?\/d\/([-\w]+)\/?(edit)?#?/
+export const gdocIdRegex = /^[0-9A-Za-z\-]{44}$/
```
ikesau, line 21: Gdoc IDs can have underscores in them also 🙂

ikesau (commented):
Hey @danyx23 

Didn't manage to review/test everything by EOD, I'd like to do some more testing in the morning, but hopefully these comments give you something to chew on for a bit 😉

adminSiteServer/mockSiteRouter.tsx
```diff
@@ -156,6 +159,15 @@ mockSiteRouter.get("/charts", async (req, res) => {
     res.send(await renderChartsPage(explorerAdminServer))
 })
 
+mockSiteRouter.get("/datapage-preview/:id", async (req, res) => {
+    console.log("previewing datapage")
+    const variableId = expectInt(req.params.id)
+    const variableMetadata = await getVariableMetadata(variableId)
+    if (!variableMetadata) throw new JsonError("No such variable", 404)
+
+    res.send(await renderDataPageV2(variableId, variableMetadata, undefined))
```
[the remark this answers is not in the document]
danyx23: I think so - would you prefer an optional parameter?

baker/GrapherBaker.tsx
```diff
@@ -185,6 +202,112 @@ export const renderDataPageOrGrapherPage = async (
...
+    const grapher = mergePartialGrapherConfigs(
+        grapherConfigForVariable as Record<string, unknown>,
+        pageGrapher as Record<string, unknown>
+    )
+
+    const faqDocs = compact(
+        uniq(variableMetadata.presentation?.faqs?.map((faq) => faq.gdocId))
+    )
+    const gdocFetchPromises = faqDocs.map((gdocId) =>
+        getDatapageGdoc(gdocId, true, publishedExplorersBySlug)
```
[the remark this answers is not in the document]
danyx23: Ah yes of course, good catch

baker/GrapherBaker.tsx
```diff
@@ -185,6 +202,112 @@ export const renderDataPageOrGrapherPage = async (
...
+export async function renderDataPageV2(
+    variableId: number,
+    variableMetadata: OwidVariableWithSource,
+    pageGrapher: GrapherInterface | undefined,
+    publishedExplorersBySlug?: Record<string, ExplorerProgram>
+) {
+    const grapherConfigForVariable = await getMergedGrapherConfigForVariable(
+        variableId
+    )
+    const grapher = mergePartialGrapherConfigs(
```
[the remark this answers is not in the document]
danyx23: Nice

baker/GrapherBaker.tsx
```diff
@@ -43,15 +52,23 @@ import {
...
     getDatapageJson,
     parseGdocContentFromAllowedLevelOneHeadings,
 } from "../datapage/Datapage.js"
 import { ExplorerProgram } from "../explorer/ExplorerProgram.js"
 import { Image } from "../db/model/Image.js"
 import { logErrorAndMaybeSendToBugsnag } from "../serverUtils/errorLog.js"
+import {
+    FaqEntryData,
+    FaqDictionary,
+} from "@ourworldindata/utils/dist/owidTypes.js"
```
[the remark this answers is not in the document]
danyx23: This was auto-added - I moved this up to the imports from utils above

datapage/Datapage.ts
```diff
@@ -75,16 +144,18 @@ export const getDatapageJson = async (
...
     isPreviewing: boolean,
     publishedExplorersBySlug?: Record<string, ExplorerProgram>
 ): Promise<OwidGdocInterface | null> => {
     // Get the google doc id from the datapage JSON file and return early if
     // none found
-    const googleDocId =
-        getLinkType(googleDocEditLink) === "gdoc"
-            ? getUrlTarget(googleDocEditLink)
-            : null
+    const isPlainGoogleId = gdocIdRegex.exec(googleDocEditLinkOrId)
```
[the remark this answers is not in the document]
danyx23: Instead of having the full edit urls in the ETL? Yeah I'm pretty sure we want the ids only. But I'm open to passing this explicitly - i.e. have a googleDocId field that can be undefined and a goolgeDocEditLink that can be undefined and then use the former if it is given or the latter if no id is there. What do you think?

db/model/Gdoc/Gdoc.ts
danyx23: You mean that they should be included in the link tracking a few lines above? Can you outline what needs to be done? Then I could add it for both

db/model/Variable.ts
danyx23: Yeah but I kinda like this style a bit more

packages/@ourworldindata/utils/src/GdocsConstants.ts
danyx23: Good catch :)

site/DataPageV2.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const imageUrl: string = urljoin(baseUrl, "default-grapher-thumbnail.png")
+    const imageWidth = "1200"
+    const imageHeight = "628"
+
+    const variableIds: number[] = uniq(
+        compact(grapher?.dimensions?.map((d) => d.variableId))
+    )
+
+    const mergedGrapherConfig = mergePartialGrapherConfigs(
+        datapageData.chartConfig as GrapherInterface,
```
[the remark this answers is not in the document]
danyx23: The types are different here because we have not lifted the type definitions up to the types project yet, so this cast is still required unfortunately

site/DataPageV2Content.tsx
danyx23: Ok - but why?

db/model/Gdoc/Gdoc.ts
ikesau: I think adding them to `blocksToTraverse` in `Gdoc.tsx` ought to be all you need to do.

datapage/Datapage.ts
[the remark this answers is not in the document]
ikesau: Okay. 🙂 

In that case I think `googleDocEditLinkOrId` is fine, but it would be nice if we didn't have to write these 6 lines of control flow each time we work with it (assuming we'll have other places we'll ingest it). Maybe a `getGoogleIdFromString: (x: string) => string | undefined` function (which, for adoption's sake, we could also then call in `getUrlTarget` ?)

e.g.

```ts
const googleDocId = getGoogleIdFromString(googleDocEditLinkOrId)
```

and in `getUrlTarget`

```ts
if (url.isGoogleDoc) {
  return getGoogleIdFromString(urlString) as string
}
```

Sorry this is quite hair-splitting stuff! 😬 

I've just spent a decent amount of time trying to be consistent throughout all the Gdocs code so we can call the same couple of utility functions over and over, so I'd like to continue to strive for that. 🙏

datapage/Datapage.ts
```diff
@@ -13,9 +14,76 @@ import {
...
+import { GrapherInterface } from "@ourworldindata/grapher"
+
+interface ETLPathComponents {
+    channel: string
+    publisher: string
+    version: string
+    dataset: string
+}
+
+const getETLPathComonents = (path: string): ETLPathComponents => {
```
ikesau, line 34: Components*

ikesau (approved):
Huge effort! Really cool to see how far the ETL is getting us now.

PR #2480, opened 2023-07-24
:sparkles: Disable variable metadata fields in admin

Marigold:
Disable all variable metadata fields. These fields should be only set in ETL / fast-track, otherwise we risk overwriting manual updates by automatic ETL run.

I didn't disable "Redistribution is prohibited" field for dataset. We don't have that setting in ETL (therefore we don't overwrite it) and I thought it could be useful? We could also move it to ETL and remove the `Update` button altogether.

### DIFF
--- a/adminSiteClient/DatasetEditPage.tsx
+++ b/adminSiteClient/DatasetEditPage.tsx
@@ -80,7 +80,6 @@ class DatasetEditable {
 class DatasetTagEditor extends React.Component<{
     newDataset: DatasetEditable
     availableTags: { id: number; name: string; parentName: string }[]
-    isBulkImport: boolean
 }> {
     @action.bound onSaveTags(tags: Tag[]) {
         this.props.newDataset.tags = tags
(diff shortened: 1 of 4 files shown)


danyx23 (changes requested):
Looks good and it's nice that we can do this :)

I think it would be nice to remove the corresponding code from the admin API server as well.

Also, how does archiving work again? Is this done only in the DB? Or in the dataset admin UI? I don't think it is affected but just to sanity check :)

Once the admin part is removed (variable) or restricted to only the necessary fields (dataset) this is good to merge from my side

Marigold:
> I think it would be nice to remove the corresponding code from the admin API server as well.

Done!

> Also, how does archiving work again? Is this done only in the DB? Or in the dataset admin UI? I don't think it is affected but just to sanity check :)

You can archive a dataset from admin UI. When you archive a dataset in ETL (i.e. move it `archive/dag.yml`) you also have to archive it manually in admin.

danyx23 (approved):
Nice! :shipit:

Issue #2486, opened 2023-07-24
Tasks: Redesign of Grapher's Configuration Popup


samizdatco:
### Background
- [Cycle 2023.07+08 (Christian, Sophia): Grapher Redesign](https://www.notion.so/owid/Cycle-2023-07-08-Christian-Sophia-Grapher-Redesign-d32d47cb95f14e8fb74b7eae40ec7141)

### See Also
- #2432

## To Do:
- [ ] Create popup-style container for controls triggered by 'gear' button in the main grapher view
- [ ] Implement within-container controls for the full set of graph-specific setttings currently handled on an ad-hoc basis:
   - [ ] alignAxisScalesToggle
   - [ ] facetControl
   - [ ] noDataAreaToggle
   - [ ] relativeToggle
   - [ ] xLogLinearSelector
   - [ ] yLogLinearSelector
   - [ ] zoomToggle
- [ ] support selection of controls via archieML keywords, including main-view controls:
   - [ ] timeline
   - [ ] entitySelector
- [ ] on `/grapher/…` pages, use a fly-in 'drawer' rather than a pop-up menu
- [ ] update the settings immediately upon change, but hide pop-up/drawer only when dismissed

pablo-mayrgundter:
Hi!

I have some suggestions For the Choose Data to Show/Country chooser panel, and found this issue.  Hope these help!

For context, I was just comparing countries in https://ourworldindata.org/grapher/child-stunting-ihme and didn't see any groupings available, tho this would help gain insight.

Also, I'm happy to prepare the lists I propose.  I can also try a PR but my first attempt to build the project failed.

1) Provide (static) common country groupings for all datasets, incl. those currently without e.g. Child Stunting by IHME & WHO, as noted.
 - Continents & Sub-regions (e.g. North Africa, Sub-Saharan, Middle East, Central America, Caribbean, Oceania)
 - Geographical: Size by quintile, Landlocked, Island, Desert
 - Climate (Köppen system): Tropical, Arid/Desert, Temperate, Continental, Polar, Highland
 - Major Demographics: Size by quintile, Young, Aging, Stable
 - Health & Education: High/Low of Literacy/Life Expectancy/Fertility
 - Economic Level: GDP by quintile, WDI or Developed, Emerging, Least Developed
 - Major political-economic groupings: EU, G20, NATO, BRICS, NAFTA, ASEAN, OPEC, APEC, AU, SAARC
[...]

PR #2511, opened 2023-08-06
Redesigned Settings Menu

samizdatco:
This PR moves the collection of one-off controls that are contextually inserted above the chart into a pop-up dialog box which is summoned by a button in the upper right corner labeled Settings. On grapher and explorer pages, a `nav#grapher-settings-drawer` element has been added to the markup outside of the grapher instance. When present, the settings menu will use a portal to render its contents into this `<nav>` (which slides in from the right edge of the viewport). If the `<nav>` is not present, the menu is presented as a pop-up whose bounds are fit to the containing grapher instance.

Tracking issue: #2486


danyx23:
@samizdatco is this PR still relevant or has this work gone elsewhere? Stalebot just closed it...

samizdatco:
It's still active. Reopening…

PR #2516, opened 2023-08-07
:hammer: disable chunking of images as a band aid for baking issues

danyx23:
This should hopefully fix the staging server baking issues we've had lately

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -598,7 +598,9 @@ export class SiteBaker {
             "published"
         )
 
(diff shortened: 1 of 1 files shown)


ikesau (approved):
At this point, you could also get rid of the chunking logic entirely and just iterate through `images` 

Given the TODO it's nbd either way 👍

danyx23:
Yeah I considered it but I thought maybe we want this again in a few days as another stop-gap measure and then just setting to 1 is more straightforward

PR #2540, opened 2023-08-11
feat(explorer): add duplicate transform

### LINKED ISSUE #2514: Enable indicator name selection for explorers on the chart level
## Core problem

In explorers, `name` in the `columns` table sets default indicator names across all charts in an explorer. But sometimes, we would like to use a different name depending on the chart to make them easier to understand.

For example, in [this chart](https://owid.cloud/admin/explorers/preview/history-of-war?facet=none&country=~OWID_WRL&Dataset=Uppsala+Conflict+Data+Program&Metric=Conflict+deaths&Conflict+type=All+armed+conflicts&Sub-metric=by+type) we want to display estimates across several conflict types, so the names should identify the conflict type. But in [this chart](https://owid.cloud/admin/explorers/preview/history-of-war?facet=none&country=~OWID_WRL&Dataset=Uppsala+Conflict+Data+Program&Metric=Conflict+deaths&Conflict+type=Interstate+conflicts&Sub-metric=by+type), we only look at one conflict type but three different estimates, so the names should identify the estimate type. Though currently, the best estimate has to have the same name.

## Proposed solution
Ideally, we would be able to set `name` at the chart level (such as by specifying them in the order of the indicators).

## Alternatives
Currently, we have to devise names that work well across charts, or duplicate indicators to then use different names for them.

sophiamersmann:
Resolves #2514

- Adds a `duplicate` transform that is useful in explorers
- The duplicate transform is special: It operates on data values and touches column definitions
- For example, a column with transform `duplicate my-slug` inherits all metadata (and data) from the `my-slug` column

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -544,7 +544,7 @@ export class Explorer
             [...ySlugs.split(" "), xSlug, colorSlug, sizeSlug].filter(identity)
         ) as string[]
 
-        // find all variables the the transformed columns depend on and add them to the dimensions array
+        // find all variables that the transformed columns depend on and add them to the dimensions array
         if (uniqueSlugsInGrapherRow.length) {
             const baseVariableIds = uniq(
                 uniqueSlugsInGrapherRow.flatMap((slug) =>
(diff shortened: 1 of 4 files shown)


marcelgerber (approved):
Nice work! I just added a test case covering the new transform, which thankfully passes :)

sophiamersmann:
cool, that's a good idea! thank you, Marcel :)

PR #2544, opened 2023-08-14
Use grapherConfigETL in explorers

danyx23:
The new grapherConfigETL on the variables table holds the grapher config that is written in the ETL (e.g. it is filled for the dummy indicator in this YAML file). This partial grapher config is not used yet in explorers. This PR adds this capability.


sophiamersmann (approved):
Looking good! Shipping it 🚀

PR #2551, opened 2023-08-16
Avoid repetitions in sources tab

pabloarosado:
I don't have much confidence on this PR (feel free to change it as you see fit), but the intention is the following:
* Field "Data published by" should avoid repeating entities, as it happens in, e.g. [this chart](https://owid.cloud/admin/charts/7079/edit), where we see "Data published by Fishcount, Fishcount, Fishcount" (because the indicator comes from three snapshots of the same producer).
* That field "Data published by", as far as I understand, should be the long citation of the sources, and should therefore correspond to ETL field `citation_producer`, instead of `attribution`.
* I think it's better to show citations as a list of bullet points instead of very long texts joined by "," or ";".
* I don't think we should show again `citation_producer` at the bottom of the sources tab for each variable. That information should already be displayed in "Data published by".

Please let me know if you think I misunderstood the meaning of some of the metadata fields, or if you disagree with this way of showing them in the old sources tab.

### DIFF
--- a/packages/@ourworldindata/grapher/src/sourcesTab/SourcesTab.tsx
+++ b/packages/@ourworldindata/grapher/src/sourcesTab/SourcesTab.tsx
@@ -79,11 +79,15 @@ export class SourcesTab extends React.Component<{
                 : undefined)
 
         const citationProducer =
-            column.def.origins &&
-            column.def.origins.length &&
-            column.def.origins[0].citationProducer
-                ? column.def.origins[0].citationProducer
-                : undefined
+            column.def.origins && column.def.origins.length
+                ? [
+                      ...new Set(
+                          column.def.origins.map(
+                              (origin) => origin.citationProducer
+                          )
+                      ),
+                  ]
+                : []
 
         return (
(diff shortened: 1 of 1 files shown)


packages/@ourworldindata/grapher/src/sourcesTab/SourcesTab.tsx
```diff
@@ -138,10 +142,18 @@ export class SourcesTab extends React.Component<{
...
                             </tr>
                         ) : null}
-                        {attribution ? (
+                        {citationProducer.length > 0 ? (
                             <tr>
                                 <td>Data published by</td>
-                                <td>{attribution.join(", ")}</td>
+                                <td>
+                                    <ul>
+                                        {citationProducer.map(
```
Marigold, line 150: The bullet point looks a bit weird if there's only one citation (see http://staging-site-fix-repetitions-in-sources-tab/admin/variables/791804). Perhaps it'd be better to only show bullet points with 2+ citations?

danyx23:
Thanks @pabloarosado! Shipping it now

PR #2554, opened 2023-08-16
Font weight fixes

ikesau:
Fixes a few spots where we had specified font weights that we didn't support that [now look wrong with the "correct" font weight.
](https://github.com/owid/owid-grapher/pull/2518)
e.g. 

image

### DIFF
--- a/site/owid.scss
+++ b/site/owid.scss
@@ -1202,7 +1202,6 @@ html:not(.js) {
             color: #002147;
             font-family: inherit;
             font-size: 1rem;
-            font-weight: 600;
 
             &:focus {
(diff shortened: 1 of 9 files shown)


site/owid.scss
```diff
@@ -1202,15 +1202,13 @@ html:not(.js) {
             color: #002147;
             font-family: inherit;
             font-size: 1rem;
-            font-weight: 600;
```
ikesau, line 1205: These were bold for no reason - better to remove it entirely.

samizdatco (approved):
Looks good!

PR #2555, opened 2023-08-17
Grapher redesign

sophiamersmann:
WIP (I hope to get a staging site for free by opening this pr)


packages/@ourworldindata/grapher/src/axis/Axis.ts
```diff
@@ -634,7 +634,7 @@ export class DualAxis {
 
     @computed get verticalAxis(): VerticalAxis {
         const axis = this.props.verticalAxis.clone()
-        axis.range = this.innerBounds.yRange()
+        axis.range = this.innerBounds.padTop(6).yRange() // top padding leaves room for tick labels
```
sophiamersmann, line 637: Sometimes, the vertical y-axis overflows, see [this chart](https://ourworldindata.org/grapher/dietary-land-use-vs-beef-consumption). If not accounted for, the top-most axis label would overlap with the content switchers. The axis itself is padded here because I don't want to pad every chart by default if that's not necessary. Is there a better way to do this?

site/css/chart.scss
```diff
@@ -19,7 +19,7 @@
 
     // Charts shouldn't be below this height in any circumstance, whether mobile or not.
     // Except in iframes, where we override this.
-    min-height: 480px;
+    min-height: 540px;
```
sophiamersmann, line 22: The minimum height of Grapher is increased a little since many charts did not really fit into a 480px-tall space.

packages/@ourworldindata/grapher/src/controls/ActionButtons.tsx
```diff
@@ -0,0 +1,371 @@
...
+    @computed private get hasDownloadButton(): boolean {
+        return true
+    }
+
+    @computed private get hasShareButton(): boolean {
+        return !this.manager.hideShareTabButton
+    }
+
+    @computed private get hasFullScreenButton(): boolean {
+        return !this.manager.isInIFrame
```
sophiamersmann, line 198: I don't think it makes sense to offer full-screen mode for charts that are embedded in an iframe. But that means that charts in gdocs won't show the full-screen button when previewed.. (since the gdoc preview is embedded via an iframe)

packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
```diff
@@ -1160,11 +1160,14 @@ export class LineChart
...
-                    : this.defaultRightPadding
-            ),
+            bounds: this.boundsWithoutColorLegend
+                .padRight(
+                    this.lineLegendDimensions
+                        ? this.lineLegendDimensions.width
+                        : this.defaultRightPadding
+                )
+                // top padding leaves room for tick labels
+                .padTop(6),
```
sophiamersmann, line 1170: This is one of the instances where I included some padding on top to prevent charts to overflow. Not really sure this is the best way of doing this...

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -1993,12 +2008,6 @@ export class Grapher
                 title: this.isPlaying ? `Pause` : `Play`,
                 category: "Timeline",
             },
-            {
-                combo: "f",
-                fn: (): void => this.toggleFacetControlVisibility(),
-                title: `Toggle Faceting`,
-                category: "Chart",
-            },
```
sophiamersmann, line 2001: why did you remove this shortcut?

packages/@ourworldindata/grapher/src/controls/Controls.tsx
```diff
@@ -1,29 +1,599 @@
...
+export class EntitySelectorToggle extends React.Component<{
+    manager: EntitySelectionManager
+}> {
+    static shouldShow(manager: EntitySelectionManager): boolean {
+        const toggle = new EntitySelectorToggle({ manager })
+        return toggle.showToggle
+    }
+
+    @computed get showToggle(): boolean {
+        const { isOnTableTab, isOnMapTab } = this.props.manager,
```
sophiamersmann, line 71: Grapher has a `isOnChartTab` property as well

packages/@ourworldindata/grapher/src/controls/Controls.scss
```diff
@@ -1,171 +1,483 @@
...
-    from,
-    20%,
-    40%,
-    60%,
-    80%,
-    to {
-        animation-timing-function: cubic-bezier(0.215, 0.61, 0.355, 1);
+@at-root {
+    // collapse the Settings toggle down to just an icon on small screens
+    .GrapherComponentSmall:not(.GrapherComponentMedium)
```
sophiamersmann, line 171: I don't think `:not(.GrapherComponentMedium)` is needed here?

Aside, you could also do `&.GrapherComponentSmall` instead of `@at-root` if you wanted to since `Controls.scss` is imported within `.GrapherComponent {}`

sophiamersmann:
Hey @samizdatco! I reviewed your code again and it looks good to me!

I noticed that some data pages don't use the settings drawer ([example 1](http://staging-site-grapher-redesign/grapher/share-of-population-in-extreme-poverty?tab=table), [example 2](http://staging-site-grapher-redesign/grapher/yearly-number-of-objects-launched-into-outer-space)). Is it possible that the data pages are rendered by [this component](https://github.com/owid/owid-grapher/blob/grapher-redesign/site/DataPage.tsx) where the drawer nav is missing?

The second data page example has a [All Charts section](http://staging-site-grapher-redesign/grapher/share-of-population-in-extreme-poverty?tab=table#all-charts). I'm wondering if then all charts in the All Charts section would also be using the settings drawer? If so, I think it would be better to special-case those to use the popup instead.

packages/@ourworldindata/grapher/src/controls/Controls.scss
sophiamersmann: Ah I can see how these `GrapherComponentMedium/Small/Narrow` classes are annoying in CSS if they're not mutually exclusive. If you want to you could update the [css classes](https://github.com/owid/owid-grapher/blob/e2073c69ffeb37a19aa659540c4721519fff4827/packages/%40ourworldindata/grapher/src/core/Grapher.tsx#L2297-L2299) to be mutually exclusive. I haven't used these a lot, so it should not be much work to update the few instances where I do. It would be important though not to touch `grapher.isNarrow`, `grapher.isSmall`, etc. themselves, that would mess things up in code..

packages/@ourworldindata/grapher/src/core/Grapher.tsx
samizdatco: It existed previously as an escape hatch to reveal the faceting checkbox even on charts where the author had chosen to hide it. It wasn't a shortcut to change the faceting mode, it just showed/hid the facet checkbox that used to exist in the controls row

packages/@ourworldindata/grapher/src/controls/Controls.scss
samizdatco: Ultimately I think we'll need to do the label-hiding in the entity-selector and settings buttons in js since it will depend on the number of content-switcher buttons, the text-width of the entity name and the possible presence of a 'promoted' single control (replacing the settings button when it's the only option present). 

Until then, I'm using CSS container queries with a pair of breakpoints chosen for the common (worst) case of:
- a three-segment content swticher
- "countries or regions" as the entity name 
- a visible Settings menu

packages/@ourworldindata/grapher/src/core/Grapher.tsx
sophiamersmann: If it's easy to do I think there is no harm in keeping it (I've been using it a lot when debugging faceting issues). But no worries if it's difficult to make work :)

explorer/Explorer.tsx
```diff
@@ -336,10 +339,13 @@ export class Explorer
...
         // preserve the previous tab if that's still available in the new view;
-        // and use the first tab otherwise
+        // and use the first tab otherwise, ignoring the table
+        const tabsWithoutTable = this.grapher.availableTabs.filter(
+            (tab) => tab !== GrapherTabOption.table
+        )
         newGrapherParams.tab = this.grapher.availableTabs.includes(previousTab)
             ? previousTab
-            : this.grapher.availableTabs[0]
+            : tabsWithoutTable[0] ?? GrapherTabOption.table
```
sophiamersmann, line 348: Note for the reviewers: Explorers persist the current tab when switching charts. If the current tab is not available, it used to fallback to the first tab in Grapher's `availableTabs` property which used to be the chart or the map, but in the new designs is the table. That's not usually what we want, so here I explicitly ignore the table to make sure that we get either the map or the chart if the persisted tab is not available.

packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
```diff
@@ -137,64 +155,61 @@ export class CaptionedChart extends React.Component<CaptionedChartProps> {
...
-            : PADDING_BELOW_HEADER
-        return new Bounds(0, 0, this.bounds.width, this.chartHeight)
-            .padWidth(OUTSIDE_PADDING)
-            .padTop(topPadding)
-            .padBottom(OUTSIDE_PADDING)
+    @computed protected get boundsForChartArea(): Bounds {
+        const { bounds, chartHeight } = this
+        return new Bounds(0, 0, bounds.width, chartHeight).padWidth(
+            FRAME_PADDING
+        )
```
sophiamersmann, line 166: Note for the reviewers: The vertical padding (`padTop`, `padBottom`) has been removed from the bounds of the chart area. As a result, some chart types (e.g. Scatter plot) would sometimes overflow. To counteract, padding has been applied to the axis of specific chart types when necessary (see `ScatterPlotChart.tsx`, for example).

packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
```diff
@@ -469,12 +539,25 @@ export class StaticCaptionedChart extends CaptionedChart {
...
+                    targetX={paddedBounds.x}
+                    targetY={paddedBounds.bottom - this.staticFooter.height}
+                />
+                {includeDetailsInStaticExport && this.renderSVGDetails()}
             </svg>
         )
     }
 }
+
+function VerticalSpace({ height }: { height: number }): JSX.Element {
```
sophiamersmann, line 554: Note for the reviewers: Although a bit unconventional, adding vertical space as a `<div />` makes margin collapsing impossible and makes it easier to track the space available for the chart ares (see `chartHeight` function)

packages/@ourworldindata/grapher/src/controls/ChartIcons.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    // bar charts
+    [ChartTypeName.DiscreteBar]: <FontAwesomeIcon icon={faChartColumn} />,
+    [ChartTypeName.StackedBar]: <FontAwesomeIcon icon={faChartColumn} />,
+    [ChartTypeName.StackedDiscreteBar]: (
+        <FontAwesomeIcon icon={faChartColumn} />
+    ),
+
+    // world map
+    [ChartTypeName.WorldMap]: <FontAwesomeIcon icon={faEarthAmericas} />,
```
sophiamersmann, line 22: Note for the reviewers: I think `ChartTypeName.WorldMap` is legacy code? For completeness sake, I assigned an icon...

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -1078,31 +1106,6 @@ export class Grapher
...
-        ) {
-            this.tab = desiredTab
-            this.overlay = undefined
-            return
-        }
-
-        // table tab cannot be downloaded, so revert to default tab
-        if (desiredTab === GrapherTabOption.download && this.isOnTableTab)
-            this.tab = this.authorsVersion.tab ?? GrapherTabOption.chart
-        this.overlay = desiredTab
```
sophiamersmann, line 1103: Note for the reviewers: No need to carry this logic over since that download modal only offers a csv data file when accessed from the table tab

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -1993,18 +2008,18 @@ export class Grapher
...
                 fn: (): void => this.toggleYScaleTypeCommand(),
                 title: "Toggle Y log/linear",
                 category: "Chart",
             },
+            {
+                combo: "w",
+                fn: (): void => this.toggleFullScreenMode(),
+                title: `Toggle full-screen mode`,
+                category: "Chart",
+            },
```
sophiamersmann, line 2022: Note for reviewers: A little treat for our power users hehe. Not sure how much I like the `w` (for 'window'). Suggestions welcome!

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -2278,58 +2289,70 @@ export class Grapher
...
-        const style = {
-            width: renderWidth,
-            height: renderHeight,
-            fontSize: this.baseFontSize,
+        const containerStyle = {
+            width: this.renderWidth,
+            height: this.renderHeight,
+            fontSize: this.isExportingtoSvgOrPng
+                ? 18
+                : Math.min(16, this.baseFontSize), // cap font size at 16px
```
sophiamersmann, line 2307: Note for the reviewers: This has an effect on text in modals only, since text around the chart (header, footer) ignore the base font size and only honour `grapher.isNarrow/isSmall/isMedium`, and the CaptionedChart's `<svg />` element has its own base font size set (i.e. axis labels and chart ticks are only effected by the font size set on the svg element). In modals, I don't want to scale up the font size, that's why it's capped here.

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -2360,6 +2383,25 @@ export class Grapher
         else if (renderWidth >= 1080) this.baseFontSize = 18
     }
 
+    @computed get isNarrow(): boolean {
+        if (this.isExportingtoSvgOrPng) return false
```
sophiamersmann, line 2387: Note for the reviewers: For now, exported charts use ideal bounds. That's why they can never be narrow, for example. But when @marcelgerber presented our new cloudflare workers, it seemed like we wanted static charts that could potentially have any size?

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -2377,9 +2419,28 @@ export class Grapher
...
     }
 
+    @action.bound private setUpWindowResizeEventHandler(): void {
+        const updateWindowDimensions = (): void => {
+            this.windowInnerWidth = window.innerWidth
+            this.windowInnerHeight = window.innerHeight
+        }
+        const onResize = debounce(updateWindowDimensions, 400, {
+            leading: true,
+        })
```
sophiamersmann, line 2429: For the reviewers: I'm not confident `leading=true` makes sense here?

packages/@ourworldindata/grapher/src/dataTable/DataTable.tsx
```diff
@@ -776,35 +946,59 @@ function ColumnHeader(props: {
...
+            {isActiveIcon ? (
+                <FontAwesomeIcon icon={activeIcon} />
+            ) : (
+                <span style={{ display: "inline-block", width: "max-content" }}>
+                    <FontAwesomeIcon icon={faArrowUpLong} />
+                    <FontAwesomeIcon
+                        icon={faArrowDownLong}
+                        style={{ marginLeft: "-4px" }}
+                    />
+                </span>
```
sophiamersmann, line 996: Note for the reviewers: The "↑↓" icon is not part of the free collection of FontAwesome. That's why I'm faking it here.

packages/@ourworldindata/grapher/src/footer/Footer.tsx
```diff
@@ -10,167 +10,338 @@ import {
...
-                : new Set(),
-        })
+            sourcesText,
+            actionButtonsWidthWithIconsOnly,
+        } = this
+        if (hasNote) return true
+        const sourcesWidth = Bounds.forText(sourcesText, {
+            fontSize: sourcesFontSize,
+        }).width
+        return sourcesWidth > 2 * (maxWidth - actionButtonsWidthWithIconsOnly)
```
sophiamersmann, line 207: Note for the reviewers: This is a heuristic. If the sources text is "kind of long", we render it above the action buttons using the full width. If it's short, it might be placed next to the action buttons.

packages/@ourworldindata/grapher/src/modal/DownloadModal.tsx
```diff
@@ -201,18 +218,18 @@ export class DownloadTab extends React.Component<DownloadTabProps> {
...
+            this
         const pngPreviewUrl = this.pngPreviewUrl || this.fallbackPngUrl
 
         let previewWidth: number
         let previewHeight: number
         const boundScalar = 0.17
-        if (bounds.width / bounds.height > targetWidth / targetHeight) {
-            previewHeight = bounds.height * boundScalar
+        if (tabBounds.width / tabBounds.height > targetWidth / targetHeight) {
+            previewHeight = Math.min(72, tabBounds.height * boundScalar)
```
sophiamersmann, line 229: Note for the reviewers: We don't want the preview image to be too large since the max-width of the download modal is fixed.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -358,13 +376,21 @@ const TimelineHandle = ({
...
+                        style={{ zIndex: tooltipZIndex }}
+                    />
+                    <div
+                        className="handle-label"
+                        style={{ zIndex: tooltipZIndex }}
+                    >
+                        {tooltipContent}
+                    </div>
+                </>
+            )}
```
sophiamersmann, line 393: Note for the reviewers: I'm rolling my own tooltips instead of relaying on Tippy since I wanted these to look exactly like the tooltips used for the action buttons.

packages/@ourworldindata/grapher/src/core/Grapher.tsx
sophiamersmann: Okay, cool, thanks for verifying! I always get a little confused about debounce options

packages/@ourworldindata/grapher/src/footer/Footer.tsx
```diff
@@ -194,64 +365,190 @@ export class Footer extends React.Component<{
...
+            <p className="sources" style={sources.style}>
+                {sources.renderHTML()}
+                {" - "}
+                <a
+                    className="learn-more-about-data"
+                    data-track-note="chart_click_sources"
+                    onClick={action(() => {
+                        // on data pages, scroll to the "Sources and Processing" section
+                        // on grapher pages, open the sources modal
+                        const sourcesIdOnDataPage = "sources-and-processing"
```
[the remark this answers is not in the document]
sophiamersmann: yeah, fair enough. will do so tomorrow!

packages/@ourworldindata/grapher/src/controls/Controls.scss
```diff
@@ -1,336 +1,102 @@
...
-        &.open-in-another-tab-button {
-            svg {
-                margin-left: 0.3em;
-            }
+//
+// Make buttons narrower by omitting their text labels as the viewport gets narrower
+// TODO: move the logic to js to calculate actual widths for different entity name lengths
+//       and the variable number of buttons present in the content switcher block
+//
+@container grapher (max-width:550px) {
```
danyx23, line 69: I think this is the first time that we use container queries in our codebase. Can you add a comment here on how this behaves if container (size?) queries are not supported?

danyx23 (approved):

danyx23 (approved):

danyx23 (approved):
Looks good, great work! I had a bunch of smaller items, mostly requests to add some comments but from my side this is good to ship whenever you feel it is ready

packages/@ourworldindata/grapher/src/fullScreen/FullScreen.scss
```diff
@@ -0,0 +1,15 @@
...
+    left: 0;
+    z-index: $zindex-full-screen;
+    height: 100%;
+    width: 100%;
+
+    .FullScreenContent {
+        position: absolute;
+        top: 50%;
+        left: 50%;
+        transform: translate(-50%, -50%);
```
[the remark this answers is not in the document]
sophiamersmann: It centers the container horizontally and vertically. But this piece of code has since been [removed by Marcel](https://github.com/owid/owid-grapher/pull/2700) to fix an issue in Firefox.

packages/@ourworldindata/grapher/src/footer/Footer.tsx
```diff
@@ -10,167 +10,345 @@ import {
...
++------------------------------+-----+------------------+
+|  Note                              |  Action buttons  |
++-------------------------------------------------------+
+
+If the note is missing and the sources text is not too long, the sources are placed next to the action buttons:
++------------------------------------+------------------+
+|  Sources                           |                  |
++------------------------------------+  Action buttons  |
+|  Origin URL | CC BY                |                  |
++-------------------------------------------------------+
```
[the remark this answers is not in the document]
sophiamersmann: I had a lot of fun writing it :)

package.json
```diff
@@ -139,6 +143,7 @@
         "googleapis": "^108.0.0",
         "handsontable": "^12.3.3",
         "html-to-text": "^8.2.0",
+        "indefinite": "^2.4.3",
```
sophiamersmann, line 146: ah wait, should I rather install this package into the grapher package? (it's only used in Grapher)

site/gdocs/Chart.tsx
```diff
@@ -38,11 +39,26 @@ export default function Chart({
...
+        // listed in an All Charts block. The <Chart /> component is not used to render charts in an All Charts block,
+        // so we can just set this to true here.
+        shouldOptimizeForHorizontalSpace: true,
+    }
+
+    // props passed to explorers
+    const explorerProps: Pick<
+        ExplorerProps,
+        "shouldOptimizeForHorizontalSpace"
+    > = merge({}, common)
```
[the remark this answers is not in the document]
sophiamersmann: When embedded in an article and viewed on mobile, we want to optimise for horizontal space (see screenshot in the [resp. PR](https://github.com/owid/owid-grapher/pull/2688)). In practise, when optimising for horizontal space, the horizontal frame padding is et to 0px and the left and right borders are removed, while the top and bottom borders stretch across the whole page. To this end, I have added a new property to Grapher called `shouldOptimizeForHorizontalSpace`.

Usually that works like this:
- The gdoc `<Chart />` component renders a figure with `data-grapher-src` set and the MultiEmbedder picks up the figure element and renders Grapher into it
- To pass information from the gdoc `<Chart />` component to Grapher, a stringified config is specified as `data-grapher-config` on the figure element, that data attribute is read by the MultiEmbedder, the parsed config is merged with other configs and then passed to Grapher when it's rendered. This is how, for example, the title or subtitle is passed to Grapher in narrative charts. And we use the same strategy to pass `shouldOptimizeForHorizontalSpace` (it's just another property).

That works for normal Grapher charts, but sometimes authors want to show one specific chart from an explorer. To do so, they link to a particular explorer view and hide the controls. Such charts look like ordinary Grapher charts but are in fact explorers. We want to optimize those charts for horizontal space as well. To do so, I have added a flag `shouldOptimizeForHorizontalSpace` to the explorer props.

Passing information from the gdocs `<Chart />` component to the explorer works exactly like passing information from `<Chart />` to Grapher:
- I added a new data attribute to the figure element, called `data-explorer-props`
- Whatever configuration is passed to `data-explorer-props` is picked up by the MultiEmbedder and then passed to the explorer
- For now, as a precaution, I only allow `shouldOptimizeForHorizontalSpace` to be passed
- An alternative would be to add `shouldOptimizeForHorizontalSpace` to the query string of an explorer but it would be odd, I think, to expose `shouldOptimizeForHorizontalSpace` like that

packages/@ourworldindata/grapher/src/header/Header.tsx
```diff
@@ -121,48 +132,80 @@ export class Header extends React.Component<{
                     style={{
                         fontFamily:
                             "'Playfair Display', Georgia, 'Times New Roman', 'Liberation Serif', serif",
+                        fontWeight: 500,
                     }}
                     target="_blank"
                     rel="noopener"
                 >
-                    {title.render(x, y, { fill: "#555" })}
+                    {title.render(x, y, { fill: "#4E4E4E" })}
```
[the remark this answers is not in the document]
sophiamersmann: doing this just help to discover a bug/inconsistency :)

packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
sophiamersmann: Good point!

packages/@ourworldindata/grapher/src/core/Grapher.tsx
sophiamersmann: okay, thank you, both!

packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
```diff
@@ -599,12 +604,20 @@ export class LineChart
...
     }
 
     @action.bound onLineLegendMouseLeave(): void {
-        this.hoveredSeriesName = undefined
+        clearTimeout(this.hoverTimer)
+
+        this.hoverTimer = setTimeout(() => {
+            // wait before clearing selection in case the mouse is moving quickly over neighboring labels
+            this.hoveredSeriesName = undefined
+        }, 100)
```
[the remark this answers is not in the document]
samizdatco: You're right, the timeout is being short circuited by the immediate clear happening in `onCursorLeave`…

packages/@ourworldindata/grapher/src/controls/Controls.scss
samizdatco: It's also used to step down the font size in the tooltips on small screens (where it's more just a nice-to-have feature). Here it's pretty essential though since it keeps both buttons visible, so there ought to be a fallback strategy…

samizdatco: I added one more breakpoint & width-class to grapher that will kick in when container queries aren't supported

packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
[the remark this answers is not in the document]
samizdatco: Fixed (and also tweaked to 200ms)

site/css/content.scss
```diff
@@ -256,39 +256,22 @@ figure[data-explorer-src] {
 .article-content {
     span.add-country {
         font-size: 0.9rem;
-        font-weight: bold;
-        color: $controls-color;
+        font: 400 13px/16px Lato, "Helvetica Neue", Helvetica, Arial,
+            "Liberation Sans", sans-serif;
```
marcelgerber, line 260: can we use `$sans-serif-font-stack` here instead?

packages/@ourworldindata/grapher/src/controls/SettingsMenu.tsx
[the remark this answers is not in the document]
samizdatco: Agreed

packages/@ourworldindata/grapher/src/controls/SettingsMenu.tsx
```diff
@@ -0,0 +1,456 @@
...
+    @computed get chartType(): string {
+        const { type } = this.manager
+        return type.replace(/([A-Z])/g, " $1")
+    }
+
+    @computed get drawer(): Element | null {
+        // use a drop-down menu when embedded in an iframe
+        if (this.manager.isInIFrame) return null
+        // use the drawer `<nav>` element if it exists, otherwise render into a drop-down menu
+        return this.manager.base?.current?.closest(".related-charts")
```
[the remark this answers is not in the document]
sophiamersmann: I put the `isInIFrame` check in after testing what Grapher looks like when embedded via an iFrame on foreign sites. It needs the additional check because what's embedded is `ourworldindata.org/grapher/slug`, so it does find a `nav#grapher-settings-id` on the page, but a drop-down is more sensible for iframe embedded charts, I think.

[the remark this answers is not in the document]
marcelgerber: Okay okay, I understand.
There might be a better way to detect a data page though?

Otherwise, could you please write this up as a code comment so this knowledge is not lost in the future?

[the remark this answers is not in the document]
sophiamersmann: If I'm not mistaken, `grapher.isEmbeddedInAnOwidPage` is true for all embedded charts, including the related charts blocks on data pages. (Would be good to double-check though, I'm not 100% sure)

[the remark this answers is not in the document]
samizdatco: I think we must have had dueling edits at one point too, since I don't see any mention of `isInIFrame` in the branch's current version of [`SettingsMenu.tsx`](https://github.com/owid/owid-grapher/blob/grapher-redesign/packages/%40ourworldindata/grapher/src/controls/SettingsMenu.tsx)...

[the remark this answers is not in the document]
sophiamersmann: that's weird.. I'm not sure how that would happen. It seems like it got lost in [this commit](https://github.com/owid/owid-grapher/pull/2732/commits/053fe9aa7978e61373ec8eba9ddd4df834dd24e1)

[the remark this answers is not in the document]
sophiamersmann: I'm actually very confused about this. To me it looks like you deleted it?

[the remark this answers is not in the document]
samizdatco: Yeah it's a mystery to me as well. I wonder if I had unsaved changes in my editor before pulling and then overwrote those lines when I saved? In any case, thanks for the link so I can add it back.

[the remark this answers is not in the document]
samizdatco: I'm now letting either `isEmbeddedInAnOwidPage` or `isInFrame` put the settings menu into drop-down mode. It seems to be working as intended on data pages: the main graphic uses the drawer and the Related Charts instances use the drop-down.

packages/@ourworldindata/grapher/src/core/Grapher.tsx
samizdatco: Makes sense, added it back

packages/@ourworldindata/grapher/src/controls/SettingsMenu.tsx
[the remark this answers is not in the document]
sophiamersmann: Nice!

PR #2556, opened 2023-08-17
Resize charts in gdoc articles on mobile

sophiamersmann:
Sets the height of Grapher charts to the viewport height on mobile. Sometimes, the heights of the header and navigation are taken into account so that the full chart (and sometimes the navigation) are both visible on the screen.

Updated components include:
- Stand-alone charts
- Charts in Key Insights block
- Charts in the All Charts block
- Charts in a Chart Story

Creating a new `Chart.scss` file may not be necessary; I wasn't sure where that code should live.


sophiamersmann:
I'm realising now that I want these changes to be a part of the staging server. I'll close this PR and open another one against the grapher redesign feature branch.

PR #2569, opened 2023-08-23
:sparkles: add copy to yaml button on chart admin reference tab

danyx23:
With the new metadata it is now possible to set grapher default values for each indicator. This is usually done by editing the yaml file in the grapher step where all the other metadata is also defined usually. 

Writing grapher configs as yaml from scratch is annoying, so this PR adds a button in the chart admin on the references tab to copy the YAML config of a chart (minus some fields like id) to the clipboard so it can just be pasted into the ETL.


adminSiteClient/EditorHistoryTab.tsx
```diff
@@ -60,6 +62,25 @@ export class EditorHistoryTab extends React.Component<{ editor: ChartEditor }> {
...
+        // Use the Clipboard API to copy the config into the users clipboard
+        const chartConfigObject = {
+            ...this.props.editor.grapher.object,
+        }
+        delete chartConfigObject.id
+        delete chartConfigObject.dimensions
+        delete chartConfigObject.version
+        delete chartConfigObject.isPublished
+        const chartConfigAsYaml = dump(chartConfigObject)
+        navigator.clipboard.writeText(chartConfigAsYaml)
```
Marigold, line 75: `navigator.clipboard` is undefined at HTTP site (i.e. our staging servers) and would require some workaround. We probably won't need it there, so it's fine to only make it work for production grapher, up to you.

Here's chatgpt snippet with the workaround in case
```javascript
function copyToClipboard(text) {
  if (navigator.clipboard) {
    navigator.clipboard.writeText(text).then(() => {
      console.log('Text copied to clipboard');
    }).catch(err => {
      console.error('Failed to copy text to clipboard', err);
    });
  } else {
    const textarea = document.createElement('textarea');
    textarea.value = text;
    document.body.appendChild(textarea);
    textarea.focus();
    textarea.select();

    try {
      document.execCommand('copy');
      console.log('Text copied to clipboard');
    } catch (err) {
      console.error('Failed to copy text to clipboard', err);
    }

    document.body.removeChild(textarea);
  }
}

```

Marigold (approved):
Looks good! (It doesn't work for http sites, leaving the decision to fix up to you)

adminSiteClient/EditorHistoryTab.tsx
danyx23: Nice catch!

PR #2581, opened 2023-08-28
:hammer: remove unique name-namespace constraint from datasets table

Marigold:
Fix for https://github.com/owid/etl/issues/1544.

Remove unique index `datasets_name_namespace_d3d60d22_uniq (name, namespace)`. We already have a constraint on `(shortName, namespace, version)` and want to have datasets with the same title but different versions.


pabloarosado (approved):
Looks good. Thanks for doing this!

PR #2583, opened 2023-08-28
Implement late Aug 2023 metadata updates

danyx23:
This implements the metadata updates tracked in #1550. 

* [x] Move `presentation.key_info_text`-> `indicator.description_key`.
* [x] The field `description_short` should populate chart subtitles by default (currently, no subtitle appears unless passed in grapher config).
* [x] Move `presentation.processing_info` -> `indicator.description_processing`.
* [x] Currently, the automatic attribution format is `producer year`, but it should be `producer (year)`.
* [x] Rename `version` -> `version_producer`.


Marigold (approved):
Looks good! I've added migration with renames and tested it with its [ETL counterpart](https://github.com/owid/etl/pull/1554). The staging site for this PR isn't working, but I deployed it to http://staging-site-mojmir/admin/datapage-preview/540251

PR #2591, opened 2023-08-30
Grapher Settings Drawer Redesign

samizdatco:
This PR moves the collection of one-off controls that are contextually inserted above the chart into a pop-up dialog box which is summoned by a button in the upper right corner labeled Settings. On grapher and explorer pages, a `nav#grapher-settings-drawer` element has been added to the markup outside of the grapher instance. When present, the settings menu will use a portal to render its contents into this `<nav>` (which slides in from the right edge of the viewport). If the `<nav>` is not present, the menu is presented as a pop-up whose bounds are fit to the containing grapher instance.

Tracking issue: #2486
Related issue:  #1714


packages/@ourworldindata/grapher/src/controls/Controls.tsx
```diff
@@ -1,18 +1,500 @@
...
+            showSelectEntitiesButton,
+            showChangeEntityButton,
+            showAddEntityButton,
+            entityType,
+            entityTypePlural,
+            isSelectingData: active,
+        } = this.props.manager
+
+        const [icon, label] = showSelectEntitiesButton
+            ? [faEye, `Select ${entityTypePlural}`]
```
sophiamersmann, line 64: Unfortunately, many charts don't specify the proper `entityType` and `entityTypePlural`. `entityTypePlural` in particular is often not set since it's only used for some scatter plots, and `entityType` is mainly used for faceting, so I wouldn't expect this to be set for charts that can't be faceted. We've discussed this [in this thread](https://owid.slack.com/archives/C05KKHZL8LC/p1692105774588569) in a different context. Back then I did a quick analysis and since we don't have many non-country charts, it's only around 120 charts that would need to be updated (maybe more, I don't remember what I checked specifically). Some of these charts could be updated in bulk (e.g. if "egg" is one of the entities, use "product" and "products", etc.) but others would need to be updated manually. I wanted to do that in this cycle but I didn't find the time. It's definitely something we should consider doing when overhauling the entity picker (currently, the heading is "Choose data to show"), but if it's useful for the controls as well, then we might as well do it now. If you do end up doing the data work, it would be nice to also update the heading in the entity picker modal.

packages/@ourworldindata/grapher/src/controls/Controls.tsx
```diff
@@ -1,18 +1,500 @@
...
+            StackedArea,
+            StackedBar,
+            StackedDiscreteBar,
+            ScatterPlot,
+            LineChart,
+            Marimekko,
+        ].includes(type)
+    }
+
+    @computed get showFacetControl(): boolean {
```
sophiamersmann, line 203: This one should also resepct `hideFacetControl`, no?

packages/@ourworldindata/grapher/src/controls/Controls.tsx
```diff
@@ -1,18 +1,500 @@
...
+            !this.manager.hideFacetYDomainToggle &&
+            this.manager.facetStrategy !== FacetStrategy.none &&
+            this.manager.type !== StackedDiscreteBar
+        )
+    }
+
+    @computed get showZoomToggle(): boolean {
+        // TODO:
+        // grapher passes a SelectionArray instance but programmatically defined
+        // managers treat `selection` as a string[] of entity names. do we need both?
```
sophiamersmann, line 167: Not sure why selection is sometimes treated as SelectionArray instance, sometimes as string[]. That's probably a question for @marcelgerber.

I've seen the [makeSelectionArray](https://github.com/owid/owid-grapher/blob/a3d32f96431cbc43b74d8d00b9539432c52044d5/packages/%40ourworldindata/grapher/src/chart/ChartUtils.tsx#L80) in use, so you could use that here as well?

packages/@ourworldindata/grapher/src/controls/Controls.tsx
```diff
@@ -1,18 +1,500 @@
...
+        if (!this.visible) this.shouldRender = false
+    }
+
+    @computed get manager(): SettingsMenuManager {
+        return this.props.manager
+    }
+
+    @computed get chartType(): string {
+        const { type } = this.manager
+        return type.replace(/([A-Z])/g, " $1")
```
sophiamersmann, line 259: Out of interest, why is this necessary? Shouldn't the `type` be one of [ChartTypeName](https://github.com/owid/owid-grapher/blob/a3d32f96431cbc43b74d8d00b9539432c52044d5/packages/%40ourworldindata/grapher/src/core/GrapherConstants.ts#L5)?

packages/@ourworldindata/grapher/src/controls/Controls.tsx
```diff
@@ -208,47 +649,59 @@ export class FacetStrategyDropdown extends React.Component<{
...
+            byEntity = this.strategies.includes(FacetStrategy.entity),
+            byMetric = this.strategies.includes(FacetStrategy.metric)
+
+        if (byEntity || byMetric) {
+            const facet =
+                byEntity && byMetric
+                    ? `metric, ${entityType}`
+                    : byEntity
+                    ? entityType
+                    : "metric"
```
sophiamersmann, line 663: We could also use the manager's `facettingLabelByYVariables` here if we wanted to. `facettingLabelByYVariables` is an alternative for "metric", specified by the author of a chart.

packages/@ourworldindata/grapher/src/controls/Controls.scss
```diff
@@ -1,175 +1,408 @@
...
-    font-weight: bold;
-    border: 1px solid #afb7c1;
-    border-radius: 5px;
-    padding-left: 10px;
-    padding-right: 10px;
-    cursor: pointer;
-}
+$medium: 400;
+$bold: 700;
+$lato: $sans-serif-font-stack;
```
sophiamersmann, line 17: I copied these variables into `grapher.scss` when you sent them to me over Slack. I'm realising now that wasn't a good idea because these seem to be specific to the controls menu. I think I'll remove them from `grapher.scss` and refactor my css code. Does that make sense to you?

packages/@ourworldindata/grapher/src/controls/Controls.scss
```diff
@@ -1,175 +1,408 @@
...
+        }
+    }
 }
 
-.selected .FacetStrategyPreview-split-child {
-    background: #d9eafb;
+@at-root {
+    // collapse the Settings toggle down to just an icon on small screens
+    .GrapherComponentSmall nav.controlsRow .settings-menu button.menu-toggle {
+        min-height: 32px;
```
sophiamersmann, line 82: You could be using `$controlRowHeight` here if it's available. If it's not, we could consider moving the defintion elsewhere (it's currently defined in `CaptionedChart.scss`)

packages/@ourworldindata/grapher/src/controls/Controls.scss
```diff
@@ -1,175 +1,408 @@
...
+            width: 100%;
+            height: 100%;
+            z-index: $zindex-controls-backdrop;
+        }
+    }
 }
 
-.selected .FacetStrategyPreview-split-child {
-    background: #d9eafb;
+@at-root {
```
sophiamersmann, line 79: I didn't know about @at-root. Neat!

sophiamersmann:
Hey! I hope it's okay I looked at the code today. I thought it would be good to do this before I'm off and tomorrow is a short day for me, so I worry that might not have the time tomorrow..

It's really neat to see the drawer in action and it makes me so happy that our controls and their look & feel get overhauled. It looks fantastic!

These are my notes:
- It'd be nice if the settings popup was dismissible by clicking outside of it (similar to how it works for the drawer and the modals)
- If the `<CollapsibleList />` component isn't used anymore, we could remove it?
- The drawer should probably also open on data pages (like [this one](https://ourworldindata.org/grapher/yearly-number-of-objects-launched-into-outer-space))
  - Can data pages have an "All charts" section? If yes, we would only want to open the drawer for the main chart on top?
- The settings menu should be hidden on map & table tabs, see  [this chart](http://staging-site-grapher-redesign-drawer/grapher/women-choose-where-to-live-same-way-as-men) for example
- I'm not really sure what the `data-track-note` is for and whether that's still in use. It's probably worth checking in with Marcel to see if it's currently used for analytics.
- The control row should be 32px, but is closer to 33px at the moment. You could either make it fit or adapt the height of the control bar [here](https://github.com/owid/owid-grapher/blob/bece0e603277747a7b778fbc772220d96ab1928f/packages/%40ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx#L119) and [here](https://github.com/owid/owid-grapher/blob/bece0e603277747a7b778fbc772220d96ab1928f/packages/%40ourworldindata/grapher/src/captionedChart/CaptionedChart.scss#L3) (I know it's a single pixel but it adds up...)

I think we need to do some more harmonising of our work on narrative charts. This piece of Archie syntax shouldn't show any controls:
```
{.chart}
title: Custom title
subtitle: Custom subtitle
[.tabs]
- all
[]
{}
```

And this one should show the `yLogLinearSelector` selector only:
```
{.chart}
title: Custom title
subtitle: Custom subtitle
[.controls]
- yLogLinearSelector
[]
{}
```

I'm not sure what's needed to make this happen but one piece to the puzzle is this [method](https://github.com/owid/owid-grapher/blob/bece0e603277747a7b778fbc772220d96ab1928f/packages/%40ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx#L197) that should also account for controls.

packages/@ourworldindata/grapher/src/controls/Controls.scss
samizdatco: Yeah, agreed. 

Keeping these sorts of variables local has been my approach both here and with the tooltips. There probably are a few things that could usefully be pulled up to the grapher level, but so far my preference has been to name them after their use-case within the component (rather than more generic names like blue-20, grey-80, etc.)

packages/@ourworldindata/grapher/src/controls/Controls.tsx
samizdatco: Agreed that it's probably worth fixing at the data level since we'll be able to make use of the names again in the entity-selector UI.

Just to clarify though: when the config doesn't override the entity type names, they default to `country or region` and `countries or regions`, so there will always be **a** string for this, it will just be the **wrong** string until we update those un-customized, non-country charts.

sophiamersmann: Yes, it would show the default (which for non-country entities is wrong).

PR #2599, opened 2023-08-31
:sparkles: on the variable edit admin page add link to garden meta yaml

danyx23:


Marigold:
@danyx23 it would be useful to also have a link to **grapher** metadata. Today we talked with @pabloarosado and we'll be storing data pages metadata in the grapher YAML for datasets that haven't been migrated and are still using sources.

PR #2602, opened 2023-09-01
Add "Show selection only" toggle for tables

sophiamersmann:
Adds a "Show selection only" toggle for tables.

For now, only the ability has been added, but no toggle.

To do:

- [ ] Actually add the toggle (after Christian's changes have landed)
- [ ] What if the current selection is empty?


sophiamersmann:
This PR is part of the Grapher redesign project

sophiamersmann:
This feature has been added elsewhere

PR #2603, opened 2023-09-01
fix(explorer): prevent metric from being updated when entities are sorted

sophiamersmann:
Fixes an issue where clicking the entity picker's sort icon would also update the metric to its default value ("Relevance")


marcelgerber (approved):
Nice, thank you for fixing (and sorry for breaking this earlier this week 👀)

PR #2632, opened 2023-09-12
:tada: final metadata renames

Marigold:
Migration & metadata renames from https://github.com/owid/etl/issues/1584

### DIFF
--- a/site/DataPageV2Content.tsx
+++ b/site/DataPageV2Content.tsx
@@ -189,9 +189,9 @@ export const DataPageV2Content = ({
     const [grapher, setGrapher] = React.useState<Grapher | undefined>(undefined)
 
     const sourceShortName =
-        datapageData.producerShort && datapageData.titleVariant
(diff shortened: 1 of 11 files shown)


site/DataPageV2Content.tsx
```diff
@@ -620,7 +617,7 @@ export const DataPageV2Content = ({
                                                                 label={
                                                                     source.producer ??
                                                                     source.datasetDescriptionOwid ??
```
pabloarosado, line 619: Shouldn't `datasetDescriptionOwid` be `descriptionSnapshot`?

pabloarosado (approved):
I think you may have missed changing `datasetDescriptionOwid` to `descriptionSnapshot`. I caught a few instances, maybe not all. Apart from that, looks good!

Marigold:
> I think you may have missed changing datasetDescriptionOwid to descriptionSnapshot. I caught a few instances, maybe not all. Apart from that, looks good!

I forgot to commit those changes... Thanks for catching that!

PR #2634, opened 2023-09-12
🎉 Add gdoc quick edit buttons for logged-in admins

ikesau:
Adds two buttons to make it easier to update articles from the published page, like we have for WordPress pages.

`position: absolute` (instead of `fixed`) so that it doesn't obscure the nav in the rare case you're logged in on mobile.

image
image


mlbrgl (approved):
LGTM!

Issue #2520, opened 2023-08-08
Switch grapher sources rendering to markdown


danyx23:
We used to use HTML sometimes in Sources so that the Sources tab would render links or bullet points. With the new metadata we now want to standardize on our markdown flavor. This will require the Grapher sources tab to switch from using `dangerouslySetInnerHTML` to `MarkdownTextWrap`.

This will have to be merged in sync with the ETL work to replace HTML with markdown: https://github.com/owid/etl/pull/1422

It will probably also be good to to implement #2489 before doing this issue

PR #2639, opened 2023-09-19
Render source fields as markdown instead of HTML

danyx23:
This PR switches the rendering of text from Source in the sources tab of Grapher from HTML to markdown. This is the companion PR for [this PR in the ETL](https://github.com/owid/etl/pull/1422) that converts legacy HTML to markdown in sources.

This PR implements #2520


ikesau (approved):
lgtm!

PR #2648, opened 2023-09-22
🐛 remove size suffix from migrated WP image filenames

ikesau:
Some images are saved in the WP html with dimension data appended to them.

e.g.

`causes-of-death-2048x1024.png`

Which was making its way into the archie, despite the fact that [we only uploaded the source images to WP](https://github.com/owid/owid-grapher/blob/a04b880d763d073fe21153ebc288932895785082/devTools/uploadWordpressImagesToObjStorage/upload-wordpress-images-to-obj-storage.ts#L97) when we backed them up to drive.

The fixes that.

### DIFF
--- a/db/model/Gdoc/htmlToEnriched.ts
+++ b/db/model/Gdoc/htmlToEnriched.ts
@@ -987,9 +987,13 @@ function cheerioToArchieML(
                             {
                                 type: "image",
                                 // src is the entire path. we only want the filename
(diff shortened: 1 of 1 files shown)


danyx23 (approved):

danyx23 (approved):
Looks good! Maybe add a comment at the regex to show what kind of text we are matching here and why we are removing it? Good to ship though

PR #2649, opened 2023-09-22
Entry Emulator - Stacked Columns

ikesau:
According to Joe, the authors want to move away from two-column layouts in general, because it results in writing that doesn't work well on mobile where we always have to be single column anyway.

Running this on all our posts will require quite a bit of manual touching up for all the times where a column starts with "In this chart we can see...." and then, now at the bottom of several paragraphs, the chart will finally be visible.

So even though this is a fairly rudimentary solution, my thinking is that we get this merged now, and then the leads can see how it looks during their session on Tuesday and determine if they want to prioritize a more intelligent column consolidation algorithm, or if there's something else they'd prefer I work on because they're fine with doing the touchups manually.

## Before

image

## After
image


danyx23 (commented):
I'm fine with this technically but are we sure we want this in general? cc @JoeHasell

JoeHasell:
Thanks @danyx23 – I don't have a sense of what the PR is actually implementing. Going purely off Ike's framing at the top, I agree it's too general compared with what we had in mind. See my thread [here in Slack](https://owid.slack.com/archives/C03SK89695E/p1695412927378239) in reply to Ike a few days back.

PR #2650, opened 2023-09-22
Entry Emulator - Edit Notice

ikesau:
Many of our entries and articles have notices at the beginning saying when they were first published and last updated. 

In our old articles, these sentences were lifted into the article header, but the wp migration was putting them unceremoniously into the article body.

This PR adds code which identifies these phrases and puts them in a callout.

## Before
image

## After
image

### DIFF
--- a/db/model/Gdoc/htmlToEnriched.ts
+++ b/db/model/Gdoc/htmlToEnriched.ts
@@ -736,7 +736,7 @@ function cheerioToArchieML(
         // Special handling for a unique span that we use to mark the first published date
         if (
             span.spanType === "span-simple-text" &&
-            span.text.trim().match(/^This (entry|article) was first published/)
+            span.text.trim().match(/published/)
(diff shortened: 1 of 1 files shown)


danyx23 (approved):

db/model/Gdoc/htmlToEnriched.ts
```diff
@@ -732,13 +732,29 @@ function cheerioToArchieML(
         unwrapElement(element, context)
 
     const span = cheerioToSpan(element)
-    if (span)
+    if (span) {
+        // Special handling for a unique span that we use to mark the first published date
+        if (
+            span.spanType === "span-simple-text" &&
+            span.text.trim().match(/published/)
```
ikesau, line 739: This is probably too generous now and I should make sure it only applies to the first span in the article, not any span that contains the word "published"

PR #2678, opened 2023-09-29
🐛 Exclude gdocs with future `publishedAt` dates from baking and atom feed

ikesau:
Encountered an issue where a post that was scheduled for Monday was featured in the atom feed.

This PR as is will fix that from happening (but in general I don't think we should/can schedule posts currently)

If we want to do this properly, we'll need to tighten up the way we handle times with the datepicker. 

Currently our DB is in UTC, but the times that the antd DatePicker chooses seem to be kind of random. The hour varies depending on which day you pick. If we want to schedule posts we'd want to be able to specify this or standardize it with a clearly specified explanation in the UI.

### DIFF
--- a/db/model/Gdoc/Gdoc.ts
+++ b/db/model/Gdoc/Gdoc.ts
@@ -852,7 +852,10 @@ export class Gdoc extends BaseEntity implements OwidGdocInterface {
         // the database dependent on the mapping function, which is more practical
         // but also makes it less of a source of truth when considered in isolation.
         return Gdoc.find({
-            where: { published: true },
+            where: {
+                published: true,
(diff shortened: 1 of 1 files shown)


mlbrgl (approved):
Tested 👍

Now that the instability has been resolved in https://github.com/owid/owid-grapher/pull/2698, "scheduling" posts could be a matter of: 
- exposing the (local) time in the date picker
- explaining that this should be understood as "do not publish before xx:xx" rather than "publish at xx:xx"

However, I think this distinction makes the feature much less useful (if not confusing), so keeping it under wraps until we can schedule predictably is more desirable.

PR #2702, opened 2023-10-05
enhance(admin): persist entities when changing the chart type

sophiamersmann:
### Problem

- When changing chart types (or adding y-dimensions), a random sample of entities are selected
- Apparently, that's annoying for authors; they would prefer to have entities persisted if possible (see [Slack](https://owid.slack.com/archives/C0149NKLZAM/p1696419244456579))

### Solution

- Only sample entities when none are selected

### DIFF
--- a/adminSiteClient/EditorBasicTab.tsx
+++ b/adminSiteClient/EditorBasicTab.tsx
@@ -86,7 +86,7 @@ class DimensionSlotView extends React.Component<{
         const { availableEntityNames, availableEntityNameSet } = selection
 
         if (grapher.isScatter || grapher.isSlopeChart || grapher.isMarimekko) {
-            // chart types that display all entities by default shouldn't specifically select any
+            // chart types that display all entities by default shouldn't select any by default
             selection.clearSelection()
         } else if (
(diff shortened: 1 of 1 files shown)


adminSiteClient/EditorBasicTab.tsx
```diff
@@ -86,23 +86,32 @@ class DimensionSlotView extends React.Component<{
...
             !grapher.isStackedArea &&
             !grapher.isStackedBar
         ) {
-            const entity = availableEntityNameSet.has(WorldEntityName)
-                ? WorldEntityName
-                : sample(availableEntityNames)
-            if (entity) selection.setSelectedEntities([entity])
+            // non-stacked charts with multiple y-dimensions should select a single entity by default.
+            // if possible, the currently selected entity is persisted, otherwise "World" is preferred
+            if (selection.numSelectedEntities !== 1) {
```
sophiamersmann, line 98: We could also pick the first entity if multiple entities are currently selected

sophiamersmann:
it's a low risk change, so I'll just merge this in now :)

Issue #2707, opened 2023-10-06
SVG is not properly escaped and conversion to png raises an error


Marigold:
## Description

SVG of this [chart on plastic waste](http://staging-site-mojmir/admin/charts/7221/edit) contains the following tag
```
<text x="606.129" y="497.23305925518" font-size="11.700000000000001" fill="#aaa"><1%</text>
```
the text contains `<1%`, but `<` is not escaped and we get an error when opening SVG or trying to convert it to PNG. This **breaks baking**.

## Expected behaviour

It should be escaped and SVG should be valid.

## Steps to reproduce

Download SVG from [this chart](http://staging-site-mojmir/admin/charts/7221/edit) and open it in the browser.

## TODO

* [ ] undo this [hotfix in ETL](https://github.com/owid/etl/commit/8a431b38a65570b4da81650630df6a9ff12e7ad5)

danyx23:
Fixed via #2709

PR #2708, opened 2023-10-06
hotfix(grapher): escape html entities in formatted numbers

sophiamersmann:
Hot fix for #2707


Marigold (approved):
Thanks for fixing this! Looks good (some unit tests are failing by the way)

sophiamersmann:
Ah, that makes sense. Let me quickly fix those

sophiamersmann:
oh no, why does it show up as "&lt;" in some charts??

sophiamersmann:
Investigating...

sophiamersmann:
Okay, I'm back at square one. The code that I put in really shouldn't be necessary because React (or JSX) automatically escapes html entities. That usually works (we have many charts with "<") but for this specific slope chart, it doesn't - I'm not sure why because it doesn't look special to me. I'll meet Daniel in a few minutes - I'll discuss this with him. Maybe he has an idea...

PR #2720, opened 2023-10-10
enhance(scatter): remove entities from the data table that are not displayed in the chart

### LINKED ISSUE #2692: Scatter plot: Remove entities from the data table that are not displayed in the chart
## Description

- Scatter plots often use population to size circles
- Entities that have no data for any year for the x- and y-dimension, but do have data for the size-dimension show up in the data table
- We would rather not list those entities in the table

## Expected behaviour

Don't list entities that only have data for the size-dimension in the data table.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/gini-coefficient-after-vs-before-tax-wb-scatter
2. Note that Afghanistan is not displayed in the chart and is also not listed in the entity selector dialog
3. Note that the table lists Afghanistan as an entity

## Screenshots

Screenshot 2023-10-03 at 11 04 45
Screenshot 2023-10-03 at 11 04 52

## Additional context

This came up in the Grapher redesign bug bash.

sophiamersmann:
fixes #2692

### Problem

- In scatter plot charts, entities without data along the x- and y-axis ended up being displayed in the data table because they did have data for the size dimension (often, the size dimension is population)

### Solution

This PR makes sure that only entities show up in the data table that are actually displayed in the chart:
- Entities in scatter plots that don't have data for the x- and y-dimension are removed from the table view
- Entities in scatter plots and Marimekko charts that were manually excluded by the author are also removed from the table view

### Technical details

- Grapher has a new computed property, `tableForDisplay`
- Each chart type optionally can have a `transformTableForDisplay` method that produces `tableForDisplay`
- `tableForDisplay` is passed to `<DataTable />`

### DIFF
--- a/packages/@ourworldindata/grapher/src/chart/ChartInterface.ts
+++ b/packages/@ourworldindata/grapher/src/chart/ChartInterface.ts
@@ -31,6 +31,7 @@ export interface ChartInterface {
     // Todo: should all charts additionally have a placedSeries: ChartPlacedSeries[] getter?
 
     transformTable: ChartTableTransformer
+    transformTableForDisplay?: ChartTableTransformer
 
     yAxis?: HorizontalAxis | VerticalAxis
     xAxis?: HorizontalAxis | VerticalAxis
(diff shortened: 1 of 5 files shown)


marcelgerber (approved):
Nicely done, this makes lots of sense and I like the `tableForDisplay` approach!

Issue #2379, opened 2023-06-28
Tracking issue: scaling data pages


larsyencken:
_Things we need to do to sustainably author data pages as part of our work, continuing from owid/owid-grapher#1946._

- [x] Finalise the metadata schema (see: [proposal](https://www.notion.so/owid/2023-06-26-Proposed-metadata-structure-that-enable-data-pages-83b3602d12004796a2034fc870911137?pvs=4))
- [x] Make authoring data pages possible
  - [x] https://github.com/owid/etl/issues/1286
  - [x] owid/owid-grapher#2383
  - [x] owid/owid-grapher#2481
  - [x] owid/owid-grapher#2392
  - [x] https://github.com/owid/etl/issues/1364 - adapt ETL workflow, e.g. propagate Origins when working with Table class and add Origins as default to walkthrough
- [x] Finish data pages core features
  - [x] #2736
  - [x] #2913
- [x] Make authoring data pages routine
  - [x] owid/etl#1474
  - [x] owid/owid-grapher#2482
  - [ ] owid/owid-grapher#2523
  - [x] owid/etl#1666
- [x] Improvements after initial shipping
  - [x] #2740
- [x] Cleanup
  - [x] #2741
- [ ] ~~Enable baking of public data pages via metadata that replace grapher pages~~

JoeHasell:
@larsyencken @danyx23 One thing that is missing (or not prominently featured) from this list is the work on 'tooling' needed to make it easier a) for data managers to fill out the metadata; b) for researchers and data managers to collaborate on the metadata.

Essentially, these were one of our four big bets for the year ('technical text admin'). It feels like our ambition there has implicitly been cut down quite considerably, without us fully reflecting on it.

Perhaps we could discuss this on Weds?

larsyencken:
Hey Joe! I don't think an admin's off the table, but it's more that it might be in addition to, and after, many of the things above. In particular, the Related Research & Writing and Related Data blocks are still outstanding and need to be built. There's also the desire to see what pain points emerge from use, although if we waited for than then an admin would really be post-Porto. 

Perhaps chat next week when Daniel's back?

danyx23:
There is a follow-up tracking issue for work that we decided not to do as part of the Sept/Oct/Nov cycles: #2949

PR #2739, opened 2023-10-11
:tada: Add related research and writing via content graph to data pages

danyx23:
This PR implements #2379. It adds the missing link in our db from wordpress posts to charts that are used there. It then uses this new posts_links table together with the existing posts_gdocs_links table to find the related writing for a data page by going from indciator id -> charts using this indicator -> articles using this indicator.

The posts_links table was modelled on the posts_gdocs_links table as I thought that uniformity is more important than the optimal layout here. Extracting the links is a bit crudely done ATM in that it just uses regex's on the raw html tag instead of parsing the html and querying for a tags. The latter would give us the text content of the content that establishes the links which is probably often useful, but it would complicate and slow down the script. I'd like to hear your opinions on whether this should switch to proper parsing and filling richer information into the DB.

The thumbnail rendering is also a bit ad-hoc. We have an Image component but that one is built for use in gdocs and we need to show thumbnails for both WP posts and Gdocs articles.

To rank related research and writing we use the pageviews table. This is empty by default in dev environments and so this PR adds a make command to refresh pageviews (fetched from datasette-private)

- [ ] ❗ after merging this to production, run the db/syncPostsToGrapher.js script to fill the new relationship table!

### DIFF
--- a/db/migration/1692042923850-AddPostsLinks.ts
+++ b/db/migration/1692042923850-AddPostsLinks.ts
@@ -7,7 +7,7 @@ export class AddPostsLinks1692042923850 implements MigrationInterface {
                 id int NOT NULL AUTO_INCREMENT,
                 sourceId int NOT NULL,
                 target varchar(2047) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
-                linkType enum('url','grapher','explorer') CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs DEFAULT NULL,
+                linkType enum('url','grapher','explorer', 'gdoc') CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs DEFAULT NULL,
                 componentType varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
                 text varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
                 queryString varchar(2047) COLLATE utf8mb4_0900_as_cs NOT NULL,
--- a/db/migration/1692042923850-AddPostsLinks.ts
+++ b/db/migration/1692042923850-AddPostsLinks.ts
@@ -1,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class AddPostsLinks1692042923850 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        queryRunner.query(`-- sql
+            CREATE TABLE posts_links (
+                id int NOT NULL AUTO_INCREMENT,
+                sourceId int NOT NULL,
+                target varchar(2047) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
+                linkType enum('url','grapher','explorer') CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs DEFAULT NULL,
+                componentType varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
+                text varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs NOT NULL,
+                queryString varchar(2047) COLLATE utf8mb4_0900_as_cs NOT NULL,
+                hash varchar(2047) COLLATE utf8mb4_0900_as_cs NOT NULL,
+                PRIMARY KEY (id),
+                KEY sourceId (sourceId),
+                CONSTRAINT posts_links_ibfk_1 FOREIGN KEY (sourceId) REFERENCES posts (id)
+                ) ENGINE=InnoDB;`)
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        queryRunner.query(`-- sql
+            DROP TABLE IF EXISTS posts_links;
+        `)
+    }
+}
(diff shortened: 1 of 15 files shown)


danyx23:
* master
  * **PR #2848** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2848" target="_blank">Graphite</a> 
    * **PR #2739** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2739" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2739?utm_source=stack-comment).

db/model/PostLink.ts
```diff
@@ -0,0 +1,47 @@
...
+    static createFromUrl({
+        url,
+        sourceId,
+        text = "",
+        componentType = "",
+    }: {
+        url: string
+        sourceId: number
+        text?: string
+        componentType?: string
```
marcelgerber, line 29: Since `text` and `componentType` are both `NOT NULL`, is it valid to have them as optional here?
Will TypeORM be clever about it and turn them into an empty string if missing, or will it fail?

db/syncPostsToGrapher.ts
```diff
@@ -222,6 +299,24 @@ const syncPostsToGrapher = async (): Promise<void> => {
             featured_image: post.featured_image || "",
         }
     }) as PostRow[]
+    const postLinks = await dataSource.getRepository(PostLink).find()
```
marcelgerber, line 302: Don't know if there's a difference somewhere but this should also work?

```suggestion
    const postLinks = await PostLink.find()
```

db/syncPostsToGrapher.ts
```diff
@@ -99,6 +104,78 @@ export async function buildReusableBlocksResolver(): Promise<BlockResolveFunctio
...
+            ...allSrcs.map((link) => PostLink.createFromUrl(link)),
+            ...allProminentLinks.map((link) => PostLink.createFromUrl(link)),
+        ],
+        postLinkCompareStringGenerator
+    )
+
+    const linksToAdd: PostLink[] = []
+    const linksToDelete: PostLink[] = []
+
+    // This is doing a set difference, but we want to do the set operation on a subset
```
marcelgerber, line 166: There's the slight technicality in here that the same link can appear multiple times in the same document, which is something we don't detect here.
This is fine for our current use case. And is probably also fine overall (especially seeing that WP is gonna be dead soon anyways).

If you agree, we can also change out the `groupBy` calls to `keyBy`.

db/wpdb.ts
```diff
@@ -660,6 +663,145 @@ export const getRelatedChartsForVariable = async (
...
+                    from post_tags pt
+                    join tags t on pt.tag_id = t.id
+                    where pt.post_id = p.id
+                ) as tags
+            from
+                posts_links pl
+            join posts p on
+                pl.sourceId = p.id
+            left join charts c on
+                pl.target = c.slug
```
marcelgerber, line 706: If I see this right this will only work if there's _not_ a redirect?
And otherwise, `chartSlug` above will be null, and the join of `chart_dimensions` will also not work?

db/wpdb.ts
```diff
@@ -660,6 +663,145 @@ export const getRelatedChartsForVariable = async (
...
+                cd.chartId = c.id
+            left join pageviews pv on
+                pv.url = concat('https://ourworldindata.org/', p.slug )
+            left join posts_gdocs pg on
+            	pg.id = p.gdocSuccessorId
+            left join post_tags pt on
+                pt.post_id = p.id
+            where
+                pl.linkType = 'grapher'
+                and componentType = 'src' -- this filters out links in tags and keeps only embedded charts
```
marcelgerber, line 719: I don't know what that comment means? Can we reword it somehow?

db/wpdb.ts
```diff
@@ -660,6 +663,145 @@ export const getRelatedChartsForVariable = async (
...
+                and cd.variableId = ?
+                and cd.property in ('x', 'y') -- ignore cases where the indicator is size, color etc
+                and p.status = 'publish' -- only use published wp charts
```
marcelgerber, line 722:
```suggestion
                and p.status = 'publish' -- only use published wp posts
```

db/wpdb.ts
ikesau: (explaining for my own understanding as well 😊)

`linkType = 'grapher'` is for URLs with paths that match on `/^\/grapher\/[\w]+/`

`componentType = 'src'` is for the links that match the `anySrcRegex` (and not the `anyHrefRegex` or `prominentLinkRegex`

So this clause is filtering out any links that may be from an inline link to a grapher, because we only want to count grapher embeds (i.e. links from iframes - `<iframe src="https://ourworldindata.org/grapher/life-expectancy"`)

db/migration/1692042923850-AddPostsLinks.ts
```diff
@@ -0,0 +1,26 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class AddPostsLinks1692042923850 implements MigrationInterface {
```
ikesau, line 3: image

A tragedy of alphabetization 🥲

db/wpdb.ts
```diff
@@ -660,6 +663,145 @@ export const getRelatedChartsForVariable = async (
...
+                and coalesce(pg.published, 0) = 0 -- if the wp post has a published gdoc successor then ignore it
```
ikesau, line 723: There are cases where we have a published WP post that is succeeded by a published Gdoc, that isn't linked via `gdocSuccessorId` (e.g. Topic Pages do this)

So you might want to handle that case too and filter out WP posts that share a slug with a published Gdoc post.

db/model/PostLink.ts
marcelgerber: Ah I see! I hate how with object destructuring in function parameters, you have to do the destructuring and TS typing separately...

db/wpdb.ts
danyx23: Great catch, thanks!

danyx23:
Hey @ikesau and @marcelgerber, thanks for the feedback! I've incorporated what you mentioned. Ike, you wanted to take another look IIRC - let me know if you see anything else :)

PR #2747, opened 2023-10-12
Add description key and description processing to sources tab

pabloarosado:
We used to show plenty of information about variables (including information about the processing of the variable) via the `description` field, which is surfaced in the sources tab. But now, with the new metadata, we are only surfacing `descriptionShort`, and therefore we are currently missing crucial information in charts.

This PR adds a section in the variable description table called "Key information" (showing `descriptionKey`) and another called "Processing notes" (showing `descriptionProcessing`).

~NOTE: For some reason, `descriptionProcessing` is always empty in the database `variables` table, but I understand that this was unintended, and can be easily fixed.~ (Mojmir has fixed this)


pabloarosado:
This is an example of how the sources tab will look like:
Screen Shot 2023-10-12 at 12 17 19

One thing to note is that bullet points within the string field `descriptionProcessing` (that in the YAML file were created using `- Blah`) look different to the bullet points of `descriptionKey`. But I guess this is a minor style concern.

Marigold (approved):

lucasrodes (approved):
LGTM, thanks for the rapid implementation

PR #2785, opened 2023-10-17
Entry Emulator - hide sticky nav on linear topic pages

ikesau:
Is it a hack if you use the code as was intended? 😁 

This adds an empty array to the gdoc content object's `sticky-nav` property, which causes the sticky nav to not render.

It was either this, or some sort of flag to hide the sticky nav. This seemed like the less intrusive way to handle it.

### example:
image

### archie excerpt:
image


danyx23:
Yeah looks good! Just add it to the documentation and maybe tell authors briefly about it in some slack channel :)

danyx23 (approved):

PR #2789, opened 2023-10-18
:sparkles: speed up image sync by replacing s3cmd with awscli

Marigold:
`s3cmd` is really slow. Syncing images takes more than 3 minutes now, which is considerable time for our staging servers. Switching to `awscli` reduces the time to about 15s. (I'm not sure why is it so much faster, parallel sync perhaps?)

Moreover, using `.aws/config` is more intuitive than `~/.s3cfg`

### DIFF
--- a/devTools/docker/sync-s3-images.sh
+++ b/devTools/docker/sync-s3-images.sh
@@ -13,4 +13,4 @@ fi
 # for local development, it should be owid-image-upload/local-yourname
 # at least until we decide to instead host images locally, if ever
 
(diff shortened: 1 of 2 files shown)


Marigold:
Eh, damn. `aws s3 sync` doesn't transfer public/private permissions. We'd have to run it with `--acl public-read`, but that would make all objects public. @ikesau do we have any private images?

Marigold:
Thanks! I went over all images in that bucket and none of them was private. Merging this as it impacts staging server performance a lot.

PR #2794, opened 2023-10-18
Sources modal: simplify CoreColumnDef

sophiamersmann:
Mild refactor of metadata handling in Grapher.

- Cleans up `CoreColumnDef` to closer resemble the given metadata json hierarchy
    - In particular, `source` and a `presentation` objects are preserved

Other changes:
- Enables keyboard shortcuts on data pages, and adds a shortcut for the sources modal (clicking `s`)

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreColumnDef.ts
+++ b/packages/@ourworldindata/core-table/src/CoreColumnDef.ts
@@ -80,7 +80,7 @@ export interface CoreColumnDef extends ColumnColorScale {
 
     // Metadata v2
     origins?: OwidOrigin[]
-    presentation?: Omit<OwidVariablePresentation, "topicTagsLinks">
+    presentation?: OwidVariablePresentation
 
     // Dataset information
     datasetId?: number
--- a/packages/@ourworldindata/core-table/src/CoreColumnDef.ts
+++ b/packages/@ourworldindata/core-table/src/CoreColumnDef.ts
@@ -3,6 +3,7 @@ import {
     OwidVariableDisplayConfigInterface,
     ToleranceStrategy,
     OwidOrigin,
+    OwidSource,
 } from "@ourworldindata/utils"
 import { CoreValueType, Color } from "./CoreTableConstants.js"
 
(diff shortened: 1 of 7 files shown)


sophiamersmann:
* master
  * **PR #2794** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2794" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2794?utm_source=stack-comment).

packages/@ourworldindata/core-table/src/CoreColumnDef.ts
```diff
@@ -76,16 +75,12 @@ export interface CoreColumnDef extends ColumnColorScale {
...
-    dataPublisherSource?: string
-    retrievedDate?: string
-    additionalInfo?: string
-    attribution?: string
+    source?: OwidSource
     timespanFromMetadata?: string
 
+    // Metadata v2
     origins?: OwidOrigin[]
+    presentation?: Omit<OwidVariablePresentation, "topicTagsLinks">
```
danyx23, line 83: Why is this typed to omit the topicTagsLinks? It will have them at runtime, no?

danyx23 (commented):
Looks good but I'm confused about the type presentation - can you clarify that? Thanks!

packages/@ourworldindata/core-table/src/CoreColumnDef.ts
sophiamersmann: yeah it does have access to topicTagsLinks at run time. it's just not supposed to use it... should I simplify?

danyx23: Why is it not supposed to use it?

sophiamersmann: Don't know, I guess it can. I just didn't think topic page related info is useful to Grapher

danyx23: I think with the sources/about tab in particular we see that this is not a clear line to cut. I appreciate your instinct towards compartmentalizing but I think there is a good chance that in places like the about tab in particular we'll want to make use of all the metadata we have. E.g. tags could be used to group indicators or provide inline links to topic pages. I think I would just pass the full object through.

danyx23 (approved):
I'd say let's just use the full object without the Pick

packages/@ourworldindata/core-table/src/CoreColumnDef.ts
sophiamersmann: Makes sense!

Issue #2755, opened 2023-10-12
Data pages: fix rendering and wording issues


danyx23:
* [x] Following [Ed's suggested](https://owid.slack.com/archives/C0193RW5E2J/p1695720931545929?thread_ts=1695717004.876169&cid=C0193RW5E2J) (it seems we already decided about it at some other time in the past) scatter plots should not show population in the footer of the chart, but they should show population in the sources tab (in "Data published by").
* [x] The heading of the text section "What you should know about this data" should change to something less strong when showing something other than the description_key field and be hidden if we have no text.
* [x] Bullet points from description_key are not picked.
* [x] I have the text `How the producer of this data - undefined - describe this data?`
* [x] Collapse snapshot origins into a single one when they refer to the same data product.
* [x] The text in the full citation is faded too early (see screenshot in comments below)
* [x] Instead of mentioning Twitter (in "How to cite this data"), we should say "social media".
* [x] We should not say "OWID" anywhere (as it happens in "How to cite this data").
* [x] It seems like the link we provide in the full citation is the download link. I suppose it should instead be the main url of the source.
* [ ] ~Also, should we include URLs? (e.g. What happens if it's very long and messy? How to ensure links won't break?)~
* [x] We should show the date of next expected update. This should be based on `dataset.update_period_days` and the current date (if it's passed, increase it by a month).
* [x] Currently, the short and full citation seem to use the format "Producer 1; Producer 2" and "Producer, data product title; Producer 2, data product title 2". Maybe they should instead use the indicator's `attribution_short` and `attribution`.
* [x] Change the wording of "processed" or "adapated" by under the chart to "with minor/major processing by Our World In Data"
* [x] The slugify function that creates links from topic tags (so that clicking the topic tags in the top right links to the topic page) needs to be able to handle & and ' characters
* [x] If the time range for an indicator includes negative years (like our long run population data does), then show `10000 BCE to 2020 CE` instead of `-10000-2020`
* [x] Last updated date should render as a formatted date instead of a raw date string
* [x] If there are multiple producers, use "[Main source] and other sources - Processed by Our World in Data"

### The following issues will be fixed by switching to a new markdown engine

This work is tracked in #2758 and will fix these issues that were tracked here:
* [x] The processing description is not rendered properly (e.g. bullet points are put together without line breaks).
* [x] Jump lines and bullet points are not respected in the content of the same section (Note: this will need a new markdown renderer)

### Other work
[...]

PR #2798, opened 2023-10-19
Show only important sources in summaries

danyx23:
This PR introduces two different ways to iterate over the sources of a chart - a `columnsWithSourcesExtensive` and `columnsWithSourcesCondensed`. When rendering the sources at the bottom of the chart we want to only show the most relevant ones (i.e. hide stuff like population used for sizing a scatter) - for this we use the `Condensed` version. When showing the sources tab we use the Extensive so that all our sources are rendered there.

This is related to #2755 and fixes these points:
* [x] Following [Ed's suggested](https://owid.slack.com/archives/C0193RW5E2J/p1695720931545929?thread_ts=1695717004.876169&cid=C0193RW5E2J) (it seems we already decided about it at some other time in the past) scatter plots should not show population in the footer of the chart, but they should show population in the sources tab (in "Data published by").

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -1450,7 +1450,7 @@ export class Grapher
         const { yColumnSlugs, xColumnSlug, sizeColumnSlug, colorColumnSlug } =
             this
 
-        const columnSlugs = excludeUndefined( [
+        const columnSlugs = excludeUndefined([
             ...yColumnSlugs,
             xColumnSlug,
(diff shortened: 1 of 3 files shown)


sophiamersmann (approved):
Looks good to me!

Nit: Maybe it would be nice to always specify columns in the same order? (probably the order you're using for the extended list of source columns)

danyx23:
### Merge activity

* **Oct 24, 12:28 PM**: @danyx23 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2798).

PR #2807, opened 2023-10-19
Add package for shared UI components

danyx23:
This PR adds a 5th folder to our packages/@ourworldindata folder to be used for shared UI components (shared between Grapher and the site). The current motivation for this is that there is quite a bit of overlap now between some elements on the site (e.g. parts of data pages) and some parts of grapher (e.g. the new about tab).

This folder could also be a place for storing shared styles. As a proof of concept I moved the util.scss and variables.scss files into ui-components/src/styles but please everyone let me know if you think this is a bad idea and we should cut the styles up some other way to make re-use of css variables and/or styles easier


sophiamersmann:
Hey, thank your for looking into this!

The Markdown components are good candidates for sharing. Moving `utils.scss` and `variables.scss` into a shared folder also makes sense to me (although Grapher would only make use of bits of it). `typography.scss` and `colors.scss` would also be helpful to share. It would be good to do a minimal refactor of Grapher to actually use the shared sass variables (e.g. font stacks and colors in `grapher.scss`). But obvs that doesn't need to be part of this PR.

On the name, I think I prefer `@ourworldindata/components` over `@ourworldindata/ui-components` (no good reason, just a matter of taste).

If we move datapage components into `@ourworldindata/components`, then it also makes sense to move `SharedConstants.ts` (it contains datapage section ID's).

[I haven't actually run this code because `make up` is broken on this branch]

PR #2810, opened 2023-10-20
:honeybee: update lerna

Marigold:
We've recently had some issues with Lerna when it froze for no reason. I had to reinstall it to make it work. I'm bumping up its version to 6.0.x, hoping it would fix it. (>=6.1 wouldn't build)


Marigold:
Merging as a hotfix, this issue happens on all staging servers.

PR #2811, opened 2023-10-20
fix(modal): don't dismiss when clicking on an interactive element

sophiamersmann:
- In the chart revision tool, authors look at two charts side by side
- If one chart has a modal open, then clicking on the second chart closes the modal of the first one
- This makes it impossible to have both modals open at the same time
- This happens because any outside click is registered and dismisses the modal
- As a fix, I put in an additional check that checks whether the target that is clicked on is interactive
- If it is interactive, we assume the user didn't intend to dismiss the modal, so we keep it open

Is this a sensible solution?

<details>
<summary>Video of the bug:</summary>
https://github.com/owid/owid-grapher/assets/12461810/547a1ae8-4edc-4291-b6a6-36f5d33d5607
</details>

### DIFF
--- a/packages/@ourworldindata/grapher/src/modal/Modal.tsx
+++ b/packages/@ourworldindata/grapher/src/modal/Modal.tsx
@@ -33,10 +33,14 @@ export class Modal extends React.Component<{
     }
 
     @action.bound onDocumentClick(e: MouseEvent): void {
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
Nice solution, makes sense!

PR #2822, opened 2023-10-23
chore(line-legend): remove unnecessary code

sophiamersmann:
I stumbled over this when working on something else.

I don't quite understand the comment that explain why it was necessary to take into account the "Add country" button, but I assume that it's not relevant anymore since we have rewritten Grapher controls from scratch.


marcelgerber (approved):
Good catch!

PR #2827, opened 2023-10-23
🎉 Native gdocs tables

ikesau:
Adds support for tables in our articles via the Gdoc table element.

The `template` property determines which rows/columns get highlighted in the table. Its possible values are: `header-column`, `header-row`, and `header-column-row`.

This is the Gdoc content for the first table in the preview:

image

The inline table gets parsed and converted into an Archie representation:

```
{.table}
template: header-row
[.+rows]
    {.table-row}
        [.+cells]
            [.+table-cell]
                Deadhead
            []
            [.+table-cell]
                Danfan
            []
        []
    {}
    {.table-row}
        [.+cells]
            [.+table-cell]
                rectangular granny glasses
            []
            [.+table-cell]
                steely dan t-shirt
            []
        []
    {}
    {.table-row}
        [.+cells]
            [.+table-cell]
                sense of oneness
            []
            [.+table-cell]
                LA Eyeworks clipons
            []
        []
    {}
    {.table-row}
        [.+cells]
            [.+table-cell]
                beat poetry
            []
            [.+table-cell]
                MacMall catalog
            []
        []
    {}
[]
{}
```

And this then gets parsed as a regular archie component in our `rawToEnriched` pipeline.

tables

Known remaining work:
- [x] Fix list parsing
- [x] Design QA with Marwa


larsyencken:
Hey @ikesau, can you add a test article onto the staging server for this for quicker QA?

danyx23:
I added the env vars for gdocs to the staging server

http://staging-site-gdocs-table/admin
I'm playing around with this doc (which may at any time be broken :) )
http://staging-site-gdocs-table/admin/gdocs/1ImYJfRy_cXvsRTsI1lYgXmh1vG1mMtvnXWrGuDtrGLw/preview

danyx23 (changes requested):
The code and structure look good. When using it I found that any kind of archieml component breaks because apparently the properties are not preserved. Arguably we don't want too many kinds of blocks to appear there anyhow but I think the parsing issue should be resolved. 

Of the block content types that we do support, I think lists are pretty likely to be legit and probably images as well. Using gdoc bullet points broke the rendering of the page.

I think once these issues are resolved it should be good to ship from my side.

ikesau:
Lists broke? They shouldn't! I'll do some more testing

Thanks 🙂

ikesau:
Thank you @sophiamersmann 

I hadn't pushed up the min-width changes on this remote 🤦 

Definitely had the scopes the wrong way around. Need to be more vigilant to stop [automation blindness](https://pluralistic.net/2023/08/23/automation-blindness/) with copilot. 😱 

@danyx23 the table issue should be fixed now. Usually `paragraphToString` takes care of closing tables because there's always a new line before the `[.body]` closing tag, but that wasn't the case inside the table context, so I needed to add a manual check to ensure it gets closed properly before moving onto the next cell.

i.e. we were effectively getting this:

```
{.table-row}
[.+cells]
[.+table-cell]

[.list]
* one

* two
[]
[]
{}
```

(A missing `[]`)

And now we properly get

```
{.table-row}
[.+cells]
[.+table-cell]

[.list]
* one

* two
[]
[]
[]
{}
```

danyx23 (approved):
:shipit:

PR #2830, opened 2023-10-23
Fixed the line seperator bug between colorLegend and SizeLegend in scatterplot charts

Anubhav-2003:
Issue it fixes: #2735 

What kind of change does this PR introduce:
It fixes the line seperator bug between ColorLegend and SizeLegend in the scatterplot charts.

### DIFF
--- a/packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx
+++ b/packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx
@@ -759,13 +759,16 @@ export class ScatterPlotChart
                 <VerticalColorLegend manager={this} />
                 {sizeLegend && (
                     <>
-                        <line
-                            x1={bounds.right - sidebarWidth}
-                            y1={sizeLegendY - 14}
-                            x2={bounds.right - 5}
-                            y2={sizeLegendY - 14}
-                            stroke="#ccc"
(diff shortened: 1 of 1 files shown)


sophiamersmann (changes requested):
Hey @Anubhav-2003

Thank you for your contribution!

The check that you added almost gets us there. In a few instances, the separator line is also removed for charts that have both a colour and a size legend. Check out [this chart](https://ourworldindata.org/grapher/comparison-extreme-poverty-rich-countries), for example. It would be better to rely on `legendItems` or even `legendDimensions` itself to determine whether the colour legend is present or not.

Additionally, it would be nice to remove the extra padding that is only needed when the colour legend is present. The y-position of the size legend is defined here:

https://github.com/owid/owid-grapher/blob/91ad2fcdaa2e5ed6bea54184e83d1827e47850c3/packages/%40ourworldindata/grapher/src/scatterCharts/ScatterPlotChart.tsx#L742

It would be great if you could update `sizeLegendY` such that the size legend is rendered at the very top if the colour legend is not displayed.

Anubhav-2003:
@sophiamersmann I have tried to implement the changes according to your instructions. Please review my changes. Thank You.

sophiamersmann (approved):
Looking good! I'll merge this code into master now. Thanks again for your contribution!

PR #2833, opened 2023-10-23
Keep time axis constant while playing the timeline animation

sophiamersmann:
Whoever reads this, feel free to pick it up :)

---

Keeps the horizontal time axis constant while playing the timeline animation

- Works for line charts and stacked area charts
- Stacked bar charts need more work because they don't use our `HorizontalAxis` component

Other changes:
- Removed the timeline's `userHasSetTimeline` property (it wasn't currently in use)
- Small UX win: when a user starts dragging while an animation is playing, the animation is stopped

To do:
- [ ] Make it work for StackedBar charts
- [ ] (optional): Keep y-axis domain constant for LineCharts, StackedArea and StackedBar charts


sophiamersmann:
I had pinned this because I thought I had time to get back to this after Porto, but I don't think I'll be working on this any time soon..

Issue #2577, opened 2023-08-25
Stacked area charts should not be facetable by metric


sophiamersmann:
## Core Problem

In stacked area charts, we stack multiple indicators together. Whenever we do that, we've determined that that aggregation makes sense. But, if we facet by metric and stack across entities, it seems to not make sense.

For example:

- When the indicators use % as a unit, we cannot facet by metric (see [Slack example](https://owid.slack.com/archives/C03NV9Z3YSV/p1692913141364249))
- Stacking a mix of aggregates and countries on top of each other doesn't make sense (e.g. stacking France, Europe, and World means that France is counted three times, https://ourworldindata.org/grapher/above-or-below-extreme-poverty-line-world-bank?facet=metric&country=FRA~Europe+and+Central+Asia~OWID_WRL)

## Proposed Solution

Disallow faceting stacked area by indicator.

larsyencken:
Another [example](https://ourworldindata.org/grapher/above-or-below-extreme-poverty-line-world-bank?facet=metric&country=Sub-Saharan+Africa~BGD) reported by Este in [Slack](https://owid.slack.com/archives/C46U9LXRR/p1697844233815379)

danyx23:
@sophiamersmann can you write a follow up issue please? Thanks!

sophiamersmann:
This issue has been resolved, with the exception of forbidding non-sensical stacking, which is transferred to [this follow-up issue](https://github.com/owid/owid-grapher/issues/3608)

PR #2835, opened 2023-10-24
fix(faceting): disable faceting by metric if percentages are stacked

sophiamersmann:
- Disables faceting by metric for stacked area/bar charts if percentages are stacked
- Example: https://ourworldindata.org/grapher/men-weight-categories?facet=metric&country=DEU~ESP~FRA
- Relates to #2577

### DIFF
--- a/packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
+++ b/packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
@@ -336,8 +336,8 @@ export class AbstractStackedChart
 
         if (
             hasMultipleYColumns &&
-            // Stacking percentages doesn't make sense
-            !hasPercentageUnit &&
+            // Stacking percentages doesn't make sense unless we're in relative mode
+            (!hasPercentageUnit || this.manager.isRelativeMode) &&
             // Stacking countries/regions doesn't make sense if aggregates are present
             (!isAggregateEntitySelected || !areMultipleEntitiesSelected)
         )
--- a/packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
+++ b/packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
@@ -15,7 +15,6 @@ import {
     flatten,
     guid,
     max,
-    isAggregateEntity,
 } from "@ourworldindata/utils"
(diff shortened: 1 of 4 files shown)


packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
```diff
@@ -310,25 +311,37 @@ export class AbstractStackedChart
...
+        if (areMultipleEntitiesSelected && !hasMultipleUnits)
             strategies.push(FacetStrategy.entity)
 
-        if (hasMultipleYColumns) strategies.push(FacetStrategy.metric)
+        if (
+            hasMultipleYColumns &&
+            // Stacking percentages doesn't make sense
+            !hasPercentageUnit &&
+            // Stacking countries/regions doesn't make sense if aggregates are present
+            (!isAggregateEntitySelected || !areMultipleEntitiesSelected)
```
sophiamersmann, line 342: this is a little too strict, because stacking continents on top of each other, for example, is okay. the problem is that the is-this-an-aggregate-check is true for all aggregates, including continents/regions/"World". so, if we wanted to allow stacking aggregates, we would need to write additional checks to ensure that, for example, "World" is not stacked on top of a country or region. theoretically, we do have that information (`regions.json`) states members for each aggregate, but I don't think this affects a lot of charts and I didn't want to go all-in

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
```diff
@@ -310,25 +311,37 @@ export class AbstractStackedChart
...
 
-        if (hasMultipleEntities && !hasMultipleUnits)
+        if (areMultipleEntitiesSelected && !hasMultipleUnits)
             strategies.push(FacetStrategy.entity)
 
-        if (hasMultipleYColumns) strategies.push(FacetStrategy.metric)
+        if (
+            hasMultipleYColumns &&
+            // Stacking percentages doesn't make sense
+            !hasPercentageUnit &&
```
sophiamersmann, line 340: this is a little too strict, I think. currently, relative mode is disabled when faceted by metric, but - in theory - it should be possible to stack percentages when we're in relative mode (and can be sure that the percentages add up to 100)

danyx23 (approved):
LGTM!

PR #2843, opened 2023-10-24
Fixed the error related to placement of x-axis tick labels consistently

Anubhav-2003:
Fixes: #2690 

The following PR fixes the placement of x-axis tick labels for the extreme positions to center.

### DIFF
--- a/packages/@ourworldindata/grapher/src/axis/Axis.ts
+++ b/packages/@ourworldindata/grapher/src/axis/Axis.ts
@@ -526,14 +526,6 @@ export class HorizontalAxis extends AbstractAxis {
         let xAlign = HorizontalAlign.center
         const left = x - width / 2
         const right = x + width / 2
-        if (left < this.rangeMin) {
-            x = this.rangeMin
-            xAlign = HorizontalAlign.left
-        }
-        if (right > this.rangeMax) {
-            x = this.rangeMax
-            xAlign = HorizontalAlign.right
-        }
         return {
             value,
             formattedValue,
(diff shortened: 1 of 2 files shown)


Anubhav-2003:
@sophiamersmann Kindly review my PR. Thank You.

marcelgerber:
Hi @Anubhav-2003, very nice work!

When looking at this I also looked back at #975, which is the PR that introduced this code back in 2021. Back then, the main problem was that in some cases when faceting, axis labels would overlap.

And while this PR works great in most cases, and the result indeed looks much better, there are also a few cases where it causes overlapping or cut-off text:

<details>
<summary>A bunch of screenshots</summary>

CleanShot 2023-10-25 at 14 22 35
CleanShot 2023-10-25 at 14 22 57
CleanShot 2023-10-25 at 14 23 05
CleanShot 2023-10-25 at 14 23 12
CleanShot 2023-10-25 at 14 23 19

</details>

However, I agree that we can do better in most cases, and in my head I'm currently thinking that the following would be a good solution:
- Place the x axis tick labels centered below the ticks, except if
	1. the chart is in faceting mode, or
	2. the axis labels are "particularly long" (some heuristic, but maybe more than 6 or 7 characters?)

Another (potentially more scalable) alternative would also be to pass to an axis a "safe zone" where it can render text without causing overlap - and where it would instead default to the right/left-aligning if the space is not enough.

I checked the `AxisConfig` interface, and sadly there's currently not a good way to determine (1).

---

So, overall, this requires some more work.
Feel free to continue working on this, Anubhav, but also don't feel obliged to do so. This is not quite that easy to solve any more and will require some hours.
Also, do feel free to reach out to me or Sophia whenever you need our input. Thank you!

marcelgerber (changes requested):
See the comment above ⬆️

marcelgerber:
Yes that sounds great!
Again, this issue actually turns out to be not super easy to solve in a good way. So, don't get too hung up on this one and feel free to pick out another issue - or reach out to us.

PR #2844, opened 2023-10-24
:bug: don't leave buildkite handing for skipped builds

Marigold:
Skipped builds (that happen when there's more than one build in the queue) were in an infinite loop, which eventually led to connection timeouts for those old builds. This PR adds all build states and handles them appropriately.

Additionally:
* await buildkite builds
* put back local bake when not baking on buildkite (enables baking on staging servers and local)

### DIFF
--- a/baker/BuildkiteTrigger.ts
+++ b/baker/BuildkiteTrigger.ts
@@ -4,6 +4,19 @@ import {
     BUILDKITE_BRANCH,
 } from "../settings/serverSettings.js"
 
+type BuildState =
+    | "running"
+    | "scheduled"
+    | "passed"
+    | "failing"
+    | "failed"
+    | "blocked"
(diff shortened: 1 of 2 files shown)


danyx23 (commented):
IIUC there are 11 states and you check for 8 explicitly in the code - i.e. you throw an error if it is failed, not_run or blocked. I think it would be good to spell that out in a comment, maybe towards the end where you throw the error?

Marigold:
@danyx23 good point, added a comment. I've also fixed two more bugs when testing this on staging:

* await buildkite builds
* put back local bake when not baking on buildkite (enables baking on staging servers and local)

danyx23 (approved):
Great! I haven't run it, only read the code. If you want me to try it out please say so but otherwise I think this could be merged

PR #2853, opened 2023-10-25
Data page: About this data (& refactor shared components)

sophiamersmann:
### Description

Implements the redesigned About this data section and refactors shared data page components.

**About this data section:**
- Updating the design meant adding a few pieces of metadata, including `unit`, `additionalInfo` and the indicator's `title`

**Refactor:**
- Component-specific styles are moved into a scss file next to the component file
- Grid-related styles and classes are removed (shared components simply take up the full width of the parent container)
- `<IndicatorBrief />` is split into two components, `<IndicatorKeyData />` and `<IndicatorDescriptions />` (makes it a little easier for the sources modal)
- Metadata helper functions are moved into `metadataHelpers.ts` in the utils folder

**Other changes:**
- I didn't touch any of the logic; with some exceptions (hehe)
    - Origins are deduplicated based on their label (name of producer, title of data product)
    - The heading of a source shows the producer and title of the data product (needs Marwa's input) 
- I fixed minor visual glitches 

### Open questions

- Should we keep referring to indicators in code as "variables", or do we want to start using both terms interchangeably?


packages/@ourworldindata/components/src/IndicatorSources/IndicatorSources.scss
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    .key-data__title {
+        color: $blue-90;
+    }
+
+    .key-data--hide-overflow div:last-child {
+        white-space: nowrap;
+        text-overflow: ellipsis;
+        overflow: hidden;
```
danyx23, line 38: This looks interesting but exceeds my CSS understanding. What does this do?

danyx23 (approved):
Very nice work! The components are so much tidier now, love it :)

I have one comment about prefering not to switch to an enum - otherwise this is good to ship!

danyx23 (commented):
Very nice work! The components are so much tidier now, love it :)

I have one comment about prefering not to switch to an enum - otherwise this is good to ship!

packages/@ourworldindata/components/src/IndicatorSources/IndicatorSources.scss
sophiamersmann: I think this is code I copied over. It makes it so, that the content doesn't wrap into a new line. Instead, if the content overflows, three dots "..." are added at the end of the line

sophiamersmann:
Ah, I think for now we're referring to indicators as "variables" in code. Should I rename the components to use "Variable" instead of "Indicator" or should we just start using both terms interchangeably?

danyx23:
Yeah this is a bit tricky. I think when communicating to the user it should always be Indicator. In the database etc it will be Variable for a long time probably. For cases like components it's a bit open. I'd say let's use both and for entirely new components like here use Indicator?

sophiamersmann:
sounds good to me!

thanks for the quick review, Daniel!

danyx23:
### Merge activity

* **Nov 27, 6:24 AM**: @danyx23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2853).

PR #2856, opened 2023-10-25
enhance(scatter): don't animate scatter plots as time scatter

### LINKED ISSUE #2855: Don't animate a scatter plot as a time scatter
Animating a time scatter can be overwhelming. It would be better if we animated the dots only.

Example of a scatter plot animation:

https://github.com/owid/owid-grapher/assets/12461810/557c3d11-9b5e-4394-ae7b-4f67f7afe2c9

More videos in this Slack thread: https://owid.slack.com/archives/C0149NKLZAM/p1697710951929099

sophiamersmann:
fixes #2855

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -1031,6 +1031,8 @@ export class Grapher
     }
 
     @computed private get onlySingleTimeSelectionPossible(): boolean {
+        // scatter plots should not be animated as time scatter
+        if (this.isPlaying && this.isScatter) return true
         return (
             this.isDiscreteBar ||
             this.isStackedDiscreteBar ||


marcelgerber (approved):
Yeah this works, but now http://staging-site-enhance-scatter-animation/grapher/life-expectancy-vs-gdp-per-capita is _incredibly_ jumpy, sadly.

sophiamersmann:
uff... that's difficult to watch. Demonstrates really well that it'd valuable to keep the axis domains constant while animating.

Let's bank this pr then until on of us has investigated how tricky it would be to keep the axis domains fixed while playing the animation. If we find that it's difficult to do (and we'll likely won't do it or won't to it soon), we can still revisit this pr...

sophiamersmann:
Do you have an intuition of how many scatter plots have sparse data like that? (there is no easy way to query for that, right?)

danyx23:
Hm, I think this is pretty useful and would be nice if we could find a reasonable compromise for fixing the axis. I'll add another idea on #2701

sophiamersmann:
I had pinned this because I thought I had time to get back to this after Porto, but I don't think I'll be working on this any time soon..

Issue #2831, opened 2023-10-23
Tables do not respect tolerance and show misleading data


CGiattino:
Below is Charlie's originally reported issue but we have since found that the problem is actually quite a bit worse and more urgent. In the [Objects launched into space chart](https://ourworldindata.org/grapher/yearly-number-of-objects-launched-into-outer-space?tab=table) there is no tolerance but every country that has had an object launched into space in any year is shown with a value of one in 1957. Clearly we should fix this quickly to only show values if they are within the tolerance range (or only at the exact year if tolerance is not set or 0)

## Description

The year tolerance is not being enforced for tables, even when it's set for that indicator (on the Grapher admin under Basic > opening the drop-down for that indicator). When the tolerance has not been set manually, it would be useful to have a sensible default.

As you can see in the screenshots below, this makes it so that country-years are shown in the table when they are hundreds or even thousands of years different from the year shown in the table column.

## Expected behaviour

Authors should be able to set the tolerance in the Grapher admin of which years should be shown if the target year is missing. When it's not set manually, there should be a sensible default tolerance value — but it's important to ensure that the default doesn't cause problems of its own.

Country-year values outside of the tolerance range should not be shown in the table (nor should associated "Absolute" and "Relative" change be shown for these outside-tolerance values)

## Steps to reproduce

Example charts:
- https://ourworldindata.org/grapher/population?tab=table
- https://ourworldindata.org/grapher/maternal-mortality?tab=table

## Screenshots

![image](https://github.com/owid/owid-grapher/assets/7489781/2cea5699-64f3-4afb-9383-3c0a15afdc4a)

![image](https://github.com/owid/owid-grapher/assets/7489781/d3473192-6229-4eec-ae13-100b148bf881)

![image](https://github.com/owid/owid-grapher/assets/7489781/6fc5a308-5ebf-4ec9-a0e0-9e51e00da5a8)

danyx23:
I just realized that the issue is quite a bit worse - if no tolerance is set, the table shows data with tolerance matching indicators.

On the [objects launched into space chart](https://ourworldindata.org/grapher/yearly-number-of-objects-launched-into-outer-space?tab=table) there is no tolerance but we show the closest value (1) in the first year (1957) for every country

PR #2859, opened 2023-10-26
fix(data-table): respect tolerance

sophiamersmann:
- fixes https://github.com/owid/owid-grapher/issues/2831

- there are two code paths: "range" mode and "point" mode
  - in "range" mode, a start and end year are selected and we show four columns: Start year, end year, absolute change, relative change
  - in "point" mode, a single year is selected and we show a single column for that year
- in "point" mode, all is good
    - tolerance is appropriately applied and no hundreds-years-away data points are used to fill an empty cell
- the issue is with the "range" mode
  - here, we take the range of available data points within the current selection, and then apply min/max to get the values for the start and end point, i.e. if the user-selected time range is [1000, 2000] but we only have data for [1990, 2000], then we'll end up showing the 1990 value for the year 1000
  - the tolerance defined by the author is not taken into account, as far as I can see
- to fix the issue, we simply grab the closest available data point within tolerance for both the start and the end year

I can see that the special code path for "range" mode was a deliberate choice, but I'm not sure why (to show the delta columns as often as possible?). What am I missing?

### DIFF
--- a/packages/@ourworldindata/grapher/src/dataTable/DataTable.tsx
+++ b/packages/@ourworldindata/grapher/src/dataTable/DataTable.tsx
@@ -675,7 +675,6 @@ export class DataTable extends React.Component<{
                     : TargetTimeMode.range
 
             const prelimValuesByEntity = this.preliminaryDimensionValues(
-                targetTimeMode,
                 sourceColumn,
                 targetTimes
             )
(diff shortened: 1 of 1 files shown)


packages/@ourworldindata/grapher/src/dataTable/DataTable.tsx
```diff
@@ -784,7 +768,11 @@ export class DataTable extends React.Component<{
                     start !== undefined &&
                     end !== undefined &&
                     typeof start.value === "number" &&
-                    typeof end.value === "number"
+                    typeof end.value === "number" &&
+                    // sanity check: start time should always be <= end time
+                    start.time !== undefined &&
+                    end.time !== undefined &&
+                    start.time <= end.time
```
sophiamersmann, line 775: arguably, this could also be `<`

danyx23 (approved):
Thanks a lot for the quick fix, @sophiamersmann! A few rambling comments to your question of why this behaviour was different:
* Visually this used to be a lesser problem because in the past the table showed the year the data point was from. With the redesign we made this more subtle and only show the circled i indicator. It used to be that there were two lines in such cases, where the first one was in gray the year and the second was the value. This I think was acceptable but now it is not obvious enough anymore that the data is not from the year the column says above
* The functionality of this was never great. If you sort by values in the first column in a long run time series like population for year -10000, you used to get the highest value for whatever the first datapoint was for any entity (go to https://ourworldindata.org/grapher/population?tab=table&country=CHN~IND~USA~BRA~NGA~IRN~ZAF and sort by first column - West Germany in 1800. Obviously 😵‍💫 )
* I think the two different code paths where a performance tuning. Computing the relevant values on the first and last year only is obviously better for this case and the current solution might use substantially more RAM for long run time series - but I'd say lets wait and see if this causes any issues (or maybe add a comment that maybe this should be optimized again for the table case where we want to compute changes)

If you compare the 12 most important charts then I think the new behaviour is clearly better in all of them:

Current behaviour:
https://ourworldindata.org/12-key-metrics

Your fixes: 
http://staging-site-fix-data-table-tolerance/12-key-metrics

Old grapher (pre redesign):
https://staging.owid.cloud/admin/posts/preview/24902

So all in all: :shipit:

sophiamersmann:
thank you for taking the time to write all of this down! shipping it 🚀

PR #2863, opened 2023-10-26
🐛 Don't trigger lightning updates if dod document is updated

ikesau:
The DoD document requires we rebake dods.json and grapher embeds that reference them. 

So this PR disqualifies it from ever being a lightning update candidate.


mlbrgl (approved):
👍

Issue #2714, opened 2023-10-09
Tooltip content not nicely formatted when variable name is long and a time notice is displayed


sophiamersmann:
The tooltip content is weirdly formatted when the variable name is long and a time notice is displayed.

## Steps to reproduce

1. Go to https://ourworldindata.org/explorers/coronavirus-data-explorer?zoomToSelection=true&facet=none&country=BHR~BRA~CHL~FRA~DEU~HUN~IND~ISR~SRB~TUR~GBR~USA~URY~ESP~ITA~ARE~MEX~RUS&pickerSort=desc&pickerMetric=population&hideControls=true&Interval=Cumulative&Relative+to+Population=true&Color+by+test+positivity=false&Metric=People+vaccinated+%28by+dose%29
2. Hover over the "Italy" bar
3. Note that the tooltip content is not formatted nicely

## Screenshots

<img width="971" alt="Screenshot 2023-10-09 at 16 11 57" src="https://github.com/owid/owid-grapher/assets/12461810/d08ca53e-ca51-46e3-ac2c-12b506656a60">

ikesau:
@samizdatco is this just a matter of some max-widths here and there? famous last words but that seems simple enough?

sophiamersmann:
Sure!

parthiv360:
Hey @sophiamersmann can you please suggest how to traverse to this page in localhost. Thank you

ikesau:
If your [localhost route](http://localhost:3030/explorers/coronavirus-data-explorer?zoomToSelection=true&facet=none&country=BHR~BRA~CHL~FRA~DEU~HUN~IND~ISR~SRB~TUR~GBR~USA~URY~ESP~ITA~ARE~MEX~RUS&pickerSort=desc&pickerMetric=population&hideControls=true&Interval=Cumulative&Relative+to+Population=true&Color+by+test+positivity=false&Metric=People+vaccinated+%28by+dose%29) isn't loading for you (or any explorer, for that matter) you might be missing https://github.com/owid/owid-content 

Clone it into the same directory that contains this repo and then restart your local server 🙂

parthiv360:
Okay Thanks.

parthiv360:
Hey @ikesau I have gone through this issue and was trying to correct it. I have tracked the pages were the potential changes need to be done. 

I have a confusion that do I need to change the container size to adjust the time or adjust the time or something keeping the size of container fixed.

Thank you.

parthiv360:
[...]

PR #2865, opened 2023-10-27
Tooltip content is formatted

parthiv360:
I have updated the tooltip content by making changes in the tooltip.scss property for issue no #2714.


sophiamersmann:
Hey @parthiv360,

Changing the time notice's text indent is not likely what we want to do here. I haven't looked into it in detail, but it would probably be better to maybe set a maximum width on the series name, or a similar approach.

parthiv360:
@sophiamersmann ok thank you , will give it a try. Can you please help me pinpoint the location if possible? As i Did try to find out different approaches to do it.

PR #2868, opened 2023-10-27
🎉 Entry Emulator - support dateline in topic page header

ikesau:
Adds support for rendering a string dateline in topic page headers, which we'll use for linear topic pages.

Thus I've removed the code that extracted this text and turned it into a callout.

About 30 out of 50 of the entries we wish to migrate feature this text, so instead of writing a weird path that allows us to return metadata out of the `cheerioToArchieML` function, I figure I'll just manually copy-paste it 30 times.

image


ikesau:
Superseded by https://github.com/owid/owid-grapher/pull/2883

PR #2795, opened 2023-10-19
Sources modal: first version using metadata v2 (outdated)

sophiamersmann:
> **Note**
> Work in progress

First version of the sources modal for metadata v2, to be shared with data managers

PR #2877, opened 2023-10-30
Build sources modal from shared components

sophiamersmann:
### Description

Builds sources modal from shared components.

**Handling of metadata:**
- Source link(s) from old metadata
	- If a single source link is given, it is displayed in the "Link" field of the Key data block
	- If multiple source links are given, and they can be matched to their respective source, then the attributions are linkified (and no "Link" field is needed)
	- If multiple source links are given that cannot be matched to the sources, they're displayed in the "Links" field of the Key data block
- If given, old metadata is displayed as a single entry in the "This data is based on the following sources" Section
	- It's a single entry rather than multiple because the source `description` refers to all sources

**Other changes:**
- don't assume processing level is "minor" if no processing level is given

**Known bugs:**
- [x] "Last updated" date is invalid because the catalog's version is "latest" ([example chart](http://localhost:3030/grapher/covid-vaccine-by-age))


sophiamersmann:
* master
  * **PR #2794** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2794" target="_blank">Graphite</a> 
    * **PR #2853** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2853" target="_blank">Graphite</a> 
      * **PR #2877** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2877" target="_blank">Graphite</a>  👈
    * **PR #2795** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2795" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2877?utm_source=stack-comment).

packages/@ourworldindata/components/src/IndicatorSources/IndicatorSources.tsx
```diff
@@ -1,109 +1,150 @@
...
+                const content = (
+                    <SourceContent
+                        source={source}
+                        isEmbeddedInADataPage={isEmbeddedInADataPage}
+                    />
+                )
+                return source.description ||
+                    source.citation ||
+                    source.dataPublishedBy ? (
+                    <ExpandableToggle
```
danyx23, line 31: I wonder if we should show this section expanded if we are in the "About this data" overlay? The data page is pretty long and shows a lot of stuff so the collapsed section makes sense, but if you want to see the sources then why make users click once more? I'm not 100% sure this is the best way forward but this is my hunch.

danyx23 (commented):
Very nice! I had a few detail questions, but overall this is ready to go IMHO. 

I saw that you used css variables in a few places which I don't think we use much so far - I have no opinion on this, but maybe worth chatting about in a grapher hour or so just to make sure we use the same tools for the same kinds of problems.

danyx23:
### Merge activity

* **Nov 27, 6:24 AM**: @danyx23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2877).

PR #2875, opened 2023-10-30
:sparkles: add react-remark library, test it in SimpleMarkdownText

danyx23:
This PR adds the [react-remark](https://github.com/remarkjs/react-remark) library. This seems to be the best library for parsing markdown given the following requirements:
- We want parsing to be separate from rendering, with the ability to modify the AST in between
- We'd like extensions to be straightforward and ideally a nice ecosystem of plugins for things like mathematical formulas, tables etc
- We'd like to avoid `dangerouslySetInnerHTML` in React

This PR then replaces the rendering path of the SimpleMarkdownText component with the new parser. 

Bundle size increases by about 40KB minified + gzipped but I think this is an ok trade-off

PR #2874, opened 2023-10-30
:hammer: fix minor rebase issues

danyx23:

PR #2873, opened 2023-10-30
markdown-transform-dod-links

danyx23:

PR #2872, opened 2023-10-30
:sparkles: add react-remark library, test it in SimpleMarkdownText

danyx23:
:sparkles: add react-remark library, test it in SimpleMarkdownText

:sparkles: replace our svg markdown parser with mdast parser

PR #2884, opened 2023-11-01
Markdown: turn plain URLs into links

danyx23:
By default, plain link like https://domain.com are not shown as links in markdown (see [commonmark spec](https://spec.commonmark.org/0.30/#autolinks)) and instead have to be written in the `[text](https://domain.com)` style. 

This PR adds autolinking to plain links. This is done in two different way
- once for the SimpleMarkdown component that uses react-remark via a remark transformation plugin
- and once for the MarkdownTextWrap component by doing the translation as part of the mapping to IRNodes

In theory it should be possible to use the remark plugin in both cases, but navigating the set of interdependent libraries at the state two years ago is a pain and so I went for the two bespoke versions. When we upgrade our setup to be ESM everywhere we can upgrade to the latest versions of the unified, react-remark et al libraries and try again to use the same code in both cases.

### DIFF
--- a/packages/@ourworldindata/components/src/markdown/remarkPlainLinks.ts
+++ b/packages/@ourworldindata/components/src/markdown/remarkPlainLinks.ts
@@ -1,20 +1,55 @@
 import findAndReplace from "mdast-util-find-and-replace"
 
-export const urlRegex = /https?:\/\/([\w-]+\.)+[\w-]+(\/[\w\- .\+/?:%&=~#]*)?/
+// This regex matches:
+//   "http"
+//   an optional "s"
+//   two / characters
+//   The subdomains and hostname: Any word or numeric character or "_" or "-" one or more times followed by a period
+//   The TLD: Any word or numeric character or "_" or "-" one or more times
+//   The path, query string and fragment: A forward slash followed by any word or numeric character (unicode classes so umlauts like ö match
+//       as well as any of the following: .+?:%&=~#) zero or more times. Note that we exclude space even though that is valid in a URL but it tends
+//       to make the match too greedy.
+//   Note that this URL will tend to match too much at the end - if there is a URL at the end of a sentence, the period will
+//   be matched by the regex. We accept this and handle it below when we create the link.
+//   This URL could be made to not match trailing characters like .?: but AFAIK this requires a negative lookbehind which are not supported in
+//   Safari before 16.4 which came out in Spring 2023
+export const urlRegex =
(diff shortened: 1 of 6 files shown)


danyx23:
* master
  * **PR #2875** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2875" target="_blank">Graphite</a> 
    * **PR #2872** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2872" target="_blank">Graphite</a> 
      * **PR #2884** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2884" target="_blank">Graphite</a>  👈
        * **PR #2873** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2873" target="_blank">Graphite</a> 
          * **PR #2874** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2874" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2884?utm_source=stack-comment).

marcelgerber (approved):

packages/@ourworldindata/components/src/markdown/remarkPlainLinks.ts
```diff
@@ -0,0 +1,23 @@
+import findAndReplace from "mdast-util-find-and-replace"
+
+export const urlRegex = /https?:\/\/([\w-]+\.)+[\w-]+(\/[\w\- .\+/?:%&=~#]*)?/
```
marcelgerber, line 3: Ah, now that I think of it, including `.?:` can be annoying in cases like `Learn more at https://ourworldindata.com/poverty.`.

I know that some other sites do something like "match `.?:` unless they are at the very end of the URL", which seems sensible.

danyx23: Good catch. I took this opportunity to also make sure we match urls with unicode word characters in the path. 

AFAIK to properly match this with a regex you need to use negative lookbehind assertions and they are not supported in Safari before 16.4 which came out in Spring this year so I instead handle trailing punctuation characters in the replacement function.

danyx23:
### Merge activity

* **Nov 7, 3:55 PM**: @danyx23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2884).

PR #2885, opened 2023-11-01
Blockquote component

ikesau:
Adds a [blockquote](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) component for linear topic pages. Previously we were (mistakenly) converting these to pullquote components which are enormous serif text blocks intended for extreme emphasis.

## Archie

```
{.blockquote}
citation: – Bastian Herre
[.+text]
Measuring the state of democracy across the world helps us understand the extent to which people have political rights and freedoms.
[]
{}
```

or

```
{.blockquote}
citation: https://ourworldindata.org/democracies-measurement
[.+text]
Measuring the state of democracy across the world helps us understand the extent to which people have political rights and freedoms.
[]
{}

```


packages/@ourworldindata/utils/src/owidTypes.ts
```diff
@@ -1226,6 +1226,20 @@ export interface EnrichedBlockTableCell {
...
+    value: {
+        text?: RawBlockText[]
+        citation?: string
+    }
+}
+
+export type EnrichedBlockBlockquote = {
+    type: "blockquote"
+    text: EnrichedBlockText[]
+    citation?: string
```
sophiamersmann, line 1240: Is it right that `citation` is marked as optional here?

site/gdocs/ArticleBlock.tsx
```diff
@@ -587,6 +587,34 @@ export default function ArticleBlock({
...
+                            d={textBlock}
+                            key={i}
+                        />
+                    ))}
+
+                    {isCitationAUrl ? null : (
+                        <footer>
+                            <cite>{block.citation}</cite>
+                        </footer>
+                    )}
```
sophiamersmann, line 614: If `citation` is truly optional, this might produce an empty `<footer><cite></cite></footer>` (although it doesn't at the moment because it errors out if no citation is given)

site/gdocs/centered-article.scss
```diff
@@ -70,6 +70,19 @@ $banner-height: 200px;
     margin-top: 0;
 }
 
+.article-block__blockquote {
+    margin-top: 0;
+    p {
+        font-style: italic;
+        &:last-of-type {
+            margin-bottom: 4px;
```
sophiamersmann, line 78: If we only need this margin if the footer/cite block is present, it might be better to make it the footer's responsibility to apply this margin?

sophiamersmann (approved):
Looks good to me!

I'm a little confused about the citation - not sure if it's supposed to be optional or not.

A more general note: Maybe it's just an example, but if we want the citation to always start with a "-" like "- Bastian Herre", then it might be better to automatically insert it rather than requiring the author to type it?

ikesau:
Thanks @sophiamersmann - that's what I get for rushing out the door before testing each state properly. I fixed the code and added a comment in `ArticleBlock` to properly explain the logic 👍

PR #2886, opened 2023-11-02
Sources modal: Multiple indicators

sophiamersmann:
Displays expandable tabs for multiple indicators.

I haven't been very disciplined about pushing changes to the branches they belong to. So, this branch also contains various bug fixes that touch on data pages and the sources modal.

### DIFF
--- a/packages/@ourworldindata/components/src/Tabs/Tabs.tsx
+++ b/packages/@ourworldindata/components/src/Tabs/Tabs.tsx
@@ -1,0 +1,77 @@
+import React, { useLayoutEffect, useRef } from "react"
+import cx from "classnames"
+
+export const Tabs = ({
+    labels,
+    activeIndex,
+    setActiveIndex,
+    horizontalScroll = false,
+}: {
+    labels: string[]
+    activeIndex: number
+    setActiveIndex: (label: number) => void
+    horizontalScroll?: boolean
+}) => {
+    const container = useRef<HTMLDivElement>(null)
+
+    // roving tabindex for keyboard navigation
+    function getNextIndex(eventKey: string): number {
+        switch (eventKey) {
+            case "Home":
+                return 0
+            case "End":
+                return labels.length - 1
+            case "ArrowRight":
+                return activeIndex === labels.length - 1 ? 0 : activeIndex + 1
+            case "ArrowLeft":
+                return activeIndex === 0 ? labels.length - 1 : activeIndex - 1
+            default:
+                return activeIndex
+        }
+    }
+
+    // enable keyboard navigation
+    function handleKeyDown(event: React.KeyboardEvent<HTMLButtonElement>) {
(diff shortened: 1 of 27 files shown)


sophiamersmann:
* master
  * **PR #2794** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2794" target="_blank">Graphite</a> 
    * **PR #2853** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2853" target="_blank">Graphite</a> 
      * **PR #2877** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2877" target="_blank">Graphite</a> 
        * **PR #2886** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2886" target="_blank">Graphite</a>  👈
    * **PR #2795** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2795" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2886?utm_source=stack-comment).

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -258,11 +258,11 @@ export abstract class AbstractCoreColumn<JS_TYPE extends PrimitiveType> {
     }
 
     @imemo get name(): string {
-        return this.def.name ?? this.def.slug
+        return this.def.name || this.def.slug
     }
 
     @imemo get displayName(): string {
-        return this.display?.name ?? this.name ?? ""
+        return this.display?.name || this.name || ""
```
sophiamersmann, line 265: this is why the svg tester comes back with a few updated charts

packages/@ourworldindata/components/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,78 @@
...
+    labels: string[]
+    activeIndex: number
+    setActiveIndex: (label: number) => void
+    horizontalScroll?: boolean
+    slot?: JSX.Element
+}) => {
+    const container = useRef<HTMLDivElement>(null)
+
+    // roving tabindex for keyboard navigation
+    function getNextIndex(eventKey: string): number {
```
danyx23, line 20: ❤️

danyx23 (changes requested):
Looks good! I have a few remarks about the wording in slack and I think we should only show multiple indicator tabs if there is a difference between the content we show for each one (i.e. if all are the same we don't show any tabs; if there are differences it is fine to show all indicators, even if most are the same - i.e. no need to de-duplicate further; but all sources being the same is a common case due to our need to split multi-dimensional indicators and we should not confuse our users with this if we don't have to)

PR #2891, opened 2023-11-02
Linear topic page all charts improvements

ikesau:
Adds an anchor link to the all charts block for linear topic pages, and also updated the `LinkedA` component to not open in a new tab if it's an anchor link.

image


danyx23 (approved):

danyx23 (approved):
Looks good! I haven't tried it but the code seems simple enough :)

PR #2892, opened 2023-11-03
Smaller chromeless explorer embeds in articles & a fix for native tables crashing

ikesau:
Just two small improvements:

1. Renders explorers at 8 columns if `hideControls=true` is present in the query string.
2. Skips parsing native tables if we're not inside a `{.table}` block.
    1. The alternative was turning `{.table-rows}` into its own component which was a lot of new code and would require a migration for any tables that had already been created.


danyx23 (approved):

danyx23 (approved):
:shipit: :)

PR #2893, opened 2023-11-03
Fix thumbnails for prominent link to Graphers / Explorers

ikesau:
An idea at a fix for https://github.com/owid/owid-grapher/issues/2870

I really can't see how prominent links were working with plain Grapher URLs before, except I know they were. Something must have been removed at some point. 😓 

Anyway, this adds a little fork at the Prominent Link level, to load via the Grapher preview worker instead of via the Image attachments system. It doesn't currently take the query string into account, that but seems like an improvement over it breaking.

## Example Archie

```
{.prominent-link}
{}

{.prominent-link}
url: https://ourworldindata.org/explorers/poverty-explorer
{}
```

## Results
image

@marcelgerber do you think this is an okay way to do this? In my mind this is the same sort of dependency as `DATA_API_URL` , but maybe there are reasons we don't want to add this?

If we go with this we also might want to look at how `BAKED_GRAPHER_EXPORTS_BASE_URL` is formed

### DIFF
--- a/.env.devcontainer
+++ b/.env.devcontainer
@@ -8,7 +8,6 @@ GRAPHER_DB_USER=grapher
 GRAPHER_DB_PASS=grapher
 GRAPHER_DB_HOST=db
 GRAPHER_DB_PORT=3306
-GRAPHER_PREVIEW_URL=https://ourworldindata.org/grapher/thumbnail
 DEBUG='knex:query'
 
 GDOCS_PRIVATE_KEY=''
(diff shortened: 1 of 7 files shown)


marcelgerber (changes requested):
Hm, this is interesting.

Afaict, the only problem that we had before is that `<Image filename={BAKED_GRAPHER_EXPORTS_BASE_URL + "/SLUG.svg"} />` didn't work, because `Image.ts` couldn't find any metadata for that image (makes sense!).
On the other hand, going with `` instead works just fine!

---

As I think we've discussed before, I'm still a bit hesitant to go with the thumbnails worker in more places because while it's working great in most cases, there are some charts (e.g. covid charts) where it fails, and it sometimes takes a few seconds to render a thumbnail...

marcelgerber (approved):

PR #2895, opened 2023-11-06
fix(explorer): show data source defined in explorer configs

sophiamersmann:
Reverts some changes made in https://github.com/owid/owid-grapher/pull/2794

### Bug

- Some explorer charts define source information in the explorer tsv config
- That information is currently not taken into account, and no source is displayed (see screenshots)

### What happened

- Source properties used to be spread into the core column definition (as `sourceName`, `sourceLink`, etc)
- The explorer's column grammar defines `sourceName`, `sourceLink`, etc. and sets these properties on the core column defintion
- In https://github.com/owid/owid-grapher/pull/2794, I removed `sourceName`, `sourceLink`, etc in favour of defining a `source` object that holds those properties
- As a result, source information defined in the explorer config are ignored

Screenshot 2023-11-06 at 08 38 15
Screenshot 2023-11-06 at 09 51 55


danyx23 (approved):
Looks good! :shipit:

PR #2925, opened 2023-11-15
🎉 Collections

ikesau:
Adds static and dynamic collections.

Static collections are simple and just take a list of grapher slugs to bake.

Dynamic collections are more complicated as they depend on the browser's query parameters. 

Here's the logic:
1. Extract the slugs* from the brower's `charts` query parameter
2. Construct an `ObservableMap` that is keyed by `slug-index` where `index` is the slug's position in the query parameter, whose values are pointers to refs that will eventually exist when the MultiEmbedder does its thing
3. Attach the map to the window
5. Pass these slugs to the `DynamicCollection` component which renders a page of `<figures>`
6. Have the MultiEmbedder do its thing
7. Once each Grapher has finished mounting, attach it to the correct pointer in the `ObservableMap`
8. Add a computed property, `allGrapherSlugsAndQueryStrings`, that is listening to each Grapher in the `ObservableMap` and serializing the slug+querystring combination of each Grapher embedded in the page
9. Update the browser URL whenever this computed property changes

\* they can also contain URI-encoded query parameters but I'll just call them slugs for simplicity's sake.

### DIFF
--- a/site/collections/CollectionsPage.scss
+++ b/site/collections/CollectionsPage.scss
@@ -21,6 +21,30 @@
         width: 100%;
         height: 650px;
         margin: 0;
-        margin-bottom: 48px;
+        margin-bottom: 16px;
+    }
+
+    .remove-from-collection-button {
+        border: 1px solid $vermillion;
+        color: $vermillion;
+        background-color: transparent;
+        border-radius: 0;
+        padding: 8px;
+        line-height: 16px;
+        margin: auto;
+        display: block;
+        margin-bottom: 24px;
+        cursor: pointer;
+        svg {
+            margin-right: 8px;
+            height: 10px;
+            position: relative;
+            top: -2px;
+        }
+
+        &:hover {
+            border-color: $accent-vermillion;
+            color: $accent-vermillion;
+        }
     }
 }
--- a/site/collections/CollectionsPage.scss
+++ b/site/collections/CollectionsPage.scss
@@ -21,33 +21,6 @@
         width: 100%;
         height: 650px;
         margin: 0;
-        margin-bottom: 16px;
-        .GrapherComponent {
-            width: 100%;
-        }
-    }
-
-    .remove-from-collection-button {
-        border: 1px solid $vermillion;
-        color: $vermillion;
-        background-color: transparent;
-        border-radius: 0;
-        padding: 8px;
-        line-height: 16px;
-        margin: auto;
-        display: block;
-        margin-bottom: 24px;
-        cursor: pointer;
-        svg {
-            margin-right: 8px;
-            height: 10px;
-            position: relative;
-            top: -2px;
-        }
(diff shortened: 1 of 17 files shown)


sophiamersmann:
Hey @ikesau, are you planning to do more work on this, or do you consider this good to go?

ikesau:
@sophiamersmann yup! good to go 🙂

site/collections/CollectionsPage.scss
```diff
@@ -0,0 +1,26 @@
...
+        margin-bottom: 32px;
+
+        &:not(:last-of-type) {
+            margin-bottom: 0;
+        }
+    }
+
+    figure {
+        width: 100%;
+        height: 650px;
```
sophiamersmann, line 22: Can we use `$grapher-height` here?

Would also be nice to use the height of the viewport on mobile.

site/collections/DynamicCollection.tsx
```diff
@@ -0,0 +1,169 @@
...
+        // while still keeping the unmounted graphers in the URL in the right place
+        const slugsAndQueryStrings = new Array(this.graphers.size)
+
+        for (const [originalSlugAndUrl, { index, grapher }] of this.graphers) {
+            if (!grapher) {
+                // Strip index suffix from originalSlugAndUrl
+                const withoutIndex = originalSlugAndUrl.replace(/-\d+$/, "")
+                slugsAndQueryStrings[index] = encodeURIComponent(withoutIndex)
+            } else {
+                slugsAndQueryStrings[index] = encodeURIComponent(
```
sophiamersmann, line 71: Can we remove `encodeURIComponent` here and above? Grapher URLs should be clean and it would be nice to have a readable URL.

site/collections/StaticCollectionPage.tsx
```diff
@@ -0,0 +1,52 @@
...
+    props: StaticCollectionPageProps & { baseUrl: string }
+) => {
+    const { baseUrl, title, charts, introduction } = props
+
+    return (
+        <html>
+            <Head
+                canonicalUrl={`${baseUrl}/shared-collection/top-charts`}
+                pageTitle="Chart Collection"
+                pageDesc="View charts you've saved on Our World in Data."
```
sophiamersmann, line 22: The description doesn't really work for the static collection page?

sophiamersmann (approved):
Looks great!

Love the Top 50 page! (On a side note, when working on Grapher, I often specifically test our most viewed charts. A dedicated page for this will make my life so much easier!)

Regarding the dynamic page, I'm not sure. It took me a while to understand the code. Props for figuring it out, but the complexity of it makes me a bit uncomfortable. I also don't find the page very useful in its current form (with no easy way to add/remove charts). But I know there's been discussions around this, so please feel free to go ahead!

PR #2952, opened 2023-11-27
Revert perf optimization that breaks map charts

### LINKED ISSUE #2948: Data not shown in map view, but available in chart view
## Description

Some charts created with the [Information Capacity Dataset](https://owid.cloud/admin/datasets/6296) are not showing data in their map view, even though it seems correctly formatted. We can even see that the data is displayed on the tooltip view or chart tab.

The view is also not showing any year in the title and the time slider is not available.

## Expected behaviour

To see the data deployed in the map view, with the time slider and year in the title.

## Steps to reproduce

Open any of the draft charts below:

- https://owid.cloud/admin/charts/7397/edit
- https://owid.cloud/admin/charts/7394/edit
- https://owid.cloud/admin/charts/7396/edit
- https://owid.cloud/admin/charts/7400/edit

You'll see that the map shows no data, but the tooltip does.

## Screenshots

This is how the charts are looking now:
image

## Environment

For desktop, provide:

-   OS: macOS Sonoma 14.1.1
-   Browser Chrome
-   Version 119.0.6045.159

## Additional context

We would like to know the reason for this problem as soon as possible, since we release the new State Capacity topic page on the 30th.

sophiamersmann:
❗ **Urgent, should be done by the 30th (Thursday)**

- fixes #2948

### Summary

- Pablo A wants to publish a set of map charts on the 30th
- Some map charts are broken; the data is available in the Table and Chart tab, but not in the Map tab (see #2948)
- I traced it down to the tolerance computation in the MapChart's transformTable function, and then did a `git bisect` to find the bad commit
- The commit introducing this bug is a perf optimization: https://github.com/owid/owid-grapher/pull/2620/commits/d58bef8cbe24185d3a7284f4ca23586def996ee6
- Since it's marked as a perf optimization only, I'm inclined to remove it, which fixes the bug

@danyx23 since you were the one reviewing [Marcel's PR](https://github.com/owid/owid-grapher/pull/2620), do you think it's safe to do so?

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreTable.ts
+++ b/packages/@ourworldindata/core-table/src/CoreTable.ts
@@ -1621,7 +1621,6 @@ class FilterMask {
         for (let i = 0; i < this.numRows; i++) {
             if (this.mask[i]) keepIndexes.push(i)
         }
-        if (keepIndexes.length === this.numRows) return columnStore
 
         Object.keys(columnStore).forEach((slug) => {
             const originalColumn = columnStore[slug]


danyx23 (approved):
Ok let's do it but I find it pretty unsettling that this causes an issue. Basically this should be a no-op but it's not. I also tried to use lodash cloneDeep instead of return columnStore and that also fixes it (i.e. that code is the same as the code path below). 

What this points to is that somewhere it is important that we implicitly make a copy here and this worries me - it points to a mutation of a shared table that shouldn't take place. 

But since neither of us probably want to do a longer investigation on this rn I am ok with merging this one.

sophiamersmann:
Same! Makes me uncomfortable as well, and I'd love for one of us to get to the bottom of this, but since we are pressed for time... I'll make a note of this though, and maybe – once Marcel is back – he can have another look at this.

PR #2957, opened 2023-11-28
🔨 Data pages: ignore indicator level grapher configs on /grapher/ pages for now

danyx23:
This PR changes how grapher configs are treated in data pages. Grapher configs coming from the ETL level grapher config are no longer taken into account on /grapher/ pages, only in pure datapage previews.

The PR also streamlines the code between previews and baking a bit.

Once our grapher admin can deal with defaults in a better way (i.e. show the user where defaults are used and given them an option to disable the default value) we should change this again.

PR #2958, opened 2023-11-28
💄 render descriptionShort as markdown, fix DoD link display

danyx23:
This PR renders descriptionShort as markdown and makes sure that DoD links are not underlined and instead retain the style they have site-wide.

I don't know if there is a better way to make sure that links that have the dod-span class are not underlined - @sophiamersmann if you would prefer another way of doing this in CSS then please let me know! 🙏

### DIFF
--- a/packages/@ourworldindata/grapher/src/modal/SourcesModal.tsx
+++ b/packages/@ourworldindata/grapher/src/modal/SourcesModal.tsx
@@ -16,6 +16,7 @@ import {
 import {
     IndicatorSources,
     IndicatorProcessing,
+    SimpleMarkdownText,
 } from "@ourworldindata/components"
 import React from "react"
 import { action, computed } from "mobx"
@@ -393,7 +394,7 @@ export class Source extends React.Component<{
             <div className="source">
                 {this.renderTitle()}
                 {this.def.descriptionShort && (
-                    <p>{this.def.descriptionShort}</p>
+                    <SimpleMarkdownText text={this.def.descriptionShort} />
                 )}
                 <SourcesKeyDataTable
                     attribution={this.attributions}
--- a/site/DataPageV2Content.tsx
+++ b/site/DataPageV2Content.tsx
@@ -822,7 +822,11 @@ const KeyDataTable = (props: {
                     <div className="key-data__title key-data-description-short__title">
                         {datapageData.title}
                     </div>
-                    <div>{datapageData.descriptionShort}</div>
+                    <div>
+                        <SimpleMarkdownText
+                            text={datapageData.descriptionShort}
+                        />
+                    </div>
                 </div>
             )}
             {source && (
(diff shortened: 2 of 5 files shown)


danyx23:
* master
  * **PR #2957** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2957?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2958** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/2958?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2958?utm_source=stack-comment).

site/DataPageV2Content.tsx
```diff
@@ -822,7 +822,11 @@ const KeyDataTable = (props: {
                     <div className="key-data__title key-data-description-short__title">
                         {datapageData.title}
                     </div>
-                    <div>{datapageData.descriptionShort}</div>
+                    <div>
+                        <SimpleMarkdownText
```
sophiamersmann, line 826: The new renderer by default wraps any text into a paragraph which adds unwanted vertical space. I added an option called `useParagraphs` for single-liners specifically that – if set to false – disables that beaviour. It would be good to use that here. 

```jsx
<SimpleMarkdownText text={...} useParagraphs={false} />
```

sophiamersmann: Or, alternatively, if we expect `descriptionShort` be have multiple paragraphs, then it would be good to set the `p` margins to 0 here.

danyx23: Good to know! The ETL guidelines say that this should be a single paragraph so I'll go with useParagraphs=false. Thank you!

packages/@ourworldindata/grapher/src/modal/SourcesModal.tsx
```diff
@@ -393,7 +394,7 @@ export class Source extends React.Component<{
             <div className="source">
                 {this.renderTitle()}
                 {this.def.descriptionShort && (
-                    <p>{this.def.descriptionShort}</p>
+                    <SimpleMarkdownText text={this.def.descriptionShort} />
```
danyx23, line 397: @sophiamersmann should we add the useParagraphs stanca here as well then?

sophiamersmann: Since the old code wraps descriptionShort in a paragraph as well, I would leave it as is here.

danyx23:
### Merge activity

* **Nov 28, 9:54 AM**: @danyx23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/2958).

PR #2965, opened 2023-11-30
Gdocs TypeORM refactor

ikesau:
Splits up the `Gdoc` class into `OwidGoogleAuth`, `GdocBase`, and `GdocPost`

`OwidGoogleAuth` is just a collection of static methods for getting the authenticated auth object from google's library

`GdocBase` has all the properties and methods that every archie document we register will want to have.

`GdocPost` is an extension of `GdocBase` that has additional properties (implemented via several `_underscorePrefixedProperties`) that are only relevant for Gdoc posts (i.e. articles/topic pages)

I wrote an unused `GdocFaq` as a small proof of concept, but didn't implement it yet as it would require changes to the admin client which I didn't want to block this PR with. This means everything remains a `GdocPost` for now

Things I've tested:
- Publishing posts
- Republishing posts
- Unpublishing posts
- Validating posts
- Deleting posts
- Publishing DoDs
- Referencing DoDs
- Validating DoDs
- Republishing Faqs (@danyx23 I haven't yet tested that these updates get rendered correctly on datapages - could you check this out or tell me how to most quickly test such things?)
- Baking posts
- Baking dods
- Tracking links
- Prominent links
  - To graphers, external sources, and gdocs
- Images
- Validating images
- Tracking images
- Migrations


danyx23:
For testing FAQs the easiest thing is to look at the data page for the dummy indicator (in the admin go to indicators -> search for "Dummy" -> click "preview the datapage for this indicator at the top". The gdoc that has these FAQs is "Daniel's example FAQs".

The FAQ stuff seems to work as well.

I haven't tested this extensively. I've read through the code and it all looks good but I didn't do a thorough pass of matching all moved pieces between GdocBase and GdocPost - do you think it would be good to still do that? Basically: which level of additional safety would you like here? From my side this could be merged since you seem to have tested all the relevant paths.

On a side note: I think it would be good to have a more extensive test suite for GDocs that tests at least some of the pieces end to end or as integration tests.

PR #2984, opened 2023-12-02
Topic tag slug resolver

ikesau:
This PR replaces the `isTopic` boolean on Tags with a `slug` column and adds UI in the admin to set a Tag's slug.

This allows us to resolve datapage's `topicTagsLinks` ( `Tag["name"][]` ) into actual slugs instead of the good-enough heuristic we launched with. And it will allow us to do the same for data insights

I added some util functions that the site renderer uses to supply the correct mapping to datapages. There's a small & regrettable amount of prop drilling to get them in the right place so they can hydrate correctly.

### DIFF
--- a/baker/GrapherBaker.tsx
+++ b/baker/GrapherBaker.tsx
@@ -54,7 +54,6 @@ import {
     getMergedGrapherConfigForVariable,
 } from "../db/model/Variable.js"
 import { getDatapageDataV2, getDatapageGdoc } from "../datapage/Datapage.js"
-import { slugify_topic } from "../site/DataPageV2Content.js"
 import { ExplorerProgram } from "../explorer/ExplorerProgram.js"
 import { Image } from "../db/model/Image.js"
 import { logErrorAndMaybeSendToBugsnag } from "../serverUtils/errorLog.js"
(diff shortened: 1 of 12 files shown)


db/migration/1701450183524-TagTopicSlug.ts
```diff
@@ -0,0 +1,396 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+import fs from "fs/promises"
+import { get, set } from "lodash"
+
+/**
+ * This file includes the code I used to generate the data we're inserting into the new column
+ */
+
+// SELECT name FROM tags WHERE isTopic = TRUE
```
ikesau, line 9: I don't think any of this code needs to be reviewed. It's just here in case we need it again for some reason.

baker/GrapherBaker.tsx
```diff
@@ -270,9 +270,10 @@ export async function renderDataPageV2({
     const firstTopicTag = datapageData.topicTagsLinks?.[0]
 
     if (firstTopicTag) {
+        const slug = await getSlugForTopicTag(firstTopicTag)
```
ikesau, line 273: TODO: handle this with bugsnag

PR #3003, opened 2023-12-08
:tada: Vastly improved search over variables in the variable admin

larsyencken:
Improve search through our indicators in the admin with more expansive search and search verbs, Google style.

- `(no prefix)` searches name or path (regex)
- `name:` search the name only (regex)
- `path:` search the catalog path only (regex)
- Search part of the catalog path (like)
    - `namespace:`
    - `version:`
    - `dataset:`
    - `table:`
    - `short:`
- Search in time with `before:` and `after:`
- Negate a search term with `-` before it
- Search for private or public with `is:private` and `is:public`

## Todo

- [x] Draft search logic
- [x] Move search logic to its own module
- [x] Verify other locations where variables are listed still render nicely

### DIFF
--- a/adminSiteClient/AdminLayout.tsx
+++ b/adminSiteClient/AdminLayout.tsx
@@ -122,7 +122,9 @@ export class AdminLayout extends React.Component<{
                     </ul>
                 </nav>
                 {showSidebar && <AdminSidebar />}
-                {this.props.children}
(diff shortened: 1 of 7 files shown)


larsyencken:
Staging site: http://staging-site-admin-etl-path/admin/variables

Screenshot 2023-12-21 at 16 24 25

PR #3005, opened 2023-12-09
🎉 Add support for small filename in gdocs images

ikesau:
Adds an optional `smallFilename` property to the `{.image}` component in archie.

This small image is assumed to be smaller than the regular image, and will show when the `Image` component's `img` tag has a width that is less than or equal to the width of the small image.

e.g.

```
{.image}
filename: large.png
smallFilename: small.png
{}
```

This will generate two `<source>` elements for the Image's `<picture>` to consider. The first one will only be active when the viewport width is less than 768px, the second source will always be active.

When previewing, it only uses the Digital Ocean CDN to load the original versions of the images (i.e. a srcset with only one URL specified.)

When baking, it makes multiple resizes of each image and includes all of those in the srcset.

https://github.com/owid/owid-grapher/assets/11844404/9c56435e-9f27-463e-bd35-7fb9e25ad8c5

## Testing guidelines:
You can use the following archie to test it:
```
{.image}
filename: ike-test-life-expectancy-large.png
smallFilename: ike-test-life-expectancy-small.png
{}

{.image}
filename: ike-test-life-expectancy-large.png
{}
```
It should work as expected in the preview, with DPR 2.0 and DPR 1.0

And then if you publish it and run `node itsJustJavascript/baker/buildLocalBake.js --steps gdocPosts gdriveImages` and then `yarn startLocalBakeServer`

it should work with all the resized versions of those two images as expected  http://127.0.0.1:3000/your-article-slug-here.html

### DIFF
--- a/packages/@ourworldindata/utils/src/image.ts
+++ b/packages/@ourworldindata/utils/src/image.ts
@@ -68,7 +68,12 @@ export type SourceProps = {
     srcSet: string
 }
 
-// When we have a small and large image, we want to generate two <source> elements, one of which will only be used on mobile
+/**
+ * When we have a small and large image, we want to generate two <source> elements.
+ * The first will only be active at small screen sizes, due to its `media` property.
+ * The second will be active at all screen sizes.
+ * These props work in conjuction with a `sizes` attribute on the <source> element.
+ */
 export function generateSourceProps(
     smallImage: ImageMetadata | undefined,
     regularImage: ImageMetadata
--- a/packages/@ourworldindata/utils/src/image.ts
+++ b/packages/@ourworldindata/utils/src/image.ts
@@ -62,3 +62,29 @@ export function getFilenameWithoutExtension(
 export function getFilenameAsPng(filename: ImageMetadata["filename"]): string {
     return `${getFilenameWithoutExtension(filename)}.png`
 }
+
+export type SourceProps = {
+    media: string | undefined
+    srcSet: string
+}
+
+// When we have a small and large image, we want to generate two <source> elements, one of which will only be used on mobile
+export function generateSourceProps(
+    smallImage: ImageMetadata | undefined,
+    regularImage: ImageMetadata
+): SourceProps[] {
+    const props: SourceProps[] = []
+    if (smallImage) {
+        const smallSizes = getSizes(smallImage.originalWidth)
+        props.push({
+            media: "(max-width: 768px)",
+            srcSet: generateSrcSet(smallSizes, smallImage.filename),
+        })
+    }
+    const regularSizes = getSizes(regularImage.originalWidth)
+    props.push({
+        media: undefined,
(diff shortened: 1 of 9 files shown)


ikesau:
This requires some additional work to function correctly for higher device pixel ratios 🙂

site/gdocs/Image.tsx
```diff
@@ -81,13 +91,34 @@ export default function Image(props: {
...
+            `${IMAGE_HOSTING_CDN_URL}/${IMAGE_HOSTING_BUCKET_SUBFOLDER_PATH}/${f}`
+
+        const PreviewSource = (props: { i?: ImageMetadata; sm?: boolean }) => {
+            const { i, sm } = props
+            if (!i) return null
+
+            return (
+                <source
+                    srcSet={`${makePreviewUrl(i.filename)} ${i.originalWidth}w`}
+                    media={sm ? "(max-width: 798px)" : undefined}
```
sophiamersmann, line 104: Did you mean to write 768px?

site/gdocs/Image.tsx
```diff
@@ -81,13 +91,34 @@ export default function Image(props: {
...
-            <img
-                src={encodeURI(previewSrc)}
-                alt={alt}
-                className={cx(maybeLightboxClassName, className, "lazyload")}
-            />
+            <picture className={className}>
+                <PreviewSource i={smallImage} sm />
+                <PreviewSource i={image} />
+                <img
+                    src={makePreviewUrl(image.filename)}
```
sophiamersmann, line 117: Did you drop `encodeURI` on purpose?

packages/@ourworldindata/utils/src/image.ts
```diff
@@ -62,3 +62,34 @@ export function getFilenameWithoutExtension(
...
+ */
+export function generateSourceProps(
+    smallImage: ImageMetadata | undefined,
+    regularImage: ImageMetadata
+): SourceProps[] {
+    const props: SourceProps[] = []
+    if (smallImage) {
+        const smallSizes = getSizes(smallImage.originalWidth)
+        props.push({
+            media: "(max-width: 768px)",
```
sophiamersmann, line 85: If I understand correctly, this fixes the main concern I had when looking at the code yesterday. To make sure I understand: The media attribute makes sure that the small image will only ever show up on small screen sizes (<= 768px), regardless of its own size? (This is important because I'll probably render mobile Grapher charts at a resolution that is close to the ideal one)

sophiamersmann (approved):
Looks good!

When clicking on the small image, the viewer opens the large version. Do you think that's easy to address? Having the viewer show the small version when appropriate would be neat.

packages/@ourworldindata/utils/src/image.ts
ikesau: Yes that's correct 🙂 

Which is why the mobile export can be 1536px wide for (DPR 2) but it will still only show on mobiles.

Issue #2523, opened 2023-08-08
Show new metadatafields in admin variable and dataset pages


danyx23:
We now have a new metadata schema in the ETL and the database, but this information is not yet shown in the grapher admin. This issue is about bringing this information into the grapher admin pages.

Specifically this is about:
- [ ] On the indicator page, show new metadata fields (descriptionShort, descriptionFromProducer, the presentation fields etc)
- [ ] On the indicator page, show origins in addition to the source (and make it possible to see all origins)
- [ ] On the indicator page, show the updatePeriodDays (even though it comes from the dataset)
- [ ] On the dataset, show the origins of all the indicators that belong to this dataset
- [ ] On the dataset, show the updatePeriodDays

PR #3006, opened 2023-12-11
:sparkles: add data page metadata fields to admin

Marigold:
Implementation of https://github.com/owid/owid-grapher/issues/2523.

## Indicator page

- Add all new fields, including `updatePeriodDays`
- Add a list of origins for an indicator (all fields except license)
- Add indicator source (there's always max 1)

## Dataset page

- Add `origins` to `admin/api/datasets/X.json`
- Add `updatePeriodDays`
- Add a list of origins for all indicators from a dataset
- Add a list of dataset sources and **remove confusing fields** that use the first dataset source

## Notes

- I initially wanted to put field help text from our ETL docs into `secondaryLabel`, but eventually decided to only include link to our docs as the help would get outdated soon
- There's still the old confusing **Source** on Dataset page. Perhaps we could remove it and add sources to indicator tab instead?
- None of the fields are editable (turning editing on probably won't work for JSON fields... or maybe none of the fields)
- JSON fields like grapherConfigETL are shown as plain text
- Old indicators show empty data page metadata and empty origins (to shame us)
- Let me know if I'm misusing HTML rows / cols structure

## Examples:
- [Indicator with full metadata](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/variables/815313)
- [Indicator with some metadata](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/variables/540251)
- [Multiple origins for an indicator](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/variables/734429)
- [Old indicator without new metadata](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/variables/445674)
- [Dataset page](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/datasets/5864)
- [Dataset with multiple sources](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/datasets/888)
- [Indicator with source](http://staging-site-2523-show-new-metadata-fields-in-admin-variable-an/admin/variables/40977)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,7 +14,7 @@
     "typescript.preferences.useAliasesForRenames": false,
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
+        "source.fixAll.eslint": "explicit"
     },
     "typescript.preferences.importModuleSpecifierEnding": "js",
     "javascript.preferences.importModuleSpecifierEnding": "js",
--- a/adminSiteClient/DatasetEditPage.tsx
+++ b/adminSiteClient/DatasetEditPage.tsx
@@ -13,6 +13,7 @@ import { BindString, Toggle, FieldsRow, Timeago } from "./Forms.js"
 import { EditableTags } from "./EditableTags.js"
 import { ChartList, ChartListItem } from "./ChartList.js"
 import { OriginList } from "./OriginList.js"
+import { SourceList } from "./SourceList.js"
 import { VariableList, VariableListItem } from "./VariableList.js"
 import { AdminAppContext, AdminAppContextType } from "./AdminAppContext.js"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome/index.js"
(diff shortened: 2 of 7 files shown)


Marigold:
@danyx23  added sources to indicator and dataset admin pages.

danyx23 (commented):
Looks good! I think we should probably add a limit clause to the sources query for datasets. The reason is that for [some extreme datasets](http://datasette-private/owid?sql=with+datasetsWithSources+as+%0D%0A+%28select+datasetId%2C+count%28*%29+as+num+from+sources+group+by+datasetId+order+by+num+desc+limit+10%29%0D%0Aselect+dws.*%2C+d.name%0D%0Afrom+datasetsWithSources+dws+%0D%0Ajoin+datasets+d+on+dws.datasetId+%3D+d.id) there are hundreds or thousands of sources. Maybe limit it to 50 and add a note above or below the list?

I also removed the singular source on datasets as we no longer need it there

danyx23 (changes requested):
I think we should limit the number of sources we retrieve, otherwise this is good to ship!

PR #3016, opened 2023-12-12
🔨 change fallback chain for display.name and titlePublic

danyx23:
This PR changes the fallback chain for title-ish strings according to #3017

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreTableColumns.ts
+++ b/packages/@ourworldindata/core-table/src/CoreTableColumns.ts
@@ -23,6 +23,7 @@ import {
     PrimitiveType,
     imemo,
     ToleranceStrategy,
+    IndicatorTitleWithFragments,
 } from "@ourworldindata/utils"
 import { CoreTable } from "./CoreTable.js"
 import { Time, JsTypes, CoreValueType } from "./CoreTableConstants.js"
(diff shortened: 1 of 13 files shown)


packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -266,7 +267,22 @@ export abstract class AbstractCoreColumn<JS_TYPE extends PrimitiveType> {
...
+    }
+
+    @imemo get titlePublicOrDisplayName(): IndicatorTitleWithFragments {
+        return this.def.presentation?.titlePublic
+            ? {
+                  title: this.def.presentation?.titlePublic,
+                  attributionShort: this.def.presentation?.attributionShort,
+                  titleVariant: this.def.presentation?.titleVariant,
+              }
+            : { title: this.display?.name ?? this.name ?? "" }
```
sophiamersmann, line 285: `||` instead `??` would be better here because sometimes `display.name` is an empty string, and then the title would be empty

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -266,7 +267,22 @@ export abstract class AbstractCoreColumn<JS_TYPE extends PrimitiveType> {
...
     }
 
     @imemo get nonEmptyDisplayName(): string {
```
sophiamersmann, line 288: After your changes, this is the only place where `nonEmptyDisplayName` is used now: https://github.com/owid/owid-grapher/blob/cc7ff708a2121cab3ef7d36afc621edf67445942/packages/%40ourworldindata/grapher/src/core/Grapher.tsx#L1467

It makes sense to replace this with `titlePublicOrDisplayName` and then delete `nonEmptyDisplayName`

sophiamersmann (approved):
Looks good to me!

Just a few minor comments about how to show the extra fields in Grapher's data tabels.

For table captions, the extra fields and the unit are both shown in grey and not separated from each other, which is a bit weird:

Screenshot 2023-12-15 at 11 00 47

The extra fields make column names a bit crowded, not sure we want to show it here? If there's been a lot of discussion this, please ignore me.

Screenshot 2023-12-15 at 11 23 53

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
danyx23: Ah nice catch!

danyx23:
Thanks a lot, @sophiamersmann, very good points. I left the table as is for now since we kinda need the additional fragments with the title public but maybe we can tweak it some time soon with Marwa.

PR #3023, opened 2023-12-14
Data Insight Page

ikesau:
The view for a specific data insight (no baking yet)

image

testing data:
https://docs.google.com/document/d/1o8PtWZOUsdr_j31y6-kKI-mbSRVWqzWe5m1fsN6EPl0
https://docs.google.com/document/d/117Z5d_gF1ONg8UaOeAFjGxahv1oRxuVPM69S7lZWZKc
https://docs.google.com/document/d/1g8GiNLfo2p2nTuDzrJy96fFhfsi1FSQwS8PhweCfwvc
https://docs.google.com/document/d/1K3nc9c0WlrEjsJkJ770mVVKmbCOiZZmByDXcatl76VU
https://docs.google.com/document/d/1DdgUxo8mFC90EZ14wwOGABP90t4GTEQ_JrZXuW0Cpqc

### DIFF
--- a/db/model/Gdoc/GdocBase.ts
+++ b/db/model/Gdoc/GdocBase.ts
@@ -31,6 +31,7 @@ import {
     Tag as TagInterface,
     OwidGdocPublicationContext,
     BreadcrumbItem,
+    MinimalDataInsightInterface,
 } from "@ourworldindata/utils"
 import { BAKED_GRAPHER_URL } from "../../../settings/serverSettings.js"
 import { google } from "googleapis"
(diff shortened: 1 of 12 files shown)


db/model/Gdoc/GdocBase.ts
```diff
@@ -565,29 +565,6 @@ export class GdocBase extends BaseEntity implements OwidGdocBaseInterface {
         }
     }
 
-    async loadRelatedCharts(): Promise<void> {
```
ikesau, line 568: Only GdocPosts have related charts, so I moved this to be subclass-only, which helps when it comes time to render the feed page.

sophiamersmann (approved):
Looks good to me 👌🏻 

Seeing it in code now, I think you're right in assuming that it should be fairly obvious that `owid.org/data-insights#most-recent-data-insight` is not a stable link and sharing it doesn't make sense

PR #3024, opened 2023-12-14
🎉 Data insights index page

ikesau:
data-insights-feed

A feed page with some sense of pagination in the mock site router

Will implement the actual pagination tomorrow

### DIFF
--- a/packages/@ourworldindata/utils/src/owidTypes.ts
+++ b/packages/@ourworldindata/utils/src/owidTypes.ts
@@ -1383,6 +1383,8 @@ export interface OwidGdocDataInsightContent {
     type: OwidGdocType.DataInsight
 }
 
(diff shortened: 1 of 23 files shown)


packages/@ourworldindata/utils/src/owidTypes.ts
```diff
@@ -1383,6 +1383,8 @@ export interface OwidGdocDataInsightContent {
     type: OwidGdocType.DataInsight
 }
 
+export const DATA_INSIGHTS_INDEX_PAGE_SIZE = 2
```
ikesau, line 1386: This should be 20 but I've made it artificially low while we're testing pagination

sophiamersmann (approved):
Looks good to me!

Some changes are a little over my head, but I tested it, and everything works well!

Btw, it's nice how easy to read the code is although I'm not very familiar with this part of the codebase. I really like how you break things apart :)

PR #3031, opened 2023-12-18
:tada: lightning chart deploys

Marigold:
## Problem

Updating a chart enqueues a full deploy that can take >15min and is prone to errors (due to high load on external servers like R2 or Spaces). Deploying only a single chart could significantly speed up most deploys.

## Solution

* Add the `bakeCharts` command that is very similar to `bakeGDocs`, which takes care of GDocs lightning deploys
* Trigger a deployment after clicking on `Update chart` or `Publish` buttons
* Lightning build with `chartSlug` goes to deploy queue which triggers BuildkiteJob [OWID - deploy content](https://buildkite.com/our-world-in-data/owid-deploy-content-master) with env variable `LIGHTNING_CHART_SLUGS`
* Buildkite runs `bakeCharts --slugs LIGHTNING_CHART_SLUGS` on the server, syncs `grapher/exports` with generated png & svg to R2 and then deploys the site to Cloudflare

## Potential issues

What other things need to happen when we publish or update a chart? Perhaps updated redirects might require full rebuild or at least rebuilding of redirects file in Buildkite.

Things that could be affected by updated chart (need to check)
- explorer based indicators
- prominent links

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -114,7 +114,7 @@ const triggerStaticBuild = async (user: CurrentUser, commitMessage: string) => {
     })
 }
 
-const enqueueLightningChange = async (
+const enqueueLightningGDocChange = async (
     user: CurrentUser,
     commitMessage: string,
     slug: string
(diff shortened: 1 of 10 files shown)


Marigold:
Longer discussion [on slack](https://owid.slack.com/archives/CQQUA2C2U/p1702998865746039). It's not clear whether this is worth merging. We should ask authors if they're even interested in this.

ikesau:
From Gdocs, prominent links to gdocs are already broken which means they're probably not used anywhere.

https://github.com/owid/owid-grapher/issues/2870

There are still WP articles that use (working) prominent links for graphers though 😢

danyx23:
@Marigold we'll try to do a product hour about gdocs at some point soon and then ask authors how much this is bothering them. Suggest we keep this open until then.

Marigold:
It turned out authors don't really need this. It could still be useful for approving charts that trigger chart updates and it could eliminate freezing of admin, but it's probably not worth at this point.

PR #3032, opened 2023-12-19
StackedArea: Don't show interpolated values in tooltips

### LINKED ISSUE #2866: StackedArea: Tooltips should not show interpolated values
## Description

Stacked area charts shows interpolated values that are not in the data.

## Expected behaviour

Stacked area charts, like line charts, linearly interpolate between consecutive values. From a data viz perspective, it's unclear whether this is wanted or not (and requires discussion). But they should never show interpolated values as if they were real data values.

## Steps to reproduce

1. Go to this URL https://ourworldindata.org/grapher/energy-consumption-by-source-and-country?country=~BLR
2. Hover over any point between 2000 and 2019.
3. See that nuclear shows some data values, but they shouldn't, since there's no data for Belarus' nuclear consumption between 2000 and 2019.

**NOTE**:  I changed the data so that Belarus's nuclear consumption between 2000 and 2019 is zero instead of nan. This "fixes" the problem. With nans, what you'd see is the linearly interpolated values between 2000 and 2019 as if they were real values. So currently I don't have any example to reproduce the behaviour (and if I see one, I manually fix the problem in the data, to avoid spurious interpolated values).

## Additional context

See [discussion on Slack](https://owid.slack.com/archives/C46U9LXRR/p1698401249045499).

sophiamersmann:
- fixes #2866

### Problem

- StackedArea charts show interpolated (i.e. made-up) values in the tooltip
- I don't have a live example, but Pablo created a [buggy chart](http://staging-site-fix-tooltip-interpolated-values/grapher/energy-consumption-by-source-and-country?country=~BLR) for me on staging
    - You won't be able to see the bug since it's fixed on this branch now, but still:
    - Select Belarus and hover over any data point between 2000 and 2019
    - Now it shows "No data" for Nuclear (which is correct), but we used to show interpolated values here (I should have screenshotted it before pushing a fix 🤦‍♀️)

### Solution

- The fix is not trivial because linear interpolation overwrites the original column. We thus lose information on which values are real and which have been interpolated
- To keep track of this, the interpolation step appends a new column `<slug>-originalValue` to the table that stores the original values (`<slug>` stores the interpolated ones)
    - Alternatively, I could have created a new boolean column like `<slug>-interpolated`, but I went with the `<slug>-originalValue` approach because its application is less narrow, and it also aligns with the `<slug>-originalTime` pattern for tolerance
- A column's `owidRows` are then extended to optionally also hold the original value
- Given the (possible interpolated) value and the original value, the `AbstractStackedChart` then marks interpolated values as `fake`, which we don't show in the tooltip

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreTableColumns.ts
+++ b/packages/@ourworldindata/core-table/src/CoreTableColumns.ts
@@ -30,7 +30,10 @@ import { Time, JsTypes, CoreValueType } from "./CoreTableConstants.js"
 import { ColumnTypeNames, CoreColumnDef } from "./CoreColumnDef.js"
 import { EntityName, OwidVariableRow } from "./OwidTableConstants.js" // todo: remove. Should not be on CoreTable
 import { ErrorValue, ErrorValueTypes, isNotErrorValue } from "./ErrorValues.js"
-import { getOriginalTimeColumnSlug } from "./OwidTableUtil.js"
+import {
+    getOriginalTimeColumnSlug,
+    getOriginalValueColumnSlug,
+} from "./OwidTableUtil.js"
 
 interface ColumnSummary {
     numErrorValues: number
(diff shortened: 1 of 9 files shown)


danyx23 (approved):
Nice, this looks good!

PR #3050, opened 2023-12-29
Fix broken links in sources modal

sophiamersmann:
- Links were sometimes interpreted as Markdown
- Example: https://ourworldindata.org/grapher/distribution-of-population-between-different-poverty-thresholds-historical

Screenshot 2023-12-29 at 11 54 12


danyx23 (approved):
Nice - thanks for the quick fix!

Issue #3052, opened 2023-12-29
Indicator's `description` is used as a source's description instead of an indicator's description


pabloarosado:
## Description

An indicator's old `description` field does not appear next to the indicator's title, and instead it appears as a source's description.

## Expected behaviour

If there is no `descriptionShort`, but there is a `description`, it should appear next to the title (this is important for indicators using the old metadata). However, currently, `description` does not appear next to the title.

Instead, `description` is mistakenly shown as the source's description, when the indicator has sources. And, when the indicator does not have sources (because it has origins), the indicator's `description` does not appear anywhere.

## Steps to reproduce

1. Check this indicator: https://owid.cloud/admin/variables/734511
2. See that it has a `description`, but, in the chart preview (or in [this chart](https://owid.cloud/admin/charts/890/edit)) the description does not appear next to the title. Instead, it appears as a source description.

## Screenshots
![Screenshot 2023-12-29 at 13 51 42](https://github.com/owid/owid-grapher/assets/12246978/f38b25af-cf84-40f1-86c9-fce60068025a)
![Screenshot 2023-12-29 at 13 52 01](https://github.com/owid/owid-grapher/assets/12246978/a875ce53-c7ca-4e09-b07b-c78fef6ed642)

## Additional context

See discussion [on Slack](https://owid.slack.com/archives/C46U9LXRR/p1703853341938489).

PR #3053, opened 2023-12-29
Fix misplaced indicator's `description` field

pabloarosado:
Fixes https://github.com/owid/owid-grapher/issues/3052

NOTE: I am not very familiar with this code and have not looked deeply into it, so there may be cases I haven't considered.

This PR:
* Shows `description` next to the indicator's title, when `descriptionShort` does not exist but `description` does exist.
* Removes indicator's `description` from the sources.


sophiamersmann (approved):
Looks good to me! Thank you!

PR #3057, opened 2024-01-02
Add globe option to map view

JohnGinger:


packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -1,77 +1,77 @@
-import React from "react"
+import { CoreColumn, EntityName, OwidTable } from "@ourworldindata/core-table"
```
JohnGinger, line 1: My editor autosorted the imports into alphabetical order. I can revert that if it is confusing.

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -933,15 +1067,16 @@ class ChoroplethMap extends React.Component<{ manager: ChoroplethMapManager }> {
...
+                                        d={this.getPath(feature)}
                                         strokeWidth={
                                             (isFocus ? focusStrokeWidth : 0.3) /
                                             viewportScale
                                         }
                                         stroke={stroke}
                                         strokeOpacity={strokeOpacity}
                                         cursor="pointer"
-                                        fill={`url(#${Patterns.noDataPatternForMapChart}-${this.manager.projection})`}
+                                        // Hashing looks weird as it doesn't rotate with the rest of the globe
```
JohnGinger, line 1078: I think this looks clearer than the hashing lines, but could see the arguments for both ways.

sophiamersmann:
Hey, @JohnGinger, this looks great! Thank you so much for your contribution!

We will have an internal discussion about whether adding a Globe view fits into the roadmap for Grapher. If we do want to go ahead with this change, would you be open to working some more on this PR to tweak it according to our needs?

JohnGinger:
Thanks @sophiamersmann - yes happy to tweak it if you tell me what you need tweaking.

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
sophiamersmann: I don't mind it :)

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
sophiamersmann: I think the hashing lines are the right approach here. Any solid colour (including the gray that you have chosen) could be used to indicate a data value, in which case countries with that data value and countries without data would look the same.

JohnGinger:
Hi @sophiamersmann 

Sorry for the delay. I think I've resolved all those issues. I've removed the spinning as I was concerned about getting the performance correct when using mobX (I'm sure there is an easy idiomatic way - but I couldn't work out what it was)

Uploading Screen Recording 2024-01-15 at 16.46.04.mov…

As a heads-up. I wasn't able to work out a way to test this on mobile.

sophiamersmann:
Looks great! I'll have a closer look at the end of the week. Thanks, @JohnGinger

packages/@ourworldindata/grapher/src/mapCharts/MapChart.sample.ts
```diff
@@ -25,18 +25,86 @@ export const legacyMapGrapher: GrapherProgrammaticInterface = {
             3512,
             {
                 data: {
-                    years: [2000, 2010, 2010],
-                    entities: [207, 15, 207],
-                    values: [4, 20, 34],
+                    years: [...Array(300).fill(1).map((x,i) => 2000)],
```
sophiamersmann, line 28: Changing the sample data here makes some of our tests fail. Can you revert these changes?

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -866,17 +963,51 @@ class ChoroplethMap extends React.Component<{
...
+                    this.enableDragging = true
+                    ev.preventDefault() /* Without this, title may get selected while shift clicking */
+                }}
+                onMouseMove={(ev: React.MouseEvent<SVGGElement>) => {
+                    if (this.enableDragging) {
+                        const rotate = this.globeProjection.rotate()
+                        const sensitivity = 1.25
+                        const rotateX = rotate[0] + ev.movementX * sensitivity
+                        let rotateY = rotate[1] - ev.movementY * sensitivity
+                        rotateY = Math.max(-90, Math.min(90, rotateY))
```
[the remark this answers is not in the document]
JohnGinger: The maths of rotating gets confusing if you go all the way round the globe (and everything goes upside down). I thought it was easier and also a better user experience to keep things the 'right' way up.

JohnGinger:
Thanks for your detailed comments @sophiamersmann - will have a look at them early next week.

sophiamersmann:
Great! Thank you :)

I almost forgot... our designer came back with a sketch. The Globe view switcher should sit next to the Continent switcher in the control bar. It should be fairly easy for you to add. We have a [LabeledSwitch](https://github.com/owid/owid-grapher/blob/master/packages/%40ourworldindata/grapher/src/controls/LabeledSwitch.tsx) component you can use, and the Continent switcher is rendered [here](https://github.com/owid/owid-grapher/blob/6c1fd424abd00427e382ad508ce55c02d359699f/packages/%40ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx#L294).

Screenshot 2024-01-18 at 14 38 21

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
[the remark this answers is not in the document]
sophiamersmann: Ahh, good point!

JohnGinger:
https://github.com/owid/owid-grapher/assets/2624571/f030ac61-c6ad-44e2-b0a7-7c6cbe084c0e

Made those markups - thanks for the tips - performance is much better now!

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -269,14 +273,14 @@ export class MapChart
         return this.props.manager
     }
 
-    @computed private get entityNamesWithData(): Set<EntityName> {
+    @computed private get entityNamesWithData(): Set<any> {
```
JohnGinger, line 276: EntityName seems to have disappeared from the repo?

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -269,14 +273,14 @@ export class MapChart
...
         // We intentionally use `inputTable` here instead of `transformedTable`, because of countries where there is no data
         // available in the map view for the current year, but data might still be available for other chart types
         return this.inputTable.entitiesWith([this.mapColumnSlug])
     }
 
     // Determine if we can go to line chart by clicking on a given map entity
-    private isEntityClickable(entityName?: EntityName): boolean {
+    private isEntityClickable(entityName?: any): boolean {
```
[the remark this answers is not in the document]
JohnGinger: I agree code rabbit - but EntityName seems to have been removed from the repo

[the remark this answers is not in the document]
JohnGinger: Ok my mistake, I have no idea what was going on with it not building in my codebase before

sophiamersmann:
Great work! That's much better.

We're almost there... The last missing feature to be implemented is automatic spinning of the globe. Do you want to have a go at that? I remember you saying that you ran into performance issues the first time you tried. If that's the case, try repeatedly calling `requestAnimationFrame` instead of using `setInterval`, like so:

```js
const loop = () => {
    // do something

    // then request another frame
    requestAnimationFrame(loop)
}
requestAnimationFrame(loop)
```

Small note: I think it would be better to dismiss tooltips while dragging the globe. They're a little distracting.

JohnGinger:
Hi,

Happy to. Do you have a preference for where you want the spinning button to be?

Best,
John

sophiamersmann:
The globe should automatically spin – there shouldn't be a button.

toni-sharpe:
**non-blocking**

Cool thing. 

I have observations:

* what if there are several maps on the page? Should rotating be on by default? How is performance in this case? 
-- is the globe loaded in the background or on user choice?
-- maybe see what some throttling / an old phone does?

* I hate to be negative, but this view shows me a lot less data than the flat map. When I've used OWID it's been to use trusted data to back-up some argument that really matters.

* You are right that the flat map is distorted and could be misleading, I don't know how much that matters when looking at vaccine rates for example, food for thought; but, there are [flat projections that solve this problem](https://en.wikipedia.org/wiki/List_of_map_projections) - loads of efforts!

* What happens when this thing is triggered at the same time?

image

* Maybe consider that it's a bit white Western focused (there are six choices, eight billion people and two of those six cover Europe and N America, 1/10th or so) - maybe it could be bulked out to 20 choices or something that would balance things, capital cities of the [top 20 countries for population maybe](https://en.wikipedia.org/wiki/List_of_countries_and_dependencies_by_population)? With a minimum number of points per continent, just because Oceania.

* And a random option (!sea) just for fun.

sophiamersmann:
Hey @JohnGinger,

I hope you're good!

I just wanted to check in with you about this PR. If you intend to do some more work on it, I can make the stale bot go away, and we can keep the PR open for longer. If you have moved on, that's okay, too. In that case, I would want to merge this PR into a feature branch on our repo so that one of us can pick up the work.

Thank you,
Sophia

JohnGinger:
Hi, sorry for the delay - life has got a bit busy. 

I'm still happy to work on it - Is it just the rotation left? I have some time I can spend on this weekend.

sophiamersmann:
Great! Yes, the rotation is the last missing piece. There are some other details to be figured out (e.g. what the starting point of the rotation should be). Questions like these need some discussion, which always results in some back and forth, so it might be best for one of us to take over at that point. After you have implemented the rotation, I'd give it a review, and then we'd merge it into a feature branch so that one of us can figure out the last details. Does that sound good to you?

JohnGinger:
Sounds good to me

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -725,6 +785,45 @@ class ChoroplethMap extends React.Component<{
         )
     }
 
+    @computed get globeWidth(): number {
```
toni-sharpe, line 788: Where is this used? I did a browser search for these in the PR screen and they didn't come up. Is something hooking in and getting them?

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -725,6 +785,45 @@ class ChoroplethMap extends React.Component<{
...
+        return (this.bounds.width - 300) / 2
+    }
+
+    @computed get globeHeight(): number {
+        return (this.bounds.height - 150) / 2
+    }
+
+    @computed get globeScale(): number {
+        if (this.bounds.height < this.bounds.width) {
```
toni-sharpe, line 797: Noting on line [773](https://github.com/owid/owid-grapher/pull/3057/files#diff-1e9d0a7908d5769c4657f34c3286d9f969805a0b9e432f09eb889103eb5b9756R773) that `this` is destructured and that pattern is in the file elsewhere, could be followed here

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -725,6 +785,45 @@ class ChoroplethMap extends React.Component<{
...
+            return this.bounds.height / 2.3
+        } else {
+            return this.bounds.width / 2.3
+        }
+    }
+
+    @computed get globeProjection(): any {
+        const worldView = this.manager.projection === MapProjectionName.World
+
+        const scale = worldView ? this.globeScale : this.globeScale * 1.8
```
toni-sharpe, line 807: A few [magic numbers](https://stackoverflow.com/a/47902) here? Better as labelled consts? 

The `1.8` and `2.3` getting used above - I think this is probably more important when the contributor isn't a core team member, though I'd flag it in anyone's code

packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
```diff
@@ -361,6 +372,15 @@ export class CaptionedChart extends React.Component<CaptionedChartProps> {
             height: chartHeight,
         }
 
+        const globeSwitcher: React.CSSProperties = {
```
toni-sharpe, line 375: Again I couldn't see this used. Not familiar with the repo really, so of course it's likely me, [but I did search](https://github.com/search?q=repo%3Aowid%2Fowid-grapher%20globeSwitcher&type=code).

toni-sharpe (commented):
A few comments, I noticed some things that aren't referenced in the PR, but perhaps they are hooked into by other stuff? Or relate to the PR comments about refactoring and providing functions for the globe, no worries there.

Anyway, I had to break the review duck, even if I misunderstood a lot.

MobX seems easy to understand though, going to run through he docs now

These are comments and are meant to be **non-blocking** 

One thing, you use CloudFlare? Me too, and I get a branch deployed for me with the code on it. Do you get that? I'm just on their dev pages.

larsyencken:
Hi @toni-sharpe, thanks for your efforts to give feedback here. We'd prefer however if only Our World In Data staff give feedback on pull requests and code that contributes to our work. 

That's because from the outside, it's easy to lack context on our codebase, our standards and where we're trying to take both of those things. Historically we've found those things hard to communicate quickly. Since staff are the ones who will maintain any contributions made, it's important to leave it to them to make these calls.

toni-sharpe:
@larsyencken ok, I can leave that be. 

Submitting PRs however is OK?

Or involving myself in discussions, but maybe avoiding more code related issues?

larsyencken:
> @larsyencken ok, I can leave that be.
> 
> Submitting PRs however is OK?
> 
> Or involving myself in discussions, but maybe avoiding more code related issues?

Thanks Toni! Opening issues and submitting PRs is ok, with the understanding that not every issue can be addressed, or contributed change merged.

toni-sharpe:
@larsyencken

It was my intention to really jump into this one, I strongly approve of the mission, I'm doing similar things, I'm out of work and have time, etc. Therefore I will need to ask specifically here. 

Opening PRs and issues is fine
Reviewing PRs should be left to the employed team

But I have [also done things like this](https://github.com/owid/owid-grapher/issues/3240#issuecomment-1962735314) - commented on an issue with a solution. It's not really in fitting with your pattern though. Is this OK? Or should commenting on issues be left to the team also?

Sorry to be so specific here, but I would still like to contribute, and I really wanted to contribute properly to this project.

JohnGinger:
@sophiamersmann - sorry for taking ages to get to this. I may still get to it, but I wouldn't be offended if you took it internally to finish it off.

sophiamersmann:
Hey @JohnGinger, no worries, thank you for letting me know!

The best way forward is probably to merge this PR into a feature branch on `owid-grapher`, so that we can take ownership of it. I created a branch called [`globe-view`](https://github.com/owid/owid-grapher/tree/globe-view). Could you change the base branch of this PR to `globe-view`, or maybe open another PR against `glove-view` (whichever is easiest). Thank you!

JohnGinger:
Thanks so much @sophiamersmann - sorry I've dropped this for ages

PR #3061, opened 2024-01-04
:honeybee: (admin) remove WorldMap from chart type list

sophiamersmann:
- `WorldMap` is removed from the admin's chart type list but continues to be used as chart type internally
- None of our charts are of type `WorldMap` currently, so no migration necessary
- Technically, this is a breaking change for our schema, but I don't think there is a need to increase the schema version etc since there is no code change


danyx23 (approved):

danyx23 (approved):
Nice, thanks for the cleanup!

PR #3067, opened 2024-01-04
🐛 Fix missing featured images

ikesau:
`getPrefetchedGdocAttachments` is a function that queries the DB a couple of times to get all possible attachment data and caches it, so that we can pick the attachments from memory instead of querying the DB for each gdoc bake.

There was a bug in my implementation where linked documents' featured images weren't getting picked, because `linkedDocuments` is a dependency of `linkedImageFilenames`, but both are undefined when we call the function.

This PR fixes the bug by appending the featured images of any linked documents to the list of linkedImageFilenames when we're doing the picking.

**Before:**
image

**After:**
image

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -343,14 +343,21 @@ export class SiteBaker {
                 linkedGrapherSlugs,
                 linkedExplorerSlugs,
             ] = picks
+            const linkedDocuments = pick(
+                this._prefetchedAttachmentsCache.linkedDocuments,
+                linkedDocumentIds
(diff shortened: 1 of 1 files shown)


danyx23 (approved):
Let's merge this because it fixes something that is currently visibly broken but I'd like to use this opportunity to clean this up. I find the code that deals with the contexts hard to understand and as is demonstrated by this bug it seems to have internal dependencies that are not expressed in the types. 

Can you prepare an overview of the attachments flow for the next site meeting and we discuss how this could be improved?

PR #3090, opened 2024-01-11
:hammer: sync master-1 with master on every push to master

Marigold:
Every foundation server needs its master server as a template to create new staging servers. Copying `staging-site-master` across hosts is too slow (>45min and it stops `staging-site-master` container), so we rather create `staging-site-master-1` that will be kept in sync with `master` and automatically updated by Buildkite.

### DIFF
--- a/.github/workflows/sync-master.yml
+++ b/.github/workflows/sync-master.yml
@@ -1,0 +1,28 @@
+# Each foundation server that hosts staging servers needs a master server that is copied
+# to staging servers. foundation-2 uses `master` and foundation-1 uses `master-1`. This
+# action syncs the master branch to master-1 branch which triggers its update.
+# This is much faster than copying `staging-site-master` across our servers with LXC.
+name: Sync Master-1 with Master
+
+on:
+    push:
+        branches:
+            - master
+
+jobs:
+    sync-branch:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v3
+              with:
+                  ref: master-1
+                  fetch-depth: 0 # fetches all history for all branches and tags
+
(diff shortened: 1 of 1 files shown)


Marigold:
@larsyencken ignore this for now, please. After [weighing pros and cons again](https://github.com/owid/ops/pull/126/files#diff-3c603137ad675da17b24ae7c5827c1f6b61cf347f151eafc5665d248c10439d1R33), I've decided to start with copying `staging-site-master` in the morning. If it works, I'll close this PR.

larsyencken:
Ok! I converted this to draft then.

If you did end up wanting to do this approach, I'd suggest doing something more symmetrical for all the LXC hosts. E.g. ensuring there was a base image on `foundation-{1,2,3}`, and that we had a good pattern in case we got more servers later.

Marigold:
Copying server early in the morning didn't work that well. Even moving `staging-site-master-1-tmp` to  `staging-site-master-1` on the same server was timing out (for unknown reasons).

Anyway, with the recent changes to managing envs & secrets with `ansible-vault`, the second option - keeping branch `master-1` around synced with `master` looks much more appealing. This is the last PR that needs to be merged to make it work. (We could also have a Buildkite job running every hour or so and doing the same thing as this github action, but this seemed easier)

> If you did end up wanting to do this approach, I'd suggest doing something more symmetrical for all the LXC hosts. E.g. ensuring there was a base image on foundation-{1,2,3}, and that we had a good pattern in case we got more servers later.

Agree, I called it `staging-site-1` for this reason. I'm just not sure if it's worth keeping the exact copy `staging-site-master-2` on `foundation-2`. Maybe once we have more servers, but it'd be a bit wasteful now.

PR #3096, opened 2024-01-12
🎉 add `allGraphersLoaded` Event

ikesau:
Takes https://github.com/owid/owid-grapher/pull/3087 further by adding an event that will dispatch on any page to confirm that **all** graphers have loaded.

This is necessary for this [site-screenshots](https://github.com/owid/site-screenshots/pull/4) PR so that we can reliably take a screenshot once the page is ready instead of sleeping an arbitrary amount of time that will either be too long or too short for most pages.

I can imagine we'll forget this code exists and accidentally break it from time to time, but once we have a proper site screenshots test on every page type, we'll be able to make our CI tests fail until it's resolved (or confidently vetoed, like we do with the SVG tester)


danyx23 (approved):
Huh, I thought I had marked this already as approved - sorry for the delay, LGTM!

PR #3102, opened 2024-01-15
🐛 Fix mistaken HTML character encoding grapher urls

ikesau:
Given the following (subselection o) archieML JSON in the posts table:
```
"body": [
      {
        "type": "text",
        "value": [
          {
            "text": "I am some text with an ampersand &",
            "spanType": "span-simple-text"
          }
        ],
        "parseErrors": []
      },
      {
        "url": "https://ourworldindata.org/grapher/annual-deforestation?stackMode=absolute&region=World",
        "type": "chart",
        "parseErrors": []
      }
    ]
```

Here's how it was getting inserted into Gdocs:

Before:
image

After:
image

I opted to make the fix inline just before the gdoc creation, because an arbitrary HTML code escape done earlier in the pipeline would then cascade into a bunch of other places that we'd need to make sure was handled correctly.

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -2378,7 +2378,11 @@ apiRouter.post("/posts/:postId/createGdoc", async (req: Request) => {
     const tags =
         tagsByPostId.get(postId)?.map(({ id }) => TagEntity.create({ id })) ||
         []
-    const archieMl = JSON.parse(post.archieml) as OwidGdocPostInterface
+    const archieMl = JSON.parse(
+        // Google Docs interprets &region in grapher URLS as ®ion
+        // So we escape them here
+        post.archieml.replaceAll("&", "&amp;")
+    ) as OwidGdocPostInterface
     const gdocId = await createGdocAndInsertOwidGdocPostContent(
         archieMl.content,
         post.gdocSuccessorId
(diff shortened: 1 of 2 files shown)


mlbrgl (approved):
Tested and works as expected on a test document with a few possible edge cases, including HTML entities.

> because an arbitrary HTML code escape done earlier in the pipeline would then cascade into a bunch of other places that we'd need to make sure was handled correctly.

I see what you mean. Replacing "&" across the content feels unnatural though. Maybe double-escaping issues don't apply here since there shouldn't be any HTML coming from WP after `migrateWpPostsToArchieMl`, and so no HTML entities left in the content, but I wouldn't be entirely sure at this point (e.g. could we not technically create Archie HTML blocks from a migration?).

Maybe it is ok to be more surgical and only replace `&reg` to `&amp;reg` and expand the search/replace if we find other problematic cases?

Also, it might be better to put that somewhere at the bottom of `migrateWpPostsToArchieMl`, which feels like a more natural place to perform this kind of transformation in bulk rather than just-in-time in the createGdoc route?

_As a side note, I encountered some weird behaviours while trying to run these tests along a full editing loop, from WP UI to syncWpPostsToGrapher and migrateWpPostsToArchieMl, encoutering DataSource errors, and Archie gdocs only partially imported into the DB. I also noticed authors were rendered as "[object Object]" in the admin UI, at which point I decided my set-up was probably messed up._

PR #3106, opened 2024-01-17
:tada: Metadata editing from admin

Marigold:
Make some indicator fields editable. Call [ETL API](https://github.com/owid/etl/pull/2190) when those fields change. Please check out that PR, there are instructions how to test it.

Notes:
* We only send changed fields in the body and trigger ETL run through API.
* One of the fields, "Description key", is actually an array of strings. I created a new component, `BindStringArray`, that renders this array as a bulleted list and edits it properly on change. I didn't implement `onBlur` method that `BindString` has (I wasn't sure how it is used).
* The user must hit the "Update button" to apply changes. If the request is successful, nothing is shown. If ETL API returns validation errors (4xx), we show them as a modal window with a request body & response. (ETL API sends all requests as notifications to Slack, some as warnings)
* ETL API also updates JSON in Data API on S3, so any changes should persist after refreshing the page.

## Todo

- [x] 🐛 Fix defaults for running it locally
- [x] ✨ ~Remove~ Fix bullet point from `descriptionKey`
- [ ] 🐛 You cannot delete a field by blanking it out
- [x] ✨ Make `processingLevel` a drop-down

## Summary by CodeRabbit

- **New Features**
  - Introduced a `BindStringArray` component for managing arrays of strings in a textarea field with various user interface enhancements.
  - Added new utility functions for creating and parsing bullet lists.
  - Implemented new links and buttons for updating indicators on the Variable Edit Page.
  - Integrated new asynchronous methods for ETL API health checks and data saving.

- **Enhancements**
  - Updated form submission process to include improved object difference calculation and API interaction.

- **Configuration**
  - Updated default URLs for system settings and added a new `ETL_API_URL` for production ETL API integration.

### DIFF
--- a/adminSiteClient/Forms.tsx
+++ b/adminSiteClient/Forms.tsx
@@ -772,6 +772,40 @@ export class BindString extends React.Component<{
 }
 
 @observer
+export class BindStringArray extends React.Component<{
+    field: string
+    store: Record<string, any>
+    label?: React.ReactNode
+    secondaryLabel?: string
+    placeholder?: string
+    helpText?: string
+    softCharacterLimit?: number
+    disabled?: boolean
+    rows?: number
+    errorMessage?: string
+    buttonContent?: React.ReactChild
+    onButtonClick?: () => void
+}> {
+    @action.bound onValue(value: string = "") {
+        this.props.store[this.props.field] = parseBulletList(value)
+    }
+
+    render() {
+        const { field, store, label, ...rest } = this.props
+        const values = store[field] as string[] | []
+        return (
+            <TextAreaField
+                label={label === undefined ? capitalize(field) : label}
+                secondaryLabel={this.props.secondaryLabel}
+                value={createBulletList(values || [])}
+                onValue={this.onValue}
+                {...rest}
+            />
+        )
+    }
+}
+
+@observer
 export class BindAutoString<
     T extends { [field: string]: any },
     K extends Extract<keyof T, string>,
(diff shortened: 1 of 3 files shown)


larsyencken (commented):
Overall I'm happy with this one, feedback is only minor, and mostly it will just need updating based on any API changes.

larsyencken (changes requested):
See Todo list added to the PR body

Marigold:
Thanks for review, Lars! We should discuss these two points
> ✨ Remove bullet point from descriptionKey

We store it as a list of keys, and so bullet points seemed natural. What would be a better way? (newlines? dynamically adding fields for individual points?)

> 🐛 You cannot delete a field by blanking it out

That's a "feature". In case you want to overwrite a field by **empty string**, you can't remove it from override YAML (because of fallback to ETL YAML file). We'd have to distinguish two cases - overwriting by an empty string and removing (thereby falling to default). I don't have a sense of importance of deleting field, which is why I only went for overwrite.

Marigold:
@larsyencken I fixed it all (and left blanking the field as setting the empty string) and got feedback from Bastian, who said it works well. He also suggested merging two "Update indicator" buttons (from two forms) into a single one, which I did (there's a small glitch when submitting a form with Enter, but nothing serious). 

On [the ETL side](https://github.com/owid/etl/pull/2321), I've added integration tests and made a couple of fixes.

larsyencken:
I've tried a few times to QA this on staging, but it appears to be broken there. Any click on "Update indicator" does nothing, and there's no errors in the logs at `/home/owid/.pm2/logs/etl-api-out.log` or the equivalent error log.

Any ideas what's happening?

larsyencken (approved):
As discussed, merge and improve after

PR #3111, opened 2024-01-18
:bug: fix broken color scheme selection by reverting interface change

danyx23:
This PR reverts one change that was done in the types extraction PR a week ago. The interface `ColorScaleConfigInterface` used to be defined as a type alias to a class (unlike what the name suggests). I changed this and with it also the way the ColorScale.config field was implemented but this then broke the color scale dropdowns.

I'd like to revert this PR at some point and figure out why this breaks the drop downs but I don't have the time rn to investigate this in detail.


sophiamersmann (approved):
yap, that fixes it. the things we have to do...

Issue #3116, opened 2024-01-19
🐛 MySQL docker failure during dev and CI


larsyencken:
MySQL in Docker is failing in the `db-load-data` step due to a new failure in the upstream `mysql:latest` image.

## Example in CI

https://github.com/owid/owid-grapher/actions/runs/7570304773/job/20653772497

<details><summary>See CI log</summary>

```
==> Starting & initializing Mysql instance for testing
Creating network "owid-grapher_default" with the default driver
Creating volume "owid-grapher_mysql_data_testing" with default driver
Pulling db (mysql/mysql-server:latest)...
latest: Pulling from mysql/mysql-server
Digest: sha256:d6c8301b7834c5b9c2b733b10b7e630f441af7bc917c74dba379f24eeeb6a313
Status: Downloaded newer image for mysql/mysql-server:latest
Building db-load-data
#0 building with "default" instance using docker driver

#1 [internal] load .dockerignore
#1 transferring context: 2B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 185B done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/mysql/mysql-server:latest
#3 DONE 0.0s

#4 [1/2] FROM docker.io/mysql/mysql-server:latest
#4 DONE 0.0s

#5 [2/2] RUN microdnf -y update  && microdnf install -y     libpwquality     curl     rsync     unzip     gzip
#5 0.255 Downloading metadata...
#5 9.737 Downloading metadata...
#5 19.57 Downloading metadata...
#5 20.85 Downloading metadata...
#5 21.37 Downloading metadata...
#5 24.81 Package                                                                             Repository                Size
#5 24.81 Installing:                                                                                                       
#5 24.81  acl-2.2.53-1.el8.x86_64                                                            ol8_baseos_latest      82.9 kB
#5 24.81  cpio-2.12-11.el8.x86_64                                                            ol8_baseos_latest     271.9 kB
#5 24.81  cracklib-2.9.6-15.el8.x86_64                                                       ol8_baseos_latest      95.4 kB
#5 24.81  cracklib-dicts-2.9.6-15.el8.x86_64                                                 ol8_baseos_latest       4.1 MB
[...]

PR #3117, opened 2024-01-19
:bug: Try resolving #3116 by pinning MySQL to 8.0

larsyencken:
Since everything was working until recently, it seems likely that the upstream MySQL docker image on `:latest`, which is currently `8.3`, is broken.

This PR tries to fix the issue by pinning to `8.0` to see if that works.


larsyencken:
Did not solve the issue. It's more of a package issue than an image issue.

PR #3125, opened 2024-01-22
🎉 add ARIA labels to all site buttons

ikesau:
Just a quick and easy thing I noticed when reviewing our Lighthouse audits. Effectively, procrastination in the face of our other, much harder to improve metrics like Largest Contentful Paint 😬 

## Summary by CodeRabbit

- **Accessibility Improvements**
	- Enhanced screen reader support with added `aria-label` attributes to various interactive elements across the site.

### DIFF
--- a/site/CookieNotice.tsx
+++ b/site/CookieNotice.tsx
@@ -42,6 +42,7 @@ export const CookieNotice = ({
                 </div>
                 <div className="actions">
                     <button
+                        aria-label="Reject cookies"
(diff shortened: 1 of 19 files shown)


mlbrgl (approved):
Good idea!

There are a couple of cases (search, donation), where the aria-label and the default accessible name could be considered redundant (e.g. "Search" vs "Submit search"), but not a major concern.

PR #3128, opened 2024-01-23
:sparkles: migrate gdrive images from Spaces to R2

Marigold:
Migrate GDoc images from DigitalOcean Spaces to Cloudflare R2. Use `rclone` instead of `aws cli` for syncing images. We're doing this to have a **unified rclone & R2 experience**.

The challenge will be to make everyone switch to new R2 credentials. This involves:
1. Installing `rclone`
2. Generating their personal R2 tokens or receiving new ones
3. Creating `~/.config/rclone/rclone.conf`
4. Adding the following to their local `.env`
```
IMAGE_HOSTING_R2_CDN_URL=https://images-staging.owid.io
IMAGE_HOSTING_R2_BUCKET_PATH=owid-image-upload-staging/dev-name
```
5. (Optional) Removing old `IMAGE_HOSTING_*` values from env file

This PR uses new names for env variables to ensure devs update them. Images are split into two buckets now - `owid-image-upload` and `owid-image-upload-staging` with different permissions (the structure stays the same, i.e. `owid-image-upload/production`).

I only plan to sync production images, devs will have to run `make sync-images` first to populate the staging bucket.

## TODO - change envs
- [ ] Run a script to update `.env` of all staging servers (we already have that)
- [ ] Update envs of non-staging servers (ideally by cleaning their `init.sh`, adding new envs to `ansible-vault` and run `init.sh` to sync it)
- [ ] Update envs of non-foundation servers (`owid-live`)

## TODO

- [x] create new `owid-image-upload` and `owid-image-upload-staging` buckets in R2 with proper permissions
- [ ] sync all assets with `rclone sync s3:owid-image-upload/production/ r2:owid-image-upload/production/ --transfers=64 --checkers=64 --fast-list --verbose`
- [ ] install `rclone` and add config on all staging servers
- [ ] cleanup `spaces-nyc3:owid-image-upload/staging-servers`


Marigold:
@larsyencken, what do you think about this? It's going to be annoying, but most of our stuff is already on R2 so this transition makes sense.

Marigold:
This is still an issue. It's a bit better after increasing number of retries when fetching from DigitalOcean, but we're still running into issues every now and then.

devTools/docker/sync-s3-images.sh
```diff
@@ -13,12 +13,12 @@ fi
...
   echo
-  echo '  aws configure --profile=owid-spaces'
+  echo '  aws configure --profile=owid-r2'
   echo
   exit 1
 fi
 
-aws --profile=owid-spaces --endpoint=https://nyc3.digitaloceanspaces.com s3 sync s3://owid-image-upload/production/ s3://$IMAGE_HOSTING_BUCKET_PATH/ --acl public-read
\ No newline at end of file
+aws --profile=owid-r2 --endpoint=https://078fcdfed9955087315dd86792e71a7e.r2.cloudflarestorage.com s3 sync s3://owid-image-upload/production/ s3://$IMAGE_HOSTING_BUCKET_PATH/ --acl public-read
```
larsyencken, line 24: Based on our discussion, and tool preference, we should use rclone instead, and guide on how to set up r2 credentials in rclone.

We should assume a multi-user setup, which means our config for rclone should use a profile called `owid-r2` rather than `r2`.

larsyencken (approved):
Based on the HTTP 500s we've been getting, and keep getting, I'm all in favour of removing Spaces entirely.

I think you can manage the prod credentials to do this move, but I'm slightly wary of changing things whilst @ikesau is pushing on homepage this week.

First thing next week?

## Meta: PR feedback

- In the PR, try to only describe the change and the motivation, not the issue and the commentry
  - Move that and any discussion to an issue instead
  - Alternatively, it's fine to put discussion about the value of the PR in comments
- If you do this, the PR head will be much simpler, and understandable at a glance

danyx23:
Hey @Marigold, what is the status of this PR? I think you managed to solve the retry issues but this still seems like a good idea, no? I agree with Lars that unless there is a good reason for the aws tool it seems to make sense to switch to rclone. Giving everyone their own R2 keys would also be great!

Marigold:
Retry issues are gone (though having to retry so often is not optimal), but I agree that having unified `rclone` and R2 would be a good long-term investment. @larsyencken do I have a green light to start the transition?

danyx23:
Hey @Marigold - yes this looks good. Just a few questions for clarification:
- Will data managers have to run this too? My guess would be that they mostly care about a running admin, much less a working site preview, and as such would not need to do this setup?
- How will we do credential generation? Will everyone do their own?
- If you run the ETL do you already have legitimate R2 credenitals or are those different?

danyx23 (approved):

Marigold:
> Will data managers have to run this too? My guess would be that they mostly care about a running admin, much less a working site preview, and as such would not need to do this setup?

That's right. I'd say that `rclone` could be useful to them, but they don't need any of that to run Admin. I'll say it's optional. They'll have to change a couple of env variables in `etl/.env` though.

> How will we do credential generation? Will everyone do their own?

Only devs (people with Cloudflare access). I'll create tokens for others under the shared Cloudflare user and send them their credentials (shared as a secured note in 1password).

> If you run the ETL do you already have legitimate R2 credenitals or are those different?

Data managers have R2 credentials with read & write access to all buckets. Their new credentials will only have access to staging buckets. (And I'll revoke the current ones after the migration)

(I'm writing a more detailed [guide in Notion](https://www.notion.so/owid/2024-03-06-Rclone-guide-9175170f53484cfcbcb11201c18e39c3) which contains more detailed information)

danyx23:
Great, thanks for answering my questions!

PR #3074, opened 2024-01-08
🔨 Convert SuggestedChartRevision db access to knex

danyx23:
This PR attempts to change use of the `suggested_chart_revisions` table to knex. It is WIP at this point as the code changes led to some more restructuring

PR #3064, opened 2024-01-04
🔨 add types for database rows and prepare use of knex instead of TypeORM

danyx23:
This PR add helper types for all db tables that are still in reasonable use, even if they are not touched by the Grapher codebase ATM.

The naming scheme for the types is as follows:
- Db**Plain**TableNameSingular
- Db**Raw**TableNameSingular
- Db**Enriched**TableNameSingular
- Db**Insert**TableNameSingular

TableNameSingular is the singular case for the table name, e.g. `Variable` for the `variables` MySQL table or `PostGdoc` for `posts_gdocs`.

The DbPlain* variant is for cases where all columns have a primitive type that does not need to be parsed. 

The DbRaw* variant is for tables that have complex types (e.g. json columns) that need to be parsed. The resulting type once they have been (manually) parsed is the DbEnriched* variant of the type. Next to the type definitions in ./packages/@ourworldindata/types/src/dbTypes are also the parsing/serializing functions. These are currently explicit functions for each column even if they are for now all implemented as simple `JSON.parse` calls - but in theory they might want to add validation.

Finally, the DbInsert* variant is for inserting rows into the database. All fields that correspond to columns that are either nullable or have default values are optional so that the minimal object that is valid in the DB can be expressed correctly but all optional fields can be specified as well.

This PR also adds setup for the `sql-ts` tool that generates typescript types from MySQL tables. This is useful if a substantial number of new tables are added and you want to get a reasonable starting point for the corresponding TS type.

PR #3056, opened 2023-12-30
Extract grapher and core table types

danyx23:
This PR extracts all types from the `grapher `and `core-table` packages and moves them into the `types `package. This will allow us to freely re-use types finally, e.g. so that the `ChartsRow `type defined in types can specify that the `content `field is a `GrapherInterface`.

PR #3054, opened 2023-12-29
⚒️ Move types out of utils and into their own types package

danyx23:
This PR breaks the messy `owidTypes.ts` file up into around 20 files and moves them into the `types` package. This PR has some issues on the tests that were fixed in the next PR in the stack (#3056) - it is only supposed to be merged as part of the entire stack.

PR #3131, opened 2024-01-24
:hammer: (db) migrate Dataset and Source to knex

sophiamersmann:
- Makes the `Dataset` and `Source` TypeORM classes obsolete by switching db access to knex
- This PR does not do any clean-up, e.g. SQL queries in the router files are not moved, although they ultimately should live in the `db` folder

- I added two helper functions that can be used as drop-in replacements for `queryMysql` and `mysqlFirst` that are often used in our codebase, `knexRaw` and `knexRawFirst` (both live in `db.ts`)

- Routes that I could easily test are tested, but not all of the code is!

@danyx23 let me know if this is how you imagined it, or if I should do anything differently.

@coderabbitai ignore

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -50,15 +50,20 @@ import {
     OwidGdoc,
 } from "@ourworldindata/utils"
 import {
+    DbPlainDatasetTag,
     GrapherInterface,
     OwidGdocType,
+    DbPlainTag,
     grapherKeysToSerialize,
+    DbRawVariable,
+    DbRawOrigin,
+    parseOriginsRow,
 } from "@ourworldindata/types"
 import {
     getVariableDataRoute,
     getVariableMetadataRoute,
 } from "@ourworldindata/grapher"
-import { Dataset } from "../db/model/Dataset.js"
+import { getDatasetById, setTagsForDataset } from "../db/model/Dataset.js"
 import { User } from "../db/model/User.js"
 import { GdocPost } from "../db/model/Gdoc/GdocPost.js"
 import { GdocBase, Tag as TagEntity } from "../db/model/Gdoc/GdocBase.js"
@@ -1589,7 +1594,10 @@ apiRouter.get(
 )
 
 apiRouter.get("/datasets.json", async (req) => {
-    const datasets = await db.queryMysql(`
+    await db.knexInstance().transaction(
+        async (trx) => {
+            const datasets = await db.knexRaw<Record<string, any>>(
+                `
         WITH variable_counts AS (
             SELECT
                 v.datasetId,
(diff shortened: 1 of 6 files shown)


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3131?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #3054** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3054?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #3056** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3056?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #3064** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3064?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #3131** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3131?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
        * **PR #3074** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3131?utm_source=stack-comment).

adminSiteServer/apiRouter.ts
```diff
@@ -1949,27 +1954,24 @@ apiRouter.put("/datasets/:datasetId", async (req: Request, res: Response) => {
...
-            commitName: res.locals.user.fullName,
-            commitEmail: res.locals.user.email,
-        })
-    } catch (err) {
-        logErrorAndMaybeSendToBugsnag(err, req)
-        // Continue
-    }
+        try {
+            await syncDatasetToGitRepo(t, datasetId, {
+                oldDatasetName: dataset.name!,
```
sophiamersmann, line 1966: The TypeORM code just assumed that a dataset `name` was given. That's why I felt it was safe to sometimes ignore a possible undefined value via Typescript's `!`. Let me know if you prefer for this to be handled properly.

adminSiteServer/apiRouter.ts
```diff
@@ -2450,16 +2451,16 @@ apiRouter.post("/posts/:postId/unlinkGdoc", async (req: Request) => {
 
 apiRouter.get("/sources/:sourceId.json", async (req: Request) => {
     const sourceId = expectInt(req.params.sourceId)
-    const source = await db.mysqlFirst(
+    const source = await db.knexRawFirst<Record<string, any>>(
```
sophiamersmann, line 2454: When it wasn't particularly important, I didn't use specific types. Let me know if you prefer to have strict types everywhere.

danyx23 (changes requested):
Thanks a lot for test driving the knex approach! This looks very good overall. There are two things that I would change on the db-types PR learning from this and two things I'd ask you to change.

The two changs I'd like to see to this PR is:
- not make the knex instance optional in the knexRaw and knexRawFirst helper functions
- and to wrap all API functions we touch in a transaction scope even if we only read data as we go through them. The reasoning here is that we want reads to be consistent and it's probably better to just implement a standard pattern everywhere rather than having to evaluate if it is necessary each time.

The two things I'll change on the other PR are:
- bump the knex version so that we can create readyOnly transactions which are a bit more performant than generic ones
- add a migration to make name on datasets mandatory rather than nullable

danyx23 (changes requested):
Thanks a lot for test driving the knex approach! This looks very good overall. There are two things that I would change on the db-types PR learning from this and two things I'd ask you to change.

The two changs I'd like to see to this PR is:
- not make the knex instance optional in the knexRaw and knexRawFirst helper functions
- and to wrap all API functions we touch in a transaction scope even if we only read data as we go through them. The reasoning here is that we want reads to be consistent and it's probably better to just implement a standard pattern everywhere rather than having to evaluate if it is necessary each time.

The two things I'll change on the other PR are:
- bump the knex version so that we can create readyOnly transactions which are a bit more performant than generic ones
- add a migration to make name on datasets mandatory rather than nullable

sophiamersmann (commented):
Good to go I think!

Before the `name` field was made non-nullable in the db, I added a check to make sure a dataset `name` is given. We could get rid of that now :)

PR #3138, opened 2024-01-25
biomejs experiment

ikesau:
biome's a new kid on the block that does formatting and linting very quickly. 

Benchmarks
```
/usr/bin/time yarn testLint // 22.31s

/usr/bin/time yarn testPrettierAll // 7.46s

/usr/bin/time yarn biome lint * // 0.78s

/usr/bin/time yarn biome format * // 0.78s
```

This isn't a fair comparison as prettier is also linting our scss, html, and yaml files.

Unfortunately, while biome is currently working on adding CSS support, they've said they don't plan to add SCSS support anytime soon.

At best, we'd be living in a two-formatter world by running [stylefmt](https://github.com/masaakim/stylefmt) on SCSS changes.

Will leave this around as a draft if anyone wants to have a play to work out next steps. Worst case we close this and have it as a reference if we ever want to return to it in the future.


biome.json
```diff
@@ -0,0 +1,113 @@
...
+	"$schema": "https://biomejs.dev/schemas/1.5.3/schema.json",
+	"organizeImports": {
+		"enabled": true
+	},
+	"linter": {
+		"enabled": true,
+		"rules": {
+			"recommended": true,
+			"a11y": {
+				"useKeyWithClickEvents": "off",
```
ikesau, line 11: These are all the lint rules I had to disable so that it wouldn't complain about any of our current code.

Many of them are very sensible and I think we'd do well to fix most of the things it complains about (especially in the non-style related rule categories)

ikesau:
Might try to hack on scss/html handling again during cooldown

Issue #3137, opened 2024-01-25
Accessibility considerations for year sliders


CAYdenberg:
## Description

A number of charts feature year sliders that allow the user to pan back and forth in time. While not formally broken, the behaviour of these elements is mysterious (actions have unexpected consequences), and are also inaccessible by keyboard control.

## Expected behaviour

There should be clear and well labelled ways to convert between range and single-year mode, and controls should be accessible to keyboard users.

## Steps to reproduce

Go to https://ourworldindata.org/grapher/co2-emissions-vs-gdp and mess with the slider.

The only way to convert from single-year to range mode is to click one of the labels. The only way to convert back to single-year is to slide the handles on top of each other. These consequences are unexpected. Proper HTML form controls would be much simpler, easier to understand, and usable by keyboard.

## Environment

The same experience is present on all devices.

danyx23:
There is a half-finished branch by @sophiamersmann that we'd like to finish at some point that adds an explicit tab to switch between single-year and multi-year view

danyx23:
Switching between single year chart view (bar chart) and multiple year view (line chart) will likely be added in the next few months (related to #3231 and #1162). 

The accessibility concern is a very good point and we should implement that part of the issue soon as well.

PR #3162, opened 2024-02-01
✨ (a11y) make time slider keyboard accessible

sophiamersmann:
### Summary

- Grapher's time slider is inaccessible to keyboard users and screen readers (as described in #3137)
- This PR doesn't change fundamentals but improves the status quo by implementing the Slider pattern defined [here](https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb/examples/slider-multithumb/):
    - Adds the start and end thumbs to the tab sequence
    - Adds aria information that is accessible to screen readers
    - Adds support for keyboard navigation via arrow keys

Important limitation:
- Keyboard navigation only allows updating both sliders separately. It is currently impossible – by keyboard navigation only – to enter "single-year" mode if a chart has range mode

### Other a11y improvements
- Use a `<button />` when appropriate (instead of links or divs)
- Add aria labels for buttons that may not contain a textual label
- Add focus state to the LabelledSwitch component

## Summary by CodeRabbit

- **New Features**
	- Enhanced accessibility across various components with the addition of `aria-label` attributes.
	- Improved keyboard navigation for the timeline component with new keyboard event handling methods.
- **Refactor**
	- Changed tab elements from anchor (`<a>`) tags to button (`<button>`) tags in the `ContentSwitchers` component for better semantic HTML and accessibility.
	- Updated the `ActionButton` component to hide tooltips on mouse leave.
- **Style**
	- Adjusted CSS selectors in `ContentSwitchers` to apply styles to button elements instead of anchor tags.

### DIFF
--- a/packages/@ourworldindata/grapher/src/controls/ActionButtons.tsx
+++ b/packages/@ourworldindata/grapher/src/controls/ActionButtons.tsx
@@ -347,6 +347,7 @@ export function ActionButton(props: {
                 onMouseLeave={(): void => {
                     setShowTooltip(false)
                 }}
+                aria-label={props.label}
             >
                 <FontAwesomeIcon icon={props.icon} />
                 {props.showLabel && (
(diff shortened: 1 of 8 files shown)


danyx23 (approved):
Looks great!

PR #3189, opened 2024-02-13
:honeybee: set the dev tools to use a profile for s3 credentials

larsyencken:
## Motivation

Previously, our dev tool used the default s3 credentials when syncing to DigitalOcean Spaces. This is a little problematic, since we are now using multiple object stores.

This change suggests setting up a profile for OWID and explicitly using that. In this way, it's compatible with using multiple object stores, and in fact having multiple sets of S3 credentials on your machine.

## Setting up a profile

`aws configure --profile=owid-spaces`


ikesau (approved):
Worked as expected once I configured a profile. 👍 

Could you add a comment to the bash file that includes that aws command?

PR #3202, opened 2024-02-15
:honeybee: Dev tooling: log all MySQL queries in dev

larsyencken:
Sometimes it's helpful to understand what queries are actually hitting the database as you work.

This change mounts the .gitignored `logs/` folder into the Docker MySQL container, and writes queries to `logs/queries.log` as they happen. This means you can `tail -f logs/queries.log` in your local environment and see queries happen live.


larsyencken:
@Marigold Added slow query logging as you suggested

PR #3208, opened 2024-02-16
:hammer: Remove workerpool from baking charts and use async

Marigold:
Counterintuitively, using multiple workers (16 currently) is slower than baking all charts in async. In addition to being faster, async processing might mean more transparent stacktraces and lower CPU usage.

* 16 processes - 100.5s
* async - 63s

Tested on `owid@staging-site-master` with
```
yarn buildTsc && yarn buildLocalBake http://staging-site-master /home/owid/live-data/bakedSite --steps charts
```

I hope I didn't miss anything.


mlbrgl (approved):
makes sense, let's give this a try ([related conversation](https://owid.slack.com/archives/CQQUA2C2U/p1708072084140799))

PR #3211, opened 2024-02-19
🐛 commit svg tester diffs to different branches

sophiamersmann:
### Problem

- Both SVG tester actions commit to the same branch, but force-push their changes, which means that they overwrite changes on remote
- As a result, the last action that pushes wins and overwrites the commit of the other action

### Solution

- We could attempt to fix this, but I never felt strongly about having both actions commit to the same branch
- It's easier and less error-prone to just have both actions commit to different branches
    - If on branch `<my-branch>`, then the first action now commits to a branch of the same name, and the other action commits to `<my-branch>__all-views`

### Tests

[The follow-up PR](https://github.com/owid/owid-grapher/pull/3215) makes the SVG tester actions fail on purpose. Note that both actions commit to different branches:

SVG tester:

Screenshot 2024-02-19 at 13 59 55
Screenshot 2024-02-19 at 14 00 24

SVG tester (all views):

Screenshot 2024-02-19 at 14 00 35
Screenshot 2024-02-19 at 14 00 49

### DIFF
--- a/.github/workflows/svg-compare-all-views.yml
+++ b/.github/workflows/svg-compare-all-views.yml
@@ -16,14 +16,14 @@ jobs:
               if: ${{ github.repository_owner == 'owid' }}
               shell: bash
               run: |
-                  echo "PUSH_BRANCH_NAME=${{ github.ref_name }}" >> $GITHUB_ENV
+                  echo "PUSH_BRANCH_NAME=${{ github.ref_name }}__all-views" >> $GITHUB_ENV
                   echo "TOKEN=${{ secrets.GITHUBPAT }}" >> $GITHUB_ENV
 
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* `master`
  * **PR #3211** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3211?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3211?utm_source=stack-comment).

danyx23 (approved):
Looks like a good approach!

PR #3226, opened 2024-02-22
Updated images and text to match what I see on a fresh install

toni-sharpe:
Back again with a few updates to the new user stuff.

Small things keeping it up to date and making it clearer for folk who've not got the best eyes.

These things sometimes get out of date, so I thought I'd update - we've got 5 TMUX panels now and one for the DB, rather than having it in docker. 

I updated the image, and provided some text because data just in images is bad, plus, upgraded the headiong of "trouble shooting" so it doesn't look like a sub of the section before, really, troubleshooting is it's own special thing.

Updated a bit of text - do check this, I'm new.

### Screen shots

This is the updated visual I added. I added the URLs and TMUX as text, though possibly could have used the alt tag.

image

~Looking forward to my conversation with the rabbit haha~ they're hiding ...


toni-sharpe:
As a general aside, I still don't have a DB working here, even having followed the instructions in the discussions,

I'll look at those more closely, and I'm wondering if something to do with either copying across my ZSH/bash settings from my old computer might have borked something, or if [my other project](https://brainstem-research.pages.dev) which uses python, PostGreSQL and so-on might have caused issues.

I'll dig into this more, current status is jobless and hard to even get anywhere, so **I'd love** to offer some super-cheap or even free work on this, so I'll persevere (and document, help the next one out).

toni-sharpe:
OK, I think this can go, I do have some further looks into Docker, but that's not done.

I think I'm on the right lines, I have docker things running from earlier projects (you might not have encountered this on new machines for the job, and, you get Docker for Mac for free being non-profit/open source, last time, I used a (very good) command line tool). 

Whereas I have a MySQL and even using the NUKE option isn't there yet.

```
docker stop $(docker ps -a -q)
docker rm $(docker ps -a -q)
```
which is pulling things out that Docker for Mac doesn't know about.

Docker is persistent. And it's a matter of understanding the problem and being sufficiently articulate to describe it, all when devops speciated from development some years ago.

But I think this PR is atomic and can go and be useful.

toni-sharpe:
Bumping this as I htought there was some value in it, also the comment, as this could be a common occurence with outside contributors

toni-sharpe:
Closing in favour of: https://github.com/owid/owid-grapher/pull/3523

Issue #3007, opened 2023-12-11
Show at least 2 y-axis labels in charts with multiple panels


saloni-nd:
## Description

In charts where there are multiple panels, Grapher sometimes only shows 1 y-axis label (e.g. 0). 

I think there should be at least 2 axis labels, so people can have a sense of proportions.

Example [chart](https://owid.cloud/admin/charts/7161/edit), after switching to 'split by cause'

<img width="683" alt="image" src="https://github.com/owid/owid-grapher/assets/31684905/ee158df4-72c4-4f5b-8617-ca3e122a2273">

## Environment

For desktop, provide:

iOS Chrome Version 119.0.6045.199

PR #3239, opened 2024-02-23
🐛 (grapher) show a minimum of two tick labels if possible

sophiamersmann:
- Fixes https://github.com/owid/owid-grapher/issues/3007

### Problem

- On smaller screens, Grapher sometimes displays a single tick label due to space constraints, which makes it impossible to read the chart
- Example: https://ourworldindata.org/grapher/death-rate-by-cause-who-mdb?facet=metric (on a smaller screen)

### Solution

- Given a number of base ticks, we want to find a subset of ticks whose labels are not overlapping:
    - The original strategy figured out which tick labels to hide so that we don't end up with overlapping tick labels
    - The updated strategy does the same but retries if we end up with a single visible tick label:
        1. Hide overlapping ticks, allowing for some padding (3px) between tick labels (this is the original strategy)
        2. If we end up with a single tick label, try again but reduce the allowed padding between tick labels (1px) – note that for vertical axes, padding between tick labels is not essential
        3. If we still end up with a single tick label, then there really isn't enough space to show a second label – in such cases, try to show a "useful" tick label; in other words, if the zero tick is visually marked by a solid line, deprioritise it, such that a non-zero tick label is chosen to be displayed

### Example (from the original issue)

Before             |  After
:-------------------------:|:-------------------------:
Screenshot 2024-02-26 at 10 20 17  |  Screenshot 2024-02-26 at 10 20 22
Only zero ticks are shown, which makes it impossible to read the chart | We don't have space for a second tick, but we choose to show a non-zero tick, which gives useful information

### SVG tester

Screenshot 2024-02-26 at 10 12 10

Adds an additional grid line (makes no difference really). (The second chart in the top row is broken in another way, which is fixed in [this PR](https://github.com/owid/owid-grapher/pull/3244))

Screenshot 2024-02-26 at 10 11 17

Displays some more tick labels (which is a good thing)

### DIFF
--- a/packages/@ourworldindata/grapher/src/axis/Axis.ts
+++ b/packages/@ourworldindata/grapher/src/axis/Axis.ts
@@ -8,6 +8,7 @@ import {
     max,
     numberMagnitude,
     sortedUniqBy,
+    clamp,
     Bounds,
     DEFAULT_BOUNDS,
     AxisAlign,
@@ -18,6 +19,7 @@ import {
     TickFormattingOptions,
     Tickmark,
     ValueRange,
+    cloneDeep,
 } from "@ourworldindata/utils"
 import { AxisConfig } from "./AxisConfig"
 import { TextWrap } from "@ourworldindata/components"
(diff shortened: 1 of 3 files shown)


sophiamersmann:
* **#3239** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3239?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3239?utm_source=stack-comment).

danyx23 (approved):
Neato!

PR #3243, opened 2024-02-26
🔨 (grapher) refactor mobile full width mode for embedded graphers

### LINKED ISSUE #3010: Grapher's modal backdrop doesn't cover the whole frame when embedded in an article and viewed on a small screen
## Description

- On small screens, Grapher is optimised for horizontal space
    - Grapher bleeds onto the edges horizontally and the top and bottom borders stretch across the entire page (see screenshot)
- When a modal is open, the backdrop should also stretch across the entire page, but it doesn't

## Expected behaviour

- Backdrop always covers the entire frame

## Steps to reproduce

1. Go to on a mobile phone: https://ourworldindata.org/how-are-incomes-adjusted-for-inflation
2. Scroll to the "Consumer price index" chart
3. Open a modal
4. See that the backdrop doesn't cover the entire Grapher frame

## Screenshots

Screenshot 2023-12-11 at 11 12 39
Screenshot 2023-12-11 at 11 12 45

sophiamersmann:
### Summary

- On small screens, we want embedded Graphers to use the full screen width and bleed onto the edges horizontally (see screenshots below)
- This used to be implemented via a Grapher property `shouldOptimizeForHorizontalSpace`, which was awfully complicated and also led to the following bug: #3010 
- This PR removes the `shouldOptimizeForHorizontalSpace` property and simply sets negative margins on embedded charts
    - It's not guaranteed that the chart title aligns with the text. It just so happens that both the grid-gap and Grapher's frame padding are set to 16px
    - As Grapher's frame padding and the grid-gap are not likely to be updated frequently, I think it's okay to rely on that for now

### Screenshots

| Not using the full width  | Using the full width  |
| ------- | ------ |
| Screenshot 2024-02-26 at 13 59 11 | Screenshot 2024-02-26 at 13 57 13 |

### Related issue

- fixes #3010

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -89,7 +89,6 @@ export interface ExplorerProps extends SerializedGridProgram {
     isPreview?: boolean
     canonicalUrl?: string
     selection?: SelectionArray
-    shouldOptimizeForHorizontalSpace?: boolean // only relevant for explorers with hidden controls
 }
 
 const renderLivePreviewVersion = (props: ExplorerProps) => {
@@ -212,17 +211,6 @@ export class Explorer
     componentDidMount() {
         this.setGrapher(this.grapherRef!.current!)
         this.updateGrapherFromExplorer()
-
-        // Optimizing for horizontal space makes only sense if the controls are hidden
-        // and the explorer in fact looks like an ordinary grapher chart.
-        // Since switching between charts is not possible when the controls are hidden,
-        // we only need to run this code once.
-        if (
-            this.queryParams.hideControls &&
-            this.props.shouldOptimizeForHorizontalSpace
-        ) {
-            this.grapher!.shouldOptimizeForHorizontalSpace = true
-        }
 
         let url = Url.fromQueryParams(this.initialQueryParams)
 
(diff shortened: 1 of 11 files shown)


sophiamersmann:
* **#3243** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3243?utm_source=stack-comment).

ikesau (approved):
I didn't spend a ton of time parsing the old code and thinking about potential regressions, but the new stuff looks good (tested on Safari, Chrome and Firefox) 👍 

I see this is currently only implemented in Chart and Key Insight components - but are the other places _intentionally_ unchanged?

Explorer pages, datapages, key-indicators, all-charts blocks, etc, I feel like they'd all benefit from the extra horizontal width on mobile.

sophiamersmann:
Good point! The all-charts block and key-indicator blocks are intentionally unchanged. I didn't think about explorer pages and data pages, to be honest. It would be a nice touch on data pages, I think. On explorer pages, it would kind of separate the chart from the controls and break the alignment with the other boxes, if you know what I mean.

Looking at data pages in more detail, I'm noticing that data pages are padded by 24px on both sides, although it should be 16px according to the designs. I guess we'd need to fix that before we could use full-width Graphers on mobile :(

PR #3246, opened 2024-02-26
🎉 (test page) add explorer test pages

sophiamersmann:
### Background

- It would be great if we had some testing in place for explorers
    - In recent months, two explorer bugs happened that we should have caught before they went live:
        1. The source line disappeared for charts in an explorer 
        2. The entity selector disappeared for an explorer 
     - Both of these bugs weren't a result of work that was done on explorers, they were the result of code changes that were made far away from explorers
     - Additionally, these bugs only surfaced for one (or few) explorers, not all of them
     - (What I'm trying to say is that, without any testing in place, it is quite difficult at the moment to catch these bugs before shipping because, in theory, we would need to manually check all explorers for every grapher code change...)
 - In the long term, I'd love for the site screenshotting tool to include a few carefully selected explorers (ideal would be all of them, but I realise that that is probably not feasible...)
- For now, I figured that a relatively simple thing to do that would help me check explorers before shipping is to add explorer test pages in the admin that allow me to quickly scroll through before/after views of all explorers

### Summary

Adds:
- explorer test page that lists before/after views of all published explorers
- three more test pages that split explorer by chart creation mode, i.e.
    - Grapher id based explorers
    - CSV based explorers
    - Indicator based explorers

### Notes

- I didn't add pagination because I prefer scrolling and I think lazy loading should be enough to make the page performant
- Test pages don't currently work on staging sites because the [embed script](https://ourworldindata.org/grapher/embedCharts.js) isn't correctly served on staging. It's something that Marcel will look into once he's back

### DIFF
--- a/adminSiteClient/TestIndexPage.tsx
+++ b/adminSiteClient/TestIndexPage.tsx
@@ -172,6 +172,42 @@ export class TestIndexPage extends React.Component {
                             </Link>
                         </li>
                     </ul>
+
+                    <h2>Test Explorer Embeds</h2>
+                    <ul>
+                        <li>
+                            <Link native target="_blank" to="/test/explorers">
+                                All explorers
+                            </Link>
+                        </li>
+                        <li>
+                            <Link
+                                native
+                                target="_blank"
+                                to="/test/explorers?type=grapher-ids"
+                            >
+                                Grapher ID-based explorers
+                            </Link>
+                        </li>
+                        <li>
+                            <Link
+                                native
+                                target="_blank"
+                                to="/test/explorers?type=csv-files"
+                            >
+                                CSV file-based explorers
+                            </Link>
+                        </li>
+                        <li>
+                            <Link
+                                native
+                                target="_blank"
+                                to="/test/explorers?type=indicators"
+                            >
+                                Indicator-based explorers
+                            </Link>
+                        </li>
+                    </ul>
                 </main>
             </AdminLayout>
         )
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#3246** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3246?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3246?utm_source=stack-comment).

danyx23 (approved):

danyx23 (approved):
Looks good! There seems to be a bit of a style bleed or something, the explorer header in the localhost version has a different font for the heading (top left box) which then shifts the layout a bit which makes comparision maybe a little less nice. But not a block, this is a good thing to add!

sophiamersmann:
Interestingly, the difference in font size is due to the fact that the font size is based on the viewport width (`vw`). The iframed explorer on the right is contained within a smaller viewport, so its font size is smaller. Bit weird to specify a font size like this but I'll ignore it for now...

sophiamersmann:
### Merge activity

* **Mar 13, 10:09 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3246).

Issue #3238, opened 2024-02-23
Switch baked images to PNG instead of webp


ikesau:
WebP supported puttered out across all OS's and so despite smaller filesizes, it's impacting the shareability of our inline images.

We should bake the images as PNG instead.

https://github.com/owid/owid-grapher/blob/f8f12b14b9f767fbee17562b6b1707313927f915/baker/GDriveImagesBaker.tsx#L89-L94

And make sure any place that's assuming webp is changed to PNG instead.

Issue #3220, opened 2024-02-19
Featured image og:image doesn't work with non-PNG images


ikesau:
https://github.com/owid/owid-grapher/blob/5dc05158d4d54b262ca04bb4b55bbc6aeab4f41b/baker/GDriveImagesBaker.tsx#L114-L118

this fallback code bakes the image as its own filetype

but

https://github.com/owid/owid-grapher/blob/5dc05158d4d54b262ca04bb4b55bbc6aeab4f41b/packages/%40ourworldindata/utils/src/image.ts#L93-L98

assumes that the image will always exist as a PNG

We should either always bake the fallback as a PNG, or detect which file extension is used for the og:image.

PR #3248, opened 2024-02-26
🔨 bake gdoc images as PNG instead of WebP

ikesau:
>  _Lossless WebP images are typically 26% smaller than the same images in PNG format._ 

\- [MDN](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#webp_image)

Unfortunately, WebP is not as well-supported as PNG in most operating systems and programs, which Max and Este feel is more important when it comes to people using our graphics in their documents. 😢

Resolves https://github.com/owid/owid-grapher/issues/3238
Resolves https://github.com/owid/owid-grapher/issues/3220


danyx23 (approved):
Looks good - in Image.tsx:107 there is a reference to image/webp in an if branch that checks if we are previewing. Is that fine or should that change as well?

PR #3254, opened 2024-02-27
🎉 add reusable Button component

ikesau:
Makes a new component for the `~~Button~~` that we're using everywhere

Contemplated hiding the icon when `onClick` is defined but decided to cross that bridge when we come to it.

## Mobile
[red-button-mobile](https://github.com/owid/owid-grapher/assets/11844404/8a8b7e15-3910-488d-bc39-3f8e5b0165e3)
[master-mobile](https://github.com/owid/owid-grapher/assets/11844404/0fa75c75-bde9-40f4-905d-0e00fe460375)

## Desktop
[master-desktop](https://github.com/owid/owid-grapher/assets/11844404/5ea8cfb8-cd6e-4c79-8e60-dfab12b4fd02)
[red-button-desktop](https://github.com/owid/owid-grapher/assets/11844404/104f3933-12f2-4b93-b6fa-897c1fe73fc2)

### DIFF
--- a/packages/@ourworldindata/components/src/Button/Button.tsx
+++ b/packages/@ourworldindata/components/src/Button/Button.tsx
@@ -1,0 +1,56 @@
+import React from "react"
+import cx from "classnames"
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome/index.js"
+import { IconDefinition, faArrowRight } from "@fortawesome/free-solid-svg-icons"
+
+type ButtonCommonProps = {
+    text: string
+    className?: string
+    theme: "solid-vermillion" | "outline-vermillion"
+    /** Set to null to hide the icon */
+    icon?: IconDefinition | null
+}
+
+type WithHrefProps = {
+    href: string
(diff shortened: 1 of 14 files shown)


sophiamersmann (approved):
Looks good to me! Nice effort 👍🏻

packages/@ourworldindata/components/src/Button/Button.tsx
```diff
@@ -0,0 +1,56 @@
...
+    theme = "solid-vermillion",
+    className,
+    href,
+    onClick,
+    text,
+    icon = faArrowRight,
+}: ButtonProps) => {
+    const classes = cx("owid-btn", `owid-btn--${theme}`, className)
+
+    if (href) {
```
toni-sharpe, line 38: Not so sure about this bit,

[Article from MDN explains more fully](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role).

Maybe this is something to raise for future work if there is time pressure.

PR #3269, opened 2024-02-29
🚧 (slope) use line chart like entity selection

sophiamersmann:
❌ not to be merged. just an experiment.

PR #3270, opened 2024-02-29
✨ (slope) improve hover interaction

sophiamersmann:
Improves hover interaction on slope charts.

### Summary

- Improves the "algorithm" to figure out which slope is currently closest to the mouse (see below for an explanation)
- Adds support for clicking anywhere inside the Grapher frame to dismiss the current selection
- Uses the "default" cursor for value and entity labels since they're clickable (default is the text selection cursor)

<details><summary>How do we figure out which slope is currently closest to the mouse?</summary>
<p>

| Before  | After  |
| ------- | ------ |
| Screenshot 2024-03-07 at 10 27 26 | Screenshot 2024-03-07 at 10 27 34 |
| We used to measure the distance to the orange line that is defined by the start and end point of a slope but extends to both sides. | Now we measure the distance to the line _segment_ that is the slope or, if the mouse is outside of the chart area, we measure the distance to a label that is represented by a straight line |

</p>
</details> 

### Caveats

- If a Grapher chart has entities selected by default, then clicking anywhere inside the Grapher frame will dismiss the default selection ([Example](http://staging-site-slope-hover-interaction/grapher/change-in-female-height-slope))
    - I do believe that if it's possible to select entities by clicking on slopes or clicking the legend, then it should also be possible to dismiss a selection by clicking into dead space (it's almost expected behaviour, I think?)
    - But dismissing a default selection before a user has even interacted with a chart is not ideal
    - We have very few slope charts though that allow entity selection and have a set of entities selected by default
    - All in all, I think it's still net positive but can be convinced otherwise

### Videos

<details><summary>Before/After</summary>
<p>

https://github.com/owid/owid-grapher/assets/12461810/3e013eae-3981-492b-908d-a940411578db

https://github.com/owid/owid-grapher/assets/12461810/dbfcbcd3-66df-4dce-99a2-c2fe14cc6d7f

</p>
</details> 

### Context

Part of a larger body of work. See stack 👇🏻 

### SVG tester

All slope charts are updated because of the `cursor` change.

### DIFF
--- a/packages/@ourworldindata/grapher/src/slopeCharts/SlopeChart.tsx
+++ b/packages/@ourworldindata/grapher/src/slopeCharts/SlopeChart.tsx
@@ -15,6 +15,7 @@ import {
     minBy,
     maxBy,
     exposeInstanceOnWindow,
+    PointVector,
 } from "@ourworldindata/utils"
 import { TextWrap } from "@ourworldindata/components"
 import { observable, computed, action } from "mobx"
@@ -58,11 +59,15 @@ import { CoreColumn, OwidTable } from "@ourworldindata/core-table"
 import { autoDetectYColumnSlugs, makeSelectionArray } from "../chart/ChartUtils"
 import { AxisConfig, AxisManager } from "../axis/AxisConfig"
 
+export interface SlopeChartManager extends ChartManager {
+    isModalOpen?: boolean
+}
+
 @observer
 export class SlopeChart
     extends React.Component<{
         bounds?: Bounds
-        manager: ChartManager
+        manager: SlopeChartManager
     }>
     implements ChartInterface, VerticalColorLegendManager, ColorScaleManager
 {
@@ -119,12 +124,8 @@ export class SlopeChart
     }
 
     @action.bound onSlopeClick() {
-        const { manager, hoverKey } = this
-        if (
-            manager.addCountryMode === EntitySelectionMode.Disabled ||
-            !manager.addCountryMode ||
-            hoverKey === undefined
-        ) {
+        const { hoverKey, isEntitySelectionEnabled } = this
+        if (!isEntitySelectionEnabled || hoverKey === undefined) {
             return
         }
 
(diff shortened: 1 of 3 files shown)


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3270?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3270** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3270?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3269** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3270?utm_source=stack-comment).

packages/@ourworldindata/grapher/src/slopeCharts/SlopeChart.tsx
```diff
@@ -1004,26 +1043,63 @@ class LabelledSlopes
...
+                            const labelBox = s.rightLabelBounds.toProps()
+                            // points define a "strike-through" line that stretches from
+                            // the end point of the slopes to the right side of the right label
+                            const y = labelBox.y + labelBox.height / 2
+                            p1 = new PointVector(endX, y)
+                            p2 = new PointVector(labelBox.x + labelBox.width, y)
+                        }
+
+                        // calculate the distance to the slope or label
+                        const dist = PointVector.distanceFromPointToLineSegment(
```
marcelgerber, line 1086: I'm almost inclined to keep using `distanceFromPointToLineSegmentSq` (emphasis on `Sq`) here, since it's gonna be _a bit_ more performant by not running `Math.sqrt` hundreds of times, and we don't care about the actual distances, just the smallest one (and they are all positive).

Let me know what you think.

If we were to change this, then we would also need to change the tolerance below accordingly, i.e. to `10**2` and `20**2`, respectively.

marcelgerber (approved):
Nice work! I left two comments, but otherwise this is great and a definite improvement. 🙌🏻

packages/@ourworldindata/grapher/src/slopeCharts/SlopeChart.tsx
```diff
@@ -401,10 +405,41 @@ export class SlopeChart
         return sizeByEntity
     }
 
+    // click anywhere inside the Grapher frame to dismiss the current selection
+    @action.bound onGrapherClick(e: Event): void {
+        const tagName = (e.target as HTMLElement).tagName
+        const isTargetInteractive = ["A", "BUTTON", "INPUT"].includes(tagName)
```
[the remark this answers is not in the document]
sophiamersmann: Nice catch!

packages/@ourworldindata/grapher/src/slopeCharts/SlopeChart.tsx
sophiamersmann: yes, that sounds reasonable. I switched to the non-squared version because I wanted an easy-to-think-about tolerance value but didn't think about perf. let's use the squared version!

marcelgerber (approved):
Nice!

PR #3259, opened 2024-02-28
🔨 migrate PostLink to knex

danyx23:
Migrates the access to the post_links table from TypeORM to knex

PR #3258, opened 2024-02-28
💄 reorder parameters of knexRaw and knexRawFirst to have knex in first pos

danyx23:

Issue #3159, opened 2024-02-01
MySQL deadlocks caused by chart approval tool


Marigold:
When someone uses chart approval tool, admin tends to die on [MySQL deadlocks](https://app.bugsnag.com/our-world-in-data/grapher-admin/errors/647600454a89c90008b6d751?event_id=65bb5ddb00d2871af6920000&i=sk&m=oc). We turned on slow query log (`ssh owid@owid-live-db 'sudo cat /var/log/mysql/slow-query.log'`) and the output around the downtime is below
```
# Time: 2024-02-01T09:01:14.312549Z
# User@Host: live_grapher[live_grapher] @  [209.97.185.49]  Id: 8754319
# Query_time: 120.505255  Lock_time: 120.504254 Rows_sent: 0  Rows_examined: 0
SET timestamp=1706777953;
UPDATE `users` SET `lastSeen` = '2024-02-01 08:59:13.798' WHERE `id` IN (62);
# Time: 2024-02-01T09:01:14.312549Z
# User@Host: live_grapher[live_grapher] @  [209.97.185.49]  Id: 8754204
# Query_time: 120.239866  Lock_time: 120.238823 Rows_sent: 0  Rows_examined: 0
SET timestamp=1706777954;
UPDATE `users` SET `lastSeen` = '2024-02-01 08:59:14.064' WHERE `id` IN (62);
# Time: 2024-02-01T09:01:14.312680Z
# User@Host: live_grapher[live_grapher] @  [209.97.185.49]  Id: 8754207
# Query_time: 120.016265  Lock_time: 120.015208 Rows_sent: 0  Rows_examined: 0
SET timestamp=1706777954;
UPDATE `users` SET `lastSeen` = '2024-02-01 08:59:14.203' WHERE `id` IN (62);
# Time: 2024-02-01T09:01:14.312709Z
# User@Host: live_grapher[live_grapher] @  [209.97.185.49]  Id: 8754203
# Query_time: 120.532144  Lock_time: 120.530995 Rows_sent: 0  Rows_examined: 0
SET timestamp=1706777953;
UPDATE `users` SET `lastSeen` = '2024-02-01 08:59:13.755' WHERE `id` IN (62);
# Time: 2024-02-01T09:01:14.312602Z
# User@Host: live_grapher[live_grapher] @  [209.97.185.49]  Id: 8754168
# Query_time: 120.585468  Lock_time: 120.584527 Rows_sent: 0  Rows_examined: 0
[...]

PR #3271, opened 2024-02-29
✨ add transaction types and api route helpers

danyx23:
This PR adds custom types for read only transactions and read/write transactions. Through some typescript magic this is set up so that functions that need a read only transaction (because they only run `select` statements) can also be given read/write transactions but functions that need read/write transactions do not accept a readonly transaction. 

This PR also adds helper functions for API request handlers that also set up a transaction (which means that instead of writing a callback that takes a (request, response) tuple you write one that works with a (request, reqsponse, transaction) tuple).

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -51,6 +51,8 @@ module.exports = {
         "react/no-render-return-value": "warn",
         "react/no-unescaped-entities": ["warn", { forbid: [">", "}"] }],
         "react/prop-types": "warn",
+        // TODO: consider adding this and whitelisting all promises that don't need to be awaited with "void"
+        // "@typescript-eslint/no-floating-promises": "error",
     },
     settings: {
         "import/resolver": {
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -14,7 +14,7 @@ import {
 } from "../settings/serverSettings.js"
 import { expectInt, isValidSlug } from "../serverUtils/serverUtil.js"
 import { OldChart, Chart, getGrapherById } from "../db/model/Chart.js"
-import { Request, Response, CurrentUser } from "./authentication.js"
+import { Request, CurrentUser } from "./authentication.js"
 import {
     getMergedGrapherConfigForVariable,
     fetchS3MetadataByPath,
@@ -36,7 +36,6 @@ import {
     OwidGdocPostInterface,
     parseIntOrUndefined,
     parseToOperation,
-    DbEnrichedPost,
     DbRawPostWithGdocPublishStatus,
     SuggestedChartRevisionStatus,
     variableAnnotationAllowedColumnNamesAndTypes,
@@ -60,7 +59,9 @@ import {
     DbRawVariable,
     DbRawOrigin,
     parseOriginsRow,
-    AnalyticsPageviewsTableName,
+    PostsTableName,
+    DbRawPost,
+    DbPlainDataset,
 } from "@ourworldindata/types"
 import {
     getVariableDataRoute,
@@ -100,7 +101,12 @@ import { Link } from "../db/model/Link.js"
 import { In } from "typeorm"
 import { logErrorAndMaybeSendToBugsnag } from "../serverUtils/errorLog.js"
 import { GdocFactory } from "../db/model/Gdoc/GdocFactory.js"
-import { Knex } from "knex"
+import {
+    getRouteWithROTransaction,
+    deleteRouteWithRWTransaction,
+    putRouteWithRWTransaction,
+    postRouteWithRWTransaction,
+} from "./routerHelpers.js"
 
 const apiRouter = new FunctionalRouter()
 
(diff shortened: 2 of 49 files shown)


danyx23:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3271?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3271** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3271?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3259** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3258** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3271?utm_source=stack-comment).

.eslintrc.cjs
```diff
@@ -51,6 +51,8 @@ module.exports = {
         "react/no-render-return-value": "warn",
         "react/no-unescaped-entities": ["warn", { forbid: [">", "}"] }],
         "react/prop-types": "warn",
+        // TODO: consider adding this and whitelisting all promises that don't need to be awaited with "void"
+        // "@typescript-eslint/no-floating-promises": "error",
```
danyx23, line 55: I think we should do this once the whole stack is converted to knex. This will be a bit annoying because there are about 250 promises in our codebase that are not explicitly awaited, many of them totally fine (e.g. react event handlers that don't await the result of the handler); but now that we switch to transactions on all api router handlers, failing to await a promise could potentially try to access the database transaction after it has been commited or rolled back. Cases where this happens were an error before already (e.g. errors in these promises would not be reported back) and could be related to #3159 and other similar issues.

adminSiteServer/mockSiteRouter.tsx
marcelgerber, line 1: I'm wondering why you're not using `routerHelpers` in here, and in adminRouter?

adminSiteServer/apiRouter.ts
```diff
@@ -497,15 +495,13 @@ apiRouter.get(
...
+apiRouter.get("/topics.json", async (req, res) => ({
     topics: await DEPRECATEDgetTopics(),
 }))
-apiRouter.get(
-    "/editorData/variables.json",
-    async (req: Request, res: Response) => {
-        const datasets = []
-        const rows = await db.queryMysql(
-            `SELECT
+apiRouter.get("/editorData/variables.json", async (req, res) => {
```
marcelgerber, line 501: This one should be relatively easy to convert to knex, if you want to tackle that one.

adminSiteServer/apiRouter.ts
```diff
@@ -1967,14 +1946,14 @@ apiRouter.post(
 )
 
 // Get a list of redirects that map old slugs to charts
-apiRouter.get("/redirects.json", async (req: Request, res: Response) => ({
+apiRouter.get("/redirects.json", async (req, res) => ({
```
marcelgerber, line 1949: Also super easy to convert to trx style; but then I also don't really get how you're choosing which routes to convert and which are staying unchanged.

marcelgerber (approved):
Nice, hard work!

What I did as part of this review:
- Went over (pretty much) everything and checked that it makes rough sense
- For `apiRouter` in particular, checked that we never use a transaction with the wrong kind of permissions
- As a sanity check, I made the RO and RW types incompatible to see all the places where a RW transaction is passed to a method requiring a RO transaction - and all of these places make sense, i.e. the caller actually writes something.

It would be great if we could somehow also remove the `.insert`, `.update` etc. methods from the RO transaction altogether, but I don't believe this to be possible given the query builder nature of knex. A man can dream...

adminSiteServer/mockSiteRouter.tsx
danyx23: I wrote those helpers for our FunctionalRouter wrapper type which is a new class type that wraps the express Router type. We could maybe unify this but I thought the bulk use is with the FunctionalRouter in apiRouter and so I just handrolled the stuff in mockSiteRouter for now.

db/tests/basic.test.ts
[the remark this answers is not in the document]
danyx23: Nice idea, added this

adminSiteServer/apiRouter.ts
danyx23: Jup, true, added this

adminSiteServer/apiRouter.ts
danyx23: I started with roughly one entity type per PR and then upgraded the routes to the new style that were related to that entity. Over the course of the full stack most routes are converted and then the remaining (except for the ones related to the chart revision tool) I'll do in a cleanup PR at the top of the stack

danyx23:
### Merge activity

* **Mar 26, 10:04 AM EDT**: @danyx23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3271).

PR #3280, opened 2024-03-01
🎉 use ts-pattern for OwidGdocPage attributes

ikesau:
Fixes prod's homepage's canonical URL being https://ourworldindata.org/owid-homepage at the moment.

I think `ts-pattern` saves us from overlooking lots of things and that in general we should use it for most Gdoc attributes so that we don't miss anything whenever we add additional gdoc templates.


danyx23 (approved):
Looks good, shipping this!

PR #3317, opened 2024-03-08
✨ (static chart) remove link underlines

### LINKED ISSUE #3295: Static charts: Remove link underlines (if possible, only for the PNG version)
## Description

- Underlining links in static charts is misleading because, in most environments, the links aren't actually clickable
    - One exception are static SVGs opened in a browser. In that case, the links are clickable

## Approach

Ideally, we want to keep underlines for SVGs, but remove them for PNGs.

- Try to manipulate the SVG string in the rasterizer, as described [here](https://owid.slack.com/archives/C01FB9HLZQQ/p1709661061549529?thread_ts=1709658339.751229&cid=C01FB9HLZQQ)
- Or try to introduce PNG-specific behaviour when rendering by making `isExportingToSvgOrPng` into an enum, as described [here](https://owid.slack.com/archives/C01FB9HLZQQ/p1709659528460439?thread_ts=1709658339.751229&cid=C01FB9HLZQQ)
- If it turns out to be complicated, it's ok to remove underlines for all static charts including SVGs

## Additional context

[Slack discussion](https://owid.slack.com/archives/C01FB9HLZQQ/p1709658339751229)

sophiamersmann:
Resolves #3295 

I decided to remove link underlines for all static charts, including SVGs.

- Clicking on links within a static chart SVG on our website is not possible since we usually include the SVG via an `` tag (aside, we also often wrap the static chart image itself in a link, which means the whole image is clickable)
- Thus, link underlines for static charts used on our website are never relevant
- They're only relevant when someone downloads the SVG version of a chart and then looks at it in their browser
    - But if they do anything else with the chart, e.g. include it in a doc or presentation slides, share it on social media, etc., again, the link underlines would be misleading since they're not clickable in most environments

All in all, I don't think it's worth the complication of having a PNG-specific version. So, I decided to remove them from all static charts.

### DIFF
--- a/packages/@ourworldindata/grapher/src/footer/Footer.tsx
+++ b/packages/@ourworldindata/grapher/src/footer/Footer.tsx
@@ -675,10 +675,7 @@ export class StaticFooter extends Footer<StaticFooterProps> {
     @computed protected get licenseAndOriginUrlText(): string {
         const { finalUrl, finalUrlText, licenseText, licenseUrl, textColor } =
             this
-        const textDecoration = this.manager.isStaticAndSmall
-            ? "none"
-            : "underline"
-        const linkStyle = `fill: ${textColor};  text-decoration: ${textDecoration};`
+        const linkStyle = `fill: ${textColor};`
         const licenseSvg = `<a target="_blank" style="${linkStyle}" href="${licenseUrl}">${licenseText}</a>`
         if (!finalUrlText) return licenseSvg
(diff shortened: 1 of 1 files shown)


sophiamersmann:
* **#3317** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3317?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

marcelgerber (approved):
Agreed, this reasoning makes sense to me.

PR #3323, opened 2024-03-08
💄 (grapher) remove isExplorable option

sophiamersmann:
> [!WARNING]  
> Db tests are failing.

Just a little bit of cleanup. Removes the `isExplorable` property from Grapher's config and the `isExplorable` column from the `charts` table.

The config schema states that the `isExplorable` property is not in use and should be removed. I then noticed that the `charts` table has a column called `isExplorable` that also isn't in use, so I decided to remove it too.

Technically, this is a breaking schema change, but there is no real code change, so I guess it's okay not to increase the schema version etc.

### DIFF
--- a/adminSiteClient/ChartList.tsx
+++ b/adminSiteClient/ChartList.tsx
@@ -26,7 +26,6 @@ export interface ChartListItem {
     lastEditedBy: string
     publishedAt: string
     publishedBy: string
-    isExplorable: boolean
 
     tags: DbChartTagJoin[]
 }
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -273,16 +273,15 @@ const saveGrapher = async (
     const now = new Date()
     let chartId = existingConfig && existingConfig.id
     const newJsonConfig = JSON.stringify(newConfig)
-    // todo: drop "isExplorable"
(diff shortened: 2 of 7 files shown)


sophiamersmann:
* **#3323** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3323?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

marcelgerber (approved):
Nice one!

For reference, I'm pretty sure that this property was used [for the circa-2019 old explorer](https://explorer-owid.netlify.app/explore), and is therefore absolutely safe to remove at this point.

sophiamersmann:
Ah, that's good to know! It's so cool that the link to the everything explorer is still working :)

Do you have any idea why the db test might be failing? It doesn't make sense to me...

marcelgerber:
Ah, sadly the DB tests are still failing.

Let's summon @danyx23!

danyx23:
A wild Daniel appears!

Ok this is one of these cases where, after you fix it, you really don't understand how this can ever have worked before. I don't get it. Whatever, the issue is fixed now! 🎉 ... 😬

sophiamersmann:
haha, thanks a lot for looking into this!

Issue #2629, opened 2023-09-11
Long StackedDiscreteBar labels should wrap


samizdatco:
## Core problem

Long labels will compress the amount of space left over for the graph itself (see [slack discussion](https://owid.slack.com/archives/C5BDCB2R3/p1694440168482789)):

![image](https://github.com/owid/owid-grapher/assets/469523/e05d4765-4454-4c5a-a81f-740f9925987f)

## Proposed solution

A max-width cutoff of, say 25%, could be chosen and any labels wider than that could be broken into multiple lines when rendered by TextWrap. The cutoffs could also use `px` sizes with breakpoints detected using the new `isSmall`, `isMedium`, and `isLarge` grapher attributes.

## Alternatives

We could leave line-breaking for authors to control manually and parse strings for a `<br>`, `\n`, or other newline symbol.

----

## Progess

- https://github.com/owid/owid-grapher/pull/3341
- (to do) label wrapping for `DiscreteBar` charts

PR #3341, opened 2024-03-13
🐛 (stacked discrete bar) wrap long labels

sophiamersmann:
Related: #2629 

We should also do this for DiscreteBar charts, but I don't have time now. I added a note to the issue above 👆🏻 

| Before  | After  |
| ------- | ------ |
|  Screenshot 2024-03-13 at 20 42 53 | Screenshot 2024-03-13 at 20 42 41 |
| Screenshot 2024-03-13 at 20 42 07 | Screenshot 2024-03-13 at 20 42 22 |

### DIFF
--- a/packages/@ourworldindata/grapher/src/stackedCharts/StackedDiscreteBarChart.tsx
+++ b/packages/@ourworldindata/grapher/src/stackedCharts/StackedDiscreteBarChart.tsx
@@ -62,6 +62,7 @@ import { easeQuadOut } from "d3-ease"
 import { bind } from "decko"
 import { CategoricalColorAssigner } from "../color/CategoricalColorAssigner.js"
 import { getElementWithHalo } from "../scatterCharts/Halos.js"
+import { TextWrap } from "@ourworldindata/components"
 
 const labelToBarPadding = 5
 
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#3341** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3341?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/stackedCharts/StackedDiscreteBarChart.tsx
```diff
@@ -364,12 +390,18 @@ export class StackedDiscreteBarChart
         }))
     }
 
+    // useful if `this.barHeight` can't be used due to a cyclic dependency
+    // keep in mind though that this is not exactly the same as `this.barHeight`
+    @computed private get approximateBarHeight(): number {
+        return (0.8 * this.boundsWithoutLegend.height) / this.barCount
+    }
```
sophiamersmann, line 397: There is a circular dependency here that I can't get rid of easily. If I remember correctly, it's something like: `labelFontSize` -> `barHeight` -> `innerBounds` -> `labelWidth` -> `labelStyle` -> `labelFontSize`.

Breaking the cycle by using `boundsWithoutLegend` instead of `innerBounds` is not new, but I added a computed property that makes it explicit.

marcelgerber (approved):
Nice, this looks great!

It would be awesome if we had some way to have more balanced lines (i.e. that each line is roughly the same length), but I realize that that's not easy to do. And this is already soo much better than what it was before.

PR #3420, opened 2024-03-27
Search events in GTM

ikesau:
Adds 3 new events:
- `owid.site_search_click`
  - Tracks any click on the search results page, both with its position in the section and the global position
- `owid.site_search_filter_click`
  - Tracks when a filter tab is clicked on (or "Show all results" button)
- `owid.site_instantsearch_click`
  - Tracks when an instantsearch result is clicked

These will all get imported into BQ per the `owid\..*` RegEx in GTM:
image

The results click events track the query, so we can see how each query currently performs.

If we wanted to start analyzing specific searches, we could look at ingesting `page_view` events from GTM as they track the history pushes that algolia makes.

## Demo
https://github.com/owid/owid-grapher/assets/11844404/47c2dbf9-3b75-4cc3-9f68-b24df5b0e267


site/search/SearchPanel.tsx
```diff
@@ -281,22 +285,48 @@ const SearchResults = (props: SearchResultsProps) => {
...
+                    // starts from 1 at the top of the page
+                    const globalPosition = allVisibleHits.indexOf(target) + 1
+                    // starts from 1 in each section
+                    const positionInSection = target.getAttribute(
                         "data-algolia-position"
                     )
                     const index = target.getAttribute("data-algolia-index")
-                    if (objectId && position && index) {
+                    const href = target.getAttribute("href")
+                    const query = getWindowQueryParams().q
```
marcelgerber, line 303: Might be cleaner to pass the query (`this.inputValue`) in as a prop to `SearchResults`.

site/search/SearchPanel.tsx
```diff
@@ -281,22 +285,48 @@ const SearchResults = (props: SearchResultsProps) => {
...
+
+                    if (
+                        objectId &&
+                        positionInSection &&
+                        index &&
+                        href &&
+                        query
+                    ) {
                         logSiteSearchClick({
```
marcelgerber, line 312: Maybe rename:
```suggestion
                        logSiteSearchClickToAlgolia({
```

packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -1831,3 +1831,14 @@ export function cartesian<T>(matrix: T[][]): T[][] {
         [[]]
     )
 }
+
+export function isElementHidden(element: Element | null): boolean {
+    if (!element) return false
+    const computedStyle = window.getComputedStyle(element)
```
marcelgerber, line 1837: Does `getComputedStyle` also take into account whether parent elements are hidden?

marcelgerber (approved):
Nice work!

packages/@ourworldindata/utils/src/Util.ts
ikesau: image

Nope, hence the recursive to the element's parent

marcelgerber: Oh dang, yeah I totally didn't see that call

PR #3445, opened 2024-04-04
🐛 (grapher) don't validate dods

### LINKED ISSUE #3439: Datapage grapher exports are missing DoDs
1. Load https://ourworldindata.org/grapher/life-expectancy
2. Click "Download"
3. See if "Include terminology definitions at bottom of chart" checkbox is present

Some percentage of the time, it's not. It seems to be related to a race condition occurring between the loading of dods.json and the grapher metadata. In firefox, if the Transferred status is "raced":

raced

Possibly also related to https://github.com/owid/owid-grapher/issues/3088

sophiamersmann:
- Fixes #3439
- We recently started validating DoDs in Grapher
- We need validation when we automatically check for DoDs with a specific prefix, like `grapher_indicator_my-indicator-title` (we never shipped automatic DoD checking though)
- `window.details` must be populated before Grapher loads for validation to work (if it's not there, Grapher can't validate and assumes all given details are invalid)
- Since we never shipped automatic DoD checking, we also don't currently need validation and can just remove the code around it
- Let's just do so for now and figure out how to do bug-free details validation in Grapher when we really need it
- This is "tested" by refreshing a Grapher page a number of times and verifying that the DoD checkbox in the Download modal is present

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -1273,12 +1273,7 @@ export class Grapher
             ...noteDetails,
         ])
 
-        const validDetails = uniqueDetails.filter((detailId: string) => {
-            const detail = window.details?.[detailId]
-            return detail !== undefined
-        })
-
-        return validDetails
+        return uniqueDetails
     }
 
     @computed get detailsMarkerInSvg(): DetailsMarker {
--- a/site/Head.tsx
+++ b/site/Head.tsx
@@ -1,9 +1,6 @@
 import React from "react"
 import { viteAssetsForSite } from "./viteUtils.js"
-import {
-    GOOGLE_TAG_MANAGER_ID,
-    BAKED_BASE_URL,
-} from "../settings/clientSettings.js"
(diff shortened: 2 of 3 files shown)


sophiamersmann:
* **#3445** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3445?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

ikesau (approved):
I'm not quite sure what you mean by "automatic DoD checking" but I can confirm this fixes the bug and will be easy to restore if/when we get around to whatever is meant by it 😛

sophiamersmann:
Sorry, that's probably data-viz language. We toyed with the idea of automatically checking for DoDs with a specific prefix for some Grapher UI strings. For example, if we wanted DoDs for region definitions, then instead of specifying a DoD for "Africa (UNICEF)" in every chart that uses UNICEF regions, we could instead automatically check for a DoD with the name `grapher_region_africa-unicef`. That's what I mean when I say "automatic DoD checking". Hope that makes more sense!

Thanks for the review!

PR #3453, opened 2024-04-05
🐛 use `utf8mb4` for knex connections

ikesau:
The live grapher database's charset is `utf8mb4`, but the default connection charset for knex is `utf8mb3` (utf8, multibyte 3) which is [deprecated](https://dev.mysql.com/blog-archive/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/) and was causing the emoji in our dods to render as `?` characters.

This PR fixes that.

Further reading: [a great blog post on unicode](https://tonsky.me/blog/unicode/)

### DIFF
--- a/db/db.ts
+++ b/db/db.ts
@@ -35,6 +35,7 @@ export const knexInstance = (): Knex<any, any[]> => {
             password: GRAPHER_DB_PASS,
             database: GRAPHER_DB_NAME,
             port: GRAPHER_DB_PORT,
+            charset: "utf8mb4",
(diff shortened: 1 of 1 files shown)


danyx23:
Urgs - why oh why MySQL, do you have to specify this on the connection as well ?! 😭 

Please merge this.

Could any of the content in the db now have been written as utfmb3? I hope not?

ikesau:
@danyx23 I don't think so 🤞 

The emoji were being correctly saved into the DB, it was only when retrieving them that they were collated

PR #3465, opened 2024-04-09
Always add tab param to embed URL

### LINKED ISSUE #2805: Always set `tab` as query param in embed code
## Description

- The embed code we offer to users (see screenshot below) sets the iframe's source to the current URL
- But the current URL relies on defaults
    - For example, when the chart is configured to be the default view, then the canonical url without query params shows the chart; if the map is configured to be the default view, it shows the map
- As a result, if an author changes the config of a chart, embeds on foreign sites might change unexpectedly; for example, we encountered an issue where Jason embedded chart views, that silently changed into map views when authors changed the config

## Expected behaviour

~Don't rely on defaults for query params when constructing an embed code.~

Always specify the `tab` query param in the URL that is included in the embed code.

## Screenshots

Screenshot 2023-10-19 at 14 53 58

rakyi:
Fixes #2805.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.jsdom.test.ts
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.jsdom.test.ts
@@ -465,6 +465,16 @@ describe("urls", () => {
         expect(grapher.canonicalUrl?.includes("country")).toBeFalsy()
     })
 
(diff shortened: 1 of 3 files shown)


sophiamersmann (approved):
Looks great! Thanks a lot :)

Feel free to 🚢 it

PR #3452, opened 2024-04-05
🎉 (entity selector) highlight user location

sophiamersmann:
[Cycle 2024.2: Entity selector](https://github.com/owid/owid-grapher/issues/3349) | [Designs](https://www.figma.com/file/X5mOEX8zULS6qyHocUYdmh/Grapher-UI?type=design&node-id=2523%3A6266&mode=design&t=7edFp79OOjz6RENz-1)

## Summary

Detects the user's location and displays it at the top of the list. Also shows a little icon next to the user's country/region with a tooltip rendered on hover.

## Details

- The user's location is displayed at the top if it's not currently selected (if it's already selected, then I don't think there is a need to move it to the top)

## SVG tester

The SVG tester fails due to the changes in #3373

PR #3447, opened 2024-04-04
🎉 (entity selector) sort by name or value

sophiamersmann:
[Cycle 2024.2: Entity selector](https://github.com/owid/owid-grapher/issues/3349) | [Designs](https://www.figma.com/file/X5mOEX8zULS6qyHocUYdmh/Grapher-UI?type=design&node-id=2523%3A6266&mode=design&t=7edFp79OOjz6RENz-1)

## Summary

Adds a dropdown menu to sort by name or indicator, as well as a button to toggle the sort order. Draws a simple bar chart when sorted by an indicator.

## Details

- Indicators considered for sorting are all indicators of the chart that can be formatted numerically (includes `NumberOrStringColumn`) and that are not projections
- Names are sorted asc by default, while numerical values are sorted desc by default
- If sorted by an indicator, then values without data are listed at the end (and are marked with a "No data" label)

## Caveats

- The (display) name of a column that is used as dropdown option is not always nice :( (but we surface it in other places already..)
- We can't draw a bar chart for negative values. For now, we opted to simply not draw anything in that case

## SVG tester

The SVG tester fails due to the changes in #3373 

## Data

- [x] Go through the top 50 charts and make sure all columns have appropriate display names

PR #3443, opened 2024-04-03
✨ (grapher) improve sizing on grapher pages

sophiamersmann:
- In theory, it would be nice if the chart on Grapher pages would be aligned with the header (see screenshot below)
- This PR attempted to do this but made the already-very-complicated sizing logic on Grapher pages even more complicated
- So, I ultimately decided that it wasn't worth it

PR #3436, opened 2024-04-02
✨ (entity selector) redesign

sophiamersmann:
[Cycle 2024.2: Entity selector](https://github.com/owid/owid-grapher/issues/3349) | [Designs](https://www.figma.com/file/X5mOEX8zULS6qyHocUYdmh/Grapher-UI?type=design&node-id=2523%3A6266&mode=design&t=7edFp79OOjz6RENz-1)

## Summary

Visual overhaul of Grapher's entity selector (but no bigger new features added).

## Known issues

I addressed most of the feedback that I got so far, but I am still figuring these two things out:
- There is a tiny gap above the search bar on mobile
- The animation sometimes feels too fast, esp. for entities that are at the very bottom of a long list

## SVG tester

The SVG tester fails due to the changes in #3373

PR #3466, opened 2024-04-09
🎉 (entity selector) sort by external indicators

sophiamersmann:


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3466?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3466** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3443** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3443?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3459](https://github.com/owid/owid-grapher/pull/3459) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3459?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3452** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3452?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3447** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3436** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3435** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3434** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3434?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3373** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3373?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3353** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3353?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#3356](https://github.com/owid/owid-grapher/pull/3356) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3356?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3363](https://github.com/owid/owid-grapher/pull/3363) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3363?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -216,20 +235,121 @@ export class EntitySelector extends React.Component<{
...
+        ])
+    }
+
+    @computed private get populationColumnUsedInChart():
+        | CoreColumn
+        | undefined {
+        const activePopulationColumn = this.numericalChartColumns.find(
+            (column) =>
+                isPopulationVariableId(column.slug) ||
+                !!makeColumnLabel(column).match(/\bpopulation\b/i)
```
marcelgerber, line 299: Hm, this regex will also match "Population density", "Female population", "Share of population in poverty", etc.

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -216,20 +235,121 @@ export class EntitySelector extends React.Component<{
...
+        return this.populationIndicatorId !== undefined
+            ? this.populationIndicatorId.toString()
+            : undefined
+    }
+
+    @computed private get gdpPerCapitaColumnUsedInChart():
+        | CoreColumn
+        | undefined {
+        const activeGdpPerCapitaColumn = this.numericalChartColumns.find(
+            (column) => !!makeColumnLabel(column).match(/\bgdp per capita\b/i)
```
marcelgerber, line 329: Similar here: This'll also match any variable that contains `% of GDP per capita`, for example.

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
marcelgerber, line 1: What I don't quite like here is how set-in-stone these two variables are: There are many methods that have "population" or "gdp" in their name, and all of that makes it quite hard to add any other such indicators in the future.

More ideally, we would have a more configurable set; something like
```ts
const INDICATORS_TO_LOAD = [{
    key: "population",
    name: "Population",
    indicatorId: POPULATION_INDICATOR_ID_USED_IN_ENTITY_SELECTOR,
    columnNameRegex: /\bpopulation\b/i
}, // ...
]
```
could work.

marcelgerber (approved):
Nice, this works well!

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
sophiamersmann: For population, I think it is best to rely on `isPopulationVariableId` only and not use a regex since we'd get many spurious matches ([Datasette](https://datasette-public.owid.io/owid?sql=select%0D%0A++distinct+v.name,%0D%0A++v.id%0D%0Afrom%0D%0A++chart_dimensions+cd%0D%0A++join+variables+v+on+cd.variableId+%3D+v.id%0D%0Awhere%0D%0A++v.name+like+'%25population%25'))

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
sophiamersmann: GDP per capita is different because we have lots of GDP per capita variations like `GDP per capita (constant 2015 US$)` or `Real GDP per capita (2011$)` that are often simply referred to as "GDP per capita" in charts ([Datasette](https://datasette-public.owid.io/owid?sql=select%0D%0A++distinct+v.name,%0D%0A++v.id%0D%0Afrom%0D%0A++chart_dimensions+cd%0D%0A++join+variables+v+on+cd.variableId+%3D+v.id%0D%0Awhere%0D%0A++v.name+like+'%25gdp+per+capita%25'); [example chart](https://ourworldindata.org/grapher/agriculture-value-added-per-worker-vs-gdp-per-capita)). In such cases, using a second, slightly different GDP per capita version in the entity selector would be confusing. So, I think matching the name makes sense here. But I updated the regex to make sure that we're not matching "GDP per capita" if it's in parens, like `Government expenditure per student, primary (% of GDP per capita)`. 

There are some indicators that we'll match incorrectly (like `Growth rate of GDP per capita` or `Affordability of cigarettes: percentage of GDP per capita required to purchase 2000 cigarettes of the most sold brand`), which means that we'll fail to offer "GDP per capita" as a sorting option, which I think is acceptable.

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
marcelgerber: Yes, agreed.
Or otherwise, match only `/Population( \(historical estimates\))?/`, which currently finds exactly these four.

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
marcelgerber: Yes, agreed. Good on you for pointing out that the outcome of a false positive is simply that we're not offering GDP per capita for sorting in that case, which is totally accepting.

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
marcelgerber: Earlier today, I was thinking that we will probably unify the explorer entity selector with the grapher entity selector at some point.
It's a bit weird at the moment that the explorer entity selector is shown on the left, that it is also visible on the map and table tabs, etc.

Explorers currently have their own options for sorting, which can be set in the spreadsheet. Basically, they can be any column of any csv file that's being used in the explorer, including ones that haven't yet been fetched.

CleanShot 2024-04-18 at 12 43 51@2x

I'm not sure what the future of explorers entails, exactly - I think we want to move away a bit from csv's as a data source.
But still, supporting the current explorer feature set would be easier if it was easier to set/extend the set of sort indicators.

marcelgerber: Not something for now, though, just something to keep in mind for the future.

sophiamersmann: Good point! That would def make it easier :)

sophiamersmann:
### Merge activity

* **May 3, 4:04 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3466).

PR #3469, opened 2024-04-10
Bump yarn to 4.1.1

rakyi:


marcelgerber (changes requested):
Hi! We're currently still using a [`yarnPath`](https://yarnpkg.com/configuration/yarnrc#yarnPath)-based Yarn installation.

Yarn nowadays recommends to use [Corepack](https://yarnpkg.com/corepack), which ships with recent versions of Node, but is disabled by default (can be enabled using `corepack enable`). We should probably switch over to Corepack at some point in the near future (and thereby tell the team to enable it), but for now let's stay with `yarnPath`.

For that to work correctly, you'll need to also run `yarn set version latest` in order to update the Yarn release that's then included with the repo. Thank you!

rakyi:
Thanks, I ran the command to update Yarn properly.

rakyi:
After thinking about this more, we don't really need Volta when we are using the `yarnPath` and we also have `.nvmrc` that tools like nvm and fnm use, so I decided to remove the Volta config. I'll keep the PR open for a bit in case someone from the team will object to the removal in Slack.

marcelgerber (approved):
Great thanks!

PR #3502, opened 2024-04-16
Add admin page for site redirects

rakyi:
I got heavily inspired by `RedirectsIndexPage.tsx`. I used `react-hook-form` since I had good experience with it before, and it makes the form handling nice.

Server validation error leads to a generic error popup on the frontend, which is bad UX, but it seems this is what we do also elsewhere. Let me know if there is a good way to avoid that.

### DIFF
--- a/adminSiteClient/SiteRedirectsIndexPage.tsx
+++ b/adminSiteClient/SiteRedirectsIndexPage.tsx
@@ -1,0 +1,184 @@
+import cx from "classnames"
+import React, { useContext, useEffect, useState } from "react"
+import { useForm } from "react-hook-form"
+
+import { AdminAppContext } from "./AdminAppContext.js"
+import { AdminLayout } from "./AdminLayout.js"
+
+const VALID_PATTERN = /^\/$|^.*[^\/]+$/
+const INVALID_PATTERN_MESSAGE =
+    "URL cannot end with a slash unless it's the root."
+
+type Redirect = {
+    id: number
+    source: string
+    target: string
+}
+
+function RedirectRow({
+    redirect,
+    onDelete,
+}: {
+    redirect: Redirect
+    onDelete: (redirect: Redirect) => void
+}) {
+    return (
+        <tr>
+            <td className="text-break">{redirect.source}</td>
+            <td className="text-break">
+                <a href={redirect.target} target="_blank" rel="noopener">
+                    {redirect.target}
+                </a>
+            </td>
+            <td>
+                <button
+                    className="btn btn-danger"
+                    onClick={() => onDelete(redirect)}
+                >
+                    Delete
+                </button>
+            </td>
+        </tr>
+    )
+}
+
+type FormData = {
+    source: string
+    target: string
+}
+
+export default function SiteRedirectsIndexPage() {
+    const { admin } = useContext(AdminAppContext)
+    const [redirects, setRedirects] = useState<Redirect[]>([])
+    const [error, setError] = useState<string | null>(null)
+    const {
+        register,
+        formState: { errors, isSubmitting },
+        handleSubmit,
+        watch,
+    } = useForm<FormData>()
+    const source = watch("source")
+
+    useEffect(() => {
+        admin
+            .getJSON("/api/site-redirects.json")
+            .then((json) => {
+                setRedirects(json.redirects)
+            })
+            .catch(() => {
+                setError("Error loading redirects.")
+            })
+    }, [admin])
+
+    function validateTarget(value: string) {
+        if (value === source) {
+            return "Source and target cannot be the same."
+        }
+        return undefined
+    }
+
+    async function onSubmit(data: FormData) {
+        setError(null)
+        const json = await admin.requestJSON(
+            "/api/site-redirects/new",
+            data,
+            "POST"
+        )
+        if (json.success) {
+            setRedirects([json.redirect, ...redirects])
+        } else {
+            setError("Error creating a redirect.")
+        }
+    }
+
+    async function handleDelete(redirect: Redirect) {
+        if (!window.confirm(`Delete the redirect from ${redirect.source}?`))
+            return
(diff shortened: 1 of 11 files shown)


adminSiteServer/apiRouter.ts
```diff
@@ -1754,6 +1761,61 @@ getRouteWithROTransaction(
...
+    "/site-redirects.json",
+    async (req, res, trx) => ({ redirects: await getRedirects(trx) })
+)
+
+postRouteWithRWTransaction(
+    apiRouter,
+    "/site-redirects/new",
+    async (req: Request, res, trx) => {
+        const { source, target } = req.body
+        if (await redirectWithSourceExists(trx, source)) {
```
rakyi, line 1775: Should we maybe make source unique in the DB?

adminSiteClient/SiteRedirectsIndexPage.tsx
```diff
@@ -0,0 +1,184 @@
+import cx from "classnames"
+import React, { useContext, useEffect, useState } from "react"
+import { useForm } from "react-hook-form"
+
+import { AdminAppContext } from "./AdminAppContext.js"
+import { AdminLayout } from "./AdminLayout.js"
+
+const VALID_PATTERN = /^\/$|^.*[^\/]+$/
```
danyx23, line 8: I would be stricter here and have a different pattern for source and target. Sources have to start with a slash. Targets have to start either with a slash or http://.

I think it is also nice to document regexps a bit more since they are so dense - I tried one form of this e.g. [here](https://github.com/owid/owid-grapher/blob/02bd8ba00448254bb89a70bba110371e9dae68a9/packages/%40ourworldindata/components/src/IndicatorKeyData/IndicatorKeyData.tsx#L117)

adminSiteClient/SiteRedirectsIndexPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+            setError("Error deleting the redirect.")
+        }
+    }
+
+    return (
+        <AdminLayout title="Site Redirects">
+            <main>
+                <form onSubmit={handleSubmit(onSubmit)}>
+                    <div className="form-row">
+                        <div className="form-group col-12 col-md-4">
```
danyx23, line 115: I think we should explain to the admin users a bit what is expected of them here. Yes, our admin UIs are very barebones but our authors are not very versed in the technical details of HTTP redirects so we should explain a bit what is happening here.

I think it is important to tell them that source has to start with a slash and should not contain query params or fragments (you may also want to strip those if they are given). The target can contain full urls or relative urls starting with a slash and here both query params and fragments are allowed. Maybe give an example for query param and fragment in parens as it might not be obvious what these things are.

adminSiteClient/SiteRedirectsIndexPage.tsx
```diff
@@ -0,0 +1,184 @@
...
+                            <label>Source</label>
+                            <input
+                                className={cx("form-control", {
+                                    "is-invalid": errors.source,
+                                })}
+                                {...register("source", {
+                                    required: "Please provide a source.",
+                                    pattern: {
+                                        value: VALID_PATTERN,
```
danyx23, line 124: In addition to checking for a leading slash I think we should also either disallow query params/fragments via the regex or strip them out before sending the redirect to the API.

adminSiteClient/SiteRedirectsIndexPage.tsx
danyx23: It is also worth calling out that to redirect to our homepage they should use a single slash as the target

danyx23 (changes requested):
Really nice work, @rakyi! Maybe I'll warm up to hooks after all :).

I have a few comments, mostly around explaining things a bit more to the user and being a bit more strict in the validation.

db/model/Redirect.ts
```diff
@@ -3,13 +3,68 @@ import * as db from "../db"
...
+            -- Recursive step: Follow the chain of redirects.
+            SELECT r.source, r.target
+            FROM redirects r
+            INNER JOIN redirect_chain rc ON rc.target = r.source
+        )
+         -- Check if the new source appears in the redirect chain starting from
+         -- the new target.
+         SELECT COUNT(*) > 0 AS cycleExists
+         FROM redirect_chain
+         WHERE source = ? OR target = ?`,
```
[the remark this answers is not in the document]
rakyi: I reworked the validation.

adminSiteServer/apiRouter.ts
rakyi: Added.

adminSiteClient/SiteRedirectsIndexPage.tsx
rakyi: Changed. I made the validation messages more specific, which should also serve as docs for the regexps. Let me know if you think it's not sufficient.

adminSiteClient/SiteRedirectsIndexPage.tsx
rakyi: Added the instructions.

rakyi:
@danyx23 do you have an idea why [the refresh-etl job](https://buildkite.com/our-world-in-data/grapher-automated-staging-environment/builds/5075#018ef67b-1dba-4933-9851-2f18ae9bcc8a) would be failing? Rebasing on master didn't help.

danyx23 (approved):
Very nice, thanks for the changes! I had some ideas for slight tweaks and thought it's better to just add them in a commit of my own. Feel free to tweak my styling ideas, I just felt that using concrete examples form our site works better than generic a=b examples. If you are happy with then please ship it! :)

danyx23:
@rakyi No I don't know what is causing the ETL update to fail but it's not relevant for commits in this repo (it is only used when the grapher infrastructure is used as part of PRs in the ETL repository). I think it's good to ship!

PR #3523, opened 2024-04-20
Improve docs/output, helpful for 3rd parties

toni-sharpe:
## Summary

Docker is persistent, as it should be, but, it can be problematic when one switches around projects which was what I did at my previous job. The instructions in the PR should conclusively "factory reset" docker so that OWID has no problem doing it's things.

When I first everything up it was more complicated and I feel this problem was because Docker for Desktop (D4D) doesn't report existing Docker images running via other tools. At my last place, which was a corporate company, we didn't pay the liscense for D4D and were given command line options that worked first time with little thought, whereas [OWID clearly qualifies for free accounts](https://www.docker.com/community/open-source/application/).

`new-session` throws a `duplicate session` error if the TMux screen is closed without quitting, so an added line checks for the session and closes the `grapher` session before `new-session` is run. This means the TMux session can be closed properly with `Ctrl+b, Q` or by closing the terminal window in some other way. If sessions are found, it will only kill off grapher, so if grapher doesn't exist, that is reported but doesn't stop the file.

The MySQL infinite dot output might confuse people a bit. Looking [at the message source](https://github.com/owid/owid-grapher/blob/master/devTools/docker/wait-for-mysql.sh#L18) and finding the `dev/null` destination suggests the result should be immediately be true but isn't for me for some reason. I'm looking at it while I'm here but not currently intending to change anything.

## Testing

~I _think_ all one has to do is stop all the TMux stuff then try and run `make` again. Without the steps to remove _all_ containers I think the same thing will happen, whether those containers are OWID D4D ones or some thing left over on a 3rd party machine.~ No, in fact running `make up` does work, and I think I may have improved it.

For that `make up` change: to test, just remove that commit and see how it works without.

I do have an old contracting laptop which is probably sitting with Dockers ready to pop up again, this would be a true test environment as it will be a closer analogue of my situation as a 3rd party contributor, pre-start. But it's one shot, so I've been [incubating](https://academic.oup.com/sleep/article/41/10/zsy155/5065174) this problem while doing other things.

## Condition of my local set-up 

It works basically, as proven by the PRs I've submitted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ up: require create-if-missing.env ../owid-content tmp-downloads/owid_metadata.sq
 
 	@echo '==> Starting dev environment'
 	@mkdir -p logs
+
+	@if tmux has-session; then \
+		tmux kill-session -t grapher; \
+	fi
 	tmux new-session -s grapher \
 		-n docker 'docker compose -f docker-compose.grapher.yml up' \; \
 			set remain-on-exit on \; \
@@ -77,6 +81,10 @@ up.devcontainer: create-if-missing.env.devcontainer tmp-downloads/owid_metadata.
 
 	@echo '==> Starting dev environment'
 	@mkdir -p logs
+
+	@if tmux has-session; then \
+		tmux kill-session -t grapher; \
+	fi
 	tmux new-session -s grapher \
 		-n admin \
 			'devTools/docker/wait-for-mysql.sh && yarn startAdminDevServer' \; \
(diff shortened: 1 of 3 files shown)


toni-sharpe:
I think the restarting is probably too much and want to test a bit more, but also, fix a few things first, in case I break my local env. 

Really I want a document which describes the simplest example of where I can, at will, destroy and rebuild the basic environment, _as an outside contributor_, so IMO it is best done by me, then reviewed, as I represent the sort of person it will help. 

Some metrics like active state, last update, PR/issue counts, looking at the [forks section in insights](https://github.com/owid/owid-grapher/forks) suggest a possible of loss of interest/contribution unknown reasons. Look and decide/debunk at will! I haven't been highly analytic with it yet.

If I can come up with better instructions that's a good thing though.

So I'm switching it to draft for now and will handle any problems that might occur if the existing file I touched change.

toni-sharpe:
Closing this, I think I was mostly learning stuff and that's led to understanding but no contribution.

The two docker commands for removing everything at the command line I think are handled more specifically by `make down` which is better. An update to the current instructions might be needed but I'll look at that separately.

PR #3541, opened 2024-04-27
Add pre-commit hook

rakyi:


ikesau (approved):

ikesau:
Okay cool, this works (though I think we probably all run prettier on save so I'm not if this'll catch much)

The really nice thing would be if we could catch unused imports before committing, though we should check with the team first because it might be annoying for people who have a more WIP approach to their commits.

PR #3553, opened 2024-04-30
🐛 (stacked discrete bar) remove negative values

### LINKED ISSUE #3464: StackedDiscreteBar: Negative values lead to incorrect bar placement
## Description

- [This chart](https://ourworldindata.org/grapher/food-emissions-supply-chain?country=Apples~Bananas) looks broken because one of the indicators (Land use change) has negative values that stacked discrete bar charts don't support
- This leads to incorrect bar placements

## Expected behaviour

- In the long term, we should probably support negative values for (discrete) stacked bar charts (see #1166)
- For now, we should make sure that negative values don't break the chart (probably by ignoring them)

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/food-emissions-supply-chain?country=Apples~Bananas
2. Note that bars overflow
3. Hover over one of the bars
4. Note that Land use change has a negative value

## Screenshots

Screenshot 2024-04-09 at 09 54 36

sophiamersmann:
fixes #3464 by dropping negative values (excluding zero) in `transformTable`

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreTable.ts
+++ b/packages/@ourworldindata/core-table/src/CoreTable.ts
@@ -1177,6 +1177,14 @@ export class CoreTable<
         )
     }
 
+    replaceNegativeCellsWithErrorValues(columnSlugs: ColumnSlug[] = []): this {
+        return this.replaceCells(columnSlugs, (val) =>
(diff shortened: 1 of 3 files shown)


marcelgerber (approved):
nice!

PR #3589, opened 2024-05-07
🎉 (grapher) add a globe view / TAS-516

sophiamersmann:
Merged from #3057 

To do:
- [x] Choose a default projection (for when we can't detect the user's country); decided to go for an approach that rotates based on day light
- [x] Ike noticed a memory leak – after an hour or so, the page became unresponsive
- ~Add Antarctica to the globe (by adding an entry in `MapTopology.ts`, see https://raw.githubusercontent.com/alexabruck/worldmap-sensitive/master/dist/world.topo.json)~ To be done at another time (after building a pipeline for our topology)
- [x] Always render globe toggle into the Settings popup

Default projection suggestion:
00:00 - 07:59 GMT - Asia & Oceania :earth_asia:
08:00 - 15:59 GMT - Europe & Africa :earth_africa:
16:00 - 23:59 GMT - Americas :earth_americas:

Future work:
- Pinch to zoom (for maps and the globe)


sophiamersmann:
We want to implement pinch-to-zoom before shipping. To be continued...

PR #3595, opened 2024-05-09
New gdoc type announcements

danyx23:
This is a prototype branch for testing the concept of an announcements/update feed


danyx23:
Doesn't look like this is still wanted, closing it.

PR #3606, opened 2024-05-15
Install packages only when relevant files change

rakyi:
`node_modules` is an existing target, that depends on changes to `package.json`,  `yarn.lock` and `yarn.config.cjs` to run `yarn install`.


marcelgerber (approved):
Nice!

mlbrgl (commented):
looks good! That `make.up` is so fast now, wow!

PR #3609, opened 2024-05-15
🐛 fix research-and-writing links still rendering on bake even when content is invalid

ikesau:
The research and writing block had an incorrectly implemented fallback for when linked documents were invalid. 

If there's an error and we're previewing, we show a red warning block. If we're not previewing (i.e. baking), we should return null.

This was noticed because an unpublished gdoc that was being linked to in the [CO₂ topic page](https://ourworldindata.org/co2-and-greenhouse-gas-emissions#research-writing) was showing up as an empty tile.

error


mlbrgl (approved):
Makes sense, thanks!

Confirmed locally (dev and bake)

Screenshot 2024-05-16 at 19 15 25
Screenshot 2024-05-16 at 21 29 30

PR #3610, opened 2024-05-16
Run migrations with TS sources

rakyi:


ikesau:
Nice! Could you update the Makefile migrate command as well? I think it just needs to be `yarn && yarn runDbMigrations`

PR #3615, opened 2024-05-17
Make image description optional

### LINKED ISSUE #3201: Make image description/alt text for static visualizations non-mandatory
## Context

It is currently mandatory to set image descriptions to static images stored in GDocs before they can be added and published in an article. This is done to ensure that all images have alt-text.

## Problem

Authors commonly iterate on static images, and re-adding the alt text each time can be quite painful for them, as reported by Max.

## Impact

We know that having the alt text mandatory adds friction to the authoring flow. We have mixed intuitions about the costs/benefits of the alt text, we would love to experiment with a screen reader to see how valuable the alt text is.

## Proposed solution

1. Image descriptions/alt text should not be mandatory – an article should be publishable without one. As a middle ground, a warning should be raised in the settings draw when no text is provided.

We expect that it will then become more common for authors to make use of the override within the archie block itself. i.e. 

```
{.image}
filename: default-featured-image.png
alt: HERE
{}
```

2. In particular, get rid of this requirement for featured images. (It's hard to see the value of alt text for featured images – they aren't really doing anything than advertising the article, often in quite an abstract way.)

## Context

The problem was raised by Max, and the solution was proposed within a discussed between Lars, Max, Daniel and Joe.

## Technical notes

- Featured images are presentational, they shouldn't need alt text
  - We can use a special keyword in the alt text to mark them as skipped
- Static visualisations should have alt text or a warning

rakyi:
Joe suggested not requiring `alt` on featured images at all, but not sure if: 

* it's worth the added complexity
* they are always only presentational

Long term, we might switch to auto-generating the description as you suggested before, but that is a much bigger project we didn't scope out, yet.

Fixes #3201

### DIFF
--- a/db/model/Gdoc/GdocBase.ts
+++ b/db/model/Gdoc/GdocBase.ts
@@ -708,7 +708,7 @@ export class GdocBase implements OwidGdocBaseInterface {
                     errors.push({
                         property: "imageMetadata",
                         message: `${filename} is missing a default alt text`,
-                        type: OwidGdocErrorMessageType.Error,
+                        type: OwidGdocErrorMessageType.Warning,
                     })
                 }
                 return errors


ikesau (approved):
Let's see how we go 🙂

PR #3619, opened 2024-05-20
More commas on homepage chart count, `hasOutline` on archie images

ikesau:
## Changes

- Allows authors to add a grey border around images if desired (See [slack](https://owid.slack.com/archives/C05JT7FUVRS/p1712746010518449))

image

- Adds [commafication](https://www.youtube.com/watch?v=YlUKcNNmywk) to another place in the homepage

image
image


packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -1883,6 +1883,13 @@ export function roundDownToNearestHundred(value: number): number {
     return Math.floor(value / 100) * 100
 }
 
+/**
+ * Example: 120000 -> "12,000"
+ */
+export function commafyNumber(value: number): string {
```
rakyi, line 1889: It might be worth to:

* Name this `formatNumber` or `formatQuantity` to be less specific, we might decide to change the format in the future
* Use the `Intl.NumberFormat` API if we are gonna keep reusing this and define the formatter at the top scope to avoid initializing it every time we call the function, see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString):

  > Every time toLocaleString is called, it has to perform a search in a big database of localization strings, which is potentially inefficient. When the method is called many times with the same arguments, it is better to create a [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) object and use its [format()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format) method

rakyi (approved):
LGTM, except one small comment and the build errors.

PR #3636, opened 2024-05-22
garden: Improve TMux/make.up commands

toni-sharpe:
* if the TMux window is closed in some other way than Ctrl+b, Q then `make up` will error, throwing duplicate session - this change checks for a session and removes the grapher session if it's found before going back to where the contributor was.
* This also happens if `master` is updated on a fork and `make up` is re-run to bring things up to date.

Thinking from the perspective of an open source contributor, perhaps with less experience, and wanting to make things like TMux/Docker pain free. Some input from those with better code knowledge might see issues with the change, maybe I missed an instruction or somesuch. I am also a Sublime Text user, as it happens, I note some instructions are VSCode-centric.

### Current master

The contributor gets stuck with a `duplicate session` error when things are broken by something (some error, closing and re-opening tab). DBeaver showing things are broken. They can fix things with some investigation or they could be time constrained/inexperienced/imposter syndrome/etc. and leave.

https://github.com/owid/owid-grapher/assets/10499070/be1e0055-3036-4047-a2a5-3a7a72542712

Screenshot 2024-05-23 at 00 24 41

### With change

The user just needs to run `make up` twice rather than getting stuck with a duplicate session. DBeaver showing the DB. Things like errors in the UI about TS disappear, and UI refreshes.

https://github.com/owid/owid-grapher/assets/10499070/1e81f99f-9422-4b84-aef3-4ba065412f25

Screenshot 2024-05-23 at 00 40 18

### Closing TMux terminal window (likely) with PR change applied

https://github.com/owid/owid-grapher/assets/10499070/eec37dac-94e7-4bd6-b4a2-b6e843c47418

Also sees DB and UI refresh nicely.

### Finally, changes make the "waiting for MySQL" TMux screen give clearer output for new devs

Screenshot 2024-06-18 at 09 49 18

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,8 @@ up: require create-if-missing.env ../owid-content tmp-downloads/owid_metadata.sq
 
 	@echo '==> Starting dev environment'
 	@mkdir -p logs
+
+	@make kill-if-exists
 	tmux new-session -s grapher \
 		-n docker 'docker compose -f docker-compose.grapher.yml up' \; \
 			set remain-on-exit on \; \
@@ -77,6 +79,8 @@ up.devcontainer: create-if-missing.env.devcontainer tmp-downloads/owid_metadata.
 
 	@echo '==> Starting dev environment'
 	@mkdir -p logs
+
+	@make kill-if-exists
 	tmux new-session -s grapher \
 		-n admin \
 			'devTools/docker/wait-for-mysql.sh && yarn startAdminDevServer' \; \
@@ -100,6 +104,8 @@ up.full: require create-if-missing.env.full ../owid-content tmp-downloads/owid_m
 	yarn lerna run build
 
 	@echo '==> Starting dev environment'
+
+	@make kill-if-exists
 	tmux new-session -s grapher \
 		-n docker 'docker compose -f docker-compose.grapher.yml up' \; \
 			set remain-on-exit on \; \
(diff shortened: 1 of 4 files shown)


toni-sharpe:
@danyx23 just pinging you as principle here, and I pinged Lars a lot

I feel this makes it a bit easier for new open source folk, it should be as smooth as possible for them.

Maybe there's a better way of acheiving the same

toni-sharpe:
Possibly fixes:

https://github.com/owid/owid-grapher/discussions/3075
https://github.com/owid/owid-grapher/discussions/3122

toni-sharpe:
@larsyencken curious that you've taken this one over, happy to answer questions

larsyencken:
Hi @toni-sharpe, thanks for reporting this. I made a new issue for it:

- #3840 

Then I simplified your PR into a new PR for it:

- https://github.com/owid/owid-grapher/pull/3841

Some things that made this linger, as feedback:

- The PR was actually two issue descriptions (and two fixes)
  - Better to report both issues minimally, get the ok, and fix them minimally too if the issue is accepted
- The PR text was really long, which puts a burden on the reader to digest the train of thought
  - That's fine for your own personal notes, but for collaboration it's better to really condense it down
  - Contrast to #3840, which gets right to the point
- The problem was a "nice to have"
  - This means that the total investment we want to do is only a few minutes, so anything that gets in the way of that (long text, multiple issues in one) just makes it less likely we can sort it out quickly

toni-sharpe:
nice to see this merged, even via another PR, make files were new to me :-)

PR #3646, opened 2024-05-24
Update CodeQL config

rakyi:
Fix strange warning about missing config in GitHub actions.


rakyi:
It doesn't fix the warning.

Issue #3351, opened 2024-03-15
LineChart/StackedArea: Improve labelling when space is tight


sophiamersmann:
## Description

- When space is tight, the labelling algorithm used in line charts and stacked area charts doesn't do a good job
- This has been noted by Pablo A when trying to prepare a mobile static chart for data insights (see screenshots below)
- This is also the reason why we don't re-use the labelling algorithm for slope charts

## Expected behaviour

- Show as many labels as possible, even when space is tight

## Steps to reproduce

1. Go to https://owid.cloud/admin/charts/383/edit
2. Click on the 'Export' tab
3. Switch to the mobile version

## Screenshots

When space is tight:

<img width="563" alt="Screenshot 2024-03-15 at 12 02 04" src="https://github.com/owid/owid-grapher/assets/12461810/24f77321-8515-4823-840f-1deb590b1604">

If there is more space (here achieved by removing text elements):

<img width="563" alt="Screenshot 2024-03-15 at 12 02 29" src="https://github.com/owid/owid-grapher/assets/12461810/8e07a85f-83f6-4e53-85bb-b5e21ee7279a">

danyx23:
Discussed this in issue triage - because this now comes up more in thumbnails and mobile chart exports we think it's important to fix this. 

To fix it, we should probably drop entities from the list of labelled entities if there is insufficient space until there is enough space. It's a bit unclear if we should drop entities by lowest value in the last year first or by area in the visible chart space.

PR #3647, opened 2024-05-24
✨ (grapher) improve line legend algorithm

sophiamersmann:
Resolves https://github.com/owid/owid-grapher/issues/3351

### Current implementation

The current implementation uses two different placing algorithms:
  1. If all labels fit into the space, then labels are placed in a non-overlapping way using connector lines
  2. If the labels don't fit into the available space, then labels are simply placed at their y-value and overlapping labels are greyed out and rendered into the background

### Updated implementation

I got rid of the second placing algorithm. Instead, if not all labels fit into the available space, then we drop as many labels as necessary, and then the first placing algorithm is used to compute the positions of the remaining labels.

#### How labels are dropped

The default strategy, used by line charts, tries to pick labels in a balanced way so that they're roughly at equidistance from each other. Stacked area charts overwrite this behaviour and drop labels based on the area size.

The default strategy works like this: Given a set of placed labels and a set of candidates, for each candidate, we find the two closest already placed labels, one to each side, and calculate a score based on the available space between the two placed labels (the bigger, the better) and the candidate's distance to the midpoint (the smaller, the better). We then pick the candidate with the best score that fits into the available space.

### Future Work

Lines/Areas that don't get labelled should show a label on demand (previously, you could hover over the grayed-out background labels). Stacked area charts highlight the currently hovered area in the tooltip – we could do something similar for line charts. To be worked on...

| Before  | After  |
| ------- | ------ |
|  total-population-living-in-extreme-poverty-by-world-region-mobile (1) | total-population-living-in-extreme-poverty-by-world-region-mobile |
| Screenshot 2024-05-28 at 16 15 48 |Screenshot 2024-05-28 at 16 36 15|


sophiamersmann:
* **#3647** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3647?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/lineLegend/LineLegend.tsx
```diff
@@ -256,48 +262,54 @@ export class LineLegend extends React.Component<{
...
+                totalLevels: 0,
+            }
+        })
+    }
 
-                // Ensure list is sorted by the visual position in ascending order
-            }),
-            (label) => yAxis.place(label.yValue)
-        )
+    @computed get intialSeriesByName(): Map<EntityName, PlacedSeries> {
```
marcelgerber, line 296: typo: 
```suggestion
    @computed get initialSeriesByName(): Map<EntityName, PlacedSeries> {
```

packages/@ourworldindata/grapher/src/lineLegend/LineLegend.tsx
```diff
@@ -360,49 +372,163 @@ export class LineLegend extends React.Component<{
...
+                    (s) => s[1],
+                    "desc"
+                )
+
+                // pick a candidate that fits into the available space
+                let foundCandidate = false
+                for (const [candidate] of sortedCandidates) {
+                    foundCandidate = maybePickCandidate(candidate)
+                    if (foundCandidate) break
+                }
```
marcelgerber, line 509: perf: If we decide to not pick a candidate, then we can remove it from `candidates`.
Since we only ever add candidates, and don't remove any in the future, there's no way it will fit in the future.


And then, with that optimization in place, we can also use `maxBy` rather than `orderBy` to pick the highest-score candidate. Which should be more performant in most cases.

packages/@ourworldindata/grapher/src/lineLegend/LineLegend.tsx
```diff
@@ -360,49 +372,163 @@ export class LineLegend extends React.Component<{
...
+                return false
+            }
+
+            type Bracket = [number, number]
+            const findBracket = (
+                sortedBrackets: Bracket[],
+                n: number
+            ): [number | undefined, number | undefined] => {
+                const firstBracketValue = sortedBrackets[0][0]
+                const lastBracketValue = last(sortedBrackets)![1]
```
marcelgerber, line 443: At unreasonably small sizes these lines can fail, because there are no brackets at all (and hence `sortedBrackets[0]` and `last(sortedBrackets)` are both `undefined`):

http://staging-site-line-legend-algo/grapher/share-in-poverty-relative-to-different-poverty-thresholds is failing for me at size 237 x 507.

marcelgerber (approved):
Great idea!
I had a performance comment, and also found a case where this errors, but otherwise this is great, and certainly a nice improvement 🙌🏻

packages/@ourworldindata/grapher/src/lineLegend/LineLegend.tsx
sophiamersmann: Nice!

packages/@ourworldindata/grapher/src/lineLegend/LineLegend.tsx
sophiamersmann: Good catch!

packages/@ourworldindata/grapher/src/lineLegend/LineLegend.tsx
```diff
@@ -360,49 +372,160 @@ export class LineLegend extends React.Component<{
...
+                    return [candidate, score]
+                })
+
+                // pick the candidate with the highest score
+                const maxCandidate = maxBy(candidateScores, (s) => s[1])![0]
+                const picked = maybePickCandidate(maxCandidate)
+
+                // if the highest scoring candidate doesn't fit,
+                // remove it from the candidates and continue
+                if (!picked) candidates.delete(maxCandidate)
```
marcelgerber, line 507: Actually, there's room for another perf improvement here:
```suggestion
                let picked = false
                
                while (!picked && candidateScores.length) {
                    const maxCandidateArr = maxBy(candidateScores, (s) => s[1])
                    const maxCandidate = maxCandidateArr[0]
                    picked = maybePickCandidate(maxCandidate)

                    // if the highest scoring candidate doesn't fit,
                    // remove it from the candidates and continue
                    if (!picked) {
                        candidates.delete(maxCandidate)
                        candidateScores.delete(maxCandidateArr)
                    }
                }
```

... because the scores and brackets don't change until we pick a candidate.

marcelgerber (approved):

sophiamersmann:
### Merge activity

* **Jun 11, 5:48 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3647).

PR #3650, opened 2024-05-27
🐛 (grapher) don't show full-screen button when not useful

### LINKED ISSUE #2887: Mobile usability in full screen mode
## Description

Full screen mode is a nice addition which has the potential to drastically improve the experience for mobile users, however it seems to have the opposite effect, crowding screen real estate with the header and footer and leaving little/no space for the data. Turning the phone to landscape actually makes things worse. This matters because many users are instinctively trained to turn to landscape when in full screen because this is how they watch videos.

## Expected behaviour

Full screen mode should emphasize the actual data across a range of device capabilities, possibly with the header and footer accessible if needed. In general though this information isn't going to be critical at this point because the user is aware they can access it by exiting full screen mode.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/co2-emissions-vs-gdp
2. Click on "Enter full screen"
3. Turn device to landscape mode
4. Attempt to tap, scroll up and down, etc.

## Screenshots

image_123986672

## Environment

See  https://www.whatismybrowser.com/w/54YSQKM

-   iPhone SE
-   OS: iOS 16.6.1
-   Browser safari 16.6

Yes, this browser is out of date. It also occurs in Brave. I'm pretty certain it's a viewport and not a browser issue per se.

## Additional context

Loving the redesign, just a suggestion for improvement.

sophiamersmann:
Resolves #2887 

Hides the full-screen button if you'd end up with a smaller grapher frame in full-screen mode (this is the case for landscape mode on mobile, for example)

### DIFF
--- a/packages/@ourworldindata/grapher/src/controls/ActionButtons.tsx
+++ b/packages/@ourworldindata/grapher/src/controls/ActionButtons.tsx
@@ -26,6 +26,7 @@ export interface ActionButtonsManager extends ShareMenuManager {
     canonicalUrl?: string
     isInFullScreenMode?: boolean
     isDownloadModalOpen?: boolean
+    hideFullScreenButton?: boolean
 }
 
 // keep in sync with sass variables in ActionButtons.scss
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#3650** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3650?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #3667, opened 2024-05-30
Fix SimpleMarkdownText with SSR

### LINKED ISSUE #2992: Datapages do not render without Javascript
Found while investigating the archive.org issue:

image

Our regular grapher pages still work:

image

rakyi:
I looked into moving markdown rendering to the server, but dropped it because it would increase the scope too much.

Fixes #2992


ikesau (approved):
Nice :)

PR #3670, opened 2024-05-30
🎉 add commas between sibling ref superscripts

ikesau:
Different browsers render (slightly) differently but I think it's passable

Firefox:
image

Chrome:
image

Safari:
image


marcelgerber (approved):
Nice! Yes, agree that the slightly different is totally acceptable. Have tested this in Chrome, Firefox, Safari, and all look good. Go ahead!

PR #3678, opened 2024-06-03
✨ (grapher) prevent scrolling while showing tooltips / TAS-528

sophiamersmann:
Line charts and stacked area charts listen to move events on mobile to show a tooltip, but don't prevent scrolling, which makes the experience of interacting with the tooltip a bit jumpy (see video below).

This PR makes it so that scrolling is disabled when interacting with the chart area. 

Implementation detail: All React event listeners are by default passive, which means that calling `event.preventDefault()` has no effect. That's why we need to attach the event listener to the DOM ourselves (technically, only the move event needs this but I moved them all to the `componentDidMount` method so that they're all defined in the same place)

https://github.com/owid/owid-grapher/assets/12461810/d2f4ab22-fc10-44c9-8353-0cb2aabb146e

### DIFF
--- a/packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
+++ b/packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
@@ -370,9 +370,9 @@ export class LineChart
         x: number
     }>({ fade: "immediate" })
 
-    @action.bound private onCursorMove(
-        ev: React.MouseEvent | React.TouchEvent
-    ): void {
+    @action.bound private onCursorMove(ev: MouseEvent | TouchEvent): void {
+        ev.preventDefault() // prevent scrolling on touch devices
+
         const ref = this.base.current,
             parentRef = this.manager.base?.current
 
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#3679** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3679?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3678** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

danyx23 (approved):
Ah interesting, I didn't know that about react event handlers. This looks good and seems to work as advertised but I lack the expertise to know if there are any potential issues with this. I think you can safely merge this but if you want a review from someone who understands this stuff better then ping Martin :)

rakyi (commented):
I haven't done this exact thing before myself, so I'm not sure if there are any gotchas hiding here. However, I know there is a great library for floating elements I used before that has [a component for this](https://floating-ui.com/docs/floatingoverlay).

Is there a chance that whatever we are already using for tooltips has something built-in as well? Otherwise, there is also [`react-remove-scroll`](https://www.npmjs.com/package/react-remove-scroll) in case we are not happy with what we did now, or we need a more reusable/battle tested solution.

rakyi (approved):

sophiamersmann:
Thanks for your reviews! This is good to go but I want to ship it in concert with follow-up PRs, so closing it for now, but will be continued...

PR #3679, opened 2024-06-03
✨ (grapher) fixed tooltip position on mobile / TAS-520

sophiamersmann:
Makes it so that mobile tooltips stick to the bottom (see screenshot below).

Notes:
- If the tooltip content is taller than the allowed max-width, then the tooltip becomes scrollable (that's why the tooltip stays on screen even after an interaction has ended; the tooltip can be dismissed by clicking anywhere on the screen)
- For tooltips that have a "total" row (e.g. stacked area charts), the total row is moved to the top for mobile tooltips (so that it's visible by default, even if the tooltip is scrollable)
- Tooltips are automatically dismissed when the Grapher is scrolled out of view
    - This works by utilizing Grapher's intersection observer. There was a weird bug on iOS Safari where the intersection observer simply didn't fire when Grapher was out of view. For some reason, that doesn't happen if the threshold is set to some small number (instead of 0, which is the default)
    - Ideally, I would want to dismiss tooltips a bit earlier, but I didn't want to set up a second intersection observer just for that...
    - Note that previously, after scrolling a Grapher into view, we would disconnect the intersection observer. Now, we keep it around – I don't think this should be a problem, but wanted to mention it.

To do:
- Map tooltips are broken

Screenshot 2024-06-05 at 14 54 46

### DIFF
--- a/packages/@ourworldindata/grapher/src/chart/ChartManager.ts
+++ b/packages/@ourworldindata/grapher/src/chart/ChartManager.ts
@@ -39,6 +39,7 @@ export interface ChartManager {
     comparisonLines?: ComparisonLineConfig[]
     showLegend?: boolean
     tooltips?: TooltipManager["tooltips"]
+    activeTooltipId?: TooltipManager["activeTooltipId"]
     baseColorScheme?: ColorSchemeName
     invertColorScheme?: boolean
     compareEndPointsOnly?: boolean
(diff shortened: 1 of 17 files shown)


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3679?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3679** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3679?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3678** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3678?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

sophiamersmann:
Good to go once map tooltips are fixed. To be continued...

PR #3680, opened 2024-06-04
Fix heading spacing in left column of sticky-right block

### LINKED ISSUE #3240: Fix spacing between headings and columns
## Problem

We don't like the alignment of our headings, which are parallel with our content rather than above them.

## Details

This is a common layout we use:

image

The image should align with the text, not the heading.

### Manually fix in markup

There are a few ways we could do this, one of the simplest ways could be to add a new property to heading elements

```
{.heading}
level: 1
text: I am a heading
margins: sticky-left | sticky-right | left-side | right-side
{}
```
which you would use like so:

```
[+body]
{.heading}
level: 1
text: I am a heading
margins: sticky-left
{}

{.sticky-left}
[.+left]
Blah blah blah
[]
[.+right]
{.image}
filename: image.png
{}
[]
{}
[]
```

### Automatically do it

Client-side? Server-side in ArchieML rewriting?

If this is too annoying to write out we could also try and solve it programmatically by detecting when the content is like this (`left[0].type == "heading" && right[0].type === "image"`) and automatically calculating a margin-top to add to the heading based on the current height of the heading.

## Open questions

- Do we want to do this for all articles, or just for some?
  - Our intuition is that we _probably_ want this for all articles

rakyi:
Fixes #3240

### DIFF
--- a/site/css/grid.scss
+++ b/site/css/grid.scss
@@ -157,3 +157,14 @@ $grid-responsive: (lg, md, sm);
         }
     }
 }
+
+// Custom helpers
+.grid-row-start-2 {
+    grid-row-start: 2;
+}
+
+.grid-sm-row-start-auto {
+    @include breakpoint-max(sm) {
+        grid-row-start: auto;
+    }
+}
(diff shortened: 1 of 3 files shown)


rakyi:
Nice catch, I pushed a fix for the md breakpoint.

ikesau (approved):
:shipit:

PR #3684, opened 2024-06-05
:sparkles: create temp database in make refresh and swap it at the end

Marigold:
I'm having issues with refreshing MySQL on our staging servers. To keep the data up to date, we have to refresh MySQL after every commit to the ETL. The problem is that there might be a grapher baking process going on at the same time, and it will fail because MySQL is refreshing itself.

This PR makes the refresh process more "atomic" by creating a temporary database first and swapping it at the end. If this doesn't work, we'll have to start locking the database everywhere, which would be annoying.

### DIFF
--- a/devTools/docker/refresh-grapher-data.sh
+++ b/devTools/docker/refresh-grapher-data.sh
@@ -20,20 +20,27 @@ _mysql() {
 }
 
 import_db() {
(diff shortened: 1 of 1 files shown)


Marigold:
Not possible because owid user doesn't have privileges to create a new database.

PR #3687, opened 2024-06-05
:hammer: destroy algolia index together with the staging server

Marigold:


Marigold:
Don't close it! It's so close in my TODO list...

PR #3689, opened 2024-06-05
✨ (grapher) allow line breaks after hyphens / TAS-524

sophiamersmann:
### Summary

Adds the ability to break after hyphens (`-`) to `TextWrap`. This is then used for slope labels to reduce the space we need for the labels on both sides.

I considered making this the default but decided against it because many labels looked worse / less balanced with additional line breaks after hyphens.

### SVG tester

Some charts come back with a difference, but they're visually the same. The new code trims whitespace from each side:

```
Svg was different for 2649. The difference starts at character 15152.
Reference: 5" y="465.895">home) </tspan></text></g>
Current  : 5" y="465.895">home)</tspan></text></g><
```

```
Svg was different for 3552. The difference starts at character 11307.
Reference: "4.995000000000001"> Lower respiratory i
Current  : "4.995000000000001">Lower respiratory in
```

```
Svg was different for 5215. The difference starts at character 8953.
Reference: urement policies and </tspan><tspan x="1
Current  : urement policies and</tspan><tspan x="16
```

### Screenshots

| Before  | After  |
| ------- | ------ |
| Screenshot 2024-06-05 at 16 46 18  | Screenshot 2024-06-05 at 16 46 12 |

Example chart: https://ourworldindata.org/grapher/maternal-mortality-slope-chart?country=High-income+countries~Lower-middle-income+countries

### DIFF
--- a/packages/@ourworldindata/components/src/TextWrap/TextWrap.tsx
+++ b/packages/@ourworldindata/components/src/TextWrap/TextWrap.tsx
@@ -1,12 +1,7 @@
-import {
-    isEmpty,
-    max,
-    stripHTML,
-    Bounds,
-    FontFamily,
-} from "@ourworldindata/utils"
+import { max, stripHTML, Bounds, FontFamily } from "@ourworldindata/utils"
 import { computed } from "mobx"
 import React from "react"
+import { Fragment, joinFragments, splitIntoFragments } from "./TextWrapUtils"
 
 declare type FontSize = number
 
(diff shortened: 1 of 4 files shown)


sophiamersmann:
* **#3689** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3689?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

danyx23 (commented):
Looks good! I'd change the tests a bit though. I think it would be good to add two simple unit tests that just demonstrate what splitIntoFragments does with just a space and with space and hyphen. 

I think it would also be more interesting to change the existing property style test to include the hyphen, rather than just default to space (but I really like that you added this kind of test :) )

sophiamersmann:
Thanks for the review! I updated the tests :)

PR #3693, opened 2024-06-06
🐛 remove obsolete rebuild table that results in undefined variable

danyx23:
While debugging the chart issue that turned out to be an OOM problem I realized that the admin calls rebuildInputOwidTable explicitly at a point when the data for the newly added indicators isn't yet available. It is called again later, I think throuh a reaction.

This PR thus removes what I think is a redudant and semantically misplaced call.


marcelgerber (changes requested):
I found a regression in here, in that changing dimension colors doesn't immediately update the UI any more. It only does so after changing something else about the dimensions.

https://github.com/owid/owid-grapher/assets/2641501/2f6d5cfb-7139-49fe-8937-651e7ab8867a

PR #3695, opened 2024-06-06
🎉 Tag Graph

ikesau:
## Changes
- Adds a new `tag_graph` table in the DB
  - Naively prepopulated based on the current parentId structure in the `tags` table
- Adds the `@dnd-kit/core` library to our admin
- Adds a new admin route `/tag-graph` to quickly and conveniently structure the tag graph
- Updates the `/tags` admin UI to be a flat list of all tags and a simple form to add a new tag

This PR doesn't introduce any breaking changes or migrations to existing interactions with our tags table. Only once we've merged this and the data and research team has created a tag graph they're happy with, we can migrate functionality over to use the graph.

## Demo
https://github.com/owid/owid-grapher/assets/11844404/2d9ba907-95d9-4108-a14d-b42e92d687e3


adminSiteServer/apiRouter.ts
```diff
@@ -2008,30 +2011,47 @@ postRouteWithRWTransaction(
...
     async (req: Request, res, trx) => {
-        const tag = (req.body as { tag: any }).tag
+        const tag = req.body
+        function validateTag(
+            tag: unknown
+        ): tag is { name: string; slug: string | null } {
+            return (
+                checkIsPlainObjectWithGuard(tag) &&
+                typeof tag.name === "string" &&
+                (tag.slug === null || typeof tag.slug === "string")
```
danyx23, line 2021: Here it probably makese sense to also check if the slug is not empty string, just in case

db/db.ts
```diff
@@ -448,3 +453,103 @@ export const getNonGrapherExplorerViewCount = (
...
+            tagGraphRows.push({
+                parentId: child.parentId,
+                childId: child.childId,
+                weight: child.weight,
+            })
+        }
+    }
+
+    await knex("tag_graph").delete()
+    await knex("tag_graph").insert(tagGraphRows)
```
danyx23, line 528: This is an easy way to do the update but then the updatedAt timestamp will not be useful information. I think being able to know which tags have been updated might be worth the slight complication to do a select all and then do the add/update/delete operations accordingly. What do you think?

db/db.ts
```diff
@@ -448,3 +453,103 @@ export const getNonGrapherExplorerViewCount = (
...
+            tg.weight,
+            t.name,
+            t.slug,
+            IFNULL((p.type IN (:types) AND p.published = 1 AND t.slug IS NOT NULL), FALSE) AS isTopic
+        FROM
+            tag_graph tg
+        LEFT JOIN tags t ON
+            tg.childId = t.id
+        LEFT JOIN posts_gdocs p on
+            t.slug = p.slug
```
danyx23, line 482: Energy was duplicated for me - if you move the logic from isTopic into the join condition then the unpublished energy gdoc does not lead to a second row for the tag and then you don't have that problem
```suggestion
            p.slug is not null AS isTopic
        FROM
            tag_graph tg
        LEFT JOIN tags t ON
            tg.childId = t.id
        LEFT JOIN posts_gdocs p on
            t.slug = p.slug AND p.published = 1 and p.type IN (:types)
```

adminSiteClient/TagEditPage.tsx
```diff
@@ -142,49 +124,30 @@ class TagEditor extends React.Component<{ tag: TagPageData }> {
                             field="name"
                             store={newtag}
                             label="Name"
-                            helpText="Category names should ideally be unique across the database and able to be understood without context"
+                            helpText="Tag names should ideally be unique across the database and able to be understood without context"
```
danyx23, line 127: Tag names are enforced to be unique now, right?

adminSiteClient/TagGraphPage.scss
```diff
@@ -0,0 +1,113 @@
...
+        &.tag-box--dragging {
+            z-index: 2;
+            box-shadow: 4px 4px 8px 8px #bbb;
+        }
+
+        &.tag-box--hovering:not(.tag-box--dragging) {
+            border: 3px solid orange;
+        }
+
+        // 🤫
```
danyx23, line 24: 😂

db/migration/1716401298509-AddTagGraph.ts
```diff
@@ -0,0 +1,55 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class AddTagGraph1716401298509 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`-- sql
+            CREATE TABLE tag_graph (
+                parentId INT NOT NULL,
+                childId INT NOT NULL,
+                weight INT NOT NULL DEFAULT 1,
```
danyx23, line 9: Should we maybe init this to 100 and sort it so that "heavier" items "sink down"? This makes it easier to modify individual elements up or down without quickly running into negative numbers which seem a bit weird. Totally minor thing of course and feel free to ignore

db/migration/1716401298509-AddTagGraph.ts
```diff
@@ -0,0 +1,55 @@
...
+                PRIMARY KEY (parentId, childId),
+                FOREIGN KEY (parentId) REFERENCES tags (id),
+                FOREIGN KEY (childId) REFERENCES tags (id),
+                INDEX (childId),
+                CONSTRAINT chk_no_self_link CHECK (parentId != childId)
+            )`)
+
+        // create root tag
+        await queryRunner.query(`-- sql
+            INSERT INTO tags (name) VALUES ("tag-graph-root")`)
```
danyx23, line 19: This leads to tag-graph-root being a valid tag in various selection dropdowns. Probably not a big deal since our users are not out to sabotage our system but maybe nice to eventually filter it out

danyx23 (changes requested):
Very nice work! I had a few comments/questions but nothing major.

db/db.ts
ikesau: The tag graph doesn't have `updatedAt` - do you think it should?

db/migration/1716401298509-AddTagGraph.ts
ikesau: I hemmed and hawed on what the metaphor for "weight" should be - density, or weight-as-in-importance.

Ultimately I went with importance because pressing the up arrow should move the object up, and things at the top should be the most important.

I'll increase the default if for no reason other than reducing the risk of shenanigans with the number 0 😁

db/db.ts
danyx23: I think it can be really useful - e.g. when something goes wrong and you want to undo some change from yesterday or so it's really nice if that query can identify the three rows that actually changed instead of having to fetch all from a backup.

That said I'm not sure about whether it's worth changing this PR for it.

danyx23 (approved):
Looks good, thanks for the tweaks! I'm a bit on the fence about the updatedAt field - I'll leave it to you to decide if you want to change it :)

PR #3700, opened 2024-06-10
🐛 increase the resolution of homepage intro primary thumbnails

### LINKED ISSUE #3698: Images sometimes render blurry
## Problem

We have had reports from Este, Marwa, and Ed that images sometimes render blurry.

image

It appears to very from one screen to another, on the same browser, same computer.

See: [internal Slack discussion](https://owid.slack.com/archives/C46U9LXRR/p1718013402633469)

## Expected behaviour

Images should be resized and selected so as to be crisp on all screen sizes and browsers.

## Technical notes

- Is the wrong resized variant being picked for some reason?

ikesau:
Fixes #3698

All `HomepageIntro` thumbnails were using images with `containerType: "thumbnail"` which sets the source `sizes` attribute to 350px, which was inaccurate for the primary tile.

This PR fixes that by setting the containerType for the primary tile to be `span6`, which will correctly calculate the width of the image and load a sufficiently hi-res image for any given screen width/DPI

https://github.com/owid/owid-grapher/assets/11844404/6af81c2a-8225-4b02-90b0-9f38dc122994

### DIFF
--- a/site/gdocs/components/HomepageIntro.tsx
+++ b/site/gdocs/components/HomepageIntro.tsx
@@ -8,7 +8,7 @@ import { formatAuthors, groupBy } from "@ourworldindata/utils"
 import { Button } from "@ourworldindata/components"
 import { useLinkedDocument } from "../utils.js"
 import { DocumentContext } from "../OwidGdoc.js"
-import Image from "./Image.js"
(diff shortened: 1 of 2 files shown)


marcelgerber (approved):
Neat!

PR #3717, opened 2024-06-18
garden: remove unneeded DB column

### LINKED ISSUE #3169: Drop `entities.displayName` in MySQL
The `entities` table appears to have an unused field called `displayName`. It's `NOT NULL`, but is always the empty string.

We should drop it.

toni-sharpe:
Fixes #3169 

By dropping the column, any data is lost and won't be retrieved by the `down` side of the migration. If the column was truly never used then fair enough, but users find ways of doing usefuly stuff with what they find so I wanted to check that before merging

- [ ] impact of column drop on live users passed "not an impact" - preferably in the production DB, with this being 10 years of work, better be safe than in a mess with DBs.

I have checked this table locally and it's not got anything in it, **and now it's gone**

Bit of a "gotcha" having to run `yarn runDbMigrations` - I'm sure you all know, but it doesn't hurt to remind

image

### DIFF
--- a/db/migration/1720038601815-RemoveEntitiesDisplayName.ts
+++ b/db/migration/1720038601815-RemoveEntitiesDisplayName.ts
@@ -1,0 +1,12 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class RemoveEntitiesDisplayName1720038601815 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        "ALTER TABLE entites DROP COLUMN displayName"
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        "ALTER TABLE entites ADD COLUMN displayName"
+    }
+}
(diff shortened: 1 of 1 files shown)


larsyencken (changes requested):
Thanks for this gardening work Toni! 

Rather than put this in the initial schema for everything, can you please add it as a new migration?

You can find instructions for that here: https://github.com/owid/owid-grapher/blob/master/db/readme.md#creating-a-migration

toni-sharpe:
@larsyencken no problem, should get a look before Monday. Migrations as well, I should have got that!

toni-sharpe:
@larsyencken I just pushed this up, having a spot of bother locally so thought it best to keep things moving and push blind, I'll look at what my issues are and should be able to report this work as "tested locally" then. In the meantime, someone might merge it.

db/migration/1720038601815-RemoveEntitiesDisplayName.ts
```diff
@@ -0,0 +1,20 @@
...
+                DROP COLUMN
+                displayName`
+        )
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            ALTER TABLE entities
+                ADD COLUMN
+                displayName`
```
larsyencken, line 17:
```suggestion
                displayName varchar(255) COLLATE utf8mb4_0900_as_cs NOT NULL`
```

larsyencken:
Tidied up and merged now, thanks Toni!

PR #3722, opened 2024-06-19
🎉 (stacked bar) support negative values / TAS-540

### LINKED ISSUE #1166: Support negative values for stacked bar charts
## Core problem

We are trying to use stacked charts with data that contains negative values, which they simply weren't designed to handle. Right now negative values don't display correctly, which can be misleading.

## Proposed solution

Redesign a variant of ~~stacked area and~~ stacked bar charts that can handle negative values and implement it.

- [ ] Sketch out a design and validate it
- [ ] (probably) move the axes to accomodate negative values
- [ ] (probably) make a new visual mark for the totals of each stack
- [ ] (probably) might be some entity label layout changes needed

## Special concerns

With negative values, the total is no longer the top of the bars/area, instead it needs a separate and visually distinguishing mark. The tooltop might need updating, and axis will need shifting.

sophiamersmann:
Adds support for negative values for StackedBar charts, resolves #1166

## Future work

- The [follow-up PR](https://github.com/owid/owid-grapher/pull/3725) implements value-based color encoding, so that (if a single series is plotted) positive and negative values are colored differently
- We could add sub-totals for positive and negative values in the tooltip, but for now it just shows the net

## Screenshots

Example chart: http://staging-site-stacked-bar-negative-values/admin/charts/7908/edit

Screenshot 2024-06-21 at 13 32 13


sophiamersmann:
* **#3722** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3722?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
```diff
@@ -251,16 +250,16 @@ export class AbstractStackedChart
         return new AxisConfig(this.manager.yAxisConfig, this)
     }
 
+    // implemented in subclasses
+    @computed protected get yAxisDomain(): [number, number] {
```
sophiamersmann, line 254: I find it confusing if methods are overwritten in subclasses, so I use a "dummy" implementation here, and it's then the job of a subclass to overwrite it. (We can't use actual abstract class because AbstractStackedChart cannot be made abstract because of MobX)

packages/@ourworldindata/grapher/src/stackedCharts/StackedUtils.ts
```diff
@@ -28,6 +28,30 @@ export const stackSeries = <PositionType extends StackedPointPositionType>(
     return seriesArr
 }
 
+// This method shifts up positive y values and shifts down negative y values of a Series with Points in place.
+export const stackSeriesInBothDirections = <
```
sophiamersmann, line 32: We could also replace `stackSeries` with `stackSeriesInBothDirections` (`stackSeriesInBothDirections` can do everything that `stackSeries` can do)

packages/@ourworldindata/grapher/src/stackedCharts/StackedUtils.test.ts
```diff
@@ -72,5 +100,25 @@ describe(stackSeries, () => {
...
+    })
+})
+
+describe(stackSeriesInBothDirections, () => {
+    it("can stack positive values", () => {
+        const series = stackSeriesInBothDirections(
+            withMissingValuesAsZeroes(seriesArr)
+        )
+        expect(series[1].points[0].valueOffset).toEqual(10)
+        expect(series[2].points[0].valueOffset).toEqual(12)
```
marcelgerber, line 113: I took some time to grasp where these offsets are coming from; I think adding a comment here for each line will help future me :)
```suggestion
        expect(series[1].points[0].valueOffset).toEqual(10) // USA 2000
        expect(series[2].points[0].valueOffset).toEqual(12) // France 2000
```

packages/@ourworldindata/grapher/src/stackedCharts/StackedUtils.test.ts
```diff
@@ -72,5 +100,25 @@ describe(stackSeries, () => {
...
+    })
+
+    it("can stack positive & negative values", () => {
+        const series = stackSeriesInBothDirections(
+            withMissingValuesAsZeroes(seriesArrWithNegativeValues)
+        )
+        expect(series[1].points[0].valueOffset).toEqual(0)
+        expect(series[2].points[0].valueOffset).toEqual(-10)
+        expect(series[2].points[1].valueOffset).toEqual(0)
```
marcelgerber, line 122:
```suggestion
        expect(series[1].points[0].valueOffset).toEqual(0) // USA 2000
        expect(series[2].points[0].valueOffset).toEqual(-10) // France 2000
        expect(series[2].points[1].valueOffset).toEqual(0) // France 2002
```

packages/@ourworldindata/grapher/src/stackedCharts/StackedUtils.ts
```diff
@@ -28,6 +28,30 @@ export const stackSeries = <PositionType extends StackedPointPositionType>(
...
+    seriesArr.forEach((series, seriesIndex) => {
+        if (!seriesIndex) return // The first series does not need to be shifted
+        series.points.forEach((point, pointIndex) => {
+            const pointsBelowThisOne = seriesArr
+                .slice(0, seriesIndex)
+                .map((s) => s.points[pointIndex])
+            const pointBelowThisOne =
+                point.value < 0
+                    ? last(pointsBelowThisOne.filter((p) => p.value < 0))
+                    : last(pointsBelowThisOne.filter((p) => p.value >= 0))
```
marcelgerber, line 46:
```suggestion
                    ? findLast(pointsBelowThisOne, (p) => p.value < 0)
                    : findLast(pointsBelowThisOne, (p) => p.value >= 0)
```

... where [`findLast`](https://lodash.com/docs/4.17.15#findLast) is taken from lodash (it's currently not re-exported from `Util.ts`, but you can add it to the list).

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
marcelgerber: yeah, I agree that this is often confusing...

packages/@ourworldindata/grapher/src/stackedCharts/StackedUtils.ts
marcelgerber: But then stacked area charts would break in weird ways probably, right? Unless we explicitly set all negative values to zero beforehand.

packages/@ourworldindata/grapher/src/stackedCharts/StackedBarChart.tsx
```diff
@@ -81,12 +87,16 @@ class StackedBarSegment extends React.Component<StackedBarSegmentProps> {
 
     @computed get yPos(): number {
         const { bar, yAxis } = this.props
-        return yAxis.place(bar.value + bar.valueOffset)
+        return bar.value < 0
```
marcelgerber, line 90:
```suggestion
        // The top position of a bar
        return bar.value < 0
```

packages/@ourworldindata/grapher/src/stackedCharts/StackedBarChart.tsx
```diff
@@ -152,6 +162,14 @@ export class StackedBarChart
         return GRAPHER_FONT_SCALE_12 * this.baseFontSize
     }
 
+    @computed protected get yAxisDomain(): [number, number] {
+        const yValues = this.allStackedPoints.map(
+            (point) => point.value + point.valueOffset
+        )
+        const hasNegativeValues = yValues.some((value) => value < 0)
+        return [hasNegativeValues ? min(yValues) ?? 0 : 0, max(yValues) ?? 0]
```
marcelgerber, line 170: This'll go wrong if all values are negative, I think (because then `max(yValues) < 0`). And in theory, you don't need the extra `hasNegativeValues` check either:

```suggestion
        return [min([0, ...yValues]) ?? 0, max([0, ...yValues]) ?? 0]
```

I think this should work 🤞🏻

packages/@ourworldindata/grapher/src/stackedCharts/StackedBarChart.tsx
```diff
@@ -321,6 +339,25 @@ export class StackedBarChart
...
+            return {
+                seriesName: series.seriesName,
+                seriesColor: series.color,
+                point,
+            }
+        })
+        const [positivePoints, negativePoints] = partition(
+            hoverPoints,
+            ({ point }) => (point?.value ?? 0) >= 0
+        )
```
marcelgerber, line 355: Future idea: Would actually be quite cool if we could do a three-way partition into [positive, zero, negative] points here

marcelgerber (approved):
Nice! This works really well.

packages/@ourworldindata/grapher/src/stackedCharts/StackedUtils.ts
sophiamersmann: yap, that's right

sophiamersmann:
### Merge activity

* **Jul 2, 8:38 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3722).

PR #3728, opened 2024-06-20
✨  Reordering topics on the homepage 

spoonerf:
Find and implement an order that makes more sense for the links in our main menu and the "All our topics" section on our homepage - More details in https://github.com/owid/owid-issues/issues/1576


spoonerf:
Hi @rakyi! 

I started making some preliminary changes to the ordering of topic, the order has changed as expected in the 'All our topics' section at the bottom of [the homepage ](http://staging-site-homepage-reorder/), which is great! 

However, it hasn't changed in the 'Browse by topic' section at the top of the homepage, for example I would now expect Global Health to be at the top left under the Health section
Screenshot 2024-06-20 at 15 31 13

I thought both were linked from [the same code](https://github.com/owid/owid-grapher/blob/master/site/SiteNavigation.tsx#L271), but maybe I am missing something?

rakyi:
I forgot we don't bake everything when we deploy the PR branch to staging and for some reason the list in the footer is taken directly from code, while the top nav is loaded dynamically from a baked file (I don't know why, will ask).

I ran the baking/content deploy [manually](http://staging-site-homepage-reorder/admin/deploys) and now the two lists are in sync in the UI.

spoonerf:
Ah awesome, thanks for the rapid response!! That's good to know.

rakyi:
I [removed the baking variant](https://github.com/owid/owid-grapher/pull/3731), it wasn't necessary to do it that way. You can rebase on master, if you want to get that fix.

spoonerf:
Hey @rakyi, 

I am having a bit of an odd issue with the formatting of the 'All our topics' section - the placement of the vertical lines (|) and colons seems to be a bit out of sync in these two sections: 
* There shouldn't be a '|' between 'Fish and Overfishing' and 'Land Use'
* There should be a '|' between 'Lead Pollution' and 'Water'
* There should be a colon after 'Farm Size and Productivity' and 'Employment in Agriculture'
* There shouldn't be a colon after 'Agricultural Inputs' and 'Nutrition' 

Screenshot 2024-07-09 at 13 48 40

Fish and overfishing was missing from both so I've added it in, I'm not sure if that's perhaps caused an issue elsewhere? Any pointers would be appreciated!

rakyi:
It works correctly when I run this branch locally. I tried rebuilding the staging site manually in Buildkite and it didn't fix it. I'll investigate more.

spoonerf:
Ah okay @rakyi, it looks like your trigger of recreating the staging server worked, it looks as expected now, thanks!

rakyi:
That was not enough, I also [triggered baking](http://staging-site-homepage-reorder/admin/deploys) manually. I guess we can't do without it after all, so you can do that next time.

rakyi (approved):
LGTM! Just to check, we are not changing any existing URLs, are we? If we do, we'd need to set up redirects.

spoonerf:
> LGTM! Just to check, we are not changing any existing URLs, are we? If we do, we'd need to set up redirects.

No URL changes here, so should be all good, thanks!

Issue #3714, opened 2024-06-17
Enable publication dates in the future and scheduled publishing


danyx23:
## Context

In the past, our site had a relatively slow publishing rhythm where articles were written over weeks and months and then published at some point. Even in our full team our informal publishing schedule rarely had more than 2 or 3 pieces of writing go out within a week.

With data insights this has changed and we now often publish a piece of content a day or more often. Currently, this means that someone has to press publish on the gdoc in question at the time when it should be published. We want to change this now so that any gdoc can be published with a publication date in the future.

## Details

The basic version of this should be very simple to do - just filter out all content during baking where the publication date is in the future. 

This very simple version would rely on other content bakes occurring on the day of publication - e.g. if a post is published on August 1 with a publication date of August 5, then some kind of content bake would have to be triggered on August 5 for that piece of content to become visible. While some kind of content bake is triggered on most days (because of chart updates or similar), on bank holidays and such this is not guaranteed. It would thus be a nice first addition to add a daily buildkite step or something like that that guarantees that we do at least one content bake a day.

The other thing that would be useful is a future publishing schedule. This could be a read only view in the admin where you see which pieces of content will be published in the future, sorted by date (maybe a table with one row per piece of content going out that links to the gdoc admin of that content). This would make it easy to spot if there are e.g. multiple DIs scheduled for one day.

Ed is the person who will most benefit from this feature so he is a good person to talk to for questions regarding details.

## Open questions

- What time should content go live? The default would be midnight (GMT?) but when I chatted with Ed it sounded like a better default for us might be 9am UK time or so. To be verified with him.

danyx23:
It might be worthwhile to consider #3228 a bit when implementing this - i.e. which timezone do we consider for these decisions in?

PR #3729, opened 2024-06-20
Add time to publishedAt date picker

rakyi:
We want to schedule publishing of posts into the future, see #3714.

Lots of existing code already takes into account that `publishedAt` might be in the future. Main problem seems to be [Algolia indexing](https://github.com/owid/owid-grapher/blob/5e19bdb2e2b82517b2d19ee9a24ac2eaf06f3494/adminSiteServer/apiRouter.ts#L2407). Could we index posts in the same cron job that will bake them? We don't have to solve it now, since the current need is about data insights, which aren't indexed in search.

### DIFF
--- a/adminSiteClient/GdocsDateline.tsx
+++ b/adminSiteClient/GdocsDateline.tsx
@@ -2,6 +2,7 @@ import { Col, DatePicker } from "antd"
 import { Dayjs } from "dayjs"
 import { dayjs, OwidGdocErrorMessage, OwidGdoc } from "@ourworldindata/utils"
 import React from "react"
+import { PUBLISHED_AT_FORMAT } from "../settings/clientSettings"
 import { getPropertyMostCriticalError } from "./gdocsValidation.js"
 import { GdocsErrorHelp } from "./GdocsErrorHelp.js"
 
@@ -31,7 +32,7 @@ export const GdocsPublishedAt = <T extends OwidGdoc>({
             <DatePicker
                 onChange={onChangePublishedAt}
                 value={publishedAt ? dayjs(publishedAt) : undefined}
-                format="ddd, MMM D, YYYY HH:mm"
+                format={PUBLISHED_AT_FORMAT}
                 id="publishedAt"
                 status={publishedAtError?.type}
                 showTime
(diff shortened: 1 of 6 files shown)


db/db.ts
```diff
@@ -256,7 +256,7 @@ export const getPublishedDataInsights = (
         FROM posts_gdocs
         WHERE type = '${OwidGdocType.DataInsight}'
             AND published = TRUE
-            AND publishedAt < NOW()
+            AND publishedAt <= NOW()
```
rakyi, line 259: To be consistent everywhere.

adminSiteClient/GdocsDateline.tsx
```diff
@@ -32,9 +32,10 @@ export const GdocsPublishedAt = <T extends OwidGdoc>({
             <DatePicker
                 onChange={onChangePublishedAt}
                 value={publishedAt ? dayjs(publishedAt) : undefined}
-                format="ddd, MMM D, YYYY"
+                format="ddd, MMM D, YYYY HH:mm"
```
ikesau, line 35: I think we should render this as UTC instead of the user's local timezone (with a tooltip/label that explains this)

That way, the authors can all use the same time for "publishes in the morning" wherever they are in the world

ikesau: Yeah I see your point that authors might get the conversion wrong and expect to have an article be published immediately, but instead be off-by-one.

But I think as long as we have the "Publish Now" button that should be okay? If we want to be super sure, we could also add a button that sets the DatePicker to "now" (with a fix for the seconds caveat described in the comment just below this code change)

rakyi: Ed likes the idea of using UTC, but I wasn't able to make it work, see https://github.com/ant-design/ant-design/issues/45880

Since we want to ship this PR this week, I propose we won't block on this and if we really want to, we can switch to a different date picker or something later.

adminSiteClient/GdocsPreviewPage.tsx
```diff
@@ -229,11 +234,26 @@ export const GdocsPreviewPage = ({ match, history }: GdocsMatchProps) => {
...
+                                                content="There might be a slight delay before content scheduled into the future becomes live."
+                                                placement="bottom"
+                                            >
+                                                <FontAwesomeIcon
+                                                    icon={faQuestionCircle}
+                                                />
+                                            </Tippy>
+                                        </>
+                                    ) : (
+                                        "Scheduled"
```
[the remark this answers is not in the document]
rakyi: Changed.

adminSiteClient/GdocsDateline.tsx
rakyi: That's not the best related issue at ant-design, but I tried several ways of making it work and failed.

ikesau (approved):

PR #3726, opened 2024-06-20
✨ (stacked bar) align x-axis of facets / TAS-548

sophiamersmann:
Aligns x-axes of faceted StackedBar charts, resolves #2160

Implemented as part of an effort to create a set of climate charts, see draft charts on [this staging site](http://staging-site-stacked-bar/admin/charts)

## Summary

- Replaces the StackedBar chart's custom x-axis implementation with `DualAxisComponent`
- This works by passing a list of `domainValues` to the axis – it's then the axis' job (i) to figure out how much space is available for a single domain value (called the `bandWidth`) and (ii) to adjust the scale in such a way that we have sufficient padding on both sides for the outer bars to fit
- We also pass custom x-ticks to the axis, based on the actual x-values in the chart
- The `domainValues` and the `ticks` are shared across all facets, so that we end up with the same x-axis for all facets

## Screenshots

| Before  | After  |
| ------- | ------ |
| Screenshot 2024-06-21 at 17 30 05  | Screenshot 2024-06-21 at 17 30 40 |

Link: http://staging-site-stacked-bar-dual-axis/grapher/natural-disasters-by-type?facet=entity&uniformYAxis=0&country=Wildfire~Glacial+lake+outburst+flood~Dry+mass+movement~Extreme+temperature

PR #3733, opened 2024-06-21
🎉 (stacked bar) share x-axis across facets / TAS-549

sophiamersmann:
Shares the x-axis across faceted StackedBar charts

### Summary

- Hiding axis labels of inner facets declutters the chart if there are many facets
    - For charts with many facets, the x-axes are relatively close to each other, so it's easy to parse, even if there is only one axis on the bottom
    - For charts with few facets, the x-axes are relatively far from each other. If then some axis labels are missing, it can be difficult to figure out the x-axis value of a bar
- I ended up with a threshold of 12 facets (partly because we want this for the climate charts that we're working on and they have 12 facets) 

### Future work

- We might want to enable this for line charts and stacked area charts as well

### Known bugs

Suffers from [this bug](https://github.com/owid/owid-grapher/issues/2127) where axes are missing if one of the bottom facets doesn't have any data (fixed in the [next PR](https://github.com/owid/owid-grapher/pull/3735))

<details><summary>Screenshot</summary>
<p>

Screenshot 2024-06-21 at 13 50 01

</p>
</details>

### DIFF
--- a/packages/@ourworldindata/grapher/src/axis/Axis.ts
+++ b/packages/@ourworldindata/grapher/src/axis/Axis.ts
@@ -520,6 +520,10 @@ export class HorizontalAxis extends AbstractAxis {
         return this.rangeSize
     }
 
+    // note that we intentionally don't take `hideAxisLabels` into account here.
+    // tick labels might be hidden in faceted charts. when faceted, it's important
+    // the axis size doesn't change as a result of hiding the axis labels, or else
+    // we might end up with misaligned axes.
     @computed get height(): number {
         if (this.hideAxis) return 0
         const { labelOffset, labelPadding } = this
@@ -635,6 +639,10 @@ export class VerticalAxis extends AbstractAxis {
             : 0
     }
 
+    // note that we intentionally don't take `hideAxisLabels` into account here.
+    // tick labels might be hidden in faceted charts. when faceted, it's important
+    // the axis size doesn't change as a result of hiding the axis labels, or else
+    // we might end up with misaligned axes.
     @computed get width(): number {
         if (this.hideAxis) return 0
         const { labelOffset, labelPadding } = this
--- a/packages/@ourworldindata/grapher/src/facetChart/FacetChart.tsx
+++ b/packages/@ourworldindata/grapher/src/facetChart/FacetChart.tsx
@@ -65,6 +65,8 @@ import {
     NumericBin,
 } from "../color/ColorScaleBin"
 
+const SHARED_X_AXIS_MIN_FACET_COUNT = 12
+
 const facetBackgroundColor = "none" // we don't use color yet but may use it for background later
 
 const getContentBounds = (
@@ -211,7 +213,11 @@ export class FacetChart
         columnPadding: number
         outerPadding: number
     } {
-        return getChartPadding(this.facetFontSize)
+        const { isSharedXAxis, facetFontSize } = this
+        return getChartPadding({
+            baseFontSize: facetFontSize,
+            isSharedXAxis,
+        })
     }
 
     @computed private get hideFacetLegends(): boolean {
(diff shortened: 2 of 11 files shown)


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3733?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3733** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3726** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3726?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3725** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3722** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3722?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/stackedCharts/StackedBarChart.tsx
```diff
@@ -185,7 +185,9 @@ export class StackedBarChart
     }
 
     @computed protected get paddingForLegendRight(): number {
-        return this.showHorizontalLegend ? 0 : this.sidebarWidth + 20
+        return this.showHorizontalLegend
+            ? 0
+            : this.sidebarWidth + (this.sidebarWidth > 0 ? 20 : 0)
```
marcelgerber, line 190: I find this clearer:
```suggestion
            : (this.sidebarWidth > 0 ? this.sidebarWidth + 20 : 0)
```

packages/@ourworldindata/grapher/src/axis/Axis.ts
```diff
@@ -518,6 +518,10 @@ export class HorizontalAxis extends AbstractAxis {
         return this.rangeSize
     }
 
+    // note that we intentionally don't take `hideAxisLabels` into account here.
+    // tick labels might be hidden in faceted charts. when faceted, it's important
+    // the axis size doesn't change as a result of hiding the axis labels, or else
+    // we might end up with misaligned axes.
```
marcelgerber, line 524: Nice comment for future us!

packages/@ourworldindata/grapher/src/facetChart/FacetChart.tsx
```diff
@@ -342,6 +346,29 @@ export class FacetChart
...
+            ].includes(this.chartTypeName)
+        )
+    }
+
+    @computed private get isSharedXAxis(): boolean {
+        return (
+            this.uniformXAxis &&
+            // TODO: do this for stacked area charts and line charts as well?
+            this.chartTypeName === ChartTypeName.StackedBar &&
+            this.facetCount >= 12
```
marcelgerber, line 368: I guess for some magic constants it's nice to extract them into a constant - can make this into `SHARED_X_AXIS_MIN_FACET_COUNT` maybe.

packages/@ourworldindata/grapher/src/facetChart/FacetChart.tsx
```diff
@@ -446,10 +455,14 @@ export class FacetChart
...
+                gridEdges,
+                cellEdges,
+            } = gridBoundsArr[i]
             let bounds = initialGridBounds
             // Only expand bounds if the facet is on the same edge as the shared axes
-            for (const edge of edges) {
+            for (const edge of gridEdges) {
                 bounds = bounds.expand({
                     [edge]: sharedAxesSizes[edge],
                 })
```
marcelgerber, line 468: I was puzzling over what this code is supposed to do, and now found out that we don't want to do this for the `bottom` edge!

This screenshot here is with shared y axis, meaning that the y axis ranges should be the same across all facets. And sadly, as you can see, they are not!
CleanShot 2024-06-27 at 12 25 39

---

For the shared y axis, we want the facets to all have the same "charting area size" (i.e. x/y axis range), but then the charts in the first column need some extra space "on the left" to account for the y axis width. That's what this code is doing.

However, for the x axis, the axis is the same height no matter whether it's hidden or not, so no need to expand the bounds here.

I think the fix is just making sure that `sharedAxisSizes["bottom"] = 0`? Not sure.

Hope that comment makes sense, it's not trivial to explain this :)

marcelgerber: Also, given this, it makes just as much sense to iterate over `cellEdges`, which is the one that is more correct for this use case anyway (and then we could also get rid of `gridEdges` entirely probably?).

It won't make a difference now, but again, seems like the more correct approach.

```suggestion
            for (const edge of cellEdges) {
                bounds = bounds.expand({
                    [edge]: sharedAxesSizes[edge],
                })
```

marcelgerber (changes requested):
Very cool! I find an easy with facet sizing that needs to be addressed, but otherwise this is great work

packages/@ourworldindata/grapher/src/facetChart/FacetChart.tsx
sophiamersmann: Good catch! Setting `sharedAxisSizes.bottom` to zero does the trick, and I also removed `gridEdges` :)

marcelgerber (approved):
Nice! I just confirmed that this now works as expected.

sophiamersmann:
### Merge activity

* **Jul 2, 8:38 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3733).

PR #3736, opened 2024-06-21
✨ use UTC in the gdocs datepicker

ikesau:
I find the language of timezones difficult to write, but hopefully this very long comment explains what's going on here.

And if not, then this video can speak for itself 😁 

https://github.com/owid/owid-grapher/assets/11844404/dd073b3d-3dc9-4e7e-8d5a-2df93b32f3cf


rakyi (approved):
Good job getting it working! I'm not sure if it's worth the added complexity, but LGTM otherwise.

Don't forget to notify authors about the change.

PR #3745, opened 2024-06-25
🔨 add migration to add chart_diff_conflicts table

lucasrodes:
In our new workflow, we need some way to track chart conflicts* between PROD and staging servers.

Currently, we define a conflict if the `updatedAt` field of a chart is greater in PROD than in STAGING.

If we take a look at the diagram below, we observe that the current method only captures scenario 1.

image

To capture scenario 2, we need to be smarter. I've been thinking of a logical expression, but I believe there is none. Therefore, the idea now is to track in a table if a conflict was resolved. A conflict will be identified by the chart ID and timestamp of edit in PROD.

*A chart conflict happens when one user edits a chart in PROD and another in STAGING, after branching out. Similar to git-conflict.

Related: https://github.com/owid/owid-grapher/pull/3584

### DIFF
--- a/db/migration/1719334981843-chart-diff-conflicts.ts
+++ b/db/migration/1719334981843-chart-diff-conflicts.ts
@@ -1,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class ChartDiffConflicts1719334981843 implements MigrationInterface {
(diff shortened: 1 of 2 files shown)


Marigold (approved):
Looks good! Don't forget to add it to `grapher_models.py` in ETL.

PR #3749, opened 2024-06-26
Smarter gdocs attachment prefetching

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/3514

During baking, we prefetch gdoc attachments so that if, for example, multiple articles embed a grapher chart, we don't have to make multiple calls to the DB to fetch that chart's config.

For charts and linked indicators, this was done unparsimoniously: we prefetched resources that weren't being used anywhere.

Fortunately, on foundation, this has a negligible effect as you can see from the build step timings for charts and indicators:

```
6270 charts 9s
2239 indicators 0s
```

But locally, when each of these requests needs to be made over https, prefetching would take ~10 minutes, often timing out.

This PR makes it so that we only prefetch resources that are actually used somewhere, which makes local prefetching take about a minute.

As you'll see in the comments, prefetching `linkedIndicators` at all is pointless at the moment (and until we add that block to a non-homepage post we'll prefetch 0 of them) but it seems better to optimize and keep the code in than to remove it entirely.

I tested this by running `yarn buildLocalBake --steps gdocPosts dataInsights specialPages` with a `console.log` statement added to the error handlers in [useLinkedChart](https://github.com/owid/owid-grapher/blob/32c266775525e6d7a43bf62cf0654dbf7410abc7/site/gdocs/utils.tsx#L103) and [useLinkedIndicator](https://github.com/owid/owid-grapher/blob/a771d33332b0691713b7984d88f60b10c54fcb0d/site/gdocs/utils.tsx#L126) - no charts were missing. I then ran the local bake server and checked out multiple articles, data insights, linear topic pages, and topic pages and ensured they all worked correctly.

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -351,8 +351,21 @@ export class SiteBaker {
                 name: `✅ Prefetched ${Object.values(publishedExplorersBySlug).length} explorers`,
             })
 
+            // Get all grapher links from the database so that we only prefetch the ones that are actually in use
+            // 2024-06-25 before/after: 6266/2194
+            const grapherLinks = await db
+                .getGrapherLinkTargets(knex)
+                .then((rows) => rows.map((row) => row.target))
+                .then((targets) => new Set(targets))
+
             // Includes redirects
-            const publishedChartsRaw = await mapSlugsToConfigs(knex)
+            const publishedChartsRaw = await mapSlugsToConfigs(knex).then(
+                (configs) => {
+                    return configs.filter((config) =>
+                        grapherLinks.has(config.slug)
+                    )
+                }
+            )
             const publishedCharts: LinkedChart[] = []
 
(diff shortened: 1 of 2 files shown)


baker/SiteBaker.tsx
```diff
@@ -351,8 +351,21 @@ export class SiteBaker {
...
+
             // Includes redirects
-            const publishedChartsRaw = await mapSlugsToConfigs(knex)
+            const publishedChartsRaw = await mapSlugsToConfigs(knex).then(
+                (configs) => {
+                    return configs.filter((config) =>
+                        grapherLinks.has(config.slug)
+                    )
+                }
+            )
```
danyx23, line 368: Arguably it would be nicer to write this as a single query but it's clearly an improvement over the status quo so totally fine to ship as is :)

danyx23 (approved):
Looks good! Sorry for the late review.

PR #3753, opened 2024-06-28
Make the gdocs datepicker start on Monday

ikesau:


rakyi (approved):
Oh, dayjs ships their own locale and apparently even for en_GB they have Sunday as the first day of the week. I guess we have no other way to fix this. Thanks!

PR #3527, opened 2024-04-23
🐛 make key indicator collection more accessible

sophiamersmann:
I dropped the ball on the accordion's accessibility, so here I am making amends 😇

This PR makes the `key-indicator-collection` component more accessible to keyboard and screen reader users.

In particular,
- Hidden content (i.e. closed accordion items) is removed from the tab sequence and the accessibility tree (uses the [inert attribute](https://web.dev/articles/inert) to do so)
- Fixes a bug where "clicking" on an accordion item by tabbing to it and then pressing Enter made the focus jump back to the first focusable element on the screen (the OWID logo in the header)

PR #3755, opened 2024-07-01
🐛 (key indicator) fix broken animation / TAS-535

sophiamersmann:
#3527 attempted to fix a bug where - after opening an accordion item using the keyboard - the focus unexpectedly jumped to the top of the page. This was fixed by programmatically focusing the content area after opening an accordion. Unfortunately, this breaks the scroll-into-view animation. To avoid interrupting the scroll-into-view animation, we wait a few hundred ms to focus the content area. But we don't really know how much time the animation needs, so the timeout time is set through experimentation. If you're extremely quick to continue navigating, you'll still jump to the top page and then back. This only happens in Safari though – I can't repro the original bug in Chrome or Firefox. 

Not really happy with this fix, but can't think of another way...

### DIFF
--- a/site/gdocs/components/KeyIndicatorCollection.tsx
+++ b/site/gdocs/components/KeyIndicatorCollection.tsx
@@ -189,9 +189,16 @@ function AccordionItem({
                                     })
                                 }
 
-                                if (contentRef.current) {
-                                    contentRef.current.focus()
-                                }
+                                // focus on content after the scroll-into-view
+                                // animation has finished (fixes a bug in Safari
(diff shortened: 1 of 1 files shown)


sophiamersmann:
* **#3755** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3755?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

ikesau (approved):
sorry for the late review! works for me `:]`

PR #3777, opened 2024-07-08
:hammer: add command for deleting Algolia index

Marigold:
Command for deleting all Algolia indexes (implements https://github.com/owid/owid-grapher/issues/3563). Ops part https://github.com/owid/ops/pull/201.

### DIFF
--- a/baker/algolia/deleteAlgoliaIndex.tsx
+++ b/baker/algolia/deleteAlgoliaIndex.tsx
@@ -1,0 +1,46 @@
+import { ALGOLIA_INDEXING } from "../../settings/serverSettings.js"
+import { ALGOLIA_INDEX_PREFIX } from "../../settings/clientSettings.js"
+import { getAlgoliaClient } from "./configureAlgolia.js"
+import { SearchIndexName } from "../../site/search/searchTypes.js"
+import { getIndexName } from "../../site/search/searchClient.js"
+
+const deleteAlgoliaIndex = async () => {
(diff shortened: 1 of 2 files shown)


baker/algolia/deleteAlgoliaIndex.tsx
```diff
@@ -0,0 +1,38 @@
+import { ALGOLIA_INDEXING } from "../../settings/serverSettings.js"
+import { getAlgoliaClient } from "./configureAlgolia.js"
+import { SearchIndexName } from "../../site/search/searchTypes.js"
+import { getIndexName } from "../../site/search/searchClient.js"
+
+const deleteAlgoliaIndex = async () => {
+    if (!ALGOLIA_INDEXING) return
```
larsyencken, line 7: Recall that most staging servers do not get their own Algolia index, only ones with `-algolia` in the branch name, which ends up setting `ALGOLIA_INDEX_PREFIX`.

I would skip or abort if `ALGOLIA_INDEX_PREFIX` is empty, so that we don't accidentally have code that deletes the prod index.

larsyencken (changes requested):
Looks good, just a safety check to add.

baker/algolia/deleteAlgoliaIndex.tsx
Marigold: Good idea, that feels much safer.

larsyencken (approved):
A thing of beauty! :shipit:

PR #3780, opened 2024-07-08
:hammer: deprecate suggested chart revisions

Marigold:
Delete Suggested Charts Revision backed by table `suggested_chart_revisions` from Admin. It is no longer needed after switching to the new `chart-diff` workflow. This PR doesn't remove the table, which could be still useful for reconstructing chart history.


danyx23 (approved):
Thanks a lot for the cleanup!

Issue #3781, opened 2024-07-08
Cycle 2024.4 Friendly Data Catalog


ikesau:
[Figma](https://www.figma.com/design/AXOlKddXgQvUnRv3FfhlyM/Data-Catalogue?node-id=1-824&t=MM5mCLWSlsNugwFu-0)

**TODO**:
- [x] Set up correct tag graph on prod
- [x] Create Algolia Staging index and update indexer tag structure that works with drilling down (3d)
- [x] Bare bones explore + search UI (search; tag drilldown; breadcrumbs; "browse by tag" section) (3d)
- [x] URL state serialization (1d)
- [x] Country faceting UI (an entity selector to begin with) (1w)
- [x] UI polish (3d)
- [x] Baking/sitemap (1d)

**Unexpected work**
- [x] Design exploration for tag/folder navigation (3d)
- [x] Pagination (3d)
- [x] Fix internal state representation
- [x] Fix query string parsing when tag names have commas in them
- [x] Mobile UI customizations (3d)

**Next cycle, probably**:
- Tag cloud faceting (needs more discussion) (1w, some uncertainty here) 
- DSL with pills (depends on country and tag faceting decisions)
- Autosuggestions
- Explorer data (grapher-based and MDMs)
- Thumbnails for explorers
- Analytics

**Things to think about**:
https://github.com/owid/owid-grapher/issues/3648 will allow us to list data downloads at a higher level

**Cooldown simpatico**:
- Update the [sitenav tag graph branch](https://github.com/owid/owid-grapher/compare/sitenav-from-tag-graph?expand=1) to use subareas and perfectly recreate the static site nav ontology in the prod tag graph so that there's no delta when merged. d&r can then update the tag graph on prod as they wish. (3d)

PR #3790, opened 2024-07-12
Include charts' parent tags in the algolia index

ikesau:
Part of https://github.com/owid/owid-grapher/issues/3781

Adds a new function that finds all ancestors of a given tag and includes them in a chart's record when we index to Algolia.

**Examples**
- A chart record that used to have a `tag` value of `["Cardiovascular Diseases"]` will now have `["Cardiovascular Diseases", "Health"]` 
- A chart record that used to have a `tag` value of `["Indoor Air Pollution", "CO2 & Greenhouse Gas Emissions"]` will now have `["Indoor Air Pollution", "CO2 & Greenhouse Gas Emissions", "Air Pollution", "Health", "Energy and Environment"]`

[Example output of `getParentTagsByChildName`](https://github.com/user-attachments/files/16199640/tagGraphByChildren.json)

### DIFF
--- a/db/db.ts
+++ b/db/db.ts
@@ -536,11 +536,9 @@ export async function getParentTagsByChildName(
     const { __rootId, ...flatTagGraph } = await getFlatTagGraph(trx)
     const tagGraph = createTagGraph(flatTagGraph, __rootId)
 
-    const tagsById = await knexRaw<Pick<DbPlainTag, "id" | "name">>(
-        trx,
-        `-- sql
-        SELECT id, name FROM tags`
-    ).then((tags) => keyBy(tags, "id"))
+    const tagsById = await trx("tags")
+        .select("id", "name")
+        .then((tags) => keyBy(tags, "id"))
 
     const parentTagsByChildName: Record<
         DbPlainTag["name"],
--- a/db/db.ts
+++ b/db/db.ts
@@ -7,7 +7,7 @@ import {
     GRAPHER_DB_PORT,
 } from "../settings/serverSettings.js"
 import { registerExitHandler } from "./cleanup.js"
-import { keyBy } from "@ourworldindata/utils"
+import { createTagGraph, keyBy } from "@ourworldindata/utils"
 import {
     DbChartTagJoin,
     ImageMetadata,
@@ -26,8 +26,10 @@ import {
     DbPlainPostGdocLink,
     OwidGdocLinkType,
     OwidGdoc,
+    DbPlainTag,
+    TagGraphNode,
 } from "@ourworldindata/types"
-import { groupBy } from "lodash"
+import { groupBy, uniq } from "lodash"
 import { gdocFromJSON } from "./model/Gdoc/GdocFactory.js"
 
 // Return the first match from a mysql query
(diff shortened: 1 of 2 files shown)


db/db.ts
```diff
@@ -525,6 +527,45 @@ export async function getFlatTagGraph(knex: KnexReadonlyTransaction): Promise<
...
+            ...preexistingParents,
+            ...newParents,
+        ])
+    }
+
+    trackParents(tagGraph)
+
+    return parentTagsByChildName
+}
+
```
larsyencken, line 568: I would let the database do this, and replace it with one sequel query. MySQL 8 supports recursive common table expressions (the thing in the `WITH` block). It looks like this:

```sql
WITH RECURSIVE flat_tags AS (
  -- base case
  SELECT * FROM tags WHERE slug in ('population-growth', 'urbanization')

  UNION ALL

  -- recursive case
  SELECT t.*
  FROM tags t
  INNER JOIN flat_tags ft ON (t.id = ft.parentId)
)
SELECT *
FROM flat_tags
```

larsyencken (changes requested):
Looks good, but suggestion to push more into SQL.

db/db.ts
ikesau: Here's the query that approxmately creates the same structure as this TS function, though it also includes tags that aren't children of the tag-graph-root.

```sql
WITH RECURSIVE ancestors AS (
SELECT
	tg.childId,
	tg.parentId,
	t_parent.name AS parent_name,
	t_child.name AS child_name
FROM
	tag_graph tg
JOIN tags t_parent ON
	tg.parentId = t_parent.id
JOIN tags t_child ON
	tg.childId = t_child.id
WHERE
	t_parent.name != "tag-graph-root"
UNION
SELECT
	a.childId,
	a.parentId,
	t_parent.name AS parent_name,
	a.child_name
FROM
	ancestors a
JOIN tag_graph tg ON
	a.parentId = tg.childId
JOIN tags t_parent ON
	tg.parentId = t_parent.id
WHERE
	t_parent.name != "tag-graph-root"
)
SELECT
	a.child_name,
	GROUP_CONCAT(a.parent_name ORDER BY a.parentId ASC SEPARATOR ', ') AS ancestors
FROM
	ancestors a
GROUP BY
	a.child_name;
```

idk if this is any easier to work with, and I don't think performance is an issue. Do you strongly feel like we should be doing this in SQL?

db/db.ts
```diff
@@ -525,6 +527,45 @@ export async function getFlatTagGraph(knex: KnexReadonlyTransaction): Promise<
...
+        // slice to remove the root node and the current node, then map them into tag names
+        const newParents = node.path.slice(1, -1).map((id) => tagsById[id].name)
+
+        parentTagsByChildName[node.name] = uniq([
+            ...preexistingParents,
+            ...newParents,
+        ])
+    }
+
+    trackParents(tagGraph)
```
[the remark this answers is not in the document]
ikesau: I think maybe https://github.com/owid/owid-grapher/issues/3800 might be adding some confusion here 😛 

A few things:
1) With the tag graph (not the tags' `parentId` column) a node can have multiple parents:

image

2) The fake root node is there was an easy way to coexist with a tags table that has non-topic/area tags in it. Anything that's a child of the root tag node is part of the graph. Things that aren't (e.g. `Abstract`) can be ignored.

3) If I hadn't already written code for the tag graph UI that constructed the flat graph with data that I needed (and the function to turn it into a tree) I _could_ have written this function with a `idToParentIds` map and a `idToNode` map, but given I had those 2 functions already (and that's where my head was at) it seemed simpler at the time to do it this way.

We should have a call about this to make sure I'm understanding you correctly though.

PR #3743, opened 2024-06-25
🐝 (schema) set missing defaults

sophiamersmann:
## Summary

I went through the current config schema and tried to find all fields where a default value is missing. In the end, I updated the following defaults:

- Sets `true` as default for `hideRelativeToggle`
- Sets `2` as default for `dimensions[*].display.numDecimalPlaces`
- Sets `3` as default for `dimensions[*].display.numSignificantFigures`
- Sets `2020-01-21` as default for `dimensions[*].display.zeroDay`

PR #3793, opened 2024-07-16
🎉 (grapher) make charts inherit indicator-level settings / TAS-567

sophiamersmann:
## Summary

Implements chart config inheritance. Inheritance is opt-out, but disabled for all existing charts.

## Database

- Adds `variables.grapherConfigIdAdmin` and `variables.grapherConfigIdETL` that point to rows in the `chart_configs` table
    - Moves configs from `variables.grapherConfigETL` to the `chart_configs` table (`variables.grapherConfigAdmin` is discarded)
    - For ETL-authored configs, `fullConfig` = `patchConfig` (since they don't have a parent)
    - For admin-authored configs, `fullConfigAdmin` = `patchConfigETL` + `patchConfigAdmin`
    - The merged config is thus either the admin's full config of the ETL's full config (in that order)
- Adds a `charts.isInheritanceEnabled` column (disabled by default)
- Adds a view `charts_x_parents` that lists every chart and the parent variable it may inherit from
    - The parent variable is the y-variable of a single y-indicator chart (scatter plots excluded)

## Admin API

- Chart configs
    - `/charts/:chartId.config.json` returns the full config of a chart
    - `/charts/:chartId.patchConfig.json` returns the patch config (needed in admin)
    - `/charts/:chartId.parent.json` returns the parent variable id and config and whether inheritance is currently enabled (needed in the admin)
- A new query parameter has been added to the routes that save new or existing charts (`/charts` and `/charts/:chartId`)  
    - `/charts/:chartId?inheritance=enable` saves a chart, with inheritance enabled
    - `/charts/:chartId?inheritance=disable` saved a chart, with inheritance disabled
    - If the `inheritance` parameter is not given then a chart is saved without changing the inheritance setting
    - (It's a bit odd to have query params for a POST/PUT route, no? But if we had standalone routes to enable/disable inheritance, then we'd need to make two API calls in the admin, which also means we'd be triggering baking twice (unless we dance around that somehow). Since both calls potentially update a chart's config, the database could also be in a weird intermediate state if the first API call goes through but the second fails. Instead of the query params, we could also add additional routes for updating charts with inheritance enabled/disabled, but seemed more complex than the query param to me...)
- Endpoints to edit `grapherConfigETL`
    - `PUT /variables/:variableId/grapherConfigETL` creates a new config or updates an existing one
    - `DELETE /variables/:variableId/grapherConfigETL` deletes an ETL-authored config
    - Note that updating the Grapher config of a variable updates all charts that inherit from a variable, and thus might trigger a bake

## Grapher schema

- `version` and `title` are now optional
    - In practice, the `version` gets set automatically by the server if missing on save
    - In practice, charts should have a title (but indicator-level charts don't need one)
- `$schema` and `dimensions` are required
    - To know the `$schema` is esp. important for merging Grapher configs
    - `$schema` is also set to a constant value, the current schema version (i.e. the schema validator fails if the `$schema` is set to any other value) – it makes sense for now, but we might want to lift this restriction in the future
- Related: Grapher serialised `details` and `topicIds`, both were legacy and have been removed

## Admin

- The chart editor has been refactored to work with different types of charts (e.g. normal Grapher charts and indicator charts)
    - The implementation for editing indicator-level charts is missing though; the resp. classes (`<IndicatorChartEditorPage />`, `<IndicatorChartEditor />`) are mostly empty for now and are not linked to 
- The chart editor is now structured as follows:
    - The `<ChartEditorPage />` fetches data, inits a `ChartEditor` instance and renders `<ChartEditorView />`, with itself as its manager
    - `ChartEditor` handles state and IO that is specific to "normal" Grapher charts; common state and functionality that likely every editable chart needs is inherited from `AbstractChartEditor`
    - `IndicatorChartEditor`, for example, will handle state and IO that is specific to indicator-level charts
- Config inheritance has been integrated into the chart editor
    - If inheritance is possible (i.e. the current config has a parent variable with a config), then a new Inheritance tab is shown
    - The Inheritance tab allows to toggle inheritance and prints the parent and full config
    - The patch config is sent to the server on save
    - Text fields with linked/unlinked default values (title, subtitle, etc) either use the parent value as a default (if it exists), or a value computed from metadata (doing this more thoroughly is part of the follow-up work)
    - The 'Data to show' section (where `selectedEntityNames`) is edited can also be linked to the parent selection
- I have removed a few pieces of code that haven't been working for some time:
    - #3875 
    - #3876

## Caveats

- It's impossible to inherit `dimensions` from a parent indicator (since the dimensions are needed to figure out the parent indicator of a chart)
- The admin might load data a bit slower. The chart editor used to load all data at once. Now, `ChartEditorPage` loads all things related to the chart, and later `ChartEditorView` loads all things related to the admin (variables data, details)
- The colour section in the Customize tab doesn't currently work, see #3819 

## Testing

- Examples:
    - Chart without parent indicator: http://staging-site-inherit-indicator-settings/admin/charts/3375/edit
    - Chart with parent indicator that doesn't have a config: http://staging-site-inherit-indicator-settings/admin/charts/4711/edit
    - Chart with parent indicator that has a config: http://staging-site-inherit-indicator-settings/admin/charts/4408/edit

## To do

- [x] Make inheritance opt-out but disable inheritance for all existing charts

## To do after merging

- [ ] https://github.com/owid/etl/pull/3084
- [ ] https://github.com/owid/automation/pull/13


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3793?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3793** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3793?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3782** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3782?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3767** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3767?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3743](https://github.com/owid/owid-grapher/pull/3743) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3743?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/core/Grapher.jsdom.test.ts
```diff
@@ -77,20 +77,31 @@ it("can get dimension slots", () => {
 it("an empty Grapher serializes to an object that includes only the schema", () => {
     expect(new Grapher().toObject()).toEqual({
         $schema: defaultGrapherConfig.$schema,
+
+        // ideally, selectedEntityNames is not serialised for an empty object
+        selectedEntityNames: [],
```
sophiamersmann, line 82: This is unfortunate, but I'm not sure how to get rid of it. Although it's not nice, I don't think it has to be a blocker for this PR since it doesn't lead to buggy behaviour (afaik).

db/model/ChartConfigs.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+): Promise<void> {
+    await db.knexRaw(
+        knex,
+        `-- sql
+            UPDATE chart_configs
+            SET ?? = ?
+            WHERE id = ?
+        `,
+        [column, JSON.stringify(config), configId]
```
danyx23, line 66: It's maybe a bit nicer to call `serializeChartConfig` from dbType.ChartConfigs instead. For now they do the same of course but in the future they might gain additional checks.

danyx23 (commented):
Wow, what a huge piece of work. I'm not through yet with the full review (adminSiteClient and adminSiteServer are still missing) but I'm sending the comments I have so far as I'll switch to my travel laptop now.

sophiamersmann:
Daniel's comments:

> Wow, what a feat! Really nicely done. I am especially delighted about the tests you added, but the whole refactor is very nicely done. Kudos!
> 
> I read through the whole source code but on the review side I'm not entirely sure that I 100% thought through all code paths. Everything seems to make perfect sense but there are probably a few bugs still lurking.
> 
> Two examples:
> 
> * when you edit the title of a chart that has automatic time added then this is added over and over again (e.g. http://staging-site-inherit-indicator-settings/admin/charts/7562/edit)
> * When you turn on and off inheritance it resets too much sometimes IMHO. For example:
>   
>   * Create a new chart
>   * add gdp per capita (world bank, ppp, used ca. 150 times) - this one has a non-trivial etl config
>   * enable has map tab
>   * save draft
>   * enable inheritance
>   * disable inheritance
>   * has map tab is (wrongly IMHO) reset to false
> 
> For the last one I'be curious to know if chatgpt can write a regression test. If you show it the example of your current inherit on/off test and then give it the outline above (with the right setting name), can it quickly produce a valid test? Could be very useful for us. I was tempted to test it myself but don't have enough time today and didn't want to delay this review because of it.
> 
> When you have fixed the above I'd do another round of user testing but if things look good then then I think we could merge this soon and iron out any remaining issues as users find them.

sophiamersmann:
Thank you for the review! And thanks for catching the title bug! 🫠

The second issue you're describing is something that I believe I showed you in our admin session. It's a result of diffing the current layer against the parent. When inheritance is enabled, `hasMapTab` is removed from the current layer and relies on the parent value. The current layer thus "forgets" that it had `hasMapTab` enabled before. I could switch off diffing in the admin (I think? I'm not sure of the consequences), but then there would be a non-trivial difference between the server and client in how the patch config is produced. Since inheritance is not likely to be switched on/off a lot, I was hoping we could live with this weirdness (or maybe wait for authors to complain about it).

db/migration/1721296631522-MoveIndicatorChartsToTheChartConfigsTable.ts
```diff
@@ -0,0 +1,240 @@
...
+        // drop `grapherConfigAdmin` and `grapherConfigETL` columns
+        await queryRunner.query(`-- sql
+            ALTER TABLE variables
+                DROP COLUMN grapherConfigAdmin,
+                DROP COLUMN grapherConfigETL
+        `)
+
+        // add a view that lists all charts that inherit from an indicator
+        await queryRunner.query(`-- sql
+          CREATE VIEW charts_x_parents AS (
```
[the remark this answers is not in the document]
sophiamersmann: Lars suggested making inheritance opt-out but disabling inheritance for all existing charts. If that's what we want to do, then I'm inclined to keep this view. I struggled to find a good name, though – do you have a suggestion?

db/migration/1721296631522-MoveIndicatorChartsToTheChartConfigsTable.ts
```diff
@@ -0,0 +1,240 @@
...
+            .map(
+                ({ variableId, uuid }) =>
+                    `(${variableId},'${uuid}',unit,coverage,timespan,datasetId,display)`
+            )
+            .join(",")
+        await queryRunner.query(`-- sql
+            INSERT INTO variables (id, grapherConfigIdETL, unit, coverage, timespan, datasetId, display)
+            VALUES ${variablesUpdateValues}
+            ON DUPLICATE KEY UPDATE grapherConfigIdETL=VALUES(grapherConfigIdETL)
+        `)
```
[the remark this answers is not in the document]
sophiamersmann: yeah... it's a weird perf thing. Using a temp table didn't occur to me, I'll try that!

sophiamersmann:
Open bugs:
- [ ] http://staging-site-indicator-chart-editor/admin/charts/7940/edit: Changing to LineChart doesn't work; progress wheel doesn't stop spinnung

db/migration/1721296631522-MoveIndicatorChartsToTheChartConfigsTable.ts
```diff
@@ -0,0 +1,242 @@
...
+            WHERE grapherConfigETL IS NOT NULL
+        `)
+
+        // early return if there are no configs to migrate
+        // (this can happen if the migration is run for an empty test database)
+        if (variables.length === 0) return
+
+        // add a temporary table that maps variable IDs to config UUIDs
+        await queryRunner.query(`-- sql
+            CREATE TABLE tmpUUIDxVariableId (
```
danyx23, line 74: Nice :)

sophiamersmann:
### Merge activity

* **Sep 5, 4:50 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3793).

PR #3794, opened 2024-07-16
✨ switch population variable identification from variable id to version agnostic ETL path

danyx23:
This PR changes how we recognize our population indicators in Grapher. This identification is mainly done so that we can hide population in the short citation note that is visible in grapher when it is only used to size bubbles in scatter plots or similar.

### DIFF
--- a/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
+++ b/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
@@ -101,7 +101,7 @@ const EXTERNAL_SORT_INDICATORS = [
         indicatorId: POPULATION_INDICATOR_ID_USED_IN_ENTITY_SELECTOR,
         isMatch: (column: CoreColumn): boolean =>
             isPopulationVariableETLPath(
-                (column as OwidColumnDef)?.catalogPath ?? ""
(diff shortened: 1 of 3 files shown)


packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -100,7 +100,9 @@ const EXTERNAL_SORT_INDICATORS = [
         label: "Population",
         indicatorId: POPULATION_INDICATOR_ID_USED_IN_ENTITY_SELECTOR,
         isMatch: (column: CoreColumn): boolean =>
-            isPopulationVariableId(column.slug),
+            isPopulationVariableETLPath(
+                (column as OwidColumnDef)?.catalogPath ?? ""
```
sophiamersmann, line 104: This should be `.def.catalogPath`

sophiamersmann (approved):
Looks good except for one small thing (see my comment)

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
danyx23: Of course! Thanks for catching that 🙏

PR #3798, opened 2024-07-17
garden: improvements to dev admin chart list

toni-sharpe:
**Note** no rush with this one, if it got as far as merge conflicts, I'd deal with them myself.

This improves things for me and therefore possibly also for folk who dev in the same way as me.

There was an issue, I think it has been edited, where  "World" blocks the search and removing that word yielded the one needed chart.

But that has no link to the embed screen which is really useful, so I made the ID output in the table into a link that goes straight through to the embed screen. Having the side by side comparison with the live site is really useful, at least for me eg: [this Marimekko labelling fix](https://github.com/owid/owid-grapher/pull/3581) of mine a couple of months ago. In fact it provided the proof and screen-shots of it.

I also changed the widths a bit, made the chart a bit bigger because it's visual, and narrowed the name and ID columns.

(I'm working on a MacBook Air.)

Here's the overall result:

image

### Final commit

I did this quick POC, it stacks the title and chart because both benefit from the wider space - easy to roll back

Screenshot 2024-07-20 at 04 35 45

### DIFF
--- a/adminSiteClient/ChartList.tsx
+++ b/adminSiteClient/ChartList.tsx
@@ -105,7 +105,6 @@ export class ChartList extends React.Component<{
             <table className="table table-bordered">
                 <thead>
                     <tr>
-                        <th></th>
                         <th>Chart</th>
                         <th>Id</th>
                         <th>Type</th>
--- a/adminSiteClient/ChartRow.tsx
+++ b/adminSiteClient/ChartRow.tsx
@@ -46,12 +46,14 @@ export class ChartRow extends React.Component<{
 
         return (
             <tr>
-                <td style={{
-                    display: 'flex',
-                    flexDirection: 'column',
-                    width: "400px",
-                }}>
-                    <div style={{ marginBottom: '16px' }}>
+                <td
+                    style={{
+                        display: "flex",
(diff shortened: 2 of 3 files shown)


toni-sharpe:
@larsyencken tag OK I thought, I've been scarce,

Not expecting anything else merged or started this month, doing a course (ML/neural networks).

danyx23:
@toni-sharpe thanks for experimenting with this! It looks nice but trades off higher rows with a nicer design and that means that fewer rows fit on screen. For us it's important to see many of these at once on a screen so we prefer the original design - but you can of course use your modifications on a local fork. Thanks again!

toni-sharpe:
@danyx23 fair enough, I can make, and/or share a bookmarklet or something,

Commit `8952aef1...` was actually to make the ID link to the embed screen, would it be worth PRing that in independently? Would that be helpfull for everyone?

danyx23:
Hi @toni-sharpe - I think changing the id to a link isn't particularily useful for us, no.

toni-sharpe:
No problem, I'll bookmark that one, then I'm fixed,

I'll leave this screen as is unless an issue comes through

PR #3799, opened 2024-07-18
Extract data page components

rakyi:
I checked what is currently in staging and designs in Figma to get a sense of what is most useful to extract. For example, multidimensional data page doesn't have a related charts section, so I didn't extract it. Let me know what you think about the code structure and whether extracting something else would make sense.

### DIFF
--- a/site/AboutThisData.tsx
+++ b/site/AboutThisData.tsx
@@ -1,0 +1,157 @@
+import * as React from "react"
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
+import { faArrowDown } from "@fortawesome/free-solid-svg-icons"
+import dayjs from "dayjs"
+
+import {
+    DATAPAGE_ABOUT_THIS_DATA_SECTION_ID,
+    SimpleMarkdownText,
+    ExpandableToggle,
+    HtmlOrSimpleMarkdownText,
+} from "@ourworldindata/components"
+import { DataPageDataV2, OwidOrigin } from "@ourworldindata/types"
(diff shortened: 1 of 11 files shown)


site/AboutThisData.tsx
```diff
@@ -0,0 +1,157 @@
...
+        ? dayjs(origin.dateAccessed, ["YYYY-MM-DD", "YYYY"]).year()
+        : origin.datePublished
+          ? dayjs(origin.datePublished, ["YYYY-MM-DD", "YYYY"]).year()
+          : undefined
+    if (year) return ` (${year})`
+    else return ""
+}
+
+export default function AboutThisData({
+    datapageData,
```
rakyi, line 27: It was easy to unwrap this object into top-level props in `MetadataSection`, but here it was more tricky, because we pass the whole `datapageData` to functions deep down that expect an object like `makeLastUpdated` and others in `KeyDataTable`, which are also used in grapher. Let me know if you think it's worth it to unwrap it also here.

In that case, our options are to construct ad-hoc objects to pass to each of those helper functions or also refactor the helper functions to take multiple args, instead of a single object.

danyx23 (approved):
I think this looks good! How much work do you think it would be to extract related research and related data into two components too? Related research has a good chance of making it into MDims soon, related data is more of an open question but if it doesn't take too long to component-ify it then it might be nice for completeness sake.

marcelgerber (approved):
This is great, thank you! I think this is a great starting point, and if I end up needing any other components then I can just make the necessary changes myself.

PR #3801, opened 2024-07-19
New data insight feed and detail page

rakyi:


ikesau (approved):
Looks good!

QA'd CSS on all screens, published DIs on different days, copy button still works, etc 👍

PR #3805, opened 2024-07-22
Fix HMR with SSR in dev

rakyi:
When we changed a React component in dev, the HMR kicked in for the client code, but then the page would fail to load and we had to refresh it manually. Adding the Vite middleware fixes it.

### DIFF
--- a/adminSiteServer/appClass.tsx
+++ b/adminSiteServer/appClass.tsx
@@ -30,6 +30,7 @@ import { getAndLoadGdocById } from "../db/model/Gdoc/GdocFactory.js"
 interface OwidAdminAppOptions {
     gitCmsDir: string
     isDev: boolean
+    isTest?: boolean
     quiet?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


adminSiteServer/appClass.tsx
```diff
@@ -154,8 +154,18 @@ export class OwidAdminApp {
         // error handler if it exists.
         if (bugsnagMiddleware) app.use(bugsnagMiddleware.errorHandler)
 
-        // todo: we probably always want to have this, and can remove the isDev
-        if (this.options.isDev) app.use("/", mockSiteRouter)
+        if (this.options.isDev) {
+            // https://vitejs.dev/guide/ssr
+            const { createServer } = await import("vite")
```
rakyi, line 159: @danyx23 dynamic imports currently don't work with Jest. Should I try to use their experimental support for ESM or is adding `isTest` option and skipping the import in that case ok?

danyx23: We don't need the mock site router atm, the only thing the dbtests care about is that the api REST endpoints can be reached - so I think guarding this with an isTest that we set when we run `make dbtest` should be fine.

danyx23 (approved):
Looks good and very nice to have this! The db-tests are failing but once those run through feel free to merge!

PR #3806, opened 2024-07-23
Add data insights newsletter

rakyi:
* Add DI newsletter banner to
  * Homepage
  * DI index page, animate after delay
  * DI detail page, animate after delay
* Change existing newsletter components to reference the DI newsletter instead of the immediate one
* Add new `TextInput` component based on design system in Figma and use it for newsletter emails


ikesau (changes requested):
Hot! 

Everything's good except the cookie dismissal tracking. I'll review that once you add it and then we'll be good to go 🙂

site/NewsletterSubscription.tsx
```diff
@@ -91,24 +90,30 @@ export const NewsletterSubscriptionForm = ({
             id="mc-embedded-subscribe-form"
             name="mc-embedded-subscribe-form"
             target="_blank"
+            onSubmit={() =>
+                analytics.logSiteFormSubmit(
```
rakyi, line 94: We decided on this with Bobbie.

ikesau (approved):

PR #3816, opened 2024-07-26
Add author images to data insights

rakyi:


site/gdocs/components/Image.tsx
```diff
@@ -52,6 +53,7 @@ const containerSizes: Record<ImageParentContainer, string> = {
     ["datapage"]: gridSpan6,
     ["full-width"]: "100vw",
     ["key-insight"]: gridSpan5,
+    ["author-byline"]: "48px",
```
rakyi, line 56: I wasn't able to make my browser use this size. @ikesau do you know why that might be?

ikesau: The dev server doesn't use 'em because we only resize the images during baking.

https://github.com/owid/owid-grapher/blob/7527f17767410896b567916e3ef431b703f314c2/site/gdocs/components/Image.tsx#L116

It's working correctly on staging 👍

site/DataInsightsIndexPageContent.tsx
```diff
@@ -88,15 +89,19 @@ export const DataInsightsIndexPageContent = (
 ) => {
     const { pageNumber, dataInsights, isPreviewing = false } = props
     // Extract all attachments from the data insights and supply them to the AttachmentsContext
-    const { imageMetadata, linkedCharts, linkedDocuments } =
+    const { imageMetadata, linkedAuthors, linkedCharts, linkedDocuments } =
         dataInsights.reduce(
             (acc, di) => ({
                 imageMetadata: merge(acc.imageMetadata, di.imageMetadata),
+                linkedAuthors: di.linkedAuthors
```
ikesau, line 96: While working on this, did you see anything that explained why linkedAuthors should be an array instead of a dictionary? Seems like we don't need order information, and it's resulted in a bit of extra boilerplate. 🤔

ikesau (approved):
well done!

site/gdocs/components/Image.tsx
rakyi: Well, my browser won't use the 48px even on staging. It uses `_100.png` even when 48px is available, and I'm not sure if the 48px will ever be used then. Does your browser pick the 48px?

ikesau: Ah that's a pixel density thing probably. Try testing it with your browser's Responsive Design Mode and set the DPR to 1. 🙂

site/DataInsightsIndexPageContent.tsx
rakyi: I'm not sure what you mean. Semantically linked authors are a collection of entities, not a mapping and while it's true their order doesn't matter, I wouldn't want to force it into an object to save a few lines of boilerplate.

But as I wrote this, I realized we should at least dedupe the array, so I'm adding that.

site/gdocs/components/Image.tsx
rakyi: Nice, that works. I somehow didn't realize browser tools allow you to set the DPR.

site/DataInsightsIndexPageContent.tsx
ikesau: Right, but given that we don't appear to need to iterate through them at any point, I'm wondering why the decision was made to store them in an array as opposed to dictionaries like we've done for all other attachments (I was on a break when the feature was developed)

I'm not suggesting you change it, I was just wondering if during this work you saw something that justified it that I've missed 🙂

PR #3822, opened 2024-07-29
Optimize latest data insights data

rakyi:
We could optimize the feed more, but I'm not sure if it's worth it. I haven't touched that part with the new features, i.e. after this fix, the feed page size is back to where it was before the new DI redesign.

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -101,7 +101,7 @@ import {
 import {
     gdocFromJSON,
     getAllMinimalGdocBaseObjects,
-    getAndLoadPublishedDataInsights,
+    getLatestDataInsights,
 } from "../db/model/Gdoc/GdocFactory.js"
 import { getBakePath } from "@ourworldindata/components"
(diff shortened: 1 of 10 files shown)


baker/SiteBaker.tsx
```diff
@@ -832,6 +831,10 @@ export class SiteBaker {
             }
             try {
                 await this.bakeOwidGdoc(dataInsight)
+                // We don't need latest data insights nor their images on the
+                // index page.
+                dataInsight.latestDataInsights = []
+                dataInsight.imageMetadata = attachments.imageMetadata
```
ikesau, line 837: This confused me just a little bit because I was expecting this adjustment to be done right next to the `renderDataInsightsIndexPage` call.

Would you mind clarifying that? 🙂

ikesau (approved):

baker/SiteBaker.tsx
rakyi: I tweaked it a bit, let me know if it's good enough. I kept it in the loop, because that's where we have the `attachments.imageMetadata` for each data insight.

PR #3837, opened 2024-08-01
🎉 add OWID_KOS flag

ikesau:
AFAICT this is the only country in `regions.ts` that doesn't have a flag (which was resulting in a 404 for the data catalog's new entity picker)

I got this from https://upload.wikimedia.org/wikipedia/commons/1/1f/Flag_of_Kosovo.svg and ran it through https://jakearchibald.github.io/svgomg/ at minimal precision (8.13k → 1.72k) to no visual loss in clarity that I could perceive.


ikesau:
BTFO by https://github.com/owid/owid-grapher/pull/3848

PR #3842, opened 2024-08-02
💄 improve styling of lists in markdown for description from producer

danyx23:


danyx23:
To review, compare e.g. the sources panel in these two charts (focus the grapher and press the S key to bring up the sources panel even on a data page)

https://ourworldindata.org/grapher/young-people-with-knowledge-on-hiv-prevention

http://staging-site-fix-markdown-bullet-points-in-description-producer/grapher/young-people-with-knowledge-on-hiv-prevention

ikesau (approved):
Nice! I thought there was a list-style rule that needed to be updated somewhere but it was just the margin 🙂

PR #3847, opened 2024-08-05
🔨 render thumbnails in CF workers using chart configs from R2

danyx23:
This PR changes how the CF thumnail worker gets the config for a chart. It used to be the case that it would fetch the HTML file of the grapher page at the given slug and extract the config from that HTML. Now it looks up the grapher config json file in an R2 bucket.

This PR only changes this for published charts accessed by slug. A later PR will enable this also by UUID.

Cloudflare is a bit weird with the intersection of support of various features between CF workers/pages functions, R2 and local/remote dev support: 
- Initially I wanted to use bindings in CF functions. For local development, wrangler dev does not support using a remote R2 bucket which is annoying
- I then thought I'd switch to fetching data from R2 buckets via HTTPS instead of bindings, similar to what we do with our variable data/metadata.json files. I realized then though that this only works if the bucket itself is accessible publicly via HTTPS. I'd like to keep the R2 bucket internal and not expose it directly.
- My next approach was to try to use the S3 API in the CF pages function. This didn't work because the default AWS library doesn't work in CF functions and the simpler replacement libraries that I could find were very barebones and made you deal with XML manually
- So finally I decided to go back to bindings and accepting the somewhat annoying dev story

PR #3846, opened 2024-08-05
🔨 Add wrangler.toml file

danyx23:
🔨 Add wrangler.toml file

🔨 move more settings to wrangler.toml and upgrade wrangler

PR #3867, opened 2024-08-13
🔨 change CF thumbnail function to use chart configs from R2

danyx23:
This PR changes how the CF thumnail worker gets the config for a chart. It used to be the case that it would fetch the HTML file of the grapher page at the given slug and extract the config from that HTML. Now it looks up the grapher config json file in an R2 bucket.

This PR only changes this for published charts accessed by slug. A later PR will enable this also by UUID.

Cloudflare is a bit weird with the intersection of support of various features between CF workers/pages functions, R2 and local/remote dev support:

- Initially I wanted to use bindings in CF functions. For local development, wrangler dev does not support using a remote R2 bucket which is annoying
- I then thought I'd switch to fetching data from R2 buckets via HTTPS instead of bindings, similar to what we do with our variable data/metadata.json files. I realized then though that this only works if the bucket itself is accessible publicly via HTTPS. I'd like to keep the R2 bucket internal and not expose it directly.
- My next approach was to try to use the S3 API in the CF pages function. This didn't work because the default AWS library doesn't work in CF functions and the simpler replacement libraries that I could find were very barebones and made you deal with XML manually
- I then decided to go back to bindings and accepting the somewhat annoying dev story
- But then I realized that once we get MDims, our data managers will want to make changes to charts and also to make changes to MDims and see their changes. If we fetch chart configs for MDims from CF functions in the future then on staging servers this would have been a painful story
- So in the end I decided to make the R2 buckets accessible via the public internet and use fetch instead of R2 bindings.

To test this PR, the main change to your local setup is to add an entry to your .dev.vars example to specify the path within the staging R2 bucket to use for thumbnail rendering:
```
GRAPHER_CONFIG_R2_BUCKET_PATH=devs/YOURNAME
```
Before you can test the thumnail rendering locally for a particular chart, you will either have to edit one chart in the admin to make sure it gets uploaded to R2, or run the sync script in devTolls/syncGraphersToR2

This PR adds a few new env vars for CF functions:
- GRAPHER_CONFIG_R2_BUCKET_URL - the primary bucket to read from
- GRAPHER_CONFIG_R2_BUCKET_FALLBACK_URL - the fallback bucket to read from (this should be the prod bucket; on prod it should be empty)
- GRAPHER_CONFIG_R2_BUCKET_FALLBACK_PATH - the path in the fallback bucket to read from (the path inside the prod bucket; on prod it should be empty)
- GRAPHER_CONFIG_R2_BUCKET_PATH - in wrangler.toml we set this only for prod as it's constant there. For local dev setups it should be used from the .dev.vars as per above. Staging servers fall back to the branch name.


danyx23:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3867?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3865** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3865?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3867** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3867?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3827** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3827?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3847](https://github.com/owid/owid-grapher/pull/3847) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3847?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3782** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3782?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3793](https://github.com/owid/owid-grapher/pull/3793) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3793?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3767** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3767?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3743](https://github.com/owid/owid-grapher/pull/3743) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3743?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3846** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3846?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

 Join @danyx23 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

functions/_common/grapherRenderer.ts
```diff
@@ -130,32 +135,86 @@ const extractOptions = (params: URLSearchParams): ImageOptions => {
...
 
-    // Fetch grapher config and extract it from the HTML
-    const grapherConfig: GrapherInterface = await env.ASSETS.fetch(
-        new URL(`/grapher/${slug}`, env.url)
-    )
-        .then((r) => (r.ok ? r : Promise.reject("Failed to load grapher page")))
-        .then((r) => r.text())
-        .then((html) => deserializeJSONFromHTML(html))
+    const url = new URL(`/grapher/${slug}`, env.url)
+    const slugOnly = url.pathname.split("/").pop()
```
marcelgerber, line 171: Why do convert `slug` to `url` and then back to `slugOnly`? That doesn't really make much sense.

functions/_common/grapherRenderer.ts
```diff
@@ -130,32 +135,86 @@ const extractOptions = (params: URLSearchParams): ImageOptions => {
...
-        throw new Error("Could not find grapher config")
+    let fallbackUrl
+
+    if (
+        env.GRAPHER_CONFIG_R2_BUCKET_FALLBACK_URL &&
+        env.GRAPHER_CONFIG_R2_BUCKET_FALLBACK_PATH
+    ) {
+        const topLevelDirectory = env.GRAPHER_CONFIG_R2_BUCKET_FALLBACK_PATH
+        const fallbackKey = excludeUndefined([
+            ...topLevelDirectory,
```
marcelgerber, line 198: Ooh, what this does is that it makes `fallbackKey = ["v", "1", ...]`, because you're destructuring a string here.

functions/_common/grapherRenderer.ts
```diff
@@ -130,32 +135,86 @@ const extractOptions = (params: URLSearchParams): ImageOptions => {
...
     }
 
-    grapherLogger.log("fetchGrapherConfig")
+    // Fetch grapher config
+    const fetchResponse = await fetchFromR2(requestUrl, etag, fallbackUrl)
+
+    if (fetchResponse.status !== 200) {
+        console.log("Failed to fetch grapher config", fetchResponse.status)
+        return null
+    }
```
marcelgerber, line 214: Have you tested what happens if the etag matches? Is the response status 200, or is it `304 Not Modified`?

marcelgerber (commented):
Overall, this makes sense to me, and I'm very happy to see `deserializeJSONFromHTML` used in one less place :)

Had a few comments, though.

Review note: I reviewed this before having a look at #3827.

marcelgerber:
A few comments about your comment above, also:

> Cloudflare is a bit weird with the intersection of support of various features between CF workers/pages functions, R2 and local/remote dev support:

Agreed!

> 
> * Initially I wanted to use bindings in CF functions. For local development, wrangler dev does not support using a remote R2 bucket which is annoying

It is annoying; however, at least for non-Pages workers, there is the option to run `npx wrangler dev --remote`, which will give access to a real R2. Sadly, it seems there is no equivalent for Pages Functions :(

> * But then I realized that once we get MDims, our data managers will want to make changes to charts and also to make changes to MDims and see their changes. If we fetch chart configs for MDims from CF functions in the future then on staging servers this would have been a painful story

Why is that? What would the painful dev story be here, in particular?

wrangler.toml
```diff
@@ -6,12 +6,19 @@ pages_build_output_dir = "./localBake"
...
+GRAPHER_CONFIG_R2_BUCKET_FALLBACK_URL = "https://grapher-configs.ourworldindata.org"
+GRAPHER_CONFIG_R2_BUCKET_FALLBACK_PATH = "v1"
+
 
 # Overrides for CF preview deployments
 [env.preview.vars]
 MAILGUN_DOMAIN = "mg.ourworldindata.org"
 SLACK_ERROR_CHANNEL_ID = "C016H0BNNB1"
 ENV = "preview"
+GRAPHER_CONFIG_R2_BUCKET_URL = "https://grapher-configs-staging.ourworldindata.org"
```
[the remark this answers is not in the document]
danyx23: Yeah, valid point. Let's chat about it next week with Lars

functions/_common/grapherRenderer.ts
danyx23: Ah damn, good catch! There used to be an array here in an earlier revision.

functions/_common/grapherRenderer.ts
danyx23: Totally, that was a leftover from an earlier iteration

functions/_common/grapherRenderer.ts
```diff
@@ -194,17 +253,24 @@ export const fetchAndRenderGrapher = async (
     slug: string,
     searchParams: URLSearchParams,
     outType: "png" | "svg",
-    env: Env
+    env: Env,
+    etag?: string
```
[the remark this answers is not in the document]
marcelgerber: Yeah, I've seen in the upstack PR that for `slug.config.json`, we obviously have a way to provide an etag. So this comment is somewhat moot now.

danyx23:
Thanks @marcelgerber! Yes I saw that for normal CF workers the R2 story would be better with bindings but as you say, for pages functions it's not there yet.

To elaborate a bit on why R2 bindings would have been painful: we have the code that runs in a CF pages function that could use R2 bindings, and we have the code in the admin that stores configs via the S3 api. CF pages functions in local dev can't connect to a real R2 bucket and so would always try to access files from a local/"fake" R2 bucket. The S3 API on the other can only work with a real R2 bucket and has no implementation to work against a local/"fake" R2 bucket. This means that the admin would have written to the real bucket and the function would have read from local. 

Changing this would have meant to either make the admin be able to write into local R2 (probably via invoking the `wrangler` tool as a process) or providing a tool and requiring users to explicitly sync from the actual R2 bucket to the local one. 

The latter I thought was ok for local dev work but then I realized that on our staging servers, data managers will want to make changes and have them reflected in what they see. Today this is only thumbnails that use these configs but very soon, indicator based explorers will do this too and then this would have been a deal breaker.

danyx23:
### Merge activity

* **Sep 10, 6:33 AM EDT**: @danyx23 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3867).

PR #3912, opened 2024-08-30
Add a way of marking articles as deprecated

rakyi:


rakyi:
This is ready for review, but there is a chance the design will change slightly.

site/gdocs/components/centered-article.scss
```diff
@@ -195,6 +195,62 @@ $banner-height: 200px;
     margin-bottom: 32px;
 }
 
+.deprecation-notice {
+    --bg-color: #fff5d8;
```
ikesau, line 199: image

need to zindex this higher, check out `variables.scss` 🙂

ikesau: The stickiness doesn't seem to work quite right when I do screen emulation. Not sure if that'll be an actual issue on mobile.

site/gdocs/pages/GdocPost.tsx
```diff
@@ -129,7 +138,23 @@ export function GdocPost({
...
+                        >
+                            Cite this work
+                        </h3>
+                        {isDeprecated && (
+                            <p className="citation-deprecated-notice">
+                                <span className="citation-deprecated-notice__highlight">
+                                    This content is outdated
+                                </span>{" "}
+                                but if you would still like to use it, here is
+                                how to cite it.
```
ikesau, line 154: I think there should be a comma before the but, and a colon instead of period at the end.

```suggestion
                            <p className="citation-deprecated-notice">
                                <span className="citation-deprecated-notice__highlight">
                                    This content is outdated
                                </span>
                                , but if you would still like to use it, here is
                                how to cite it:
```

site/gdocs/components/centered-article.scss
ikesau: image

I don't think there shouldn't be yellow space between the header and the body on small when the notice is present 🙂

ikesau (changes requested):
Looks good overall! Just a few code nitpicks really.

Do you know if we still want to index deprecated articles and show them in the sitemap?

site/gdocs/OwidGdocPage.tsx
```diff
@@ -89,18 +89,23 @@ export default function OwidGdocPage({
     const isDataInsight = gdoc.content.type === OwidGdocType.DataInsight
     const isAuthor = gdoc.content.type === OwidGdocType.Author
 
+    let imageUrl
+    if (
+        gdoc.content.type === OwidGdocType.Article &&
```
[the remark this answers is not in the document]
rakyi: I added the constraint to fix typing error. But good idea to add the validation. I added it, pls check if I did it correctly.

site/gdocs/components/centered-article.scss
rakyi: Thanks, I didn't know about the z-index variables. I added a new one for the deprecation notice.

I can't reproduce the issue with stickiness on with screen emulation.

The yellow space between the header and the body I also can't reproduce. Maybe it's because of the missing date (don't know why it would be missing)?

rakyi: Ah, now I reproduced the stickiness issue with small screen. Looking into it.

rakyi: Pushed a fix for that.

ikesau:
Hey @rakyi sorry I went ahead and yak-shaved on your work without your blessing, but I realized earlier today in the cursor demo that my suggestion of using `0 | 1` still felt a bit code-smelly to me, so I investigated turning it into a boolean instead which MySQL can't do, and well, one thing led to another 😓

The only truly critical fix I made was in [GdocFactory](https://github.com/owid/owid-grapher/pull/3912/files/ccbadf41b83cea3ecd9b070f977b7ae9ed777aa9#diff-768f65930211cd5b0f2d281ca19127202bb5a21c9ab6cf32259f7515616109f3R233) where I noticed a leading slash was there where it shouldn't be, which is what made me realize we should be using constants for these 2 filenames.

If you'd prefer, you can revert my changes (though you'll need to fix the leading slash again) and I'll add the constants in a separate PR.

Also, do we want to exclude archived posts from the sitemap and search index? 🙂

site/gdocs/components/centered-article.scss
ikesau: And yup, looks like it's only an issue when there's no publishedAt date, so no worries there 🙂

rakyi:
Thank you for the changes!

> Also, do we want to exclude archived posts from the sitemap and search index? 🙂

I forgot to reply to this before. We shouldn't exclude these posts from search and sitemap.

ikesau (approved):

ikesau:
Once you merge this, don't forget to add an example to the article documentation gdoc 🙂 

You'll probably have to write it in the body and escape it so that ts-pattern doesn't complain - otherwise the docs themselves will have the deprecation notice which we probably don't want.

rakyi:
Yeah, I was meaning to do that. Thank you for the tip!

PR #3923, opened 2024-09-04
🐛 (admin) mark unpublished charts as drafts

sophiamersmann:
- Fixes a regression in the admin where unpublished charts weren't marked as "draft" in the admin's chart list
- This happened because the admin expected `isPublished` to be a boolean when really it was a string value that is either `"true"` or `"false"`
- This went unnoticed before having full configs because `isPublished` used to be either `null` (falsy) or `"true"` (truthy)
- I'm a bit surprised that these are parsed as strings by knex, maybe there is a deeper issue to fix? The db looks fine though, as far as I can see


sophiamersmann:
* **#3923** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

danyx23 (changes requested):
Probably the nicer way to do this is to change how isPublished is queried in the `oldChartFieldList` so that it works like hasMapTab. The reason for this behaviour is that we don't have json auto-parsing enabled in mysql2 so if you query a json value we get it's string representation that we manually have to query. The `JSON_EXTRACT(chart_configs.full, "$.hasChartTab") = true AS hasChartTab` clause (could also be done with `->>` though of course) then makes sure that the value we get from mysql is a boolean and them mysql2 can just give back that boolean

PR #3926, opened 2024-09-05
🐛 (grapher) delete stackMode=null from chart configs

sophiamersmann:
In #3793, I removed `null` as an option from the Grapher schema. I thought I checked that we didn't have any nulls in the database, but apparently, I was wrong (see [the validation error in Slack](https://owid.slack.com/archives/C5JJW19PS/p1725531937760169)). This PR writes a migration to remove nulls from the database. I'm only doing this for chart configs since none of our indicator charts has the `stackMode` set to null.

I don't think null values currently have a way to get into the database. But if it happens, the validator will warn us again, and we can act then.

### DIFF
--- a/db/migration/1725540224795-CleanUpChartConfigStackMode.ts
+++ b/db/migration/1725540224795-CleanUpChartConfigStackMode.ts
@@ -1,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class CleanUpChartConfigStackMode1725540224795
+    implements MigrationInterface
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#3926** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3926?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

danyx23 (approved):
Nice, looks good!

PR #3929, opened 2024-09-05
Add active state to admin nav

rakyi:
I thought it would be nice to make it easier for users to see where they are in the admin.

image


sophiamersmann (approved):
Nice!

PR #3933, opened 2024-09-06
🎉 add admin image validation for data insights

ikesau:
A follow-up to https://github.com/owid/owid-grapher/pull/3932 to ensure authors don't mistakenly upload images in the wrong file format.

Made non-PNGs a warning because there's still a plausible case where JPEG is better for a custom graphic... maybe 🙊


rakyi (approved):
LGTM!

PR #3948, opened 2024-09-10
Update 404 page to the new design

rakyi:
image


site/NotFoundPage.tsx
```diff
@@ -1,48 +1,39 @@
...
-                            aria-label="Submit search"
-                            className="btn"
-                            type="submit"
-                        >
-                            Search
-                        </button>
-                    </form>
+                    <p className="body-2-semibold">
+                        You can search for what you were hoping to find below or{" "}
+                        <a href="/">visit our homepage</a>.
```
ikesau, line 30: This link is missing a hover style - is it meant to be `owid-link-90` ? (we should probably add a util class for that mixin, like we do for typography)

ikesau (approved):
lgtm!

the recursive 404 page illustration makes me laugh but in a good way

site/NotFoundPage.tsx
rakyi: Almost, it should be blue-60 color. I'll still use the mixin.

PR #3955, opened 2024-09-13
:bug: Fix chart copy slug issue

### LINKED ISSUE #3954: When you make a copy of the chart, the copy keeps the slug
## Problem

When you make a copy of a Grapher chart in the admin, the copy inherits all the configuration _including the slug_. This causes a range of issues:

- It means that slugs are not unique in the database
- It means you cannot delete a draft, since it refers to a slug that is being used

## Proposed solution

The slug should be removed when making a chart copy.

larsyencken:
Fixes #3954

When saving a chart as a new copy, delete the `slug`.


sophiamersmann (approved):
Thanks! We should have done this a long time ago 🫠

PR #3963, opened 2024-09-16
🐛 fix global entity selector

### LINKED ISSUE #3931: Topic-specific country profile pages don't show data for the country
## Description

On topic-specific country profile pages, charts use the default country selection instead of the specific country. For example, [here](https://ourworldindata.org/energy/country/spain) charts should show the data for Spain.

It seems to work correctly for explorers.

## Additional context

* Related [Slack discussion](https://owid.slack.com/archives/C46U9LXRR/p1725613179393189)

sophiamersmann:
Fixes #3931 

The global entity selector manages the selection of a chart from outside by passing in its own `selection` object, which should take precedence over the config's `selectedEntityNames`.  

In #3793, Grapher's `updateFromObject` method was updated to also change the current selection if `selectedEntityNames` is given. That's sensible (and is needed in the admin), but `updateFromObject` is also called in the constructor. If `selectedEntityNames` and `selection` both are given then `selectedEntityNames` always won, which broke the global entity selector (since it relies on `selection` being respected).

I think it's sensible to apply the passed-in selection only if it's non-empty, but this leads to funny behaviour of the global entity selector if the selection happens to be empty (e.g., http://staging-site-fix-global-entity-selector/energy/country/spain?country=). But, apparently, even before #3793, the global entity selector has behaved weirdly for empty selections (e.g., https://e7fd384b.owid.pages.dev/energy/country/spain?country=). It's not ideal, but I don't think it's too bad.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -65,6 +65,7 @@ import {
     getOriginAttributionFragments,
     sortBy,
     extractDetailsFromSyntax,
+    omit,
 } from "@ourworldindata/utils"
 import {
     MarkdownTextWrap,
@@ -505,7 +506,14 @@ export class Grapher
 
         if (props) this.setAuthoredVersion(props)
 
-        this.updateFromObject(props)
+        // prefer the manager's selection over the config's selectedEntityNames
+        // if both are passed in and the manager's selection is not empty.
+        // this is necessary for the global entity selector to work correctly.
+        if (props.manager?.selection?.hasSelection) {
+            this.updateFromObject(omit(props, "selectedEntityNames"))
+        } else {
+            this.updateFromObject(props)
+        }
 
         if (!props.table) this.downloadData()
 


sophiamersmann:
* **#3963** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3963?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

ikesau (approved):
Nice! Fixes the issue 👌 

I think that the empty query param thing is only be an issue if someone gets linked to it that way.

If the user did it to themselves by unselecting the selected country, then I think they would understand why the charts aren't rendering properly.

So maybe, if you care enough, you could prune the empty `?country=` query param when the page mounts, before the embedder has started doing anything, and I think that should be fine because http://staging-site-fix-global-entity-selector/energy/country/new-zealand works?

Feel free to merge without that, though 🙂

sophiamersmann:
Yeah, that's a good suggestion. Thank you, Ike! If I have time this week, I'll follow up with another PR, but I'll merge as-is for now to get the bug fixed.

sophiamersmann:
### Merge activity

* **Sep 17, 5:46 AM EDT**: @sophiamersmann started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/3963).

PR #3967, opened 2024-09-17
✨ automatically migrate outdated configs / TAS-623

### LINKED ISSUE #3936: The saveChart API at /charts should be able to accept configs with old schema versions and migrate them
In the ETL we have an updated cycle that only touches datasets once a year or less often. When writing partial grapher configs there, it is therefore normal for them to be written using an old schema version. The grapher API that the ETL uses to save the grapher configs defined in the ETL should thus be able to accept old versions and transparently update them to the new version wherever possible (usually this means dropping obsolete fields or rewriting renamed enums etc)

sophiamersmann:
Fixes #3936 (I found v3 configs in the database, so I decided to tackle this earlier rather than later)

### Summary

Adds a new `migrations` folder to the `grapher/schema` directory. `migrations/migrations.ts` contains the actual migration functions. I wrote migrations for all breaking changes that the schema has seen so far.

The exported function `migrateGrapherConfigToLatestVersion` reads the config's schema version and applies a series of migration functions if necessary. If a config's schema field is missing or invalid, it throws. (This is different from the initial implementation! We used to assume a config adheres to the latest version if the schema field was missing.)

The types are set up so that if you increase the schema version of the default grapher config and forget to add a migration, TypeScript will let you know.

When the API receives a Grapher config, we try to convert it to a valid config as soon as possible so that the rest of the code can reasonably expect a config of type `GrapherInterface`.

To add a new migration, one would edit `migrations/migrations.ts`:
- Write a migration function that updates an outdated config to the next version
- Add a clause to the `runMigration` function that calls the new migration function

### Before merging

> [!IMPORTANT]  
> Just before merging, check if configs of more outdated versions are currently in our prod database. At the time of writing, we only have a bunch of v3 configs.

### After merging

- [ ] merge https://github.com/owid/automation/pull/13


sophiamersmann:
* **#3967** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/schema/migrations/helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+const schemaVersionRegex =
+    /https:\/\/files\.ourworldindata\.org\/schemas\/grapher-schema\.(?<version>\d{3})\.json/m
+
+const isValidSchemaVersion = (version: string): version is SchemaVersion =>
+    allSchemaVersions.includes(version as any)
+
+export function getSchemaVersion(
+    config: AnyConfigWithValidSchema
+): SchemaVersion
+export function getSchemaVersion(config: AnyConfig): SchemaVersion | null
```
danyx23, line 30: Oh this is very neat - I didn't know typescript could do function overloading to narrow down types. Thanks for this!

db/migration/1726588731621-MigrateOutdatedConfigsToLatestVersion.ts
```diff
@@ -0,0 +1,33 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class MigrateOutdatedConfigsToLatestVersion1726588731621
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        // we have v3 configs in the database; turn these into v5 configs
+        // by removing the `data` and `hideLinesOutsideTolerance` properties
+        await queryRunner.query(
```
danyx23, line 9: Would it be very annoying to just run our migration code here? It seems that this is a bit fragile and while it will work on prod if we check first, if we just use our migration code then this could work also on old, outdated staging servers or similar if we ever wanted to use it in such a context.

danyx23 (commented):
This looks very good! I really like how you made the schema versions somewhat typesafe :).

I have two minor higher level conerns: my hunch is that it would be better to migrate chart configs in the saveGrapher function instead of in the individual handlers. We'd loose the ability to type saveGrapher correctly (Grapherlnterface would be a lie), but in return every time we use saveGrapher we'd get the correct migration behaviour. One codepath where this would be missing AFAICS in the current implementation is the bulk chart editor. It uses the saveGrapher function but there is no explicit migration added for that codepath. Doing this in saveGrapher seems safer?

The other concern I have is that there are quite a few partial grapher configs in the ETL that don't specify a schema and if I understand this change they would then start breaking — is that right? Or is there a simitar change in the ETL code that sets a reasonable default (like version 603 which I think is when most of the ett code was authored)?

I think it would make sense for the ETL grapher config layer update API handler to set a default schema version if none is given but also to encourage data managers to always specify a schema version or for the updater on the ETL side to specify a reasonable default.

Let me know what you think!

db/migration/1726588731621-MigrateOutdatedConfigsToLatestVersion.ts
sophiamersmann: yes, we can also just run the migration code here 👍🏻

sophiamersmann:
> I have two minor higher level conerns: my hunch is that it would be better to migrate chart configs in the saveGrapher function instead of in the individual handlers. We'd loose the ability to type saveGrapher correctly (Grapherlnterface would be a lie), but in return every time we use saveGrapher we'd get the correct migration behaviour. One codepath where this would be missing AFAICS in the current implementation is the bulk chart editor. It uses the saveGrapher function but there is no explicit migration added for that codepath. Doing this in saveGrapher seems safer?

Makes sense!

> The other concern I have is that there are quite a few partial grapher configs in the ETL that don't specify a schema and if I understand this change they would then start breaking — is that right? Or is there a simitar change in the ETL code that sets a reasonable default (like version 603 which I think is when most of the ett code was authored)?
> 
> I think it would make sense for the ETL grapher config layer update API handler to set a default schema version if none is given but also to encourage data managers to always specify a schema version or for the updater on the ETL side to specify a reasonable default.

Yeah, I'm a bit on the fence about this. My initial implementation was forgiving (it just assumed the latest version when missing), but then I got nervous about having schemas in our system that are incorrectly versioned. Maybe the safest thing to do for missing versions is to assume v1 and then run all migrations?

sophiamersmann:
I chatted with Mojmir, and he'll make it so that configs authored in the ETL set the `$schema` field, which means it's ok to return an error on our side.

danyx23 (approved):
Nice, looks good!

PR #3970, opened 2024-09-17
Fix grapher 404 page

rakyi:
This is an alternative fix for the one reverted in 24f6288.

### DIFF
--- a/functions/grapher/[slug].ts
+++ b/functions/grapher/[slug].ts
@@ -102,6 +102,7 @@ async function handleHtmlPageRequest(
     const grapherPageResp = await env.ASSETS.fetch(url, { redirect: "manual" })
 
     if (grapherPageResp.status === 404) {
+        // grapherPageResp should be a static 404 HTML page.
         return handlePageNotFound(env, grapherPageResp)
     }
 
--- a/functions/grapher/[slug].ts
+++ b/functions/grapher/[slug].ts
@@ -1,9 +1,10 @@
 import { Env } from "../_common/env.js"
 import {
-    getOptionalRedirectForSlug,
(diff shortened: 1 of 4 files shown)


ikesau (approved):
Tested locally with grapher routes that should 404 and grapher routes that should redirect and they all worked 👍

functions/grapher/[slug].ts
```diff
@@ -146,7 +102,7 @@ async function handleHtmlPageRequest(
     const grapherPageResp = await env.ASSETS.fetch(url, { redirect: "manual" })
 
     if (grapherPageResp.status === 404) {
-        throw new StatusError(404)
+        return handlePageNotFound(env, grapherPageResp)
```
Comment (ikesau) [inline functions/grapher/[slug].ts:105]: One thing I noticed while tinkering with the code, the `For local testing` commented-out code block above here could be updated to call `fetch` from `env.ASSETS`

Comment (danyx23) [reply functions/grapher/[slug].ts:105]: I think a comment here would be useful that mentions that by passing in the grapherPageResp as the fallback we get a 404 status code but also an HTML 404 page. Otherwise this subtle detail might be lost in bigger refactorings in the future.

danyx23 (approved):
I checked some of the new routes as well and it all looks good from my side too - thanks!

functions/grapher/[slug].ts
Comment (rakyi) [reply functions/grapher/[slug].ts:106]: Added a comment about `grapherPageResp`.

> commented-out code block above here could be updated to call `fetch` from `env.ASSETS`

Even if it works, I think it makes sense to leave it to a normal fetch, since we are not fetching from the assets (semantically), we are fetching from the internet.

PR #3973, opened 2024-09-18
✨ (map) left-align categorical color legend / TAS-626

### LINKED ISSUE #3988: Left-align categorical map legends
### Motivation

When a map has many categories, then centre-aligned legends are difficult to read. This is especially true for mobile chart exports.

Example: https://ourworldindata.org/grapher/forms-of-homelessness-included-in-available-statistics

| Before  | After  |
| ------- | ------ |
| forms-of-homelessness-included-in-available-statistics (1)  | forms-of-homelessness-included-in-available-statistics |

See more examples: https://github.com/owid/owid-grapher-svgs/commit/93e1b3cbb765516dbb53ccb259a67fe68ccb4a9a

### Pro/contra

- Charts legends on top of the chart are left-aligned as well
- Numerical map legends would still be centre-aligned
- Numerical+categorical map legends would also still be centre-aligned
- Left-aligning single-line map legends look off since the map is centre-aligned 

### Slack

Slack: https://owid.slack.com/archives/C04VBQ00HQR/p1726656315000319

Marwa's notes:
Thanks for bringing this up Charlie! Here are my thoughts:
- I think the best solution would be, like you said, to left-align the legend when it’s on multiple lines and to keep it centered when it’s just on one line. I don’t really see an issue with consistency there tbh… As long as we have a clear rule and aren’t aligning the legend haphazardly
- Having said that, if we could only opt for one alignment I would chose to left-align. I see what you mean by it looking funky and I think that if it were placed below the subtitle, then it would be less of an issue (but I think that’s a bigger change that we may not want to commit to right now, just food for thought)
- Another idea could also be to just left align the legend on mobile and keep it centered on desktop

We agreed to try left-align multi-line legends only first

sophiamersmann:
Experimenting with left-aligned categorical color legends.

Keeping open for some more time to keep the staging site around.

Resolves #3988

### DIFF
--- a/packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
+++ b/packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
@@ -515,6 +515,9 @@ export class MapChart
     }
 
     @computed get legendX(): number {
+        if (this.legendAlign === HorizontalAlign.left) {
+            return this.bounds.x + 1
+        }
         return this.bounds.x + (this.bounds.width - this.legendMaxWidth) / 2
     }
 
@@ -552,7 +555,9 @@ export class MapChart
     }
 
     @computed get legendAlign(): HorizontalAlign {
-        return HorizontalAlign.center
+        return this.numericLegend
+            ? HorizontalAlign.center
+            : HorizontalAlign.left
     }
 
     @computed get numericLegendY(): number {


sophiamersmann:
* **#3973** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/3973?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

sophiamersmann:
Closing for now, will revisit when I have the time!

PR #3975, opened 2024-09-18
:sparkles: Enable explorers on staging servers

Marigold:
Load `GIT_CMS_DEFAULT_BRANCH` from env if available. This will be useful for explorers on staging servers.


marcelgerber (commented):
I checked the usages of `GIT_CMS_DEFAULT_BRANCH`, and in all cases it is only used to create GH links of the sort `github.com/org/repo/commits/GIT_CMS_DEFAULT_BRANCH`.
It doesn't in any way change which branch any commits are committed to and pushed to - that's entirely based on what the git working copy of `owid-content` is checked out to.

I hope you are aware of this, @Marigold?

marcelgerber (approved):
Cool, then this is good to go

PR #3979, opened 2024-09-19
Add gdoc tombstones

### LINKED ISSUE #3902: Add the ability to mark a page as deleted
For deleted content, i.e. content that should no longer show up at its old URL, we want to have an easy way to create special 404 pages.

* Allow both new soft delete and old actual delete of a page (current behavior)
* For new soft delete, allow users to specify:
  * An optional reason why we deleted the page
  * An optional link to a related page
* Implement the new page design
* Remove the page from search and sitemap

rakyi:
Closes #3902

Without reason or related link specified:

image

With reason and related link specified:

image

Generic 404:

image

Modified existing delete modal in the admin (remaining fields are visible only when the checkbox is checked):

image


baker/SiteBaker.tsx
```diff
@@ -633,6 +655,24 @@ export class SiteBaker {
...
     }
 
+    async bakeGDocTombstones(knex: db.KnexReadonlyTransaction) {
+        if (!this.bakeSteps.has("gdocTombstones")) return
+        const tombstones = await getTombstones(knex)
+
+        for (const tombstone of tombstones) {
+            const attachments = await this.getPrefetchedGdocAttachments(knex)
+            try {
+                await this.bakeOwidGdocTombstone(tombstone, attachments)
```
ikesau, line 665: It's possible to set an invalid `relatedLink` (with either an unpublished or unregistered gdoc) in the admin

We should throw an error during the save attempt if the `relatedLink` is invalid, and then also log an error to Bugsnag here if the `relatedLink` is invalid during bake (i.e. the `relatedLink` gdoc has been unpublished)

`ProminentLink` has a graceful fallback so I think just making sure people know it's happening through Bugsnag is sufficient.

site/owid.scss
```diff
@@ -728,88 +728,46 @@ a.ref sup {
...
-        cursor: pointer;
-        outline: none;
-        text-decoration: none;
-        box-sizing: border-box;
-        touch-action: manipulation;
-        background: #1865f2;
-        color: #ffffff;
-    }
-}
-
```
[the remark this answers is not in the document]
ikesau: Oops, thanks 🙏

ikesau (commented):
Nice! I like it 👍 

I wonder if we should have some admin UI to manage tombstones with, though. Obviously I couldn't resist trying to redelete a slug that already had a tombstone (which fails) plus I can imagine people potentially wanting to read/update/delete the tombstones. Just a simple table would work well, I think?

adminSiteClient/GdocsMoreMenu.tsx
```diff
@@ -90,14 +85,143 @@ export const GdocsMoreMenu = ({
...
+                            Create tombstone page
+                        </Checkbox>
+                    </Form.Item>
+                )}
+                {shouldCreateTombstone && (
+                    <>
+                        <Form.Item
+                            name="reason"
+                            label="Reason for removing the article"
+                            extra="Will use a generic default message if left blank."
```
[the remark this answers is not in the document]
rakyi: Changed.

ikesau (approved):

baker/GDriveImagesBaker.tsx
```diff
@@ -25,7 +25,10 @@ export const bakeDriveImages = async (
     const images: Image[] = await db
         .knexRaw<DbRawImage>(
             knex,
-            `SELECT * FROM images WHERE id IN (SELECT DISTINCT imageId FROM posts_gdocs_x_images)`
+            `SELECT *
+             FROM images
+             WHERE id IN (SELECT DISTINCT imageId FROM posts_gdocs_x_images)
+                OR filename IN (SELECT DISTINCT relatedLinkThumbnail FROM posts_gdocs_tombstones)`
```
rakyi, line 31: Does this make sense, @ikesau?

rakyi:
@ikesau could you review the new comments since your last review, pls?

baker/GDriveImagesBaker.tsx
rakyi: BTW, is subquery faster than a join, or why are we doing that?

ikesau: Will finish the review tomorrow but one thing that jumps out here is that we could get duplicates in `images` now, I think? Maybe `SELECT DISTINCT *` could help.

I don't know of a particular reason to subqueries over joins in this case. Feel free to change it to whichever's more legible.

site/TombstonePage.tsx
```diff
@@ -43,10 +51,11 @@ export default function TombstonePage({
                         <p className="body-3-medium">
                             {reason || DEFAULT_TOMBSTONE_REASON}
                         </p>
-                        {!relatedLink && (
+                        {includeArchiveLink && (
                             <p className="body-3-medium">
                                 If you’d still like to view this page, you can
-                                find it{" "}
+                                find it via the Internet Archive's Wayback
+                                Machine{" "}
```
ikesau, line 58: Especially because this wraps in an unflattering way on desktop

image

it's generally preferable to use more descriptive/explanatory text for anchor tags for screen readers (so that pressing tab doesn't just say the word "here")

e.g. 

> If you’d still like to view this page, you can check the [Internet Archive Wayback Machine](https://web.archive.org/web/*/http://localhost:3030/ikes-test-article).

site/gdocs/components/centered-article.scss
```diff
@@ -1000,6 +1000,18 @@ div.raw-html-table__container {
...
+
+    .prominent-link__heading-wrapper {
+        display: flex;
+        justify-content: space-between;
+        gap: 8px;
+    }
+
+    .prominent-link__icon-external {
+        color: $blue-60;
+        margin-top: 4px;
```
ikesau, line 1014: based on my recent experience implementing such things, I imagine Marwa might like this to darken slightly when the PL is hovered. 🙂

ikesau (approved):
Nice! Tested all the cases and it all seems good.

2 teeny tiny little comments 🤏

site/TombstonePage.tsx
rakyi: Good point, changed.

PR #3980, opened 2024-09-20
Update ESlint plugins

rakyi:
I wanted to fix the warning we get when running ESlint after update to TypeScript 5.6, but latest typescript-eslint still doesn't support it: https://github.com/typescript-eslint/typescript-eslint/issues/9653

So we at least got the other updates and related lint fixes.


rakyi:
Seems to be also quite faster, almost 50% on my machine.

marcelgerber (approved):
Nice! I pushed another commit to dedupe the lockfile (`yarn dedupe [--check]`).

> Seems to be also quite faster, almost 50% on my machine.

I assume that the new version of TS-eslint is clever enough to realize that we don't currently use any rules that require typechecking.
We may want to do that, though, but that's another PR :)

PR #3997, opened 2024-10-02
✨ remove `rel="noreferrer"` from site `<a>` tags

ikesau:
## Context
We've had an unspoken convention here to omit referrer information from our links in many places throughout the codebase.

Specifically in gdocs pages, it's worked 3 ways:
1. gdoc links (linking to another article via its gdoc URL): **referrer information included**
2. grapher/explorer links (linking via a grapher/explorer url): **referrer information included**
3. url links (when we write out the complete URL): **referrer information omitted**

When linking to an OWID resource that isn't a grapher/explorer/gdoc page, we have to use url links, e.g https://ourworldindata.org/data and so this unspoken convention has hurt our ability to understand how users are navigating our site.

## Change
This PR removes the `noreferrer` attribute from url links in gdoc articles (see: the `LinkedA` component, and the `ProminentLink` component) but also in all other <a> tags on the site to prevent confusion about what our convention is.

`target="_blank"` is still present to open the links in new tabs, and `rel="noopener"` is still present because, even though it's redundant as of 2024 (see [`target="_blank" implies rel="noopener" behavior`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility)), eslint is still requiring us to set both.

### DIFF
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -863,7 +863,7 @@ export class Explorer
                     <a
                         href={this.explorerProgram.downloadDataLink}
                         target="_blank"
-                        rel="noopener noreferrer"
+                        rel="noopener"
                         className="ExplorerDownloadLink"
                     >
                         Download this dataset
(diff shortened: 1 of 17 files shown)


danyx23 (approved):
Looks good! Sorry for the late review!

PR #4003, opened 2024-10-03
✨ reordered owid distinct colors

### LINKED ISSUE #3898: Bug: colors in line charts are often very similar
## Description
Line colors in line charts colors are sometimes very similar depending on the entities selected.
 
## Expected behavior
All lines in the chart should have colors that are easy to distinguish.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/excess-mortality-p-scores-projected-baseline
2. Follow the steps displayed in the first video below

## Additional context and screenshots

After playing with selecting and unselecting entities in [this chart](https://ourworldindata.org/grapher/excess-mortality-p-scores-projected-baseline), I realized that the lines were not very distinct. I am a bit confused by the rules for the Grapher when choosing colors.

https://github.com/user-attachments/assets/8bbcf9da-291a-483e-aa60-41b8aa8dffcd

If one refresh, the chart is loaded with distinguishable colors (second video).

https://github.com/user-attachments/assets/4b7b7cb6-bdae-49e1-a793-f2f67192f71b

I think this often leads to DI charts with similar line colors.

## Environment

For desktop, provide:

-   OS: MacOS
-   Browser: Arc
-   Version: 1.57.1 (Chromium Engine Version 128.0.6613.85)

danyx23:
This fixes #3898


lucasrodes:
hey @danyx23,
Thanks for the rapid implementation! I think this fixes a substantial part of https://github.com/owid/owid-grapher/issues/3898, but the initial confusion about how colors are chosen is still not resolved. 

I think I understood how colors are picked, and wonder if this is the way to go. Elaborated on this here: https://github.com/owid/owid-grapher/issues/3898#issuecomment-2393155362

danyx23:
@lucasrodes yes, this only implements a better order of colors which makes the default color choices on many of our charts substantially better.

I see your other point about how selecting and deselecting is a bit weird but there are several other concerns that play into that part of the codebase (e.g. persistence of colors on entities coming in via facetting or explorers). It's possible that we can improve that in the future but this is not a quick fix.

Issue #3019, opened 2023-12-13
Validate and improve error reporting for baking (e.g. missing DoDs)


ikesau:
We're meant to have a baking step called `validateGrapherDodReferences` that alerts Bugsnag if we're baking a grapher that references a dod that no longer exists.

But recently, ~80 dods got eaten due to a parsing glitch and no errors were logged to bugsnag, which strains credulity.

We should double check that this step is working as intended and also add a similar check for dods referenced in gdocs.

PR #3911, opened 2024-08-29
🔨 correctly initialize bugsnag in SiteBaker

ikesau:
Errors in the site baker that are called with `logErrorAndMaybeSendToBugsnag` haven't been logging to Bugsnag this whole time because the client wasn't being initialized before we started the bake process. 🫠 

This PR fixes that, and also fixes #3019

We have 7 errors across 16 published articles in the site baker at the moment that will start coming through once we merge this:

| error                                                                                 | source(s)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
|---------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Explorer chart with slug coronavirus-data-explorer does not exist or is not published | "pandemics", "influenza", "covid-testing-data-archived", "the-future-is-vast", "metrics-explained-covid19-stringency-index", "covid-exemplar-germany", "covid-exemplar-vietnam", "covid-exemplar-south-korea", "covid-sweden-death-reporting", "covid-health-economy", "whats-new-june-july-2020", "covid-exemplar-vietnam-2020", "covid-exemplar-south-korea-2020", "covid-exemplar-germany-2020", "identify-covid-exemplars","covid-excess-mortality" |
| Article has no tags set. We won't be able to connect this article to datapages        | "the-future-is-vast", "metrics-explained-covid19-stringency-index", "covid-exemplar-germany", "covid-exemplar-vietnam", "covid-exemplar-south-korea", "covid-health-economy", "whats-new-june-july-2020", "covid-exemplar-vietnam-2020", "covid-exemplar-south-korea-2020", "covid-exemplar-germany-2020", "identify-covid-exemplars"                                                                                                                               |
| Author "Cameron Appel" does not exist or is not published                             | "covid-testing-data-archived"                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
| Author "Guest Authors" does not exist or is not published                             | "covid-exemplar-germany", "covid-exemplar-vietnam", "covid-exemplar-south-korea", "covid-exemplar-vietnam-2020", "covid-exemplar-south-korea-2020", "covid-exemplar-germany-2020", "identify-covid-exemplars"                                                                                                                                                                                                                                                                       |
[...]

PR #4005, opened 2024-10-03
🔨 correctly initialize bugsnag in SiteBaker v2

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/3019 

I wasn't able to re-open the [original PR for this branch](https://github.com/owid/owid-grapher/pull/3911) because I rebased the branch after the PR had been closed by Stalebot.

**tl;dr:** - it was tricky to robustly fix all the errors and warnings currently in prod with code. We'll be able fix some of them by updating the data and can ignore the rest in bugsnag.

In the other PR, I had proposed 3 ways to fix the existing errors that initializing bugsnag correctly would begin to expose:

1. Gdoc explorer link validation should check for redirects
2. The missing tags warnings shouldn't be logged
3. The (current) missing author errors should be ignored in bugsnag

Each had a sort of problem.

For 1: I added code that considers redirects, but the `coronavirus-data-explorer` redirect is hardcoded in the `baker/redirects.ts` file and can't be easily imported to use in gdocs validation. We could ignore all these errors, or update and republish the articles that use the old URL.

For 2: we don't really have a good way to exclude a certain class of error/warning from Bugsnag because we don't have error codes. We can't ignore _all_ warnings, because there are some warnings that we _do_ want to be notified about such as a linked gdoc no longer existing, which we can't upgrade to errors because that prevents a gdoc from being published.

For 3: all validation errors for a gdoc were being serialized together and logged, which would mean that an ignored error might resurface if its gdoc acquired another type of error. I've fixed this by logging each error/warning to bugsnag individually.

### DIFF
--- a/db/model/Gdoc/GdocBase.ts
+++ b/db/model/Gdoc/GdocBase.ts
@@ -58,7 +58,6 @@ import {
     OwidGdocContent,
     OwidGdocType,
 } from "@ourworldindata/types"
-import { getRedirectsFromDb, resolveRedirects } from "../Redirect.js"
 
 export class GdocBase implements OwidGdocBaseInterface {
     id!: string
@@ -751,8 +750,6 @@ export class GdocBase implements OwidGdocBaseInterface {
         const chartIdsBySlug = await mapSlugsToIds(knex)
         const publishedExplorersBySlug =
             await db.getPublishedExplorersBySlug(knex)
-        const redirects = await getRedirectsFromDb(knex)
-        const resolvedRedirects = resolveRedirects(redirects)
 
         const linkErrors: OwidGdocErrorMessage[] = this.links.reduce(
             (errors: OwidGdocErrorMessage[], link): OwidGdocErrorMessage[] => {
@@ -783,9 +780,7 @@ export class GdocBase implements OwidGdocBaseInterface {
                 }
 
                 if (link.linkType === "explorer") {
-                    const resolvedTarget =
-                        resolvedRedirects.get(link.target) ?? link.target
-                    if (!publishedExplorersBySlug[resolvedTarget]) {
+                    if (!publishedExplorersBySlug[link.target]) {
                         errors.push({
                             property: "content",
                             message: `Explorer chart with slug ${link.target} does not exist or is not published`,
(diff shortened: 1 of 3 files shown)


db/model/Gdoc/GdocBase.ts
```diff
@@ -750,6 +751,8 @@ export class GdocBase implements OwidGdocBaseInterface {
         const chartIdsBySlug = await mapSlugsToIds(knex)
         const publishedExplorersBySlug =
             await db.getPublishedExplorersBySlug(knex)
+        const redirects = await getRedirectsFromDb(knex)
+        const resolvedRedirects = resolveRedirects(redirects)
```
rakyi, line 755: It seems we call validate on every gdoc in a loop when baking, which means we'd make the same query many times. We could move the logic to a separate function and make the query (and resolve redirects) only once.

Similarly for other DB queries we make here, but that is out of scope.

db/model/Redirect.ts
```diff
@@ -16,6 +16,29 @@ export async function getRedirects(
     )
 }
 
+/**
+ * Takes a list of redirects and returns a map of source -> target
+ * such that if you have [A -> B, B -> C], the map will be { A -> C, B -> C }
```
rakyi, line 21: I thought there were no redirect chains in the DB. We have a validation that prevents adding new chains, and if there are any existing, it would be better if we resolved/changed them in the DB instead of always resolving them here.

db/model/Gdoc/GdocBase.ts
ikesau: I agree, though refactoring gdocs to validate in bulk would probably be best done in a separate PR.

As is, I've decided just to not do this validation - we've never changed an explorer to point to another explorer (except with `coronavirus-data-explorer`) so I think if it comes up again, we can just update the (likely fewer than 10) gdocs that are affected.

db/model/Redirect.ts
ikesau: True! I had some leftover testing data in my redirects table it seems. Not that it matters now that I've removed the explorer validation 🙂

rakyi (approved):

PR #4030, opened 2024-10-07
🎉 update default-thumbnail design and file format

ikesau:
Updates the design per a request by Marwa and also switches the format to PNG which compresses better with fewer artifacts.


rakyi:
IMO, there is not really a need for the redirect, we can keep the old file indefinitely.

ikesau:
With the preface that it doesn't matter much either way, I think I prefer getting rid of the file in following with the principle of always trying to prune unused assets / code. Think of all the milliseconds that will be saved from cloning the repo in CI! 😄

rakyi (approved):
Storing a static file indefinitely is much simpler than having to create a migration file and running it in every environment. We also change how old email/RSS feeds/etc. render. But this is not a hill I'm willing to die on, so feel free to merge it as is. 🙂

PR #4035, opened 2024-10-08
✨ (entity selector) keep selected entities in options list

### LINKED ISSUE #3920: Entity selector: don't remove selected entities from the full list
- We've had user feedback that it's annoying that a selected entity moves to the "Selection" section at the top since this makes quickly selecting/unselecting an entity cumbersome
- The animation is also suboptimal in other cases. For example, we're not currently automatically closing the entity selector modal since that looks broken because the animation gets interrupted.
- Marwa suggested keeping a full list of all entities always, i.e. if a new entity is selected, it does appear in the "Selection" section on top, but it doesn't disappear from the full list

sophiamersmann:
Resolves #3920

Keeps selected and unselected entities both in the list of entities to choose from.

I didn't remove the animation library because the animations make it look a bit smoother when the list of selected entities is in view.

Browsers like Chrome and Firefox are smart enough not to do a layout shift when a new entity is selected while the list of selected entities is not currently in view. But in other browsers like Safari, selecting a new entity always leads to a layout shift. The animation also helps here to make the layout shift a bit less jarring. 

Marcel's point below is addressed :)

### DIFF
--- a/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
+++ b/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
@@ -837,7 +837,7 @@ export class EntitySelector extends React.Component<{
 
     private renderAllEntitiesInMultiMode(): React.ReactElement {
         const { sortedAvailableEntities } = this
-        const { selected, unselected } = this.partitionedAvailableEntities
+        const { selected } = this.partitionedAvailableEntities
 
         return (
             <Flipper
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#4035** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

marcelgerber:
https://github.com/user-attachments/assets/f5972da0-4686-48d9-a1f6-3f859d229d63

One thing I did notice here is that the checkboxes are behaving weirdly when using keyboard navigation (i.e. tab and space).
Activation works fine, but they are at first still unchecked and only properly become checked once the next checkbox is toggled.

I'm seeing this in both Edge and Firefox.

sophiamersmann:
This is just a draft implementation – I just changed as few lines as possible to get a staging site I could share.

If the problem you're describing also exists on prod, then probably worth opening an issue for it.

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -1137,14 +1042,7 @@ function FlippedListItem({
...
         >
-            <li
-                className={cx("flipped", {
-                    "most-recently-selected":
-                        mostRecentlySelectedFlipId === flipId,
-                })}
-            >
-                {children}
-            </li>
+            <li className={cx("flipped")}>{children}</li>
```
marcelgerber, line 1045: tiny nit:
```suggestion
            <li className="flipped">{children}</li>
```

marcelgerber (approved):
Nice, and great that we can get rid of a load of code! 🐰 

I found a few small things, but great work otherwise :)

PR #4044, opened 2024-10-09
Fix truncated rows in refreshPageviewsFromDatasette

rakyi:
Without the _stream=on param the rows were truncated at 4000 entries.


danyx23:
Thanks :)

PR #4045, opened 2024-10-09
Add overrides for indicator metadata

rakyi:
Another place where we show the metadata is in the sources modal, triggered by clicking on "Learn more about this data" when embedded. This should only scroll to the metadata section on the mdim page, except we currently set `isEmbeddedInAnOwidPage: true`, so it opens the modal even there.

We display a list of indicators and their sources there, even for the "all" view of the poverty mdim, which contains all the indicators. I think it means we don't have a place to show the global override for metadata there, but we should still override the per indicator metadata. Is that correct?


rakyi:
Summary from a huddle with Daniel, we decided:

* We’ll remove `isEmbeddedInADataPage: true` to make mdims behave like a data page — have entity selector and “Learn more about this data” scrolls to about section (in a separate PR)
* Leave the sources modal as it is for now and think about what to do for iframe embeds

marcelgerber (approved):
Nice, yeah this makes sense to me

PR #4047, opened 2024-10-09
🐛 (admin) fix color selection for scatters

sophiamersmann:


sophiamersmann:
* **#4047** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

danyx23:
Ah nice, thanks for catching that! In the PR that this commit is part of I moved the types from all over the codebase into their own lerna package so that we can reference them cleanly from elsewhere. One of the more annoying things that were part of this was that there was a ColorScaleConfigDefaults class that used mobx properties and I didn't want to have that as a dependency in the types package - so I had to re-model this and change it to a JS object and use it in the correct way in a few places. It seems that in this place I got the use wrong.

sophiamersmann:
Makes sense! I'm glad this turned out to be such an easy fix :)

PR #4049, opened 2024-10-09
🐛 disaggregate site-baker bugsnag errors

ikesau:
Bugsnag is currently grouping _all_ errors occurring during baking as of https://github.com/owid/owid-grapher/pull/4005 being merged.

https://app.bugsnag.com/our-world-in-data/grapher-admin/errors/66d0c35a0ab81c7871ff95d5?filters[event.since]=30d&filters[error.status]=ignored

This PR allows us to set a custom `groupingHash` which overrides Bugsnag's default grouping heuristic. I've set it for gdocs baking on the specific error message so that all instances of e.g. `Author "Our World in Data team" does not exist or is not published` get grouped and can be ignored permanently, but `Author "Mxa Rsero" does not exist or is not published` would still get logged.


marcelgerber (approved):
What a nice way to solve this, let's go! 🚀

PR #4050, opened 2024-10-10
Update cache control

rakyi:
A small follow-up to our call.

* `no-cache` is equivalent to `max-age=0, must-revalidate` and recommended being used instead, see
  https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#force_revalidation
* Authenticated responses should use private cache only

### DIFF
--- a/adminSiteServer/authentication.tsx
+++ b/adminSiteServer/authentication.tsx
@@ -183,7 +183,7 @@ export async function authMiddleware(
     }
 
     // Authed urls shouldn't be cached
-    res.set("Cache-Control", "public, max-age=0, must-revalidate")
+    res.set("Cache-Control", "public, no-cache")
 
     if (user?.isActive) {
         res.locals.session = session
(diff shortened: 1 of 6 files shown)


marcelgerber (commented):
We can actually also get rid of the `public`, reading from MDN here:
> If a request doesn't have an Authorization header, or you are already using s-maxage or must-revalidate in the response, then you don't need to use public.
  -- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#public

The (very slight) advantage of that also being that a plain `cache-control: no-cache` header can be compressed efficiently in HTTP/3's [QPACK](https://datatracker.ietf.org/doc/html/rfc9204#:~:text=39-,cache%2Dcontrol,-no%2Dcache).

rakyi:
I wasn't sure about dropping `public`, but as you pointed out, it seems fine. I left it in the `public, max-age=31536000` case, because that also compresses to a one byte in QPACK.

marcelgerber (approved):
Nice!

PR #4061, opened 2024-10-11
Stop showing variable ID as a grapher title

rakyi:
When everything hasn't loaded yet, `defaultTitle` may contain variable IDs.

I tried changing how `defaultTitle` works to avoid it ever returning a title with variable IDs, which would be possible if we made `titlePublicOrDisplayName` return an empty string instead of variable ID, which we could do if we changed `AbstractCoreColumn.name` to return an empty string instead of a slug (which is the variableId), but that breaks assumptions in other places where `titlePublicOrDisplayName` is used.


marcelgerber (approved):
Nice, this works well!

PR #4074, opened 2024-10-22
🐛 replace spaces in element IDs for Figma

sophiamersmann:
Figma shows element IDs as labels. We used to make those IDs "CSS safe," which sometimes led to ugly, difficult-to-read IDs. That's why, in a recent PR, I just assigned names as is to element IDs (including CSS unsafe stuff like spaces and %, etc). I didn't notice anything weird at the time, so I thought Figma would be ok with this. Now we're dealing with a weird Figma issue where lines show up but are not coloured. We can't repro the bug, so not sure this fix helps, but it's our best guess atm.

### DIFF
--- a/packages/@ourworldindata/utils/src/Util.ts
+++ b/packages/@ourworldindata/utils/src/Util.ts
@@ -307,6 +307,10 @@ export const makeSafeForCSS = (name: string): string =>
         return "__" + ("000" + char.toString(16)).slice(-4)
     })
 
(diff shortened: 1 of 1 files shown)


sophiamersmann:
* **#4074** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #4082, opened 2024-10-25
✨ (grapher) pin tooltips to the bottom on mobile / TAS-664

sophiamersmann:
Pin tooltips to the bottom of the screen on mobile for a more pleasant UX. When pinned to the bottom, the tooltip hides less of the chart area and isn't "in the way" when interacting with a chart. This also fixes several overflow issues for very tall tooltips.

### Summary

- If the tooltip content is very tall, then the tooltip content becomes scrollable
    - This is also the reason why the tooltip stays on screen even after an interaction has ended
- A tooltip can be dismissed by clicking anywhere on the screen
- A tooltip is automatically dismissed when Grapher is scrolled out of view
- For tooltips that have a "Total" row (e.g. for stacked area charts), the total row is moved to the top, so that it's visible by default even if the tooltip is scrollable 

### Notes

- To detect when Grapher is scrolled out of view, we keep Grapher's intersection observer around (we used to disconnect it after Grapher has been first scrolled to). That might have a perf impact? I can't imagine the perf impact being too bad since we very rarely have more than 10 rendered Graphers on a site.
- After making these changes, map charts errored when trying to render a map tooltip with a MobX issue like "Dynamic observables can't be frozen". I didn't manage to understand what the problem was, but the MapTooltip/MapSparkline refactor resolves the issue 🤷🏻‍♀️
- An event handler is registered on the document to dismiss the tooltip when clicking outside of the chart area. This doesn't work if propagation is stopped. This is the case for the the entity selector toggle and the timeline. That's why the tooltip is explicitly dismissed when clicking on the timeline or the entity selector toggle

### Caveats

- If a tooltip is open and then gets dismissed because the entity selector is opened, for example, we ideally want the tooltip to disappear immediately. But for some chart types, tooltips fade out with some delay (e.g. scatter plots).

### Example charts

- Line chart: http://staging-site-mobile-tooltips-fixed-bottom/grapher/life-expectancy
- Stacked area with many variables: http://staging-site-mobile-tooltips-fixed-bottom/grapher/dietary-composition-by-country
- Map: http://staging-site-mobile-tooltips-fixed-bottom/grapher/life-expectancy?tab=map
- Scatter: http://staging-site-mobile-tooltips-fixed-bottom/grapher/carbon-emission-intensity-vs-gdp-per-capita

### Testing

This behaviour only kicks in for touchscreen devices, so it's best to look at these changes on an actual phone.

### Drive-by changes

- I removed the computed properties for frame padding – all that passing around wasn't really necessary for a couple of constants

### Screenshot

Screenshot 2024-06-05 at 14 54 46


sophiamersmann:
* **#4082** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4082?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/@ourworldindata/grapher/src/tooltip/TooltipContents.tsx
```diff
@@ -241,24 +270,26 @@ export class TooltipTable extends React.Component<TooltipTableProps> {
...
-                                            className={classnames(
-                                                "series-value",
-                                                { missing }
-                                            )}
-                                        >
-                                            {!missing &&
-                                                column.formatValueShort(
-                                                    value,
-                                                    format
+                                {zip(columns, excludeUndefined(values)).map(
```
marcelgerber, line 273: I'm not sure about this change here: could this lead to a case where we're matching a value to the wrong col?

E.g. if `cols = [col1, col2, col3]` and `values = [1, undefined, 2]`, then we would end up with `col1: 1, col2: 2`, I think.

I'm also not sure why you changed this in the first place?

marcelgerber (commented):
Really nice work here, I like the new behavior.

I tested this on desktop and mobile. On desktop I didn't notice any issues. On mobile I noticed two:
- When clicking on "Learn more about this data" on a data page (which scrolls down), the tooltip stays visible even though the chart itself leaves the viewport. Maybe the IntersectionObserver doesn't fire for programmatic scrolls?
- The tooltip also stays visible when clicking the download button and the share button

I'm wondering if there is a better way to register click events - maybe use a `{capture: true}` event handler to catch them all??

packages/@ourworldindata/grapher/src/tooltip/TooltipContents.tsx
sophiamersmann: I changed it because it gave a mobx warning (but I'm not sure which of my code changes made the mob warning appear).

Yeah, I thought about that. Although it's technically possible that we start matching the wrong columns/values, in practice, that doesn't happen if you look at how this is used. Still, it's not bulletproof, of course — maybe it's not worth the risk of getting rid of a warning.

sophiamersmann:
Ah, damn it. I've thought about using a capture event but I don't think that works because interacting with the chart shouldn't dismiss the tooltip. If we wanted to use a capture event for the document, we would also need to use capture events for chart event handlers, which I thought was too much of a weird constraint.

packages/@ourworldindata/grapher/src/tooltip/TooltipContents.tsx
marcelgerber: Hm, interesting. I kind-of-assume that the Mobx warning would go away if you use whatever way of copying that array, and it doesn't have to be `excludeUndefined` in particular.

Can you try `[...values]` instead and see if that works?

sophiamersmann:
Ah, that's not something I considered. Yes, that should work!

packages/@ourworldindata/grapher/src/tooltip/TooltipContents.tsx
sophiamersmann: `[...values]` didn't work. I think MobX's issue is that the zipped arrays don't have the same length, which means elements out of bounds are accessed. I got rid of the `excludeUndefined` now and implemented the opposite thing – if one array is shorter than the other, we fill it up with undefined values. It's a bit cumbersome, but it does get rid of the warning.

marcelgerber: Oh, that's confusing as hell!
Do you have a good sense of why `columns` and `values` can be of different length? Is there a good reason for that?
Otherwise, it would be good to get rid of that footgun tbh.

packages/@ourworldindata/grapher/src/tooltip/Tooltip.scss
```diff
@@ -1,467 +1,496 @@
...
-                line-height: 16px;
-                width: 100%;
-                .parenthetical {
-                    color: $grey;
-                }
+                    tr.total--top {
+                        td:nth-child(2),
+                        td:nth-child(3) {
+                            border-bottom: 1px solid $background-stroke;
+                            vertical-align: top;
```
marcelgerber, line 297: In what must be absolute outrage for Marwa, the "Total" line is not correctly aligned.

See this screenshot, where the ruler is touching the "T" of "Total", but not the number behind it.

Don't know why this is - but getting rid of this `vertical-align` line gets rid of it.

CleanShot 2024-11-07 at 16 28 04

marcelgerber:
I'm currently confused how dismissing the tooltip works for chart types other than line & stacked? How does it work for maps, for example - they don't seem to have a handler inside `MapChart.tsx`?

sophiamersmann:
Line charts and stacked area charts are special because they implement touch events so you can hold your thumb down, and on movement, the tooltip gets updated. Other chart types don't do this. They only register mouse enter and leave events. So, if you press down on an element on mobile, then the mouse enter event is fired (showing the tooltip), and if you click anywhere else, then the mouse leave event is fired (dismissing the tooltip).

packages/@ourworldindata/grapher/src/tooltip/TooltipContents.tsx
sophiamersmann: Line charts pass [y-value, col-value or undefined] and [y-value column]. That's why I used excludeUndefined. Instead, I could use excludeUndefined in the line chart component (it should be safe since line charts without a y-col but with a color-col don't exist)

sophiamersmann:
The code is now using a capture event handler, but for some reason it doesn't get fired when interacting with the timeline. I'm not sure why – that's why, when interacting with the timeline, the tooltip still gets dismissed explicitly.

sophiamersmann:
@marcelgerber this is good to go from my side. Do you think we can get this over the finish line? No worries if you would need too much to finish the review though :)

PR #4091, opened 2024-10-29
🐛 (explorer) merge configs correctly

sophiamersmann:
## Problem

Explorers incorrectly merge (partial) Grapher configs with explorer-level Grapher settings.

We naively merge configs like this

```js
{
   ...partialGrapherConfig,
   ...explorerProgram.grapherConfigOnlyGrapherProps
}
```

where `partialGrapherConfig` is the indicator config (of type `GrapherInterface`), but `explorerProgram.grapherConfigOnlyGrapherProps` is of type `ExplorerGrapherInterface` that includes explorer-flavoured flat config fields, like `yAxisMin`. Overwriting the y-axis min of the partial Grapher config, which is given as `yAxis.min`, thus won't work.

## Bug

I don't know of an example in the wild where this currently leads to a problem, but you can fabricate a bug locally by following these steps:

Set `yAxis.min` of an indicator that is used in the conflict explorer by running

```sql
update chart_configs cc
join variables v on cc.id = v.grapherConfigIdETL
set
  cc.patch = JSON_MERGE_PATCH(cc.patch, '{"yAxis":{"min": 1}}'),
  cc.full = JSON_MERGE_PATCH(cc.full, '{"yAxis":{"min": 1}}')
where v.id = 985431
```
against your local database.

Navigate to [this view](http://localhost:3030/explorers/conflict-data?Conflict+type=Intrastate+conflicts&Measure=Conflict+deaths&Conflict+sub-type=All+sub-types&Data+source=Uppsala+Conflict+Data+Program&Sub-measure=Regional+data&country=~OWID_WRL) of the conflict explorer. Note that
```js
// the partial grapher config sets y-axis min to 1
explorer.partialGrapherConfigsByVariableId.get(985431).yAxis.min // = 1

// the explorer overwrites this by setting yAxis.min to 0
explorer.explorerProgram.grapherConfig.yAxisMin // = 0

// the y-axis min being used in Grapher is 1, but should be 0
// i.e. the explorer settings didn't successfully overwrite the indicator settings
grapher.yAxis.min // = 1, but should be 0
```

I ran the same query against the database on staging. You'll see that on staging Grapher's y-axis min is correctly set to 0 ([staging](http://staging-site-fix-explorer-config-merge/explorers/conflict-data?Conflict+type=Intrastate+conflicts&Measure=Conflict+deaths&Conflict+sub-type=All+sub-types&Data+source=Uppsala+Conflict+Data+Program&Sub-measure=Regional+data&country=~OWID_WRL)).

## Solution

Before the explorer's Grapher config gets merged with the partial config, we translate explorer-specific fields to valid GrapherInterface fields, e.g. `yAxisMin` is translated to `{ yAxis: { min: ... } }`. This is defined by the `toGrapherObject` function of an explorer cell definition.

The `toGrapherObject` function is conceptually similar to the `parse` function. Alternatively, I could have reworked the `parse` function so that the given key/value pair is always parsed as a valid GrapherInterface object. Maybe that would have been the cleaner solution? I'm not sure. In the end, I went with the `toGrapherObject` function because I thought it would be useful to keep both Grapher row representations around:
- `explorerGrapherConfig`: explorer-flavoured settings that directly map to a Grapher row
- `grapherConfig`: corresponding chart config that is guaranteed to be a valid GrapherInterface

## Testing

Since I don't know of any bugs in the wild (doesn't mean they don't exist though), these changes should not result in any visual changes.

### DIFF
--- a/baker/siteRenderers.tsx
+++ b/baker/siteRenderers.tsx
@@ -826,12 +826,12 @@ const getExplorerTitleByUrl = async (
     if (url.queryStr) {
         explorer.initDecisionMatrix(url.queryParams as ExplorerFullQueryParams)
         return (
-            explorer.grapherConfig.title ??
-            (explorer.grapherConfig.grapherId
+            explorer.explorerGrapherConfig.title ??
+            (explorer.explorerGrapherConfig.grapherId
                 ? (
                       await getEnrichedChartById(
                           knex,
-                          explorer.grapherConfig.grapherId
+                          explorer.explorerGrapherConfig.grapherId
                       )
                   )?.config?.title
                 : undefined)
--- a/explorer/Explorer.tsx
+++ b/explorer/Explorer.tsx
@@ -9,7 +9,6 @@ import {
     GrapherInterface,
     GrapherQueryParams,
     GrapherTabOption,
-    AxisMinMaxValueStr,
 } from "@ourworldindata/types"
 import {
     OwidTable,
@@ -41,6 +40,7 @@ import {
     isInIFrame,
     keyBy,
     keyMap,
+    mergeGrapherConfigs,
     omitUndefinedValues,
     parseIntOrUndefined,
     PromiseCache,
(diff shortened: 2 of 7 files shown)


sophiamersmann:
* **#4091** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4091?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

marcelgerber (approved):
Nice, well spotted and good work!

PR #4096, opened 2024-10-31
Update grapher configs in R2 when variable config changes

rakyi:
* Fix missing updates of configs in R2 after we update the configs in the DB on variable grapher config upsert/deletion in the API
* Update mdim view grapher configs on variable grapher config upsert/deletion
* Trigger baking if a published mdim config is created/updated


rakyi:
@sophiamersmann since you worked on the related code recently, could you take a look also? Feel free to ignore the mdim specifics.

sophiamersmann (approved):
I can't talk to the mdim specifics, but on a high level, this makes sense to me 👍🏻

If you wanted to, you could add a couple of db tests. This has saved me a few times while working on this stuff :) Up to you, though!

rakyi:
Thanks! I'll try adding a few tests. After one of the existing ones failed in CI, I now see we already have some, so will start there.

db/model/Variable.ts
```diff
@@ -280,23 +285,110 @@ export async function updateAllChartsThatInheritFromIndicator(
...
+        patchConfigETL?: GrapherInterface
+        patchConfigAdmin?: GrapherInterface
+    }
+): Promise<
+    {
+        chartConfigId: string
+        patchConfig: GrapherInterface
+        isPublished: boolean
+    }[]
+> {
```
danyx23, line 355: This inline type declaration is a bit unwieldy - maybe pull out the things you need into standalone types and/or define them as partials from other types?

db/model/Variable.ts
```diff
@@ -280,23 +285,110 @@ export async function updateAllChartsThatInheritFromIndicator(
...
+            patchConfigAdmin ?? {},
+            view.patchConfig
+        )
+        await db.knexRaw(
+            trx,
+            `-- sql
+                UPDATE chart_configs
+                SET
+                    full = ?,
+                    updatedAt = ?
```
danyx23, line 373: UpdatedAt should auto-update - I would remove the explicit update from here unless there is a good reason to do it manually?

danyx23 (approved):
I had some tiny notes but they are all optional - looks nice!

db/model/Variable.ts
rakyi: I copied and tweaked existing code, and this is what we did there already. Checked just now and `updatedAt` is not set to auto-update, and I'm not sure why. @sophiamersmann do you remember why we set it manually?

adminSiteServer/apiRouter.ts
```diff
@@ -1155,6 +1175,12 @@ putRouteWithRWTransaction(
         }
         const rawConfig = req.body as MultiDimDataPageConfigRaw
         const id = await createMultiDimConfig(trx, slug, rawConfig)
+        if (await isMultiDimDataPagePublished(trx, id)) {
```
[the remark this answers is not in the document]
rakyi: Good point, added it.

db/model/Variable.ts
sophiamersmann: I initially switched to auto-update and then switched back to manual updates when Mojmir noticed that the updatedAt of the chart and the updatedAt of the config were out of sync for some charts. If it’s automatic updates then that can happen when you run a migration on one of the tables, for example. I don’t recall why Mojmir needed these to be the same, chart-diff maybe?

db/model/Variable.ts
rakyi: Ah, I know, but there are many different inline types in this file and giving them names would be even worse, IMO.

db/model/Variable.ts
rakyi: Ah yeah, I vaguely remember that. Thanks!

danyx23: Ok cool, if there is a reason that that is fine. Could you just add a comment (can just quote the vague statement of Sophia even :) ) so that we don't accidentally delete this in the future?

rakyi: Good point, added comments.

PR #4101, opened 2024-11-01
:honeybee: collect session replays via Sentry

bnjmacdonald:
Collect session replays via Sentry for 1% of cookie-consenting user sessions. Closes owid/analytics#155.

[Implementation notes and privacy considerations](https://www.notion.so/owid/Collect-session-replays-11874b71f92e804bb617ebfe69e8d0f6?pvs=4).

Remaining todos before merging:

- [x] Update privacy policy to mention Sentry usage
- [x] apply for [Sentry nonprofit discount](https://sentry.io/for/good/)
- [x] temporarily add `SENTRY_DSN` to staging env (then remove after merging – we don't need to waste precious replay budget on staging sites going forward). owid/ops#233
- [x] If Sentry nonprofit discount not approved yet, start Sentry free trial (so that we aren't capped at 50 session replays)
- [x] add `SENTRY_DSN` to prod env (`ops:templates/owid-admin-prod/admin-env.secret`)

### DIFF
--- a/site/owid.entry.ts
+++ b/site/owid.entry.ts
@@ -16,13 +16,16 @@ import {
     ENV,
     BUGSNAG_API_KEY,
     ADMIN_BASE_URL,
+    SENTRY_DSN,
 } from "../settings/clientSettings.js"
 import { Grapher, CookieKey } from "@ourworldindata/grapher"
 import { MultiEmbedderSingleton } from "../site/multiembedder/MultiEmbedder.js"
 import { CoreTable } from "@ourworldindata/core-table"
+import { isInIFrame } from "@ourworldindata/utils"
 import { SiteAnalytics } from "./SiteAnalytics.js"
 import Bugsnag, { BrowserConfig } from "@bugsnag/js"
 import BugsnagPluginReact from "@bugsnag/plugin-react"
+import * as Sentry from "@sentry/react"
 import { runMonkeyPatchForGoogleTranslate } from "./hacks.js"
 import { runSiteFooterScripts } from "./runSiteFooterScripts.js"
 import {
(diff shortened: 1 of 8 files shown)


rakyi (commented):
Sorry, if you already discussed this before, but just a quick fly by review:

* We should be probably using `@sentry/react` instead of `@sentry/browser` to get better contextual data
* If we keep using Sentry we should stop using Bugsnag, but this can wait until we confirm it works as expected, let's just not forget about it

site/owid.entry.ts
```diff
@@ -82,6 +85,43 @@ if (BUGSNAG_API_KEY) {
...
+                integrations: [
+                    Sentry.replayIntegration({
+                        maskAllText: false,
+                        maskAllInputs: false,
+                        blockAllMedia: false,
+                        mask: [".sentry-mask"],
+                    }),
+                ],
+                replaysSessionSampleRate: ENV === "development" ? 1 : 0.01,
+                replaysOnErrorSampleRate: 0,
```
rakyi, line 106: Capturing replays on error seems even more useful than non-errors. Is there a good reason to disable it?

bnjmacdonald:
> * We should be probably using `@sentry/react` instead of `@sentry/browser` to get better contextual data

thanks Martin – done ✅

site/owid.entry.ts
bnjmacdonald: It's more so the opposite for me, as I don't want our budget of session replays to get inadvertently consumed by a few pages that raise a lot of the same error over and over (e.g. we end up with 1k monthly session replays on `/search` out of our 5k budget just b/c of errors).

Obviously, keeping `replaysOnErrorSampleRate` low would help avoid this problem, but it was easier for now to just set this to 0 to get a clear baseline of how well the 5k monthly replays are distributed across our pages (without having to worry about how much errors skew the distribution of replays).

larsyencken (approved):
Looks good, no obvious issues with the integration from my end. We can obviously tweak the sampling rates later as needed.

PR #4103, opened 2024-11-01
🔨 remove default layer from chart configs / TAS-675

sophiamersmann:
Removes the default layer from the full config of charts.

This is the smaller version that removes the default layer from the full configs of all charts but keeps using it in the admin. If I find the time this week, I also want to remove the admin's dependence on the default config, but even if I don't find the time, I think this PR is okay to be merged as is since keeping the default config up-to-date is zero effort (we have a GitHub action that infers the default config from the schema and automatically commits it).

I tested this by comparing the previously baked Grapher configs with the now-baked Grapher configs. The diff was empty for most configs. There were a few that used to have `map.colorScale` set to an empty object (not sure why) that is missing now, but that makes no difference to Grapher.

The SVG tester differences are due to a parent PR.

### DIFF
--- a/adminSiteClient/AbstractChartEditor.ts
+++ b/adminSiteClient/AbstractChartEditor.ts
@@ -80,9 +80,19 @@ export abstract class AbstractChartEditor<
         return mergeGrapherConfigs(parentConfig ?? {}, patchConfig)
     }
 
-    /** live-updating full config */
+    /** live-updating config */
+    @computed get liveConfig(): GrapherInterface {
+        return this.grapher.object
+    }
+
+    @computed get liveConfigWithDefaults(): GrapherInterface {
+        return mergeGrapherConfigs(defaultGrapherConfig, this.liveConfig)
+    }
+
+    /** patch config merged with parent config */
     @computed get fullConfig(): GrapherInterface {
-        return mergeGrapherConfigs(defaultGrapherConfig, this.grapher.object)
+        if (!this.activeParentConfig) return this.liveConfig
+        return mergeGrapherConfigs(this.activeParentConfig, this.liveConfig)
     }
 
     /** parent config currently applied to grapher */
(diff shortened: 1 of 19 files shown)


sophiamersmann:
* **#4103** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4103?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

danyx23 (commented):
I like this! I have found one issue though and I'm not sure if this is because the admin does things differently now or if this is just a little bug or if I misunderstand how this is supposed to work.

Here is what I did:
- take a line chart (like [this one](http://localhost:3030/admin/charts/551/edit)) that does not currently have inheritance
- create a parent admin layer chart
- set a property to a non-default value (e.g. y column max) and save 
- in the original standalone chart, refresh and check that the parent is now set
- go to the customize tab and click "bind to data" under y column max

I would expect the parent value to be in effect now but it is the default value. Am I wrong or is this a bug?

sophiamersmann:
"Bind to data" means the y-column max is inferred from data, it essentially sets `yAxis.max` to "auto". I haven't tried but the behaviour sounds correct to me.

sophiamersmann:
The label is maybe a bit confusing – maybe "Unset" would be clearer?

sophiamersmann:
@danyx23 do you think you have time to finish this review in the next week or so? I think it would be good if my changes and especially Marcel‘s work on narrative views builds on this.

danyx23:
Hey @sophiamersmann, thanks for the ping! (btw please ping me earlier next time, I don't get a notification if you reply here without mentioning me :)). I played with it and it looks good! The only thing that would need to be changed now is that Martin has merged his work to add chart_config entries for Mdim views and they would also need to be stripped of the defaults for completeness. I think it shouldn't be much work and it would be good to include this in this PR but if you think this is too much work now then it's also fine to merge this as is and do the mdim work in cooldown.

sophiamersmann:
Thanks! Yes, I'll rebase & update the PR

sophiamersmann:
@danyx23, as far as I can see, the defaults haven't been merged into the mdim view configs. Am I missing something?

danyx23:
@sophiamersmann Ah interesting - that was sort of a bug in that work then that now is a feature :). Yes, then there is nothing left to do in that regard and this can go. :shipit:

PR #4106, opened 2024-11-04
Prioritize mdim view config schema over top level

rakyi:
To make sure it works either way and nested config take precedence.


danyx23 (approved):
Makes sense! Sorry I took so long, I had a look earlier yesteraday and wasn't in the right mindspace for understanding why that makes a difference :)

rakyi:
Np, thanks for the review!

PR #4115, opened 2024-11-06
Fix express middleware receiving a non-error

rakyi:
Handle the error with the same pattern as elsewhere in the file.

Fixes: https://app.bugsnag.com/our-world-in-data/grapher-admin/errors/672a37bfd1461623267d57d7


Marigold (approved):
Thanks for fixing this!

PR #4131, opened 2024-11-09
✨ Site accessibility improvements

ikesau:
Some QOL things I noticed + Lighthouse recommendations

A few +/-5% lightness adjustments here and there because the background colour isn't white, because that's easier than adding $blue-61

I showed the `*:focus-within` rule to Marwa and she's happy with it 👍

### DIFF
--- a/packages/@ourworldindata/utils/src/Util.ts
+++ b/packages/@ourworldindata/utils/src/Util.ts
@@ -2000,15 +2000,3 @@ export function getPaginationPageNumbers(
 
     return pageNumbers
 }
-
-/**
- * Convert 0 to "0th", 1 to "1st", etc.
- * */
-export function getOrdinalNumberString(n: number): string {
-    const suffix = ["th", "st", "nd", "rd"]
-    const modulusValue = n % 100
-    return (
-        n +
-        (suffix[(modulusValue - 20) % 10] || suffix[modulusValue] || suffix[0])
-    )
-}
--- a/packages/@ourworldindata/utils/src/Util.ts
+++ b/packages/@ourworldindata/utils/src/Util.ts
@@ -2000,3 +2000,15 @@ export function getPaginationPageNumbers(
 
     return pageNumbers
 }
+
+/**
+ * Convert 0 to "0th", 1 to "1st", etc.
+ * */
+export function getOrdinalNumberString(n: number): string {
+    const suffix = ["th", "st", "nd", "rd"]
+    const modulusValue = n % 100
+    return (
+        n +
+        (suffix[(modulusValue - 20) % 10] || suffix[modulusValue] || suffix[0])
+    )
+}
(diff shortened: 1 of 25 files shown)


packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -2000,3 +2000,15 @@ export function getPaginationPageNumbers(
 
     return pageNumbers
 }
+
+/**
+ * Convert 0 to "0th", 1 to "1st", etc.
+ * */
+export function getOrdinalNumberString(n: number): string {
```
rakyi, line 2007: Not a big deal, but consider using [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules#formatting_text_using_the_returned_tag), which would make this more readable (IMHO) and standardized.

site/blocks/ExpandableParagraph.tsx
```diff
@@ -64,6 +64,10 @@ export const ExpandableParagraph = (
         <div className={cx("expandable-paragraph", className)}>
             <div
                 style={contentStyles}
+                // inert isn't supported in react@17
+                // prevents focus on elements that are not visible
+                // ideally would only apply to elements below the fold but that's hard
+                {...{ inert: isClosed ? "true" : undefined }}
```
rakyi, line 70: Is this really what we want? What problem do we solve? It also prevents users from selecting text within the element, and the text will be excluded from browser search, which seems too harsh.

site/gdocs/components/LatestDataInsights.tsx
```diff
@@ -251,10 +255,12 @@ function useDotButton(emblaApi: EmblaCarouselType | undefined): {
 
 function DotButton({
     children,
+    index,
     ...restProps
-}: ComponentPropsWithRef<"button">) {
+}: ComponentPropsWithRef<"button"> & { index: number }) {
```
rakyi, line 260: Actually, I was meaning to change this to a div instead of a button and remove the interactivity. The buttons are too small for touch navigation — it was flagged as such by some other web test tool I used. Should we do that instead?

packages/@ourworldindata/utils/src/Util.ts
danyx23: @rakyi Just a drive by comment to say thanks for mentioning this, I didn't know about it and this seems pretty neat. Maybe post about it on slack in developers or tools-and-workflows?

site/blocks/ExpandableParagraph.tsx
ikesau: The issue I was trying to address here is that you can tab through links that are hidden, which renders poorly (see the video.)

I tried setting up a prop that would set `tabIndex="-1"` on any nested `span-link` elements but couldn't get it to work in the small amount of time I was budgeting for this PR so I went with this simpler solution instead.

I'm open to changing this though, if you think this change is an overall downgrade 🙂 

https://github.com/user-attachments/assets/7a850299-dd8b-4364-8e98-5204ee0677af

rakyi: Hmm, I'm not sure what's better. Let's go with what you did and revert/rework later, if necessary.

rakyi (approved):

PR #4137, opened 2024-11-11
Allow fetching grapher configs from all origins

rakyi:
Fixes an error when we try to load grapher configs for multi-dim preview in admin in prod, which runs on a different domain (admin.owid.io) than the Cloudflare Pages function (ourworldindata.org).


danyx23 (approved):
Thanks!

PR #4142, opened 2024-11-12
Disable MySQL logging

rakyi:
I noticed there are 20 GB of logs I never needed to look at on my machine. IMHO, we can enable logging when debugging something as needed instead of having it always on.


larsyencken (approved):
Fine with me. Maybe it's better to leave them commented out, so that it's quick to re-enable for a specific dev environment?

rakyi:
Good idea. I commented the settings out instead.

PR #4175, opened 2024-11-15
🎉 Explorers index page

ikesau:
image

A very simple explorer index page that a few of us at OWID have been missing. It also gives the homepage "See all our Data Explorers" link somewhere useful to go to.

We'll possibly update these cards to look the same as the cards on the homepage once Marwa has some time to design the rest of the tag icons that we'd need.

### DIFF
--- a/baker/siteRenderers.tsx
+++ b/baker/siteRenderers.tsx
@@ -6,6 +6,7 @@ import { StaticCollectionPage } from "../site/collections/StaticCollectionPage.j
 import { SearchPage } from "../site/search/SearchPage.js"
 import NotFoundPage from "../site/NotFoundPage.js"
 import { DonatePage } from "../site/DonatePage.js"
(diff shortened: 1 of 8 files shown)


baker/siteRenderers.tsx
```diff
@@ -677,6 +679,19 @@ export const renderReusableBlock = async (
     return cheerioEl("body").html() ?? undefined
 }
 
+export const renderExplorerIndexPage = async (
+    knex: KnexReadonlyTransaction
+): Promise<string> => {
+    const explorersBySlug = await getPublishedExplorersBySlug(knex)
+    const explorers = Object.values(explorersBySlug).sort((a, b) =>
```
rakyi, line 686: Doesn't really matter here, but in general it's usually better for perf to do the sorting in the DB in the same query.

rakyi (approved):

Issue #3308, opened 2024-03-06
Add button to archive datasets in dataset admin, and show if it is already archived


pabloarosado:
## Core problem

Currently, the dataset admin does not show if a dataset is archived, and doesn't have a button to archive a dataset.
This button used to exist for datasets updated via "import csv", but then it was removed for datasets imported from ETL. I'm not sure why, since I think it's useful to be able to archive them once there is a newer version. I currently do it manually by directly accessing the database and setting isPrivate=1, isArchived=1.

## Proposed solution

There should be a button to archive datasets (and to make them private, I suppose), and it should be clear on the admin page of a dataset if a dataset is already archived.

danyx23:
@pabloarosado can you elaborate a bit? IIRC there is also an archiving operation in the ETL - is that in any way connected to isArchived in the DB? If this is different, would you just like to show the isArchived flag status? And would this be the thing that is modified by the button? Maybe we could clarify the terminology for this as well?

pabloarosado:
@danyx23 my suggestion was simply to have a button to archive a grapher dataset, like old (non-ETL) datasets do. For example [this old dataset](https://owid.cloud/admin/datasets/5622) has a button (although, to be fair, that old dataset should probably not have a button because it's already archived, but that's another issue).

ETL steps can also be archived within ETL. But I don't think that's relevant here. What we need is indeed the flag `isArchived`, and be able to modify that from the dataset admin page.

I'm not sure if `isPrivate` should also be modified when archiving a dataset. It has never been totally clear the distinction between the two flags. As I mentioned above, when I archive a dataset, I manually set both flags to 1. But maybe it should be enough to set `isArchived=1`.

pabloarosado:
Hi @danyx23 is there any reason why this issue is blocked? I would have thought that it's trivial to implement (old datasets used to have a button to archive them, so I suppose the capability is there somewhere, but it has been deactivated for ETL datasets).
We have been accumulating many unused datasets, and it would be good to clean up, to avoid confusion. We now have [an ETL issue](https://github.com/owid/owid-issues/issues/1740) to archive many grapher datasets. Currently, the only way to achieve that is to manually write to production database, which is time-consuming (and not very safe).

To clarify, from my perspective, I think the only requirement would be to set `isArchived=1` (I think `isPrivate=1` is not necessary, but I'm not sure if there are other reasons why this would also be necessary). Thanks!

lucasrodes:
[...]

PR #4177, opened 2024-11-18
Add button to archive dataset if it contains no charts

pabloarosado:
Add a button to archive grapher datasets, if it contains no charts.
Also, remove the "Delete dataset" button, which we probably don't need.

For more context, see https://github.com/owid/owid-grapher/issues/3308

### DIFF
--- a/adminSiteClient/DatasetEditPage.tsx
+++ b/adminSiteClient/DatasetEditPage.tsx
@@ -374,14 +374,21 @@ class DatasetEditor extends React.Component<{ dataset: DatasetPageData }> {
                         </p>
                     ) : (
                         <p>
-                            <strong>
-                                Before archiving, ensure that the corresponding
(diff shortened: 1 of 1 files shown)


lucasrodes:
Thanks for working on this, Pablo!

This will be super useful in keeping our catalog cleaner.

This made me think that it'd be great if we could archive datasets in admin via an API. The API could check that the dataset is actually archivable: (i) no explorer relies on it, (ii) not used by any chart, etc. If this API was callable from ETL, we could archive datasets from more places (e.g. 'Dashboard' page in Wizard).

pabloarosado:
Hi @danyx23, do you see any problem with this implementation? I think it would be useful to have this "Archive" button regardless of our ETL archiving strategy (I created an issue specifically for that: https://github.com/owid/etl/issues/3554). Please let me know if you see any problems with this PR. Thanks!

PR #4223, opened 2024-11-27
🐛 migrate configs fetched from R2 to latest version on the fly

sophiamersmann:
Configs in R2 might be out of date. This can happen after a config migration with a breaking change.

But Grapher expects configs to be of the latest version. We also merge configs in various places and can only do so if the configs are of the same version (or we get errors like [this one](https://owid.slack.com/archives/C028P3FM8CD/p1732679638441589)).

Whenever we fetch configs from R2, we need to make sure they adhere to the latest version and if not, migrate them on the fly. I updated the Multiembedder and explorers. Is there some place else where we should also do this?

### DIFF
--- a/packages/@ourworldindata/explorer/src/Explorer.sample.tsx
+++ b/packages/@ourworldindata/explorer/src/Explorer.sample.tsx
@@ -1,7 +1,10 @@
 import React from "react"
 import { DimensionProperty } from "@ourworldindata/utils"
 import { GRAPHER_TAB_OPTIONS } from "@ourworldindata/types"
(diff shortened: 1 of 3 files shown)


sophiamersmann:
* **#4223** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4223?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

PR #4228, opened 2024-11-28
✨ add automatic retry to client fetch requests

danyx23:
This PR adds automatic retry with exponential back-off to all client side fetches that grapher or our site do. This should make our site a bit more robust on bad connections.


ikesau (approved):
Very nice!

tested locally by blocking URLs in my network tab

also checked to make sure there weren't any other site non-POST fetch calls that you missed

PR #4229, opened 2024-11-28
✨ Add filter information to readme and metadata

danyx23:
When using the download API, add information about active filters to the metadata and readme.

### DIFF
--- a/functions/_common/readmeTools.ts
+++ b/functions/_common/readmeTools.ts
@@ -262,12 +262,13 @@ export function constructReadme(
         .flatMap((col) => [...columnReadmeText(col)])
     let readme: string
     const queryString = searchParams.size ? "?" + searchParams.toString() : ""
+    const urlWithFilters = `${grapher.canonicalUrl}${queryString}`
 
     const downloadDate = formatDate(new Date()) // formats the date as "October 10, 2024"
     if (isSingleColumn)
         readme = `# ${grapher.title} - Data package
 
-This data package contains the data that powers the chart ["${grapher.title}"](${grapher.canonicalUrl}${queryString}) on the Our World in Data website. It was downloaded on ${downloadDate}.
+This data package contains the data that powers the chart ["${grapher.title}"](${urlWithFilters}) on the Our World in Data website. It was downloaded on ${downloadDate}.
 ${[...activeFilterSettings(searchParams)].join("\n")}
 ## CSV Structure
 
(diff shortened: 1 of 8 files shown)


functions/_common/urlTools.ts
```diff
@@ -0,0 +1,12 @@
+export function getGrapherFilters(
+    searchParams: URLSearchParams
+): Record<string, unknown> {
+    const params = searchParams.size
+        ? Object.fromEntries(searchParams)
+        : undefined
+    // delete url query params that the download api uses but that are not related to grapher
+    delete params.v1
+    delete params.csvType
+    delete params.useColumnShortNames
```
ikesau, line 10: Would it be worth the extra effort to use a whitelist here (and in the creation of `urlWithFilters`) ? Otherwise you can add arbitrary text to the query string and get a readme that says whatever anyone wants.

image

functions/_common/urlTools.ts
```diff
@@ -0,0 +1,12 @@
+export function getGrapherFilters(
+    searchParams: URLSearchParams
+): Record<string, unknown> {
```
ikesau, line 3: I think it's actually not possible to get anything but a `string` as one of the values here.

functions/_common/readmeTools.ts
```diff
@@ -229,9 +230,28 @@ function* columnReadmeText(col: CoreColumn) {
...
 }
 
+function* activeFilterSettings(searchParams: URLSearchParams) {
+    const filterSettings = getGrapherFilters(searchParams)
+    if (filterSettings) {
+        yield ""
+        yield `### Active Filters`
+        yield ""
+        yield `A filtered subset of the full data was downloaded. The following filters were applied:`
+        for (const entry of Object.entries(filterSettings)) {
```
ikesau, line 240: If `getGrapherFilters` was typed with `Record<string, string>` then you could do the sugary thing of 

```
for (const [key, val] of Object.entries(filterSettings)) {
```

which is 100% absolutely completely not necessary. I'm just sharing coz I think it's neat 🙂

ikesau (approved):
If you fix the TypeError this is good to go. I'll leave it up to you (and lars?) to make the call re: whitelisting.

functions/_common/urlTools.ts
danyx23: Good point! It means that when we do this for MDims we'll have to expand the whitelist but it's probably the better way to do things.

functions/_common/readmeTools.ts
danyx23: True, this is nicer :)

PR #4231, opened 2024-11-28
:sparkles: Take user to variable edit page instead of dataset when clicking on a pencil

Marigold:
Clicking on a pencil in the sources tab will now take you to variable edit page (if the variable id exists). (Joe's request, check this [slack thread](https://owid.slack.com/archives/C0149NKLZAM/p1732568053164499))

image


sophiamersmann (approved):
Thanks a lot :) I pushed one quick cosmetic commit (I hope I didn't mess it up). Should be good to go!

PR #4235, opened 2024-11-29
Add people-rows ArchieML component

rakyi:
* The new component allows us to have rows of people in two columns
* Add socials to person block

Screenshot 2024-11-29 at 17-58-24 Our Team - Our World in Data


rakyi:
* **#4235** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4235?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4235?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

rakyi:
@ikesau I tried to make use of more of the existing grid system, but I still think flex makes the most sense for switching from a two column layout on a bigger screen to a single column on a small screen, given the constraints I explained before. LMK what you think.

db/model/Gdoc/rawToEnriched.ts
```diff
@@ -808,12 +812,24 @@ const parsePerson = (raw: RawBlockPerson): EnrichedBlockPerson => {
...
         name: raw.value.name,
         title: raw.value.title,
         text: raw.value.text.map(parseText),
+        socials: raw.value.socials?.map(parseSocialLink),
+        parseErrors: [],
+    }
+}
+
+const parsePeopleRows = (raw: RawBlockPeopleRows): EnrichedBlockPeopleRows => {
+    console.log("raw", raw.value.people)
```
ikesau, line 828: ✂️

site/gdocs/components/People.tsx
```diff
@@ -3,10 +3,24 @@ import * as React from "react"
...
+    columns?: "2"
     children: React.ReactNode
 }) {
-    return <div className={cx("people", className)}>{children}</div>
+    return (
+        <div
+            className={cx(
+                "people",
+                {
+                    "people-cols-2": columns === "2",
```
ikesau, line 18: If I'm understanding the requirements correctly based on what it's currently doing when i test the article on staging, when `columns: 2` is set, we want:

lg - 1 column
md - 2 coumns
sm - 1 column

if so, could you do something like:

```
const columnClassNames = {
1: "grid-cols-1",
2: "grid-sm-cols-1 grid-md-cols-2 grid-lg-cols-1"
}
<div className={cx("people grid", columnClassNames[columns])}>
  ...children
</div>
```

And then you don't have to worry about flex or that calc statement or anything. 🙂

ikesau (approved):
nice and easy 👍 

the css grid saga continues 😂

site/gdocs/components/People.tsx
rakyi: Yep, you are totally right. I somehow got the wrong impression about how ordering of elements changes in a grid when you change from two columns to a single column. Also, it should be just `lg-up` instead of `lg-only`. Thanks for being persistent. 🙂 Fortunately it's a small change and didn't cost us too much time.

PR #4251, opened 2024-12-04
Add donors component

rakyi:
I also fixed a few unrelated things.

Screenshot 2024-12-04 at 14-30-54 How We’re Funded - Our World in Data

### DIFF
--- a/db/migration/1733228214376-CreateDonors.ts
+++ b/db/migration/1733228214376-CreateDonors.ts
@@ -1,0 +1,24 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class CreateDonors1733228214376 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`-- sql
+            CREATE TABLE donors (
+                id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+                name VARCHAR(255) NOT NULL,
+                originalName VARCHAR(255) NOT NULL,
+                shouldPublish BOOLEAN NOT NULL DEFAULT (false),
+                comment TEXT DEFAULT (''),
+                createdAt TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+                updatedAt TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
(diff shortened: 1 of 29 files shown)


db/migration/1733228214376-CreateDonors.ts
```diff
@@ -0,0 +1,24 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class CreateDonors1733228214376 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`-- sql
+            CREATE TABLE donors (
```
rakyi, line 6: This table will be read and written to from the scripts that we use to update the donor list. Previously the scripts generated HTML, now they will use this table. That part will be reviewed and QA'd by Lars.

rakyi:
You can see the new list of donors [here](http://staging-site-about-donors/funding).

site/gdocs/components/Donors.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    const donorsByLetter = groupBy(donors, (donor) => donor[0])
+    return (
+        <div className={className}>
+            <div className="col-start-2 span-cols-12">
+                <p className="donors-note">(Listed in alphabetical order)</p>
+                {Object.entries(donorsByLetter).map(([letter, donors]) => (
+                    <div key={letter}>
+                        <h3 className="donors-letter">{letter}</h3>
+                        <ul className="donor-list grid grid-cols-6 grid-lg-cols-5 grid-md-cols-4 grid-sm-cols-2">
```
ikesau, line 17: It might be better to use `columns` on the `ul` here so that multi-line names don't add a vertical gap to other columns.

https://github.com/user-attachments/assets/31c342da-d0df-45bf-9579-c6df167299f1

though you'll also wanna tweak the line-height and margins so that it's clear when a name does wrap, that it's all still one name.

ikesau (approved):
Nice! 👍

site/gdocs/components/Donors.tsx
rakyi: Thanks, I prefer the current approach or adding back list bullets. I'll ask Marwa about how she want to handle it.

rakyi: Marwa agrees with your proposal, changed.

PR #4252, opened 2024-12-04
🎉 Cloudflare images image updating

ikesau:
image

Changes:
- Functionality to upload a new version of an image
- Standardizes API route names (they're all `/images` now)
- Refactors API routes to extract common functions
- Refactors image upload button into `PostImageButton` and `PutImageButton` + utils

### DIFF
--- a/adminSiteClient/ImagesIndexPage.tsx
+++ b/adminSiteClient/ImagesIndexPage.tsx
@@ -3,18 +3,10 @@ import React, {
     useContext,
     useEffect,
     useMemo,
+    useRef,
     useState,
 } from "react"
-import {
-    Button,
-    Flex,
-    Input,
-    Mentions,
-    Popconfirm,
-    Space,
-    Table,
-    Upload,
-} from "antd"
+import { Button, Flex, Input, Mentions, Popconfirm, Table, Upload } from "antd"
 import { AdminLayout } from "./AdminLayout.js"
 import { AdminAppContext } from "./AdminAppContext.js"
 import { DbEnrichedImageWithUserId, DbPlainUser } from "@ourworldindata/types"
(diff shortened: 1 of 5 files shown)


adminSiteClient/ImagesIndexPage.tsx
```diff
@@ -152,6 +160,33 @@ function UserSelect({
     )
 }
 
+// when updatedAt changes, the image will reload the src
+// but it looks like sometimes cloudflare doesn't update in time :(
```
ikesau, line 164: This only works _sometimes_, so maybe it's not worth it and we just let authors know that it can take up to a minute for the image to update after refreshing the page. Not sure.

rakyi: If it doesn't work reliably, I'd add the disclaimer to the page. I'm not sure if I understand the intended functionality. If we want to refresh the image every time the `updatedAt` changes, I think [we could pass it as a `key` instead](https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key).

rakyi (approved):

adminSiteClient/ImagesIndexPage.tsx
ikesau: Updating the key was the first thing I tried, but it wouldn't actually make a new request, and then when I tried forcing it to re-render, it would load the cached version, so this is the only way I found to make it work. (Though it also sometimes hits the CF cache 😫)

I think I'll just keep it as it currently is and see what the authors make of it. I'll probably have to explain in slack that sometimes the preview doesn't update and that they need to click on the image to see what the actual current image is. Seems.. fine 🤷

PR #4261, opened 2024-12-06
🎉 Automatic alt text for cloudflare images

ikesau:
https://github.com/user-attachments/assets/5b081556-8b49-4eac-9d86-a1f141e762d7

Changes:
- New API route, admin UI, and helper function to generate alt text for an image from OpenAI
- Makes alt text nullable in the DB
- Fixes some other issues with the CF images migration

### DIFF
--- a/adminSiteClient/ImagesIndexPage.tsx
+++ b/adminSiteClient/ImagesIndexPage.tsx
@@ -13,17 +13,23 @@ import { DbEnrichedImageWithUserId, DbPlainUser } from "@ourworldindata/types"
 import { Timeago } from "./Forms.js"
 import { ColumnsType } from "antd/es/table/InternalTable.js"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
-import { faClose, faUpload } from "@fortawesome/free-solid-svg-icons"
+import {
+    faClose,
+    faRobot,
+    faSave,
+    faUpload,
+} from "@fortawesome/free-solid-svg-icons"
 import { RcFile } from "antd/es/upload/interface.js"
-import TextArea from "antd/es/input/TextArea.js"
 import { CLOUDFLARE_IMAGES_URL } from "../settings/clientSettings.js"
 import { keyBy } from "lodash"
+import cx from "classnames"
 
 type ImageMap = Record<string, DbEnrichedImageWithUserId>
 
 type UserMap = Record<string, DbPlainUser>
 
 type ImageEditorApi = {
+    getAltText: (id: number) => void
     patchImage: (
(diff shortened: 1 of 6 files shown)


adminSiteClient/admin.scss
```diff
@@ -1224,4 +1224,27 @@ main:not(.ChartEditorPage):not(.GdocsEditPage) {
...
     }
+
+    .ImageIndexPage__alt-text-editor {
+        textarea {
+            border-radius: 6px;
+            position: relative;
+            border: 1px solid #ccc;
+            width: 100%;
+            &.ImageIndexPage__alt-text-editor--should-autosize {
+                field-sizing: content;
```
ikesau, line 1235: This is only supported in chrome as of now, but I think that's the browser most of the authors use.

adminSiteServer/apiRouter.ts
```diff
@@ -3283,6 +3287,44 @@ getRouteWithROTransaction(
...
+
+        const src = `${CLOUDFLARE_IMAGES_URL}/${image.cloudflareId}/public`
+        let altText: string | null = ""
+        try {
+            altText = await fetchGptGeneratedAltText(src)
+        } catch (error) {
+            console.error(
+                `Error fetching GPT alt text for image ${imageId}`,
+                error
+            )
```
rakyi, line 3316: Out of scope of this PR, but it's unfortunate if we currently have to effectively log the error twice to surface/log it properly. This shouldn't be necessary, in general.

adminSiteClient/ImagesIndexPage.tsx
```diff
@@ -55,30 +61,58 @@ function AltTextEditor({
...
-            if (trimmed !== text) {
-                patchImage(image, { defaultAlt: trimmed })
-            }
-        },
-        [image, text, patchImage]
-    )
+    // Update local state if the prop updates (i.e. from the alt text suggestion)
+    useEffect(() => {
+        setValue(text)
+    }, [text])
```
rakyi, line 78: This is an antipattern, which usually means the state should be moved up in the tree.

rakyi (approved):
Cool feature! 🙂

I'm not sure how important it is to fix/improve the state synchronization with `useEffect`.

adminSiteClient/ImagesIndexPage.tsx
ikesau: Thanks for pointing it out. I was stuck in a pattern where all the other API calls called `setImages` and that was making this not work.

Having the API call return the response directly for its caller to work with (in this case, this `AltTextEditor` component) makes much more sense.

PR #4270, opened 2024-12-09
🐛 fix default gdoc thumbnails not rendering in gdocs

ikesau:
Fixes the following whoopsie

image

image

### DIFF
--- a/site/search/SearchPanel.tsx
+++ b/site/search/SearchPanel.tsx
@@ -11,6 +11,7 @@ import {
     groupBy,
     uniqBy,
     Region,
+    DEFAULT_GDOC_FEATURED_IMAGE,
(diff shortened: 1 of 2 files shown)


site/search/SearchPanel.tsx
```diff
@@ -65,6 +66,12 @@ import { ChartHit } from "./ChartHit.js"
 const siteAnalytics = new SiteAnalytics()
 
 function PagesHit({ hit }: { hit: IPageHit }) {
+    // a temporary fix for articles that have been indexed without the directory
```
ikesau, line 69: This will be removed as part of https://github.com/owid/owid-grapher/pull/4214

danyx23 (approved):
Thanks for the quick fix!

PR #4280, opened 2024-12-11
Add more docs for using images

rakyi:


adminSiteClient/ImagesIndexPage.tsx
```diff
@@ -554,6 +563,12 @@ export function ImageIndexPage() {
     return (
         <AdminLayout title="Images">
             <main className="ImageIndexPage">
+                <Alert
```
ikesau, line 566: The delete button also says this when you click it - do you think we need this extra spooky message too?

site/README.md
```diff
@@ -34,6 +38,11 @@ filename: my_image.png
 
 where `my_image.png` is an image that has been uploaded via the `/admin/images` view in the admin client, and thus exists in Cloudflare Images.
 
+> [!CAUTION]
+> Cloudflare Images don't have separate environments for production, staging and
+> dev, so be careful not to upload images that are only available in one
+> environment and even more **careful when deleteing images**.
```
ikesau, line 44:
```suggestion
> environment and even more **careful when deleting images**.
```

ikesau (approved):
I'm not sure the alert's necessary, but we can add it if you want to be extremely clear 🙂

adminSiteClient/ImagesIndexPage.tsx
rakyi: Removing the alert.

But I'm now actually wondering if we should make the admin read-only in other envs than prod. You can't really upload the images in staging/locally and then use it in other envs, because the `images` table is only synced in the direction prod → {staging, local}, right? Same for deleting.

ikesau: Yeah, that's definitely an option.

If we had a way of knowing when we're in prod vs. staging, we could hide/disable the buttons/inputs on staging too, and explain that all updates to images must be done on prod.

I didn't do that because while developing the feature, I definitely needed to be able to CRUD images, and the intention had been for this to be a half-cycle amount of work.

But I'm going to make some tickets with potential follow ups that we can at least consider for cooldowns.

rakyi: We distinguish between envs already:

https://github.com/owid/owid-grapher/blob/a7d6726ac52c5aa88a9a31073894f9392954b819/adminSiteClient/AdminLayout.tsx#L53-L66

PR #4281, opened 2024-12-11
🔨 migrate images of unpublished gdoc posts that were missing

danyx23:
The migration script used to filter for only images that are referenced in posts_gdocs_x_images. The idea was that this would keep only images that are used in any gdoc (regardless of publishing status), but it seems that there is a bug in updating the posts_gdocs_x_images table and it only happens for published posts. 

This PR just gets rid of the check on posts_gdocs_x_images. This means that it will maybe transfer a few images to CF images that are actually unused but we can clean these up later (e.g. when we have the flat gdcos components table) and it seems more important to unblock authors today.

After this is merged and deployed the sync script has to be run again.


rakyi (approved):
Makes sense to me!

PR #4282, opened 2024-12-11
Accept only images in images admin file input

rakyi:
Improves the UX a bit. See [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept).


ikesau:
lgtm!

image

PR #4286, opened 2024-12-11
✨ Add button to see which posts images are used in

ikesau:
One more feature for the images admin! Last one! I promise! 😅 

https://github.com/user-attachments/assets/5eb5a864-32cc-47e0-b8e9-ba1dce490ec3

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -3315,6 +3315,15 @@ deleteRouteWithRWTransaction(
     }
 )
 
+getRouteWithROTransaction(apiRouter, "/images/usage", async (_, __, trx) => {
+    const usage = await db.getImageUsage(trx)
+
+    return {
+        success: true,
+        usage,
+    }
+})
+
 getRouteWithROTransaction(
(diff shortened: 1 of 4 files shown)


adminSiteServer/apiRouter.ts
```diff
@@ -3138,17 +3138,6 @@ getRouteNonIdempotentWithRWTransaction(
 postRouteWithRWTransaction(apiRouter, "/images", async (req, res, trx) => {
     const { filename, type, content } = validateImagePayload(req.body)
 
-    const preexisting = await trx<DbEnrichedImage>("images")
```
ikesau, line 3141: This is a change that Daniel suggested - it's annoying to be told to rename an image, only to find out that the hash collides after.

So now we check for the has collision first.

rakyi (approved):
Cool feature! Have you tested whether the list for images that are used a lot still looks fine?

ikesau:
> Cool feature! Have you tested whether the list for images that are used a lot still looks fine?

nah, but the currently most-reused image is 5, which looked fine.

PR #4291, opened 2024-12-12
About pages

rakyi:


packages/@ourworldindata/types/src/gdocTypes/ArchieMlComponents.ts
```diff
@@ -86,6 +86,16 @@ export type EnrichedBlockChart = {
...
 } & EnrichedBlockWithParseErrors
 
+export type RawBlockCode = {
+    type: "code"
+    value: RawBlockText[]
+}
+
+export type EnrichedBlockCode = {
+    type: "code"
+    text: RawBlockText[]
```
danyx23, line 96: Just noticed this during a rebase: This should be EnrichedBlockText[]

danyx23: Ah yes, that makes sense - the approach you took in the new PR looks good!

PR #4296, opened 2024-12-12
Fix ETL staging server MDim caching issues

danyx23:
This PR fixes an issue in the MDim workflow when working in the ETL with staging servers. This is the setup that a lot of data managers use.

Lucas reported that when he makes changes in the ETL on his machine and runs the ETL update targeting a staging server, MDim views would not get updated. 

It turns out that the reason this happens (and is different from the local workflow) is that on staging servers, /grapher/by-uuid/UUID.config.json is not handled by the mock site router but instead by CF functions (atm wrangler if I understand correctly, soon probably the CF preview deploy).

These CF functions have s-MaxAge set and so cache the results for a good while.

This PR enables the IsPreviewing flag that we use in the runSiteFooterScripts script and if it is set then reqeusts to the above url append ?nocache which then leads to the CF function (either wrangler or preview deploy) to request the freshest version from R2.

While I was at it I also reduced the cache times. These small json objects don't have to be cached for a full hour, 5 min should be plenty for prod.

### DIFF
--- a/site/multiDim/MultiDimDataPageContent.tsx
+++ b/site/multiDim/MultiDimDataPageContent.tsx
@@ -108,10 +108,14 @@ const cachedGetVariableMetadata = memoize(
 )
 
 const cachedGetGrapherConfigByUuid = memoize(
-    (grapherConfigUuid: string): Promise<GrapherInterface> =>
-        fetchWithRetry(
-            `/grapher/by-uuid/${grapherConfigUuid}.config.json`
+    (
+        grapherConfigUuid: string,
+        isPreviewing: boolean
+    ): Promise<GrapherInterface> => {
+        return fetchWithRetry(
+            `/grapher/by-uuid/${grapherConfigUuid}.config.json${isPreviewing ? "?nocache" : ""}`
         ).then((resp) => resp.json())
+    }
 )
 
 const useTitleFragments = (config: MultiDimDataPageConfig) => {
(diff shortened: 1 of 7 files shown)


site/multiDim/MultiDimDataPageContent.tsx
```diff
@@ -242,7 +248,7 @@ export const MultiDimDataPageContent = ({
 
     const currentView = useView(currentSettings, config)
     const { varDatapageData, varGrapherConfig, grapherConfigIsReady } =
-        useVarDatapageData(config, currentView)
+        useVarDatapageData(config, currentView, isPreviewing ?? false)
```
ikesau, line 251: could also just set a default in the parameter destructuring 🙂

ikesau (approved):

rakyi (approved):
Thanks for the fix!

Since we are already introducing a change in behavior between prod and staging (for a good reason) would it make sense to use the environment we run in to decide on what to do? So instead of passing the flag around, we could disable caching in staging for all requests to this endpoint.

danyx23 (commented):
@rakyi yes I thought of this as well - but it can also be useful to have a test environment behave very similar to the prod setup including caching - so that when you make non-admin requests you can test if the caching is what you expect. Let's chat about it in the dev call next week - for now I'd leave it as is.

PR #4298, opened 2024-12-13
Fix latest data insights styles

rakyi:
I verified the fix works on my iPhone.


rakyi:
Merging now to get the fix out quickly. Feel free to review post-hoc.

PR #4314, opened 2024-12-16
Fix gdoc component extraction

danyx23:
This fixes a bug when gdoc components have optional properties that are enumerated by the gdoc component extraction code

### DIFF
--- a/db/model/Gdoc/extractGdocComponentInfo.ts
+++ b/db/model/Gdoc/extractGdocComponentInfo.ts
@@ -68,7 +68,7 @@ function iterateArrayProp<T extends OwidEnrichedGdocBlock>(
     parentPath: string,
     prop: keyof T
 ): ChildIterationInfo[] {
-    if (prop in parent)
+    if (parent[prop])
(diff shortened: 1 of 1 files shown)


db/model/Gdoc/extractGdocComponentInfo.ts
```diff
@@ -68,11 +68,15 @@ function iterateArrayProp<T extends OwidEnrichedGdocBlock>(
     parentPath: string,
     prop: keyof T
 ): ChildIterationInfo[] {
-    return (parent[prop] as OwidEnrichedGdocBlock[]).map((child, index) => ({
-        child: child,
-        parentPath: `${parentPath}`,
-        path: `${parentPath}.${String(prop)}[${index}]`,
-    }))
+    if (prop in parent)
```
rakyi, line 71: This will be true, if the property is present, but has `undefined` as the value. We probably want:

```suggestion
    if (parent[prop])
```

danyx23: Ah right, thanks for catching that!

Issue #4313, opened 2024-12-16
Cloudflare Images reuploads aren't working


ikesau:
When I was implementing Cloudflare Images, I originally set the IDs of the images we uploaded to be a URI-encoded version of the image's filename. Near launch, I decided against this because I noticed some weirdness around Cloudflare doing some of its own encoding of those IDs So I removed the custom ID logic and went with the Cloudflare-generated UUID instead. This needed more thinking through and testing 🙃 

## The issue:
- User adds an image via the admin
  - Upload image A to Cloudflare Images
  - Store its data (including its UUID) in the DB
- Bake the site, pointing to the image via a URL of the form https://imagedelivery.net/our-org-id/the-image-uuid/w=some-width
- User replaces image A with image B
  - Delete image A in Cloudflare Images
  - Upload image B
  - Update DB with new data (including new UUID)
 
But the site doesn't rebake, and is stuck pointing to old URLs with old IDs that now 404. Even if we trigger a bake any time an image is updated, it will be broken until the bake finishes.

## Possible solutions:
- Don't delete the original image from Cloudflare Images, and trigger a bake any time a new version is uploaded
  - Add a `replacedBy` column to the `images` table
  - This would lead to an accumulation of orphans in Cloudflare Images but we were thinking of periodically running a script to clear these anyway
- Go back to explicitly setting the ID to something we can control
  - You can't manually set a UUID (Cloudflare throws an error) but we could generate our own ID string
    - Either a UUID in a different format (e.g. underscores instead of dashes)
    - The DB INT ID (would have to update the DB write logic in the API handler)
    - The filename again (testing robustly for the encoding issues)
- Add a Cloudflare Worker proxy that can resolve IDs correctly
  - This would require live access to the prod DB or Cloudflare D1, so probably not viable

PR #4320, opened 2024-12-16
🐛 fix cloudflare image version management

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/4313

# Changes
- Adds a `replacedBy` column to the `images` table
- Adds a `version` column to the `images` table
- Removes the unique `filename` constraint on the `images` column
- Adds a unique constraint on `filename, version`
- Updates all places that select "all images" to only select images where `replacedBy` is null
- Updates images API functions to reflect these changes
  - the PUT route now sets the `replacedBy` column, and triggers a build
  - the DELETE route uses a CTE to find every image in the chain and deletes them all
    - it's possible that a user unpublishes an article and immediately deletes its images before the rebake has occurred, thus temporarily breaking the article before it's removed, but that seems unlikely
 

# To test
## CRUD
1. Create three images
2. Upload the first one, replace it with the second, and the third
3. Confirm that all 3 images are still in the DB, and that `replacedBy` has been set correctly
4. Delete the third image and confirm that all 3 images get deleted in Cloudflare and the DB

## Baking
1. Create 2 images
2. Upload one of them, and publish it in an article
3. Confirm the image works
4. Update the image, and confirm that a bake is triggered
5. Confirm that the article still works in the meantime
6. Confirm that the updated image is used once the bake is finished

### DIFF
--- a/adminSiteClient/ImagesIndexPage.tsx
+++ b/adminSiteClient/ImagesIndexPage.tsx
@@ -554,10 +554,14 @@ export function ImageIndexPage() {
                     image: DbEnrichedImageWithUserId
                 }>(`/api/images/${id}`, payload, "PUT")
                 if (response.success) {
-                    setImages((prevMap) => ({
-                        ...prevMap,
-                        [id]: response.image,
-                    }))
+                    setImages((prevMap) => {
+                        const nextMap = { ...prevMap }
+                        delete nextMap[id]
+                        return {
+                            ...nextMap,
+                            [response.image.id]: response.image,
+                        }
+                    })
                 }
             },
             postUserImage: async (user, image) => {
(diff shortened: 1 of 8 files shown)


ikesau:
Hey @danyx23 - I ended up adding a `version` column and applying `UNIQUE(filename, version)` to get around the MySQL 8 weirdness.

This way you can PUT multiple images with the same filename because the version will be incremented, but you can't POST a new image with the same filename because it will collide with the (replaced) version 0 original image.

danyx23 (approved):
Ah this is a nice idea - the version number carries a bit of additional information which is neat.

The only small change I would still aks is to expand the comment on the DB image type definition to mention that we considered using replacedBy alone and have a unique constraint but that this doesn't work with mysql. I would then also copy this comment to the migration file (since this is where you might end up when you try to find out about why the version was added).

Nice work!

PR #4326, opened 2024-12-17
🐛 fix socials gdoc component enumeration bug

danyx23:


rakyi:
Socials were listed in the union twice:

https://github.com/owid/owid-grapher/blob/948687e235399a9b92ab8c07c559a6db638de28b/db/model/Gdoc/extractGdocComponentInfo.ts#L356

Should we remove the second instance too, or did you leave it there on purpose? BTW, this is the kind of static analysis I miss from JetBrains IDEs.

danyx23:
@rakyi no the second one we need so that the match is exhaustive :)

danyx23:
I.e. one we need, the duplication was a mistake

PR #4341, opened 2024-12-19
Fix padding of headings with multiple children

rakyi:
This is the case when e.g. part of the heading is italicized.


ikesau (approved):
tested on topic pages, articles, and about pages. looks good!

Issue #3406, opened 2024-03-26
Automate more breadcrumbs


larsyencken:
## Problem

We see it as our role to put work in context; when it comes to articles, that context is often the broader topic or area that the article sits in.

Currently, one way we provide context for an article is with _breadcrumbs_ (`posts_gdocs.breadcrumbs`), but they need to be manually specified every time. This is a waste of human effort, given that every article is tagged with a topic.

We would like instead do this automatically for each article.

## Technical notes

If https://github.com/owid/owid-grapher/pull/3695/ is merged, we will soon be able to construct a tag graph.

A simplified example:
![tag graph example](https://github.com/owid/owid-grapher/assets/11844404/448a6a0a-7821-430d-8a6f-33949bc33ad1)
⭐️= a topic

The top level tags are currently all assumed to be areas and don't have an associated page. The reason we have areas at all is because of the header nav (though there is no reason why we wouldn't be able to make these pages in the future.)

As long as we ensure every article is tagged with at least one topic tag, we'll be able to use this graph to construct breadcrumbs for every article.

We'll do this by constructing a subgraph that only features topic tags and displaying the path to any given topic.

So an article about indoor air pollution would be:
**Home > Air Pollution > Indoor Air Pollution**

An article about cancer:
**Home > Cancer**

An article about nuclear energy:
**Home > Energy > Nuclear Energy**

An article about fossil fuels has two possible breadcrumb paths (because the Fossil Fuels tag has two parents that are topics.) 

It could either be:
**Home > Greenhouse Gas Emissions > Fossil Fuels**
or
**Home > Energy > Fossil Fuels**

The tag graph has weighted edges, so that we can set one edge to always be preferred over the others, but there will likely be cases where we have an article that would prefer to have the other edge highlighted.

For example, if we set the `Energy-Fossil Fuels` edge to have a higher weight, this will be fine for articles that are about energy and fossil fuels: we can show the default breadcrumbs of **Home > Energy > Fossil Fuels**,  but if we have an article that's about fossil fuels' contributions to GHG emissions, we'll want the breadcrumbs to be **Home > Greenhouse Gas Emissions > Fossil Fuels**

A decision around whether or not parent tags need to be manually set ("parent implicitness") needs to be made. Ideally parent tags can be implicit the majority of the time and we can derive the breadcrumbs from the tag graph weights. In cases where we need to deviate from the default, authors can explicitly set the parent tag that they want to take priority, on the gdoc.

[...]

PR #4343, opened 2024-12-19
🎉 automatic breadcrumbs from the tag graph

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/3406

## Changes:
- Adds programmatically generated breadcrumbs to our tagged gdocs articles, based on the tag graph
- Fixes erroneous parent tag logic to reflect the fact that a tag may have multiple paths to get to it
  - This requires a somewhat verbose new structure called a `ParentTagArray` which is an array of tags. One child tag can have multiple `ParentTagArray`s e.g. "Nuclear Energy" can be a child of "Energy" and in a different part of the graph, a child of "CO2 Emissions"
  - We want _all_ of these parent tags for Algolia indexing, so the code has been fixed to reflect this

#### Rules for the programmatic breadcrumbs:
1. For each tag, pick the root-to-leaf subgraph with the highest weighted top node (tiebreak alphabetically)
2. For each subgraph, filter out non-topic nodes
3. Pick the longest filtered subgraph

I considered a "highest average weight" strategy as well, but seeing as weighting is only really used to order nodes relative to one another at each level, I didn't think it would lead to intuitive results.

This means that the ordering of the **areas** in the tag graph will determine which breadcrumbs trump which. I think this makes the most sense because presumably whichever topic we put at the top of the nav, we think is the most important and therefore its children are more important also, in some abstract way.

#### Automatic breadcrumbs render slightly differently to manually set breadcrumbs:
image
image

Manually set breadcrumbs were made for SDGs at first, where we wanted the leaf node to unclickable. For non-SDG articles, authors have (sensibly) shortened the title for this leaf node, but that's not something we're going to be able to do programmatically, so instead we'll only show topic breadcrumbs, and not a leaf node for the article.

### DIFF
--- a/adminSiteClient/GdocsManualBreadcrumbsInput.tsx
+++ b/adminSiteClient/GdocsManualBreadcrumbsInput.tsx
@@ -70,7 +70,7 @@ export const BreadcrumbLine = ({
     )
 }
 
-export const GdocsBreadcrumbsInput = ({
+export const GdocsManualBreadcrumbsInput = ({
     gdoc,
     setCurrentGdoc,
     errors,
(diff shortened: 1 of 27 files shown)


danyx23:
Nice work! A few notes from playing with it: I think it would be great if there were an easier way in the admin to understand if breadcrumbs were set automatically or manually. 

The final leaf node not being shown is understandable but quite an inconsistency. It might work well to use the slug, split by dash and sentence cased? E.g. "us-airline-travel" -> "Us airline travel". This will clearly sometimes not be perfect but this is what manual overrides are for. What do you think?

My hunch is also that we should advertise this staging server quite prominently after the christmas break - I'm pretty sure that Joe, Ed and maybe Max will want to look at quite a few examples to get a feeling for the logic. I think it works very well (e.g. the tie breaking etc) but I anticipate some discussion there.

adminSiteServer/app.test.tsx
```diff
@@ -920,3 +933,270 @@ describe("OwidAdminApp: indicator-level chart configs", () => {
         expect(json.success).toBe(false)
     })
 })
+
+describe.only("OwidAdminApp: tag graph", () => {
```
danyx23, line 937: Really nice tests!

danyx23 (commented):
The types in getParentTagArraysByChildName should be a bit more explicit IMHO, otherwise this looks good! 

As I wrote above, I think the semantics of this PR will warrant a bit of scrutiy and discussion and this might mean a few more code changes - if so, ping me again quickly before this gets merged.

ikesau:
Hey @danyx23 🙂 

Updates:
- Renames the `posts_gdocs` `breadcrumbs` column to `manualBreadcrumbs` to disambiguate it from automatic breadcrumbs throughout the codebase
  - Fixes various overlooked admin preview bugs that were due to these getting mixed up
- Adds a migration to delete the ~30 manually set breadcrumbs for non-SDG articles to ensure "no last node" consistency for all non-SDG breadcrumbs ([Marwa's okay with this](https://owid.slack.com/archives/C03SK89695E/p1736286406486459))
- Updates CSS
- Improves admin UI to make it clearer where breadcrumbs are coming from


https://github.com/user-attachments/assets/65ea387a-714d-4152-b902-a09db5ead573

danyx23:
The automatic breadcrumbs work fine AFAICS but the manual ones have two weirdnesses about them:
- When you edit them, you don't immediately see any changes. At least when clicking on "done" I would like the preview to be refreshed. I also don't understand why we have both "done" and "save draft" at this point. I think we should fold the behaviour of "save draft" into the "done" button. Happy to chat about the details of this.
- The other one is that it seems impossible to recreate the behaviour of the current automatic breadcrumbs with manual breadcrumbs. I can leave the last field empty but it still renders a ">" to an empty string in the breadcrumbs in that case. I'd also like to have better instructions for how to fill this in when the manual breadcrumbs is active (something like "Keep the last item empty unless you are editing an SDG page. See the writing guildelines for recommendations for how to pick topic parents")

Once these are tackled this is good to merge!

danyx23 (changes requested):
The automatic breadcrumbs work fine AFAICS but the manual ones have two weirdnesses about them:

- When you edit them, you don't immediately see any changes. At least when clicking on "done" I would like the preview to be refreshed. I also don't understand why we have both "done" and "save draft" at this point. I think we should fold the behaviour of "save draft" into the "done" button. Happy to chat about the details of this.
- The other one is that it seems impossible to recreate the behaviour of the current automatic breadcrumbs with manual breadcrumbs. I can leave the last field empty but it still renders a ">" to an empty string in the breadcrumbs in that case. I'd also like to have better instructions for how to fill this in when the manual breadcrumbs is active (something like "Keep the last item empty unless you are editing an SDG page. See the writing guildelines for recommendations for how to pick topic parents")

Once these are tackled this is good to merge!

PR #4350, opened 2024-12-20
:hammer: Refactor apiRouter into smaller files

danyx23:
apiRouter.ts has become quite large and when you work with LLMs, large files are pretty annoying. This PR splits the route implementations into new files by topic.

The individual commits make it easier to follow along what happens and should hopefully also make review easier.


adminSiteServer/apiRoutes/chartViews.ts
```diff
@@ -109,30 +115,28 @@ getRouteWithROTransaction(apiRouter, "/chartViews", async (req, res, trx) => {
...
-        }
-
-        const row = await db.knexRawFirst<ChartViewRow>(
-            trx,
-            `-- sql
+}
+
+export async function getChartViewById(
+    req: Request,
+    res: e.Response<any, Record<string, any>>,
```
rakyi, line 115: Not specific to this route, just noticed it here. Why are we using `Request` from our `authentication` module, which is just — IMO — an unnecessary alias for `express.Request` and `Response` from Express, when we have our own `Response` in the same `authentication` module with additional properties? If anything it should be the other way around, right?

My preference would be to use `Request` and `Response` only from Express, if possible.

rakyi (approved):
LGTM! And I have additional suggestions.

This might be subjective, but since we are already doing the refactoring: I feel like we have too many layers of abstraction and I suggest we get rid of `FunctionalRouter`.

I think the error handling it does is wrong — we should use [an error handling middleware](https://expressjs.com/en/guide/error-handling.html) if we want to do anything custom (which should probably just be a Sentry integration).

And we can configure the same middleware that we do in the constructor either at the top `app` level, or per each nester router.

danyx23:
@rakyi good points! I'll give it a stab to clean this up as well.

PR #4385, opened 2025-01-06
Revert temporary giving season changes

rakyi:
- Show Subscribe instead of Donate button again
- Remove call to action top banner


mlbrgl (approved):
checked code and preview, LGTM!

PR #4386, opened 2025-01-06
Remove Tablepress

rakyi:
It should not be used anywhere.


mlbrgl (approved):
looks good! I'm not seeing any uses either.

PR #4387, opened 2025-01-06
new table for chart (and possibly other objects) revisions

lucasrodes:
[Tracking issue](https://github.com/owid/owid-issues/issues/1777)

In https://github.com/owid/etl/pull/3792 I'm working on a bot that will suggest a chart for revision daily. It'll be a public chart, that has few visits.

We want to avoid suggesting the same chart multiple times, so we need to track what has already been suggested. This PR adds a new table for that.

I had thought of a table named `suggested_reviews`, with columns `id`, `type`, `date`. Where `type` can be 'chart', 'dataset', etc.

### DIFF
--- a/db/migration/1736177572560-HousekeepingSuggestedReviews.ts
+++ b/db/migration/1736177572560-HousekeepingSuggestedReviews.ts
@@ -1,7 +1,8 @@
-import { MigrationInterface, QueryRunner } from "typeorm";
+import { MigrationInterface, QueryRunner } from "typeorm"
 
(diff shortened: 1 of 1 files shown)


Marigold (approved):
Looks good. If you plan to use it in ETL, it's a good idea to add it to `grapher/model.py`, too.

If you want to test the whole thing on staging, you can create a branch with the same name `db-table-housekeeping` in ETL. That staging server should then use PRs from both repos.

Issue #4397, opened 2025-01-07
Improve UI for admin images page on small screens


ikesau:
The alt text textarea is almost unusable on a 13" screen. Let's add some padding/margin overrides to tighten things up as much as possible for this case.

![Image](https://github.com/user-attachments/assets/66c2ff13-b1c3-47e6-8752-eca989c876b5)

PR #4400, opened 2025-01-07
✨ improve images admin ui on small screens

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/4397

https://github.com/user-attachments/assets/d3a646a4-5821-4b4e-b3bd-b234c49ac84d


rakyi (approved):
Thank you! Could we also use the [`size="small"`](https://ant.design/components/table#table-demo-size) for the table, probably on all screen sizes?

PR #4407, opened 2025-01-08
✨ Add wizard link for chart animation

lucasrodes:
Add a button in the chart edit page (Export tab), linking the user to Wizard app "Chart animation" to animated that chart.

(added multiple reviewers, just with one review is fine)


marcelgerber (approved):
Nice, let's go! 🚀

PR #4424, opened 2025-01-09
Add formatting for numbers in data catalog

rakyi:
Number formatting was requested by Marwa and as I was changing it I noticed we set a wrong canonical URL for the data catalog, so I fixed that too.


ikesau (approved):
ty!!

PR #4431, opened 2025-01-10
Remove functional router

danyx23:
This PR attempts to remove the FunctionalRouter and replace it with a vanilla express router. It does not work yetj.


danyx23:
@rakyi I tried your suggestion to remove FunctionaRouter but this one doesn't work anymore on the api routes. If you have some time and want to play around with it feel free, I'll probably not get around to this before the end of next week.

PR #4433, opened 2025-01-10
🐛 take table filter toggle into account for data download / TAS-796

### LINKED ISSUE #4432: Download of 'displayed data' should take filtered table into account
See [Slack](https://owid.slack.com/archives/C46U9LXRR/p1736455010350519)

If 'Show selection only' on the table tab is toggled, downloading displayed data should only download the entities that are currently visible in the data table.

Image

sophiamersmann:
Fixes #4432

If 'Show selection only' on the table tab is toggled, then 'Download displayed data' only downloads data for those entities that are displayed in the table.

If 'Show selection only' isn't toggled, then 'Download displayed data' downloads all data (since all data is displayed). Don't know if this could be confusing. I find it okay.

This PR makes non-trivial changes to the Grapher component. Let me know if you prefer to wait for your refactor before merging this PR.

👇🏻 See my comment about `grapher.inputTable` vs `grapher.table` below

---
- <kbd>&nbsp;2&nbsp;</kbd> #4447 
- <kbd>&nbsp;1&nbsp;</kbd> #4433 👈

### DIFF
--- a/functions/_common/downloadFunctions.ts
+++ b/functions/_common/downloadFunctions.ts
@@ -68,10 +68,12 @@ export async function fetchZipForGrapher(
 }
 function assembleCsv(grapher: Grapher, searchParams: URLSearchParams): string {
     const useShortNames = searchParams.get("useColumnShortNames") === "true"
+    const fullTable = grapher.inputTable
+    const filteredTable = grapher.isOnTableTab
+        ? grapher.tableForDisplay
+        : grapher.transformedTable
     const table =
-        searchParams.get("csvType") === "filtered"
-            ? grapher.transformedTable
-            : grapher.inputTable
+        searchParams.get("csvType") === "filtered" ? filteredTable : fullTable
     return table.toPrettyCsv(useShortNames)
 }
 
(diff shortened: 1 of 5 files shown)


functions/_common/downloadFunctions.ts
```diff
@@ -68,10 +68,12 @@ export async function fetchZipForGrapher(
 }
 function assembleCsv(grapher: Grapher, searchParams: URLSearchParams): string {
     const useShortNames = searchParams.get("useColumnShortNames") === "true"
+    const fullTable = grapher.inputTable // TODO: should this be table?
```
sophiamersmann, line 71: Should this be `table` instead? `inputTable` includes all data, `table` is filtered by the time range set by the author (note that it's the author's time range, not the user's).

Not sure what's intended here, but the client side uses `table`, so regardless of what we choose, server and client should probably use the same table.

danyx23 (commented):
Nice work! Feel free to merge already, the grapher state refactor will still be open for a few days and I'm pretty used to dealing with rebase conflicts by now 😬

functions/_common/downloadFunctions.ts
danyx23: I guess you can argue both ways but since this is the best way to get the full data I'm inclined to give the user everything we have in this case - i.e. stay with inputTable for the time beeing.

danyx23 (approved):

functions/_common/downloadFunctions.ts
sophiamersmann: Ok, then I'll update the client code to use `inputTable` as well!

sophiamersmann:
@danyx23, do you think there is something to do here re the Sentry issue? 👆🏻 I don't see how it would be related to the changes that I made, but I'm also no expert on this part of the codebase.

Issue #4364, opened 2025-01-02
Disallow log scale for discrete bar charts


sophiamersmann:
## Description

Plotting discrete bar charts and stacked discrete bar charts on a log scale is confusing since it's not immediately clear that what you see is plotted on a log scale. The bar lengths might be misinterpreted by a user as a result.

## Expected behaviour

We should disallow log scale for discrete bar charts and stacked discrete bar charts.

We could also make it more obvious that the data is plotted on a log scale, but I can't think of a compelling use case for allowing bar charts on a log scale.

For discrete and stacked discrete bar charts, simply don't show the log/linear toggle. We currently don't have any bar charts that are on a log scale by default.

For line charts (that turn into a discrete bar chart):
- Hide the log/linear toggle when a discrete bar chart is viewed
- When changing to a bar chart (single year mode) then we should always display the bars in linear scale. Ideally when you switch back to a line chart that would be shown in log again if that was previously active.

## Additional context

<details><summary>Queries</summary>
<p>

```sql
select c.id, cc.chartType, cc.slug
from charts c
join chart_configs cc on c.configId = cc.id
where
    (cc.chartType = 'DiscreteBar' or cc.chartType = 'StackedDiscreteBar')
    and cc.full ->> '$.yAxis.scaleType' = 'log';
```

```sql
select c.id, cc.chartType, cc.slug
from charts c
join chart_configs cc on c.configId = cc.id
where
    cc.chartType = 'LineChart'
    and cc.full ->> '$.yAxis.scaleType' = 'log' and cc.full ->> '$.minTime' = cc.full ->> '$.maxTime';
```

```sql
select c.id, cc.chartType, cc.slug
from charts c
join chart_configs cc on c.configId = cc.id
where
    (cc.chartType = 'DiscreteBar' or cc.chartType = 'StackedDiscreteBar') and
    cc.full ->> '$.yAxis.canChangeScaleType' = 'true';
```

</p>
</details>

PR #4436, opened 2025-01-10
✨ disallow log y-scale for discrete bar charts / TAS-802

sophiamersmann:
Resolves https://github.com/owid/owid-grapher/issues/4364

Shows a discrete and stacked discrete bar chart with a linear scale always.

In particular:
- Hides the 'Enable log/linear toggle' in the admin for discrete bar charts
- Hides log/linear toggle in the settings menu for discrete bar charts
- Forces discrete bar charts to use a linear scale, even if the axis config says otherwise


danyx23 (approved):
Nice work, thanks!

PR #4449, opened 2025-01-14
Fix race condition with lighting update on staging

### LINKED ISSUE #3908: Lightning update sometimes doesn't update the article content
## Description

When lightning update bakes a single GDoc, it sometimes doesn't update the content.

## Expected behaviour

Lightning update should always result in updated content.

## Steps to reproduce

1. Publish a new GDoc
2. While the change is deploying, update its body
3. Click Republish button in its admin (it should have a lightning icon) and confirm the change
4. There should be two changes in the deploy queue in admin:
   * Full deploy triggerend by the publication of the article
   * Lightning update of the article
5. Check the content of the article after both changes complete - it should have the original content before update

## Environment

Staging server

## Additional context

AFAICT, the problem seems to be that we update the GDoc and enqueue the update in the same DB transaction, which creates a race condition.

If the deploy-queue process reads the change from the file system and starts its own transaction before the previous one commits, it will bake the GDoc with the old content. In my testing, it worked when the lighting update was the only change in the queue, but it failed when it was enqueued after a full bake that was in progress. Not sure if this makes a difference.

On prod, we trigger the bake in Buildkite instead of running it on the deploy-queue process, which is another code path I didn't verify, but presumably the race condition would be much less likely there.

rakyi:
Fixes #3908

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -240,7 +240,7 @@ getRouteNonIdempotentWithRWTransaction(
     "/gdocs/:id",
     getIndividualGdoc
 )
-putRouteWithRWTransaction(apiRouter, "/gdocs/:id", createOrUpdateGdoc)
+apiRouter.put("/gdocs/:id", createOrUpdateGdoc)
 deleteRouteWithRWTransaction(apiRouter, "/gdocs/:id", deleteGdoc)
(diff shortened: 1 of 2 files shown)


rakyi:
This unfortunately doesn't seem to fix the issue. The problem is probably somewhere eles.

Issue #4416, opened 2025-01-08
Cycle 2025.1: data insights with a single image


ikesau:
Making it possible to publish data insights with only one file, and updating the designs to assume this as the default going forward

## Motivation

Having to create 2 static visualizations per DI significantly increases the overhead in giving feedback and making revisions. We want to halve this by making it so that DIs only use one square image going forward.

## Proposed scope
[Figma](https://www.figma.com/design/whorgxK9AhKOgCd6gzWccN/Data-Insights-3.0?t=7vZU0sevIcAnWIdd-0)

**Must do**
- [x] Update DI feed page to make the sections narrower
- [x] Update DI typography and spacing
- [x] Update the homepage to use the square image
- [x] Update the way images are rendered in data insights
  - [x] Ensure `og:image` is still set correctly, and the Atom feed image.

## Technical notes
I think something like a `preferSmall` option for the `Image` component will be a nice way to migrate all previous DIs that have 2 images specified, while also allowing authors to only specify `filename` in archie going forward (`filename` is mandatory whereas `smallFilename` is optional - the other potential solution would be to change this so that validation ensures that either one or both are defined and then handle that in the `Image` component, but that seems more likely to leak into other parts of the site)

Issue #4451, opened 2025-01-14
Create a Gdoc content migration workflow


ikesau:
## Core problem

We currently have no way to programmatically update the content of google documents, if ever we want to migrate an archie component to have a different shape.

## Proposed solution

Create some sort of migration tooling that:
1. Takes a migration specification
2. Iterates through all published gdocs
3. Operates on them (probably as raw blocks)
4. Converts them back to archie
5. Replaces the google document content with the new archie
6. Republishes the post

## Possible usecases
- [Migrating data insights to have only one image](https://github.com/owid/owid-grapher/issues/4416) (i.e. dropping `filename` and renaming `smallFilename` -> `filename`)
- [Updating / removing all the titles in all-charts blocks](https://github.com/owid/owid-grapher/issues/4426)

## Technical considerations
If we operate on raw blocks to update the shape of a component, we'll potentially have to create a "version 1 backup" of the TS interface 

e.g.

```ts
import { RawBlockImage } from "@ourworldindata/types"

type OldRawBlockImage = {
  // the values this block used to have, before the migration
}

function migrateImage(block: OldRawBlockImage): RawBlockImage {
  // stuff
}


```

danyx23:
It would be worth checking out if anyone has solved this problem before (there are a few high profile users of ArchieML in Google Docs)

PR #4452, opened 2025-01-14
✨ prefer smallFilename for data insights

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/4416

image

Changes:
- Adds a `documentType` property to the `DocumentContext` object
- Adds a codepath in `Image.tsx` to always use the `smallFilename` (if defined) if the image is being used in a data insight
- Updates the latest data insight block with a similar prefer-small logic
- Updates CSS (one small update remains, pending clarification from Marwa)


rakyi (commented):
Just to confirm, we want to retroactively apply this change for all existing data insights, right?

For the logic, one of my favorite maxims of software development is "I don't know, I don't wanna know". It would be better if the `Image` didn't know where it is used, and instead we passed in a prop (`preferSmall`?) to achieve the same effect. We always know we are in a data insight at the call site after all, right? That way it will stay more decoupled.

ikesau:
Yup, retroactive (hence why I contemplated [this solution](https://github.com/owid/owid-grapher/issues/4451), but decided it was too much work for this somewhat-experimental project (even though I feel 80% confident that we'll stick with single-image data insights))

I like that maxim, and I feel you on the context codepath thing. `preferSmallfilename` is just such a stupid prop that we'd never use anywhere else. Prop-drilling it from `DataInsightBody` down to `ArticleBlocks` down to `ArticleBlock` down to `Image` seemed a little bit worse.

If you feel more strongly about it than I do, I'll gladly change it, though 🙂

rakyi:
I think adding it to the image block(s) in content body would be probably nicer than prop drilling it at the top level. But I prefer either of those to the context. So please change it, unless there is something problematic lurking there that we missed.

ikesau:
@rakyi changed 🙂

rakyi (approved):
Great, thank you!

PR #4465, opened 2025-01-17
🐛 fix immutable header mutation in handleThumbnailRequest

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/4450

Our `cors` middleware mutates all requests, but cached requests are immutable (a quick perusal through Cloudflare Workers/Functions docs didn't state this anywhere, but a lot of other third-party sources did)

This PR clones the cached thumbnail requests so that `corsify` can mutate them.

These issues are still in the `grapher/thumbnail` router, but that route should be deprecated, anyway.


marcelgerber (approved):
Nice fix!

PR #4474, opened 2025-01-22
Add Related research and writing to multi-dim pages

### LINKED ISSUE #4459: Related research and writing block for multidimensional data page
* Similar to normal data page, check for reference
* Extract links from `posts_gdocs_links` links
* Show any article that uses a chart which uses any of the indicators used in the mdim on the Y axis - should be the same logic as data pages, except a union of all indicators instead of single one

rakyi:
Resolves #4459


marcelgerber (approved):
Nice!
I quickly tested this on http://staging-site-mdim-related/admin/grapher/mdd-covid-xm-models, and all works well.

PR #4478, opened 2025-01-23
Fix multidimensional data pages baking logic

rakyi:
* Don't delete mdims in GrapherBaker
* Delete old mdims in MultiDimBaker
* Improve the console output
* Document issue with TSConfig and getting ESNext types

### DIFF
--- a/baker/GrapherBaker.tsx
+++ b/baker/GrapherBaker.tsx
@@ -4,7 +4,6 @@ import { renderToHtmlPage } from "../baker/siteRenderers.js"
 import {
     excludeUndefined,
     urlToSlug,
-    without,
     uniq,
     keyBy,
     compact,
@@ -17,7 +16,6 @@ import {
     BAKED_GRAPHER_URL,
 } from "../settings/serverSettings.js"
 import * as db from "../db/db.js"
-import { glob } from "glob"
 import { isPathRedirectedToExplorer } from "../explorerAdminServer/ExplorerRedirects.js"
 import {
     getPostIdFromSlug,
@@ -50,9 +48,10 @@ import {
 import { getAllImages } from "../db/model/Image.js"
 import { logErrorAndMaybeCaptureInSentry } from "../serverUtils/errorLog.js"
 
-import { getTagToSlugMap } from "./GrapherBakingUtils.js"
+import { deleteOldGraphers, getTagToSlugMap } from "./GrapherBakingUtils.js"
 import { knexRaw } from "../db/db.js"
 import { getRelatedChartsForVariable } from "../db/model/Chart.js"
+import { getAllMultiDimDataPageSlugs } from "../db/model/MultiDimDataPage.js"
 import pMap from "p-map"
 
 const renderDatapageIfApplicable = async (
@@ -286,28 +285,6 @@ const bakeGrapherPage = async (
             imageMetadataDictionary
         )
     )
-    console.log(outPath)
-}
-
-const deleteOldGraphers = async (bakedSiteDir: string, newSlugs: string[]) => {
-    // Delete any that are missing from the database
-    const oldSlugs = glob
-        .sync(`${bakedSiteDir}/grapher/*.html`)
-        .map((slug) =>
-            slug.replace(`${bakedSiteDir}/grapher/`, "").replace(".html", "")
-        )
-    const toRemove = without(oldSlugs, ...newSlugs)
-        // do not delete grapher slugs redirected to explorers
-        .filter((slug) => !isPathRedirectedToExplorer(`/grapher/${slug}`))
-    for (const slug of toRemove) {
-        const path = `${bakedSiteDir}/grapher/${slug}.html`
-        console.log(`DELETING ${path}`)
-        fs.unlink(path, (err) =>
-            err
-                ? console.error(`Error deleting ${path}`, err)
-                : console.log(`Deleted ${path}`)
-        )
-    }
 }
 
 export interface BakeSingleGrapherChartArguments {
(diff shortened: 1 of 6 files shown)


devTools/tsconfigs/tsconfig.base.json
```diff
@@ -12,12 +12,8 @@
...
         "declarationMap": true,
 
-        // We start with https://github.com/tsconfig/bases/blob/master/bases/node14.json then drop:
-        //   "lib": ["es2020"],
-        //   "target": "es2020",
-
-        "lib": ["dom", "dom.iterable", "es2020", "es2021", "es2022", "es2023"],
-        "target": "es2019",
+        "lib": ["dom", "dom.iterable", "ESNext"],
+        "target": "ES2022",
```
rakyi, line 16: This [should be](https://github.com/tsconfig/bases/blob/main/bases/node22.json) safe for Node 22.

devTools/tsconfigs/tsconfig.base.json
```diff
@@ -12,12 +12,8 @@
         "declaration": true,
         "declarationMap": true,
 
-        // We start with https://github.com/tsconfig/bases/blob/master/bases/node14.json then drop:
-        //   "lib": ["es2020"],
-        //   "target": "es2020",
-
-        "lib": ["dom", "dom.iterable", "es2020", "es2021", "es2022", "es2023"],
-        "target": "es2019",
+        "lib": ["dom", "dom.iterable", "ESNext"],
```
rakyi, line 15: ESNext is necessary for set `intersection`.

baker/GrapherBaker.tsx
```diff
@@ -308,7 +309,6 @@ const bakeGrapherPageAndVariablesPngAndSVGIfChanged = async (
             imageMetadataDictionary
         )
     )
-    console.log(outPath)
```
rakyi, line 311: We output the slug in the progress bar already, this doesn't add anything useful on top of that.

baker/GrapherBaker.tsx
```diff
@@ -432,7 +431,7 @@ export const bakeAllChangedGrapherPagesVariablesPngSvgAndDeleteRemovedGraphers =
         )
 
         const progressBar = new ProgressBar(
-            "bake grapher page [:bar] :current/:total :elapseds :rate/s :etas :name\n",
```
rakyi, line 435: It's not clear this is an ETA in the output, which is already cluttered, and doesn't seem useful.

baker/GrapherBaker.tsx
```diff
@@ -451,11 +450,18 @@ export const bakeAllChangedGrapherPagesVariablesPngSvgAndDeleteRemovedGraphers =
                     async (knex) => await bakeSingleGrapherChart(job, knex),
                     db.TransactionCloseMode.KeepOpen
                 )
-                progressBar.tick({ name: `slug ${job.slug}` })
+                progressBar.tick({ name: job.slug })
```
rakyi, line 453: Repeating `slug` on every line doesn't add anything useful.

mlbrgl (commented):
Nice!

My main concern is about the repetition of deleteOldGrapher / deleteOldMultiDimPages:
 1- code redundancy: they largely do the same thing
 2- execution redundancy: the deletion of obsolete /grapher pages happens twice (once at the grapher step, and another time at the mdim step)

2 is more of an optimzation for full bakes, and doesn't probably count for much so I wouldn't necessarily complexify here. Also it's nice that each step is self-contained when we do partial bakes. 

The near duplication of deleteOldGrapher in 1 is a bit more cumbersome, and it would be more straightforward to have a single function that deals with the overall deletion of /grapher pages (even if it called twice, as per 2).

It would be also a good opportunity to deal with the confusion around redirects to explorers.

On the one hand, we don't delete grapher pages that have been redirected to explorers:

https://github.com/owid/owid-grapher/blob/2d026493378f4114dd25e68db2c100490b5b62f4/baker/GrapherBaker.tsx#L344-L346

On the other hand, we don't bake grapher pages that have been redirected to explorers:

https://github.com/owid/owid-grapher/blob/2d026493378f4114dd25e68db2c100490b5b62f4/baker/GrapherBaker.tsx#L377-L382

Meanwhile, deleteOldMultiDimPages does delete grapher pages that have been redirected to explorers, and effectively overrides deleteOldGrapher's safeguard. If this safeguard is no longer relevant, I'd remove it rather than having it in one place and not the other, which is confusing.

baker/MultiDimBaker.tsx
```diff
@@ -219,12 +223,42 @@ export const bakeMultiDimDataPage = async (
...
+                    .replace(`${bakedSiteDir}/grapher/`, "")
+                    .replace(".html", "")
+            )
+    )
+    const toRemove = oldSlugs.difference(newSlugs)
+    for (const slug of toRemove) {
+        console.log(`DELETING ${slug}`)
+        const path = `${bakedSiteDir}/grapher/${slug}.html`
+        await fs.unlink(path)
+        console.log(path)
```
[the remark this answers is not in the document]
rakyi: I did it that way to be consistent with `deleteOldGraphers`, but that has since been changed in master, so I'm simplifying it.

baker/MultiDimBaker.tsx
```diff
@@ -233,5 +267,11 @@ export const bakeAllMultiDimDataPages = async (
             row.config,
             imageMetadata
         )
+        progressBar.tick({ name: slug })
     }
+    const publishedSlugs = await getAllPublishedMultiDimDataPageSlugs(knex)
```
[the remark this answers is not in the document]
rakyi: Yep. I started with that, then changed the logic a few times and then forgot that was an option. 😄

rakyi:
Good point, I moved the deletion of old graphers to a single helper function called in both places. After the removal of image exports, there is no longer a distinction between normal graphers and mdims. I kept the safeguard for redirects to explorers just in case

mlbrgl (approved):
Looks good, thanks for the updates!

PR #4487, opened 2025-01-24
Sentry cloudflare

### LINKED ISSUE #4187: Cloudflare Function error logging to Sentry
When working on upgrading the donation flow, we temporarily added Sentry logging, but then removed it (See https://github.com/owid/owid-grapher/issues/3177 and https://github.com/owid/owid-grapher/pull/3732)

Now that we're doing more and more things through CF functions, it would be nice to get a better sense of what types of errors are occurring and why, because the CF analytics dashboard isn't very granular.

We should add the Sentry middleware to each route (making sure to check with @bnjmacdonald that this won't eat too much into our session replay budget)

rakyi:
Resolves #4187


ikesau (commented):
Nice! It's a shame we lose the source map 😢 

I tried visiting https://sentry-cloudflare.owid.pages.dev/deleted/deleted-article but it didn't seem to log the error in Sentry. Did you have to do something else to get it to log?

rakyi:
I added a temporary code that raised the error on that URL to test that the integration works, and then I removed it. You could do the same, if you want to double-check.

ikesau (approved):

PR #4489, opened 2025-01-27
✨ add narrative-chart and figma-url to DI frontmatter / TAS-850

sophiamersmann:
Adds two new properties to the front matter of DIs.

- `narrative-chart` is the name of a narrative chart. It's useful to know which narrative chart a DI is based on, similar to how we track the `grapher-url` a DI is based on.
- `figma-url` is a link to a Figma page. Authors are encouraged to share a Figma link if they customised the chart, but currently paste the link into the comment section of the GDoc. We double-down on Figma as an important piece of infrastructure this cycle, so I think it's justified to track it properly. 

Both properties will be surfaced in a new Data Insights index page we have in mind.

---
This is **part 1 of 9 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #4547 
- <kbd>&nbsp;2&nbsp;</kbd> #4490 
- <kbd>&nbsp;1&nbsp;</kbd> #4489 👈

### DIFF
--- a/adminSiteClient/GdocsSettingsForms.tsx
+++ b/adminSiteClient/GdocsSettingsForms.tsx
@@ -180,7 +180,12 @@ export const GdocInsightSettings = ({
         <form className="GdocsSettingsForm">
             <GdocCommonErrors
                 errors={errors}
-                errorsToFilter={["approved-by", "grapher-url"]}
+                errorsToFilter={[
+                    "approved-by",
+                    "grapher-url",
+                    "narrative-chart",
+                    "figma-url",
+                ]}
             />
             <GdocCommonSettings
                 gdoc={gdoc}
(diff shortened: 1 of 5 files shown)


ikesau (approved):
this is a perfectly implemented PR, but I worry we're starting to overload these warnings so that authors will learn to ignore them 🙈 

I think the checklist method works better for them, so hopefully your DI tool will make this a non-issue anyway because they'll have already filled everything, everytime, before hitting publish. 🙂

ikesau:
Oh, you might want to add these to the `_omittableFields` property on `GdocDataInsight` so that we don't send them to the client.

e.g. inspect `_OWID_GDOC_PROPS.content["grapher-url"]` on https://ourworldindata.org/data-insights/papua-new-guinea-has-more-living-languages-than-any-other-country

We probably don't want arbitrary figma URLs getting embedded in our HTML 🙂

sophiamersmann:
> this is a perfectly implemented PR, but I worry we're starting to overload these warnings so that authors will learn to ignore them 🙈 

That's a good point! I'm almost inclined to remove all warnings – or maybe keep the warning for the `narrative-chart` field since that's the preferred flow (although there are good reasons to specify a `grapher-url` instead, e.g. when creating DIs from explorers or MDims). Mhhh.. I think you're right in saying that this is not the right place to prompt for this information, so I think I'm in favour of removing all warnings, but I don't have a strong sense of what's right here, so please feel free to veto.

ikesau (approved):

PR #4490, opened 2025-01-27
🎉 (admin) data insight index page / TAS-845

sophiamersmann:
Adds an index page for data insights in the admin.

### Summary

- The /dataInsights endpoint returns all data insights, enriched with information that is surfaced on the index page (e.g. image, topic tags, chart type)
- The index page has a list and gallery layout (top right) and allows to filter by publication status.

### Notes

- The data insights query assumes the image block comes first
- Because we're now surfacing the `grapher-url`, I added some validation for it. As far as I know, the `grapher-url` should be a valid grapher or explorer link, but maybe I'm forgetting something?
- Search includes the body text. I just grabbed the markdown column for that because it was easiest. What's a little weird about that is that the markdown contains image filenames (and maybe other stuff), which sometimes leads to weird search results. I don't think that's too bad but if it's straight-forward to concat all text elements instead, let me know and I'm happy to rewrite
- Data insight images that are not squared get squished in the gallery (see screenshot below)

<details><summary><b>Screenshots</b></summary>
<p>
Screenshot 2025-01-30 at 10 42 08
</p>
</details>

---
This is **part 2 of 9 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #4547 
- <kbd>&nbsp;2&nbsp;</kbd> #4490 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #4489

### DIFF
--- a/adminSiteClient/AdminSidebar.tsx
+++ b/adminSiteClient/AdminSidebar.tsx
@@ -20,6 +20,7 @@ import {
     faSitemap,
     faPanorama,
     faImage,
+    faLightbulb,
 } from "@fortawesome/free-solid-svg-icons"
 
 import { ETL_WIZARD_URL } from "../settings/clientSettings.js"
@@ -36,6 +37,11 @@ export const AdminSidebar = (): React.ReactElement => (
             <li>
                 <Link to="/chartViews">
                     <FontAwesomeIcon icon={faPanorama} /> Narrative charts
+                </Link>
+            </li>
+            <li>
+                <Link to="/data-insights">
+                    <FontAwesomeIcon icon={faLightbulb} /> Data insights
                 </Link>
             </li>
             <li>
--- a/adminSiteClient/DataInsightIndexPage.tsx
+++ b/adminSiteClient/DataInsightIndexPage.tsx
@@ -1,0 +1,513 @@
+import { useContext, useEffect, useMemo, useState } from "react"
+import * as React from "react"
+import { Button, Card, Flex, Input, Radio, Select, Space, Table } from "antd"
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
+import {
+    faCopy,
+    faPen,
+    faUpRightFromSquare,
+} from "@fortawesome/free-solid-svg-icons"
+import { faFigma } from "@fortawesome/free-brands-svg-icons"
+
+import { AdminLayout } from "./AdminLayout.js"
+import { Timeago } from "./Forms.js"
+import { ColumnsType } from "antd/es/table/InternalTable.js"
+import {
+    buildSearchWordsFromSearchString,
+    filterFunctionForSearchWords,
+    highlightFunctionForSearchWords,
+    SearchWord,
+} from "../adminShared/search.js"
+import {
+    ALL_GRAPHER_CHART_TYPES,
+    DbPlainTag,
+    GRAPHER_MAP_TYPE,
+    GrapherChartOrMapType,
+    OwidGdocDataInsightIndexItem,
+} from "@ourworldindata/types"
+import {
+    copyToClipboard,
+    dayjs,
+    RequiredBy,
+    startCase,
+} from "@ourworldindata/utils"
+import {
+    BAKED_BASE_URL,
+    CLOUDFLARE_IMAGES_URL,
+} from "../settings/clientSettings.js"
+import { AdminAppContext } from "./AdminAppContext.js"
+
+type ChartTypeFilter = GrapherChartOrMapType | "all"
+type PublicationFilter = "all" | "published" | "scheduled" | "draft"
+type Layout = "list" | "gallery"
+
+const DEFAULT_CHART_TYPE_FILTER: ChartTypeFilter = "all"
+const DEFAULT_PUBLICATION_FILTER: PublicationFilter = "all"
+const DEFAULT_LAYOUT: Layout = "list"
+
+const editIcon = <FontAwesomeIcon icon={faPen} size="sm" />
+const linkIcon = <FontAwesomeIcon icon={faUpRightFromSquare} size="sm" />
+const figmaIcon = <FontAwesomeIcon icon={faFigma} size="sm" />
+const copyIcon = <FontAwesomeIcon icon={faCopy} size="sm" />
+
+function createColumns(ctx: {
+    highlightFn: (
+        text: string | null | undefined
+    ) => React.ReactElement | string
+}): ColumnsType<OwidGdocDataInsightIndexItem> {
+    return [
+        {
+            title: "Preview",
+            key: "preview",
+            render: (_, dataInsight) =>
+                hasImage(dataInsight) ? (
+                    <>
+                        <img
+                            className="border"
+                            src={makePreviewImageSrc(dataInsight)}
+                            style={{ maxWidth: 200 }}
+                        />
+                        <Button
+                            type="text"
+                            size="small"
+                            icon={copyIcon}
+                            onClick={() =>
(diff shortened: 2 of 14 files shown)


adminSiteServer/apiRoutes/dataInsights.ts
```diff
@@ -0,0 +1,294 @@
...
+        LEFT JOIN published_charts cc_grapherUrl
+            ON cc_grapherUrl.slug = SUBSTRING_INDEX(SUBSTRING_INDEX(content ->> '$."grapher-url"', '/grapher/', -1), '\\?', 1)
+
+        -- join the chart_views table to get the config of the narrative chart
+        LEFT JOIN chart_views cw
+            ON cw.name = content ->> '$."narrative-chart"'
+        LEFT JOIN chart_configs cc_chartView
+            ON cc_chartView.id = cw.chartConfigId
+
+        -- join the images table by filename (only works for data insights where the image block comes first)
```
ikesau, line 97: I don't think any author has ever _not_ done this, but this will be the first time we explicitly hinge business logic on this assumption - so maybe we should now add validation to assert this?

adminSiteServer/apiRoutes/dataInsights.ts
```diff
@@ -0,0 +1,294 @@
...
+): OwidGdocDataInsightIndexItem {
+    const content = parsePostGdocContent(
+        dataInsight.content
+    ) as OwidGdocDataInsightContent
+    const chartConfig = dataInsight.chartConfig
+        ? parseChartConfig(dataInsight.chartConfig)
+        : undefined
+
+    // check if the given grapher-url is a valid Grapher or Explorer URL
+    const grapherUrlField = content["grapher-url"]?.trim()
```
ikesau, line 130: If trimming is necessary here, we may as well fix it upstream

https://github.com/owid/owid-grapher/blob/d132fba478a7c123973a3ecaadb5b8f8a74b1d07/db/model/Gdoc/archieToEnriched.ts#L313-L319

adminSiteServer/apiRoutes/dataInsights.ts
```diff
@@ -0,0 +1,294 @@
...
+    return undefined
+}
+
+function getChartTypeFromConfig(
+    chartConfig: GrapherInterface
+): GrapherChartOrMapType | undefined {
+    return getChartTypeFromConfigAndQueryParams(chartConfig)
+}
+
+function getChartTypeFromConfigAndQueryParams(
```
ikesau, line 214: Is this the right folder for this function? I'm surprised we don't have something like this already, or that it wouldn't be better to colocate it with other config utils.

In any case, it seems a bit easy to forget to update if we add a new chart type - do you think that's likely to be a problem? And if so, would `ts-pattern`'s  `P.match().exhaustive()` help?

ikesau: Also, oh no, just realizing all of this is for the column I suggested you drop 💀

adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -0,0 +1,457 @@
...
+                        </Button>
+                    </>
+                ) : undefined,
+        },
+        {
+            title: "Title",
+            dataIndex: "title",
+            key: "title",
+            width: 300,
+            render: (title) => ctx.highlightFn(title),
```
ikesau, line 77: Might be useful to make this a link if the DI's been published? IME authors/charlie are often looking for ways to quickly get links to things to use in different places

ikesau (approved):
Really nicely done! 

I've left quite a few comments but none of them are blocking, I don't think. I _do_ think it'd be nice to get `EditableTags` working, though I realize that'll probably be a headache to fit in the table. 🙂

adminSiteClient/AdminSidebar.tsx
```diff
@@ -38,6 +39,11 @@ export const AdminSidebar = (): React.ReactElement => (
                     <FontAwesomeIcon icon={faPanorama} /> Narrative charts
                 </Link>
             </li>
+            <li>
+                <Link to="/dataInsights">
+                    <FontAwesomeIcon icon={faLightbulb} /> Data insights
```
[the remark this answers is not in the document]
sophiamersmann: I always wonder if there is a story behind the cross bones for the bulk chart editor, but it does work... I am scared of it

adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -0,0 +1,457 @@
...
+                        ` (approved by ${dataInsight.approvedBy})`}
+                </>
+            ),
+        },
+        {
+            title: "Topic tags",
+            dataIndex: "tags",
+            key: "tags",
+            render: (tags: DbPlainTag[]) =>
+                tags.map((tag) => (
```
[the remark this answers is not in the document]
sophiamersmann: yeah, that's a good idea! it hadn't crossed my mind

adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -0,0 +1,457 @@
...
+    dataInsight: RequiredBy<OwidGdocDataInsightIndexItem, "narrativeChart">
+) {
+    return `/admin/chartViews/${dataInsight.narrativeChart.id}/edit`
+}
+
+function makePreviewImageSrc(
+    dataInsight: RequiredBy<OwidGdocDataInsightIndexItem, "image">
+) {
+    const { cloudflareId, originalWidth } = dataInsight.image
+    return `${CLOUDFLARE_IMAGES_URL}/${cloudflareId}/w=${originalWidth}`
```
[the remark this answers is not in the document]
sophiamersmann: somehow I thought this has be to be the original width, lol

adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -0,0 +1,457 @@
...
+                        target="_blank"
+                        rel="noreferrer noopener"
+                        style={{ display: "block" }}
+                    >
+                        {ctx.highlightFn(tag.name)}
+                    </a>
+                )),
+        },
+        {
+            title: "Chart type",
```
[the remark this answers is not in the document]
sophiamersmann: It gets even worse in the next PR :(

Marwa is keen on the chart type being present. The use case we envision for the gallery is that authors can search for a chart type and get inspiration from past DIs.

Now that I think about it, it might make sense to drop the column and add another dropdown menu for chart types.

Adding size="small" is a good idea in any case :)

adminSiteServer/apiRoutes/dataInsights.ts
sophiamersmann: Yeah, that's a good idea! I was debating whether it's okay to make this assumption here. I finally settled on this because writing more code to cover cases outside our guidelines didn't make much sense to me. Good call to make it explicit!

adminSiteServer/apiRoutes/dataInsights.ts
sophiamersmann: I don't think it's that common, to be honest, since we usually have a Grapher instance in our hands and can ask it for its type. But you're right that we should find a nicer spot for this code. I'll move it, but I'm not sure where to – I'll think about it.

adminSiteClient/DataInsightIndexPage.tsx
sophiamersmann: Nice! Good idea. Never let an opportunity to turn text into a link pass by...

sophiamersmann:
Thanks for the review, @ikesau! Making the tags editable is a very good idea, and I hope dropping the chart type column helps to make it fit :)

PR #4494, opened 2025-01-27
✨ check there's no slug collision before publishing

### LINKED ISSUE #4493: Validate that slug is unique before publishing/republishing a gdoc
Turns out it's currently possible to publish an article on a slug that's already taken. It shouldn't be.

[Slack thread for context](https://owid.slack.com/archives/C03N1V9KXB9/p1737957835396659)

ikesau:
Resolves #4493

Compares via canonical URL in the event we want `/data-insights/some-slug` and `/some-slug`

Renamed `addImagesToContentGraph` to `setImagesToContentGraph` because I was confused when reading it to see that it also took care of deleting images from the content graph if unpublishing.

https://github.com/user-attachments/assets/496f66d0-b7a7-49d3-a2e2-8d24958d0ed4

### DIFF
--- a/adminSiteClient/admin.scss
+++ b/adminSiteClient/admin.scss
@@ -740,7 +740,8 @@ $nav-height: 45px;
 }
 
 .errorMessage {
-    z-index: 2000;
+    // 1 above .ant-modal-mask
+    z-index: 2001;
     .modal-dialog {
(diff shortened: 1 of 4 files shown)


rakyi (approved):
Nice QoL improvement!

PR #4498, opened 2025-01-28
🎉 hide footer when printing, better article layout

ikesau:
Still pretty glitchy on non-chrome browsers - this is just a best-effort attempt, timeboxed to an hour of tinkering. Feel free to add patches to the branch if you get nerd-sniped, but I recommend approving and running away as fast as you can 😁 

https://github.com/user-attachments/assets/c9209dac-9cca-4bff-99b7-a57fd9827e2f

### DIFF
--- a/site/gdocs/components/centered-article.scss
+++ b/site/gdocs/components/centered-article.scss
@@ -164,6 +164,40 @@
             }
         }
     }
+
+    // Quick and dirty layout overrides for article printing
+    // Stacks all elements into a single, full-width column
+    // There's lots to improve, but it's better than before!
+    @media print {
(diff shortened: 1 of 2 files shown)


marcelgerber (approved):
Nice, definitely an improvement! I only got very slightly nerdsniped into hiding the cookie banner in print.

I wonder if it would make sense to have `grid-print-12` or so utility classes, similar to how we have these for various breakpoints? But maybe it's also quite some effort to make sure these are actually used, then?

ikesau:
I don't think any work on this will "make sense" in terms of investment:reward 😂

PR #4499, opened 2025-01-29
Fix prominent link URL

rakyi:
- When previewing, link to the baked site. This is necessary for previews to work correctly in admin in prod, where relative URL isn't sufficient, because admin is on a different domain then the main site.
- Don't add hash and query string to gdoc URLs. Gdocs can have params like `?tab=t.0` and similar internal hashes that we shouldn't add to our own URLs.

  It seems the support for query strings and hashes was added on purpose but there are only a couple gdocs with link with seemingly custom query param `?usp=sharing`, so it seems fine to remove that support for now. When we'll need it, we should add a way to specify the query params separately from the URL (or filter out the known godcs query params, which seems too fragile).

### DIFF
--- a/site/gdocs/components/ProminentLink.tsx
+++ b/site/gdocs/components/ProminentLink.tsx
@@ -12,6 +12,7 @@ import { GRAPHER_DYNAMIC_THUMBNAIL_URL } from "../../../settings/clientSettings.
 import {
     ARCHVED_THUMBNAIL_FILENAME,
     DEFAULT_THUMBNAIL_FILENAME,
+    OwidGdocLinkType,
 } from "@ourworldindata/types"
 
 const Thumbnail = ({ thumbnail }: { thumbnail: string }) => {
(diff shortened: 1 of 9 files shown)


site/gdocs/utils.ts
```diff
@@ -85,13 +89,13 @@ export const useLinkedDocument = (
         errorMessage = `Article with slug "${linkedDocument.slug}" isn't published.`
     }
 
-    //todo replace with getCanonicalUrl
-    const subdirectory =
-        linkedDocument.type === OwidGdocType.DataInsight ? "data-insights/" : ""
     return {
         linkedDocument: {
             ...linkedDocument,
-            slug: `${subdirectory}${linkedDocument.slug}${queryString}${hash}`,
```
ikesau, line 94: I really think we'll miss query strings and hashes sooner rather than later, for topic page key insights and linking to the subscribe section on our homepage.

I guess as a workaround we have the absolute URL method, but can I make the case for keeping them around and just filtering out the `tab` param from gdocs?

site/gdocs/components/ProminentLink.tsx
```diff
@@ -63,17 +64,17 @@ export const ProminentLink = (props: {
     let title: string | undefined = props.title
     let description: string | undefined = props.description
     let thumbnail: string | undefined = props.thumbnail
-    if (linkType === "gdoc") {
-        href = `/${linkedDocument?.slug}`
-        title ??= linkedDocument?.title
-        description ??= linkedDocument?.excerpt
-        thumbnail ??= linkedDocument?.["featured-image"]
-    } else if (linkType === "grapher") {
+    if (linkedDocument && linkType === OwidGdocLinkType.Gdoc) {
```
ikesau, line 67: `&& linkType === OwidGdocLinkType.Gdoc` is redundant, I think

ikesau (approved):
Thanks for this! That was a pretty lazy todo in `useLinkedDocument` 😂 

You could merge this as is, but my preference would be for gdocs links to still support query strings and hashes.

site/gdocs/utils.ts
rakyi: It's unintuitive for the users to edit the gdoc URL and add our own hash/query params. I hope we'll find a better solution for that. And so far we didn't really use the current affordance, so let's cross the bridge when we get there.

PR #4505, opened 2025-01-30
Add admin for multidimensional data pages

### LINKED ISSUE #4458: An admin to publish mdims and set the publication slug
* Table view with a list of mdims, with columns:
  * (Optional) Thumbnail of the default view
  * Title (with a link to the preview page)
  * Editable slug
  * Button(s) to publish/unpublish
  * (Optional) published at (change existing column to be a date)
  * (Optional) updated at
* Search by title and/or slug
* Validate slug uniqueness
  * Across mdims
  * Across normal graphers
  * Across grapher redirects
  * Add the same, but in reverse, to the chart admin

rakyi:
Resolves #4458

ETL path and a new/changed admin API will be in a follow-up PR.

### DIFF
--- a/adminSiteClient/AdminApp.tsx
+++ b/adminSiteClient/AdminApp.tsx
@@ -1,3 +1,4 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
 import * as React from "react"
 import { Admin } from "./Admin.js"
 import { ChartEditorPage } from "./ChartEditorPage.js"
@@ -46,6 +47,17 @@ import { ChartViewEditorPage } from "./ChartViewEditorPage.js"
 import { ChartViewIndexPage } from "./ChartViewIndexPage.js"
 import { ImageIndexPage } from "./ImagesIndexPage.js"
 import { DataInsightIndexPage } from "./DataInsightIndexPage.js"
+import { MultiDimIndexPage } from "./MultiDimIndexPage.js"
+
+const queryClient = new QueryClient({
+    defaultOptions: {
+        queries: {
+            // This is not what we've done in the past, so the behavior might
+            // be confusing for users.
+            refetchOnWindowFocus: false,
+        },
+    },
+})
 
 @observer
 class AdminErrorMessage extends React.Component<{ admin: Admin }> {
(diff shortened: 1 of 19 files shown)


adminSiteClient/ChartViewIndexPage.tsx
```diff
@@ -1,11 +1,10 @@
 import { useCallback, useContext, useEffect, useMemo, useState } from "react"
 import * as React from "react"
-import { Button, Flex, Input, Space, Table } from "antd"
```
rakyi, line 3: We can use the public type instead of the internal one.

adminSiteClient/AdminApp.tsx
```diff
@@ -111,262 +121,284 @@ export class AdminApp extends React.Component<{
...
-                            <Route
-                                exact
-                                path={`/${EXPLORERS_ROUTE_FOLDER}/:slug`}
-                                render={({ match }) => (
-                                    <AdminLayout title="Create Explorer">
-                                        <ExplorerCreatePage
-                                            slug={match.params.slug}
-                                            gitCmsBranchName={gitCmsBranchName}
-                                            manager={admin}
+            <QueryClientProvider client={queryClient}>
```
rakyi, line 124: The changes below are mostly whitespace, except the new route and the new provider.

adminSiteServer/apiRouter.ts
```diff
@@ -264,11 +268,9 @@ deleteRouteWithRWTransaction(apiRouter, "/images/:id", deleteImageHandler)
 getRouteWithROTransaction(apiRouter, "/images/usage", getImageUsageHandler)
 
 // Mdim routes
-putRouteWithRWTransaction(
-    apiRouter,
-    "/multi-dim/:slug",
-    handleMultiDimDataPageRequest
-)
+getRouteWithROTransaction(apiRouter, "/multi-dims.json", handleGetMultiDims)
+putRouteWithRWTransaction(apiRouter, "/multi-dim/:slug", handlePutMultiDim)
```
rakyi, line 272: We will change the API in a follow-up PR (including making it plural).

adminSiteServer/validation.ts
rakyi, line 1: Most of the code moved from `adminSiteServer/apiRoutes/charts.ts`.

packages/@ourworldindata/types/src/dbTypes/ChartConfigs.ts
```diff
@@ -31,8 +31,8 @@ export function serializeChartConfig(config: GrapherInterface): JsonString {
 }
 
 export function parseChartConfigsRow(
-    row: DbRawChartConfig
-): DbEnrichedChartConfig {
+    row: Pick<DbRawChartConfig, "patch" | "full">
+): Pick<DbEnrichedChartConfig, "patch" | "full"> {
```
rakyi, line 35: To make it work for cases when we don't pass the whole config around.

packages/@ourworldindata/types/src/domainTypes/Various.ts
```diff
@@ -61,8 +61,8 @@ export enum TaggableType {
 // Exception format that can be easily given as an API error
 export class JsonError extends Error {
     status: number
-    constructor(message: string, status?: number) {
-        super(message)
+    constructor(message: string, status?: number, options?: ErrorOptions) {
+        super(message, options)
```
rakyi, line 65: To make the [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) work.

adminSiteServer/multiDim.ts
```diff
@@ -329,3 +324,57 @@ export async function createMultiDimConfig(
...
+    knex: db.KnexReadWriteTransaction,
+    multiDim: DbEnrichedMultiDimDataPage,
+    published: boolean
+) {
+    const chartConfigs = await getChartConfigsByIds(
+        knex,
+        multiDim.config.views.map((view) => view.fullConfigId)
+    )
+
+    await Promise.all(
```
marcelgerber, line 348: I'm unsure about the limitless concurrency this can run in, if the mdim has many views.
Maybe using `p-map` with `{concurrency: 10}` or something is better suited here?

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -22,9 +67,10 @@ export async function handleMultiDimDataPageRequest(
     }
     const rawConfig = req.body as MultiDimDataPageConfigRaw
     const id = await createMultiDimConfig(trx, slug, rawConfig)
+
     if (
         FEATURE_FLAGS.has(FeatureFlagFeature.MultiDimDataPage) &&
-        (await isMultiDimDataPagePublished(trx, slug))
+        (await multiDimDataPageExists(trx, { slug, published: true }))
     ) {
```
marcelgerber, line 74: This code path here should also call `validateGrapherSlug`

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -33,3 +79,35 @@ export async function handleMultiDimDataPageRequest(
...
+    if (!multiDim) {
+        throw new JsonError(`Multi-dimensional data page not found`, 404)
+    }
+    const { published, slug } = req.body
+    let action
+    if (published !== undefined && published !== multiDim.published) {
+        multiDim = await setMultiDimPublished(trx, multiDim, published)
+        action = published ? "publish" : "unpublish"
+    }
+    if (slug !== undefined && slug !== multiDim.slug) {
```
marcelgerber, line 99: A few things here:
- I think we should first handle slug, then publish state.
- In theory, we only need to call `validateGrapherSlug` if the mdim is published.
- And we should probably also call it at the point _when_ a mdim is published (i.e. publish state changes to true).

adminSiteClient/MultiDimIndexPage.tsx
```diff
@@ -0,0 +1,282 @@
...
+            sorter: (a, b) => a.title.localeCompare(b.title),
+        },
+        {
+            title: "Slug",
+            dataIndex: "slug",
+            key: "slug",
+            render: (slug, { id }) => (
+                <SlugField id={id} slug={slug} slugMutation={slugMutation} />
+            ),
+            sorter: (a, b) => a.title.localeCompare(b.title),
```
marcelgerber, line 134:
```suggestion
            sorter: (a, b) => a.slug.localeCompare(b.slug),
```

adminSiteClient/MultiDimIndexPage.tsx
```diff
@@ -0,0 +1,282 @@
...
+                field.toLowerCase().includes(query)
+            )
+        )
+    }, [data, search])
+
+    const columns = createColumns(slugMutation, publishMutation)
+
+    return (
+        <AdminLayout title="Multidimensional Data Pages">
+            <NotificationContext.Provider value={null}>
```
marcelgerber, line 263: I don't think we need this extra context, since it's always null and not antd-specific anyhow?

adminSiteClient/MultiDimIndexPage.tsx
marcelgerber, line 1: The combination of react-query and antd makes this page feel really nice! :tada:

marcelgerber (commented):
Nicely done!

db/model/Gdoc/GdocBase.ts
```diff
@@ -830,7 +830,10 @@ export class GdocBase implements OwidGdocBaseInterface {
                 case OwidGdocLinkType.Grapher: {
                     if (
                         !chartIdsBySlug[link.target] &&
-                        !(await isMultiDimDataPagePublished(knex, link.target))
+                        !(await multiDimDataPageExists(knex, {
+                            slug: link.target,
+                            published: true,
+                        }))
```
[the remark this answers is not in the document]
rakyi: Because I wanted to get rid of `isMultiDimDataPagePublished` after I needed the existence check based on other stuff than `published` and then I forgot to remove it. Removing!

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -285,50 +286,12 @@ export const saveGrapher = async (
...
-                `This chart slug was previously used by another chart: ${newConfig.slug}`
-            )
-        else if (await isSlugUsedInOtherGrapher())
-            throw new JsonError(
-                `This chart slug is in use by another published chart: ${newConfig.slug}`
-            )
-        else if (
+        if (newConfig.slug) {
+            await validateGrapherSlug(knex, newConfig.slug, existingConfig?.id)
+        }
```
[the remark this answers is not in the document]
rakyi: Ah, of course! Good catch, fixing.

adminSiteServer/multiDim.ts
rakyi: I'm happy to change it, and I'd be curious to hear if this was an issue in practice before.

adminSiteServer/apiRoutes/mdims.ts
rakyi: Hmm, true. I didn't realize we are checking only for published graphers in the validation. That's even more complex.

adminSiteServer/multiDim.ts
marcelgerber: I'm not quite sure honestly!
I think undici (Node's fetch implementation) limits the number of open connections by itself, and I believe we have run into some kind of error related to that in the past?

adminSiteClient/MultiDimIndexPage.tsx
rakyi: Good catch!

adminSiteServer/multiDim.ts
marcelgerber: Yeah, go for it! It's all just vibes-based anyway 😁

rakyi: Hmm, in that case, I'd probably leave it as `Promise.all()` and fix it if we ever hit a problem. The S3 client doesn't seem to use undici.

adminSiteServer/apiRoutes/mdims.ts
rakyi: * Why do you think we should handle the slug first? I don't see it.
* I'd prefer if we kept the mdim slug valid even before publishing. I'd change the logic when we'll have a use case for it.
* I added the additional validation — I think it's fine to validate the slug twice in when the mdim is published instead of making the logic more complex.

marcelgerber: - (1) otherwise we could, in theory, end up in a state where we mark the mdim as published (under the old slug), and then the slug rename fails, i.e. we have it published under the old slug.
  I realize that this cannot currently happen, because we either change the slug or the published state, not both at once - but just to be a tiny bit on the safer side, I still think switching these around makes sense.

marcelgerber (approved):
All good now! Thanks for addressing these.

adminSiteServer/apiRoutes/mdims.ts
rakyi: I switched the order and also moved the validation logic earlier, so it can also fail before we update stuff to R2.

rakyi:
Thanks for the review! Can you check again the last commit, pls?

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -93,19 +95,23 @@ export async function handlePatchMultiDim(
...
     if (slug !== undefined && slug !== multiDim.slug) {
-        await validateGrapherSlug(trx, slug)
+        await validateNewGrapherSlug(trx, slug)
         multiDim = await setMultiDimSlug(trx, multiDim, slug)
-        if (!action && multiDim.published) {
+        if (multiDim.published) {
             action = "publish"
         }
     }
+    // Note: Keep this change last, since we don't want to update the configs
```
marcelgerber, line 105: I don't know precisely when a DB rollback will happen, honestly.
Could it happen that we change the slug (and propagate that to R2), and then roll that back in the DB afterwards?

rakyi: It can still happen, it can roll back any time we throw an error before the helper knex function somewhere in our Express handler returns. But the later we do the R2 upload, the better.

We have this problem in many places, and it's fine. The proper way to do this would be with https://en.wikipedia.org/wiki/Change_data_capture, if we really needed the correctness, i.e. first commit the changes to the DB and then propagate the changes to external systems.

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -65,14 +65,16 @@ export async function handlePutMultiDim(
     if (!isValidSlug(slug)) {
         throw new JsonError(`Invalid multi-dim slug ${slug}`)
     }
+    if (!(await multiDimDataPageExists(trx, { slug }))) {
+        await validateNewGrapherSlug(trx, slug)
+    }
```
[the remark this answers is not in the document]
rakyi: We need the if, because this handler does an upsert. It won't fail on the unique constraint, it will update the existing row(s) if the mdim already exists.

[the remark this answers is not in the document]
marcelgerber: I see, but then we need to actually fail if an mdim exists. Which we currently don't do!

rakyi:
### Merge activity

* **Feb 26, 9:31 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/owid/owid-grapher/4505).

PR #4516, opened 2025-02-03
Capture error instead of object in explorerViews

rakyi:
Otherwise Sentry won't group the event correctly.

Follow up to #4511.


ikesau (approved):
Oops! Good to know - I _thought_ the error message looked a little strange 😅

PR #4523, opened 2025-02-05
✨ prevent site-tools from obscuring content of page

ikesau:
Fixes the issue where the `site-tools` container was obscuring content even when its contents was hidden.

If we're not okay with 94% browser support, I'll rewrite this with JS but it's slightly annoying as I'll have to use a ref and layout effect to apply the transformation to `SiteTools` parent (the div container)

https://github.com/user-attachments/assets/a8fb3079-3419-43a9-b020-51c9c5722e83


marcelgerber (approved):
nice, this works well!

and 94% browser support is totally aceeptable for this, but thanks for caring about it :)

PR #4528, opened 2025-02-05
Fix KeyIndicator not showing grapher with a redirect

rakyi:
We need to pass the unresolved URL to Chart, where we resolve it with useLinkedChart again. If we pass the already resolved URL for a chart that was redirected, it won't be found in the attachments.

See [Slack discussion](https://owid.slack.com/archives/C46U9LXRR/p1738749662622129).


ikesau (approved):
nice find 🙇

PR #4534, opened 2025-02-06
Migrate to external ETL regions dataset

lucasrodes:
We want to stop generating `grapher/regions/latest/regions` dataset, which is meant for our Grapher database and generates this dataset in Admin which no one uses: [Regions data (OWID, 2023)](https://admin.owid.io/admin/datasets/6334).

Instead, we should rely on step `data://external/owid_grapher/latest/regions`, which should just be equivalent for use-cases like this one in `owid-grapher`.

Would the change in this PR suffice? Are there other instances where we rely on this dataset?

### DIFF
--- a/devTools/regionsUpdater/update.ts
+++ b/devTools/regionsUpdater/update.ts
@@ -13,7 +13,7 @@ import _ from "lodash"
 
 const ETL_REGIONS_URL =
         process.env.ETL_REGIONS_URL ||
-        "https://catalog.ourworldindata.org/grapher/regions/latest/regions/regions.csv",
+        "https://catalog.ourworldindata.org/external/owid_grapher/latest/regions/regions.csv",
     GEO_JSON_URL =
         "https://raw.githubusercontent.com/alexabruck/worldmap-sensitive/master/dist/world.geo.json",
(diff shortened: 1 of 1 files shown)


lucasrodes:
cc. @pabloarosado since you pointed me out to this issue.

pabloarosado:
For more context:
* ETL currently supports two solutions:
  * One is to create an `export` step. This automatically **writes** to external repositories (optionally creating a PR).
  * The other option is an `external` step, which is a step we know **is read** by an external repository.

Either of those solutions would be better than directly calling a `data` step from an external repos.

But from memory, I think grapher (or cloudflare-workers?) may be requiring two different regions files, so the changes proposed in this PR may not be enough. We'd need to revisit [this a conversation](https://owid.slack.com/archives/CQQUA2C2U/p1730972885915539).

marcelgerber (approved):
Hi! Yes this makes sense.
I've just added some lines that exclude income groups from the regions file, otherwise these would start showing up as "Your current location" in the entity selector from now.
Which is something we _eventually_ want to do, but with a proper design for that case - see https://github.com/owid/owid-grapher/issues/3517.

But now this is good to go, there are no differences when run, and we indeed can get rid of the old etl step for this.

Issue #4507, opened 2025-01-30
Add data insights to search


ikesau:
Despite data insights being ephemeral by nature, we think it would be valuable to our readers to be able to search for them.

We want to add them to the `research-and-writing` index so that they show up alongside articles and topic pages. 

We may find out that we'll need to do some additional UI design work to distinguish them from our other content that has the "evergreen commitment" but can cross that bridge when we come to it.

[Figma](https://www.figma.com/design/HcwsPwvmm3azqy2a7RL5QK/Topic-Page?node-id=116-882&p=f&t=KaCdp1oN1th5DJ2T-0)

## Scope
- Add a new data insights function to `getPagesRecords`
  - Will have to experiment a little to see what kind of score is appropriate for a data insight
- Add support in the search page to render data insights (potentially showing their date, cropping the thumbnail if necessary, etc)

PR #4541, opened 2025-02-07
🎉 add data insights to algolia pages index

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/4507

[Figma](https://www.figma.com/design/gJzXIPz5KqzBq86CBcCLUT/Data-Insight-in-Search?node-id=0-1&p=f&t=ah6PzefsmArmM7Y8-0)

**Adds Data Insights as a supported gdoc type in the `indexPagesToAlgolia` script.**

Most of this PR's diff comes from updating the union type `PageType` to use the `OwidGdocType` enum so that I could call `getCanonicalURL` in the search panel (to correctly prefix the slug with `/data-insights/`)

I also added a slightly complicated `getFirstBlockOfType` implementation because my first attempt at this change needed to get both the first image and first paragraph. It's less essential for the PR on the whole, now, but I think will still be a useful util to keep around.

image

### DIFF
--- a/baker/algolia/utils/pages.ts
+++ b/baker/algolia/utils/pages.ts
@@ -17,8 +17,6 @@ import {
     DEFAULT_THUMBNAIL_FILENAME,
     DbEnrichedImage,
     OwidGdocDataInsightInterface,
-    getFirstTwoSentencesFromString,
-    spansToUnformattedPlainText,
 } from "@ourworldindata/utils"
 import { formatPost } from "../../formatWordpressPost.js"
 import ReactDOMServer from "react-dom/server.js"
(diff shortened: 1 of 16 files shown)


baker/algolia/utils/pages.ts
```diff
@@ -105,15 +106,20 @@ async function generateWordpressRecords(
     const getPostTypeAndImportance = (
```
ikesau, line 106: I wanted to rewrite the `PageType` type to use the `OwidGdocType` enum so that it would be compatible with `getCanonicalUrl` which we now need to be able to link to data insights correctly on the search page, so I had to refactor these 3 functions that each had a pattern match statement.

ikesau:
@Marigold okay so yeah, it is just exceeding the timeout duration - indexing everything takes around 20 minutes now

If/when we drop the ExplorerViews index (which is redundant now that we have ExplorerViewsAndCharts) we'll drop 6 minutes or so, but for now maybe the easiest thing is to increase the limit.

Marigold:
Fair enough, I've increased the timeout to 60 mins ([here it is](https://buildkite.com/our-world-in-data/grapher-automated-staging-environment/settings/steps) if you want to further increase it). This will apply only to `-algolia` branches, so we don't have to worry about higher baking time on the regular staging servers, right?

ikesau:
Hey @mlbrgl - I'm requesting a review from you on this, but it's also an invitation to get it over the finish line this cooldown if you'd like 🙂 

I've tagged Marwa in the P&D channel to make sure the design's okay

mlbrgl (approved):
nice, good to go! 

@ikesau I've added a couple of commits, let me know if they don't make sense to you

PR #4547, opened 2025-02-10
🎉 create data insights from the admin

sophiamersmann:
Adds a workflow to create data insights from the admin.

### Summary

- The workflow can be triggered from two places: the DI index page (top right) and the narrative chart editor (footer)
- Creating a data insight involves one or two steps:
    - (if applicable) The DI image is fetched and uploaded to Cloudflare (supported are Grapher and Explorer charts, Narrative charts, and Figma charts)
    - The DI document is created and saved into the user's DI folder
- The workflow is fairly complete but does not cover all use cases. These are the use cases I had in mind:
   - Creating a data insight for a Figma-edited chart (supported)  
   - Creating a data insight for a narrative chart (supported)  
   - Creating a data insight for a specific view of a Grapher or Explorer chart (supported, although I'd prefer authors to use narrative charts if possible)  
   - Creating a data insight for a custom visualisation edited outside of Figma (partially supported: the DI doc can be created, but the image must be uploaded manually)

#### Creating a data insight doc

- A new data insight doc is created by copying a [template document](https://docs.google.com/document/d/1MRuGAmXGCkWv7WT2iDR7Zt2p3RPgH7jleP5x3lmU1DA/edit?usp=drive_link) that has placeholders like `{{title}}`
- In a second step, all placeholders are either replaced with content or removed
- Finally, the Gdoc is imported into our database

#### Image upload

- Images can be sourced from Figma, narrative charts, or from Grapher and Explorer URLs (if more than one filed is provided, then Figma > narrative charts > Grapher/Explorer URLs)
- Figma images need to be loaded first, the other cases are trivial

### Common Errors

- The image filename already exists: The action fails and the author is prompted to choose a different filename
- The alt text suggestion timeouts out (common for Grapher thumbnails): It's suggested to retry, but retrying may or may not work (I worry that this could be frustrating for authors)
- No DI target folder found (e.g., for new colleagues): Throws an error

If the image upload fails, no GDoc is created. But if the image upload succeeds but the GDoc creation fails, an unused image may be left in Cloudflare.

### Testing

- Local testing:
    - Add `GDOCS_DATA_INSIGHT_API_TEMPLATE_ID=1MRuGAmXGCkWv7WT2iDR7Zt2p3RPgH7jleP5x3lmU1DA` to your env file
    - DIs will be written to the [Admin](https://drive.google.com/drive/u/0/folders/1adt6cz5LcK0rIjVyiq4L_rNMqzE6qgpb?ths=true) folder  
- Testing on staging:
    - I created a [DI folder](https://drive.google.com/drive/u/0/folders/1N8WAbwIITxWlyXK7vGvjudTe0RS_ey4e?ths=true) for you, Ike, so you can test this on staging  
    - A couple of things don't work on staging:
        - Alt text suggestions don't work for our thumbnails (probably because it's served over http)
        - After a DI has been created, the docs preview is opened in a new browser tab – I think this only works if you explicitly allow it (at least in Chrome) 

### Open Questions

- Should I have added a data-insights/create route to the admin?
- Is the summary box clear? (I struggle with copywriting)
- Should I remove the alt text suggestion since it often times out for our worker-generated images? (It works ok for Figma images)
- Should the modal remain open after a successful DI creation?  

### Drive-by changes

- I added a 'Download image' button to the 'Reupload' modal as well

---
This is **part 4 of 9 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #4547 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #4490 
- <kbd>&nbsp;1&nbsp;</kbd> #4489


db/OwidGoogleAuth.ts
```diff
@@ -26,6 +26,7 @@ export class OwidGoogleAuth {
                     },
                     scopes: [
                         "https://www.googleapis.com/auth/documents",
+                        "https://www.googleapis.com/auth/drive",
```
sophiamersmann, line 29: This is needed for the copy method to work.

adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -533,6 +542,20 @@ export function DataInsightIndexPage() {
                             }
                         />
                     )}
+                    {isCreateModalOpen && (
+                        <CreateDataInsightModal
+                            closeModal={() => setIsCreateModalOpen(false)}
+                            onFinish={(response) => {
```
ikesau, line 548: Can we refresh the index page also? It'd be nice to see the newly created DI in the list.

Ideally via another `GET` to `admin/api/dataInsights`

adminSiteClient/imagesHelpers.ts
```diff
@@ -57,14 +65,69 @@ export async function reuploadImageFromSourceUrl({
...
     )
 
     return response
 }
 
+export async function fetchFigmaProvidedImageUrl(
+    figmaUrl: string
+): Promise<FigmaResponse> {
+    const figmaApi = new Figma.Api({
+        personalAccessToken: FIGMA_API_KEY,
```
ikesau, line 80: Just noticing that I _didn't_ have this set, because it's not in the `.env.example-full` - we should put it there and mark it as optional 🙂

adminSiteClient/CreateDataInsightModal.tsx
```diff
@@ -0,0 +1,935 @@
...
+    | "imageAltText"
+type ImageFormFieldName = "imageFilename" | "imageAltText"
+
+type FormData = Partial<Record<FormFieldName, string>>
+type FormDataWithTitle = RequiredBy<FormData, "title">
+type FormDataWithImageFilename = RequiredBy<FormData, "imageFilename">
+
+type DataInsightCreationResponse =
+    | { success: true; gdocId: string }
+    | { success: false; errorMessage: string }
```
ikesau, line 79: I think this okay because you're actually handling them here, but possibly for consistency's sake, we should just throw them as JsonErrors (See https://github.com/owid/owid-grapher/pull/4564)

But maybe `admin.requestJSON` will intercept that and put up the ugly error modal, in which case, the way you're doing it is definitely superior 🙂

adminSiteClient/CreateDataInsightModal.tsx
```diff
@@ -0,0 +1,935 @@
...
+    )
+
+    const narrativeChart = formData.narrativeChart
+        ? allNarrativeChartsMap.get(formData.narrativeChart)
+        : props.narrativeChart
+
+    const getDataInsightImageUrl = ({ cache = false } = {}):
+        | { source: ImageSource; url: string }
+        | undefined => {
+        if (figmaImageUrl) return { source: "figma", url: figmaImageUrl }
```
ikesau, line 151: [me reading these 4 lines](https://www.youtube.com/watch?v=rQgXkNYHhKU&t=15s) 😹 

I think deconstructing the options param makes this a little more verbose than it needs to be, especially because it defaults to false in `makePngUrlForNarrativeChart` anyway.

In fact, can cache _ever_ be set to true?

ikesau (commented):
Having to leave this review as a WIP

From my cursory testing it looks extremely good!

adminSiteClient/CreateDataInsightModal.tsx
sophiamersmann: In almost all cases we want an uncached version. I added the cache option later because the alt text suggestion sometimes times out for our Grapher thumbnails – using the cached version helps

db/migration/1739195264083-AddDataInsightFolderIds.ts
```diff
@@ -0,0 +1,103 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+const dataInsightFolders = [
+    {
+        fullName: "Veronika Samborska",
```
ikesau, line 5: I've mentioned this new dependency to Ed so that he can add it as a TODO to his onboarding checklist 🙂

adminSiteClient/CreateDataInsightModal.tsx
```diff
@@ -0,0 +1,935 @@
...
+                                form={{
+                                    data: formData,
+                                    setField: setFormField,
+                                }}
+                                initialValue={initialValues.narrativeChart}
+                                show={showField("narrativeChart")}
+                                style={{ flex: "1 1 0px", maxWidth: "48%" }}
+                            />
+
+                            <FormField
```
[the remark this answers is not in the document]
ikesau: In fact I think all of these fields (maybe except title) could do with a placeholder. 🙂

ikesau (approved):
Another really excellent PR! I tested each of the flows' happy paths and they worked flawlessly. Really good work!

As a result, half of my comments are just code style opinion things that stuck out to me. Please don't feel obligated to make any of those changes, they're just to share ideas.

The rest are small QoL things which hopefully make sense to you 🙂

adminSiteClient/CreateDataInsightModal.tsx
sophiamersmann: I don't expect it to be unreliable on the images index page because there, we serve from Cloudflare, which is presumably much faster than our workers (especially if we don't rely on the cache, as then the workers need to render Grapher first)

db/migration/1739195264083-AddDataInsightFolderIds.ts
sophiamersmann: Thanks!

adminSiteClient/CreateDataInsightModal.tsx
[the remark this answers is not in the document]
sophiamersmann: done 👍🏻

adminSiteClient/CreateDataInsightModal.tsx
```diff
@@ -0,0 +1,935 @@
...
+    formData: FormData
+): formData is FormDataWithImageFilename {
+    return !isEmpty(formData.imageFilename)
+}
+
+function isValid(formData: FormData): formData is FormDataWithTitle {
+    return !isEmpty(formData.title)
+}
+
+function useProgress(...tasks: Task[]) {
```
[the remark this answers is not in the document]
sophiamersmann: I didn't cross my mind. It probably would have been a bit nicer!

sophiamersmann:
All good points, thank you, Ike!

adminSiteClient/imagesHelpers.ts
```diff
@@ -1,10 +1,18 @@
+import * as Figma from "figma-api"
 import { RcFile } from "antd/es/upload/interface.js"
 import { Admin } from "./Admin"
 import { DbEnrichedImageWithUserId } from "@ourworldindata/types"
-import { CLOUDFLARE_IMAGES_URL } from "../settings/clientSettings"
+import {
+    CLOUDFLARE_IMAGES_URL,
+    FIGMA_API_KEY,
+} from "../settings/clientSettings"
```
ikesau, line 8: Aha so actually, all `clientSettings` are bundled in `owid.mjs` which means that yeah, we _should_ proxy this via the admin server - the potential risk for this key leaking seems low risk as it's read-only, but if you can refactor this in less than a day I think it'd be worth it 👍

sophiamersmann: oh, good to know! glad we caught this now :) 9 out of 10 ten times my anxieties are in the way, but sometimes...

Issue #3495, opened 2024-04-15
Add "New" labels in thumbnails for our Featured Work in the homepage


mrwbkrm:
## Core problem

[Feature request](https://owid.slack.com/archives/C06CWPLE3QE/p1710158403965549) initiated by @maxroser 
We currently have no way of distinguishing between old content on our site and new content within the Featured Work block on the homepage. We want to make it clear to our users when a piece of featured content is new. 

## Proposed solution

A visible text label that can be added by @matt-mun when curating the Featured block to indicate "new" content. This label can be added to primary, secondary and tertiary spots in the Featured work block.

## Design
<img width="1018" alt="Screenshot 2024-04-15 at 17 32 57" src="https://github.com/owid/owid-grapher/assets/113343623/78b11ea6-64b6-451c-939d-6c5eccf1ad0c">

[Figma](https://www.figma.com/file/m4MLbbAmIiQLwPSG1OtA0I/%5BProject%5D-Homepage?type=design&node-id=1211%3A8431&mode=design&t=HmpteyWWSqyLWjHW-1)

danyx23:
We could either add an `isNew: true` field to the archie or parse the string of the kicker.

Either way it shouldn't be a big thing

PR #4550, opened 2025-02-10
🎉  "new" label for articles on the homepage

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/3495

## Changes:
1. Adds an `isNew` property to featured article cards in the `homepage-intro` component
2. Adds a util to validate raw boolean properties, and uses it in all relevant locations in `rawToEnriched.ts`

### Preview
image

## To test
Copy-paste the `homepage-intro` component from the OWID Homepage document into your own testing gdoc, and update the `type` front-matter to `homepage`. Add `isNew: true` to as many cards as you'd like.

### DIFF
--- a/packages/@ourworldindata/utils/src/Util.ts
+++ b/packages/@ourworldindata/utils/src/Util.ts
@@ -1808,28 +1808,6 @@ export function lowercaseObjectKeys(
     )
 }
 
-export function validateRawBoolean<
-    T extends string,
-    U extends { [K in T]?: any },
->(
-    key: T,
-    rawBlock: U
-): { isValid: true } | { isValid: false; message: string } {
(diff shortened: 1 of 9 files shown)


packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -1807,6 +1807,28 @@ export function lowercaseObjectKeys(
     )
 }
 
+export function validateRawBoolean<
```
rakyi, line 1810: This is quite specific to what we do in `rawToEnriched`, so it seems more natural to put the function into that module. Feel free to keep it here, if you disagree.

rakyi (approved):

Issue #4316, opened 2024-12-16
Data catalog double counts Google Analytics search events


bnjmacdonald:
## Description

Data catalog sends 2+ `owid.data_catalog_search` events – which appear to be duplicates – following a user search or topic filter click. Should only send a single `owid.data_catalog_search` event.

## Steps to reproduce

https://github.com/user-attachments/assets/cd706561-8346-4d86-8642-47cf658b1ce1

## Environment

Happening in at least Chrome + Safari on Mac.

rakyi:
It's possible that this is related to a state sync between a URL and a component state. Ideally, we'd send the analytics events on an even listener.

danyx23:
Should probably be time-boxed. We estimated it at three hours, if it starts taking longer then maybe worth revisiting if we want it badly enough.

PR #4553, opened 2025-02-11
🐛 fix double-logging data catalog searches

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/4316

## Context
The data catalog was logging searches to google analytics twice. You can see this on prod by going to https://ourworldindata.org/data, entering a query, and then opening the dev console and typing `dataLayer.at(-1)` and `dataLayer.at(-2)`

## Problem
The data catalog ran all of its business logic through a single `useEffect`, which meant it called anytime any of its dependencies changed. Analytics was one of the last things I added and I didn't stop to think that this callback would fire anytime any of the dependencies updated, including updates to the cache or even no-op calls to update the state.

## The fix
Extracts the analytics logic into an independent effect, which only runs when `stateAsUrl` changes (which always corresponds to a search running, even when the results are restored from the cache)

## How to test
Add a `console.log` to [`logDataCatalogSearch`](https://github.com/owid/owid-grapher/blob/65c42cbfc1f11893afbbae5f2416ccdd5be18c36/site/SiteAnalytics.ts#L96) and run some searches locally, noting that a log call only happens when a new query is run.
```ts
logDataCatalogSearch(state: DataCatalogState) {
        console.log("logging search", state);
        this.logToGA({
            event: EventCategory.DataCatalogSearch,
            eventAction: "search",
            eventContext: JSON.stringify({
                ...state,
                topics: Array.from(state.topics),
                selectedCountryNames: Array.from(state.selectedCountryNames),
            }),
        })
    }
```

### DIFF
--- a/site/DataCatalog/DataCatalog.tsx
+++ b/site/DataCatalog/DataCatalog.tsx
@@ -855,18 +855,25 @@ export const DataCatalog = ({
     const currentResults = cache[cacheKey].get(stateAsUrl)
 
     useEffect(() => {
+        // Reconstructing state from the `stateAsUrl` serialization to avoid a `state` dependency in this effect,
+        // which would cause it to run on every state change (even no-ops)
+        const url = Url.fromURL(stateAsUrl)
+        const state = urlToDataCatalogState(url)
+        analytics.logDataCatalogSearch(state)
+    }, [stateAsUrl])
+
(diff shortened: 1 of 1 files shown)


rakyi (approved):
LGTM! Thanks for the very clear description!

PR #4560, opened 2025-02-14
Fix scroll on pages with ChartList

rakyi:
Also fixes query params handling. See [Slack discussion](https://owid.slack.com/archives/C46U9LXRR/p1739468896959499).


sophiamersmann (approved):
Nice!

PR #4564, opened 2025-02-14
🐛 fix admin API error messages not showing in client

ikesau:
## Problem

`return { success: false, error: "Oh no!" }` was being used in a few places in our API, but the admin client didn't understand what to do with that response. 

image

## Fix
I grep'd `return {` in the API folder, and updated any instance of the non-working pattern with `throw new JsonError(` instead.

## To test
Check this branch out, go to `http://localhost:3030/admin/charts/create` and click "Create Draft"
The error that throws (due to missing data) will render correctly in the client.

### DIFF
--- a/adminSiteServer/app.test.ts
+++ b/adminSiteServer/app.test.ts
@@ -193,11 +193,10 @@ async function makeRequestAgainstAdminApi(
         body,
     })
 
-    expect(response.status).toBe(200)
-
     const json = await response.json()
 
     if (verifySuccess) {
+        expect(response.status).toBe(200)
         expect(json.success).toBe(true)
     }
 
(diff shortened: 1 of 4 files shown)


adminSiteServer/app.test.ts
```diff
@@ -193,11 +193,10 @@ async function makeRequestAgainstAdminApi(
         body,
     })
 
-    expect(response.status).toBe(200)
-
     const json = await response.json()
 
     if (verifySuccess) {
+        expect(response.status).toBe(200)
```
ikesau, line 199: this seems like a reasonable change to make, but it kind of begs the question: why have `success = true` and `status = 200`

🤔

PR #4568, opened 2025-02-17
Add site redirects for recent chart redirects

rakyi:
We take Grapher redirects into consideration only when a route returns 404 and Cloudflare caches assets up to a week in the CDN, so it's possible to still get the old page after a redirect is created for it.

Authors are confused why they don't see chart redirects working after they push changes. It also means different users can see different charts when clicking on the same link, depending on whether the CDN data center that handles their request had the old page cached. I tested this case with Pablo A.

We take a similar approach as for deleting articles and add redirects in the special `_redirects` file - that Cloudflare always takes into consideration - for Grapher redirects that are less than a week old.

### DIFF
--- a/baker/redirects.ts
+++ b/baker/redirects.ts
@@ -77,14 +77,46 @@ export const getRedirects = async (knex: db.KnexReadonlyTransaction) => {
         (row) => `${row.source} ${row.target} ${row.code}`
     )
 
+    const recentGrapherRedirects = (await getRecentGrapherRedirects(knex)).map(
+        (row) => `/grapher/${row.source} /grapher/${row.target} 302`
+    )
+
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
Nice! Good idea, and I think this'll work well.

The only other consideration (which is harder, though) is about deleted Grapher pages. For gdocs, we're putting tombstones there, and are adding temporary redirects. Should we be doing a similar thing for grapher pages, too?

rakyi:
Good point, although I'd like to know if it's a problem in practice. Do you have a sense of how often we delete a chart without creating a redirect for it?

If we are not sure, I'd wait until it becomes a problem in practice before doing anything about it.

PR #4569, opened 2025-02-17
🎉 Link from admin indicator search to ETL semantic search

larsyencken:
We have a great semantic search over indicators in the ETL, but people may not be aware of it or have used it.

This change adds a link from the existing indicator search, making it easy to discover.


larsyencken:
Since it's low risk, I'll just merge it.

PR #4584, opened 2025-02-20
✨ Privacy Policy front-matter

ikesau:
Adds 2 new features to our about pages

1. `hide-nav` in the front-matter, to hide the about page nav
2. `override-title` in the front matter, to support existing About pages, and also the upcoming Privacy Policy page (which needs an actual title)

It's hard to imagine a scenario where we'd want one without the other, but given that they were both trivial to add and we won't be making many of them, I figured it was better to be explicit.

## To test
Create a gdoc with the following front-matter:
```
title: My custom about page title
type: about-page
hide-nav: true
override-title: true
```

### DIFF
--- a/site/gdocs/pages/AboutPage.tsx
+++ b/site/gdocs/pages/AboutPage.tsx
@@ -8,10 +8,11 @@ import { ArticleBlocks } from "../components/ArticleBlocks.js"
 import Footnotes from "../components/Footnotes.js"
 
 export default function AboutPage({ content, slug }: OwidGdocAboutInterface) {
+    const shouldOverrideTitle = content["override-title"]
     return (
         <main className="about-page centered-article-container grid grid-cols-12-full-width">
             <h1 className="about-header col-start-2 col-end-limit display-2-semibold">
-                {content.title || "About"}
+                {(shouldOverrideTitle && content.title) || "About"}
             </h1>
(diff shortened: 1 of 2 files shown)


site/gdocs/pages/AboutPage.tsx
```diff
@@ -8,12 +8,13 @@ import { ArticleBlocks } from "../components/ArticleBlocks.js"
...
 export default function AboutPage({ content, slug }: OwidGdocAboutInterface) {
+    const shouldOverrideTitle = content["override-title"]
     return (
         <main className="about-page centered-article-container grid grid-cols-12-full-width">
             <h1 className="about-header col-start-2 col-end-limit display-2-semibold">
-                About
+                {(shouldOverrideTitle && content.title) || "About"}
             </h1>
-            <AboutNav slug={slug} />
+            <AboutNav slug={slug} hideNav={content["hide-nav"]} />
```
rakyi, line 17: I think it would be nicer, if we simply not render the `AboutNav` here, instead of passing in the arg. After all, that's the main way to decide whether something should be in the JSX or not.

```suggestion
            {!content["hide-nav"] && <AboutNav slug={slug} />
```

rakyi (approved):
Nice, just one small comment. 🙂

PR #4585, opened 2025-02-21
Preserve selected tab when switching mdim dimensions

### LINKED ISSUE #4519: Selected tab is reset to default when multi-dim dropdown changes
## Description

When you select a different tab than the default (e.g. chart or map) in the multidimensional data page and then then change any of the dimensions in the dropdown, the tab is reset to the default.

## Expected behaviour

The tab selection should be preserved.

## Additional context

This might be caused by incorrect query params manipulation.

rakyi:
The state in `MultiDimDataPageContent` is getting ridiculous. I hope the Grapher state refactor will help us there.

Fixes #4519

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -3866,3 +3866,22 @@ export const getErrorMessageRelatedQuestionUrl = (
               undefined
         : undefined
 }
+
+export function getNextTabParamWhenSwitchingGrapher(
+    grapher: Grapher,
+    previousTab: GrapherTabName
+): string {
+    if (grapher.availableTabs.includes(previousTab)) {
+        // preserve the previous tab if that's still available in the new view
+        return grapher.mapGrapherTabToQueryParam(previousTab)
+    } else if (grapher.validChartTypes.length > 0) {
+        // otherwise, switch to the first chart tab
+        return grapher.mapGrapherTabToQueryParam(grapher.validChartTypes[0])
+    } else if (grapher.hasMapTab) {
+        // or switch to the map, if there is one
+        return GRAPHER_TAB_QUERY_PARAMS.map
+    } else {
+        // if everything fails, switch to the table tab that is always available
+        return GRAPHER_TAB_QUERY_PARAMS.table
+    }
+}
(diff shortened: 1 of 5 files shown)


site/multiDim/MultiDimDataPageContent.tsx
```diff
@@ -181,7 +184,9 @@ const useVarDatapageData = (
                         `Fetching variable by uuid failed: ${grapherConfigUuid}`,
                         { cause: datapageData.reason }
                     )
-
+                // Avoid a race condition where the earlier fetch completes
+                // after the user has already switched to a different view.
+                if (ignore) return
```
rakyi, line 189: This was an unrelated bug I noticed when working on this.

site/multiDim/MultiDimDataPageSettingsPanel.tsx
```diff
@@ -123,9 +123,9 @@ const DimensionDropdown = (props: {
 export const MultiDimSettingsPanel = (props: {
     config: MultiDimDataPageConfig
     currentSettings: MultiDimDimensionChoices
-    updateSettings?: (currentSettings: MultiDimDimensionChoices) => void
+    onUpdateSettings: (currentSettings: MultiDimDimensionChoices) => void
```
rakyi, line 126: Renamed this to follow the React convention for naming event handler params. And we always pass the arg at the single call site, so it doesn't have to be optional.

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -3866,3 +3866,22 @@ export const getErrorMessageRelatedQuestionUrl = (
               undefined
         : undefined
 }
+
+export function getNextTabParamWhenSwitchingGrapher(
```
rakyi, line 3870: I struggled to find a place for this function. Let me know if you think this isn't good enough.

sophiamersmann:
Have you considered updating Grapher's `populateFromQueryParams` such that it chooses sensible fallbacks when presented with a tab that doesn't exist? If not, can you try that? I think, it'd be nice if the tab logic would be in a single place. It would also make Grapher a bit more robust generally. In that case, you could get rid of `getNextTabParamWhenSwitchingGrapher` and just need to make sure that the tab param gets passed in for explorers and mdims, even if it's the default.

rakyi:
Great idea! Turns out we can simplify that logic by reusing what we already have with a slight modification. Closing in favor of #4593.

PR #4593, opened 2025-02-26
Rework state management in MultiDimDataPageContent

### LINKED ISSUE #4519: Selected tab is reset to default when multi-dim dropdown changes

rakyi:
* Preserve selected tab when switching between dimensions (similar to explorer)
* Fix race condition when loading data after switching dimensions
* Reduce state duplication and usage of useEffect
  * Store mdim settings state in query params

See https://github.com/owid/owid-grapher/pull/4585 for related discussion.

Fixes #4519


marcelgerber (approved):
Nice! I tested this a bunch, and didn't notice any regressions.

PR #4614, opened 2025-02-28
Switch to ErrorBoundary from Sentry

rakyi:
This way we get also the React component stack attached to the error in Sentry.

### DIFF
--- a/site/gdocs/components/ArticleBlock.tsx
+++ b/site/gdocs/components/ArticleBlock.tsx
@@ -47,7 +47,7 @@ import Person from "./Person.js"
 import NarrativeChart from "./NarrativeChart.js"
 import { Container, getLayout } from "./layout.js"
 
-export default function ArticleBlock({
+function ArticleBlockInternal({
     b: block,
     containerType = "default",
     toc,
@@ -70,7 +70,7 @@ export default function ArticleBlock({
             />
         )
     }
-    const content: any | null = match(block)
+    return match(block)
         .with({ type: "aside" }, ({ caption, position = "right" }) => (
             <figure
                 className={cx(
@@ -225,9 +225,7 @@ export default function ArticleBlock({
                 />
             )
         })
-        .with({ type: "simple-text" }, (block) => {
-            return block.value
-        })
+        .with({ type: "simple-text" }, (block) => <>{block.value.text}</>)
         .with({ type: "heading", level: 1 }, (block) => (
             <h1
                 className={cx(
@@ -684,10 +682,27 @@ export default function ArticleBlock({
             />
         ))
         .exhaustive()
+}
 
+export default function ArticleBlock({
+    b: block,
+    containerType = "default",
+    toc,
+    shouldRenderLinks = true,
+}: {
+    b: OwidEnrichedGdocBlock
+    containerType?: Container
+    toc?: TocHeadingWithTitleSupertitle[]
+    shouldRenderLinks?: boolean
+}) {
     return (
         <BlockErrorBoundary className={getLayout("default", containerType)}>
-            {content}
+            <ArticleBlockInternal
+                b={block}
+                containerType={containerType}
+                toc={toc}
+                shouldRenderLinks={shouldRenderLinks}
+            />
         </BlockErrorBoundary>
     )
 }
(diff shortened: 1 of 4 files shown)


site/gdocs/components/ArticleBlock.tsx
```diff
@@ -225,9 +225,9 @@ export default function ArticleBlock({
                 />
             )
         })
-        .with({ type: "simple-text" }, (block) => {
-            return block.value
-        })
+        .with({ type: "simple-text" }, (block) => (
+            <span>{block.value.text}</span>
```
mlbrgl, line 229: I think the point of the simple-text block is to [opt out of formatting altogether](https://github.com/owid/owid-grapher/blob/ce36882c8ac95d76fbad5d518cad8075ec855bea/db/model/Gdoc/rawToEnriched.ts#L1214-L1218), and leave that responsibility to the caller. "Regular" simple texts are already [wrapped in span in span-simple-text](https://github.com/owid/owid-grapher/blob/ce36882c8ac95d76fbad5d518cad8075ec855bea/site/gdocs/components/SpanElement.tsx#L15-L17).

site/gdocs/components/ArticleBlocks.tsx
```diff
@@ -29,13 +30,17 @@ export const ArticleBlocks = ({
     <>
         {blocks.map((block: OwidEnrichedGdocBlock, i: number) => {
             return (
-                <ArticleBlock
+                <BlockErrorBoundary
```
mlbrgl, line 33: moving the error boundary outside of articleblock means that we don't get the benefit of the error boundary when articleblock is used on its own (and not through articleblocks). 
For instance, an error thrown by [a social block in author pages](https://github.com/owid/owid-grapher/blob/ce36882c8ac95d76fbad5d518cad8075ec855bea/site/gdocs/pages/Author.tsx#L52) wouldn't be caught.

mlbrgl (approved):
it's really nice that you're adding more and more reporting to Sentry!

A couple of things might need attention, up to you.

rakyi (commented):
I only now realized I didn't submit the self-review so my comments were not showing. Ah well, doing that now.

site/gdocs/components/ArticleBlocks.tsx
rakyi: I explained [here](https://github.com/owid/owid-grapher/pull/4614#discussion_r1975793288) why I moved it outside. The error should still be captured in Sentry, but it's true that a bigger part of the page would break, potentially the whole page.

To make it work properly for all `ArticleBlock`s, we'd need to introduce another layer of indirection. I guess it might be worth it, so I've added that.

site/gdocs/components/ArticleBlock.tsx
rakyi: Good point, changed.

site/gdocs/components/BlockErrorBoundary.tsx
```diff
@@ -45,12 +55,12 @@ export const BlockErrorFallback = ({
             {debug ? (
                 <>
                     <div>{error.message}</div>
-                    {resetErrorBoundary ? (
+                    {resetError ? (
                         <div>
                             <button
                                 aria-label="Reload content"
                                 style={{ margin: "10px" }}
```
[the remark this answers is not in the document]
rakyi: Switched to the `Button` component instead.

site/gdocs/components/ArticleBlock.tsx
```diff
@@ -684,10 +684,4 @@ export default function ArticleBlock({
             />
         ))
         .exhaustive()
-
-    return (
-        <BlockErrorBoundary className={getLayout("default", containerType)}>
```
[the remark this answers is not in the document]
mlbrgl: hey @rakyi, I'm a hard time coming up with an example error in this case. I'm guessing you're not talking about [errors in children components](https://github.com/owid/owid-grapher/pull/4647), nor errors that would throw during SSR. Invalid props in `window._OWID_GDOC_PROPS` somehow?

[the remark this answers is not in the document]
rakyi: Before my change, if there was an error in a child component, it would be thrown when the `with` block executed, i.e. during the assignment to the `content` variable. So the `<BlockErrorBoundary>` line would never be reached, and thus the error wouldn't be caught in the error boundary.

Does that make sense? If not, I'm happy to have a quick call to chat about it. 🙂

PR #4615, opened 2025-02-28
Fix style leak in KeyInsights

### LINKED ISSUE #3872: Grapher download menu: Style leak in Key insights component
The wrong font is used here:

https://ourworldindata.org/life-expectancy#key-insights

image

rakyi:
We were overriding h4 style in Grapher download modal.

Fixes #3872

### DIFF
--- a/site/gdocs/components/KeyInsights.scss
+++ b/site/gdocs/components/KeyInsights.scss
@@ -143,14 +143,16 @@ $slide-content-height: $grapher-height;
         &[data-active="true"] {
             display: block;
         }
+    }
+}
 
-        h4 {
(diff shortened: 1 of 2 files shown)


sophiamersmann (approved):
Thanks!

While you're at it could you also make the description text a little smaller? It's 16px when embedded in the KeyInsights block, but should be 14px.

| KeyInsights | Data page |
|--------|--------|
| Screenshot 2025-03-04 at 10 07 06 | Screenshot 2025-03-04 at 10 06 53 |

rakyi:
Good catch, added that too!

PR #4632, opened 2025-03-05
🐛 (admin) prevent infinite loop when loading variables / TAS-859

sophiamersmann:
Invalid dimensions lead to an infinite loop where variables are loaded again and again in the admin

~This appears to be fixed in the Grapher state refactor, so I'm closing this one down.~ Might still makes sense to fix this now since the refactor only ships in 6 to 8 weeks?

Example: [live](https://admin.owid.io/admin/charts/1250/edit) / [staging](http://staging-site-fix-admin-infinite-loading/admin/charts/1250/edit)


marcelgerber (approved):
nice, thank you!

PR #4637, opened 2025-03-06
:bug: Memoize email to user lookup from Slack

Marigold:
Memoize `getSlackMentionByEmail` to avoid rate limiting. If this keeps happening, we can also fetch email -> slack mapping from `users` table.


rakyi (approved):
LGTM, except the duplicate import.

PR #4643, opened 2025-03-07
🐝 (svg tester) prettify SVG files

sophiamersmann:
When working with the SVG tester, it is sometimes not clear what changed from looking at the pictures. It can then be helpful to look at the text diff, but this is not straightforward since the SVGs are on a single line.

This PR prettifies SVGs before comparison. The SVG tester doesn't fail on this branch because it prettifies both, the original and the new one, and then compares them.

We should update the SVG tester anyway because of the `includedEntities` -> `includedEntityNames` change, so I'll do that after merging this PR.

---
⧓ Review in [Butler Review `#7TLklK152`](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7TLklK152)

**🐝 prettify SVG files**

1 commit series (version 1)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 1/1 | [🐝 prettify SVG files](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7TLklK152/commit/711c085b-e239-4a8d-bd92-7ca5f384f5bf) | 💬 |  |

_Please leave review feedback in the [Butler Review](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7TLklK152)_

### DIFF
--- a/devTools/svgTester/utils.ts
+++ b/devTools/svgTester/utils.ts
@@ -26,6 +26,7 @@ import util from "util"
 import { getHeapStatistics } from "v8"
 import { queryStringsByChartType } from "./chart-configurations.js"
 import * as d3 from "d3"
+import prettier from "prettier"
 
 // the owid-grapher-svgs repo is usually cloned as a sibling to the owid-grapher repo
 export const DEFAULT_CONFIGS_DIR = "../owid-grapher-svgs/configs"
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
Real nice, thank you! I thought to do this a few times already, but then never ended up doing it.

PR #4685, opened 2025-03-23
Add pull request template

rakyi:
Based on [the previous discussion](https://www.notion.so/owid/2025-02-12-PR-review-checklist-19874b71f92e80c8891ef6ff0e88c1ad).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,6 +18,8 @@ Step-by-step instructions on how to test this change
 
 (delete all that do not apply)
 
+### Before merging
+
 - [ ] Google Analytics events were adapted to fit the changes in this PR
 - [ ] Changes to CSS/HTML were checked on Desktop and Mobile Safari at all three breakpoints
 - [ ] Changes to HTML were checked for accessibility concerns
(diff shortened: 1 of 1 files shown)


danyx23 (approved):
This looks really good, thanks for pushing this again!

The only thing that comes to mind that we might still want to add is a section for production rollout checkboxes. We sometimes have tasks like "run the sync db to r2 script" after a PR has been merged - it would be nice to have a standardized heading for these things as a reminder. Either way it is good to go IMO :)

rakyi:
Thanks for checking!

I'm not sure what you mean. I added this item under the "Checklist" heading:

https://github.com/owid/owid-grapher/blob/bc5d91de7f8f009871f9a3c47dbf77fb8da87c7c/.github/pull_request_template.md?plain=1#L31

Do you want it to have a separate heading? Or do you mean some other sync script?

danyx23:
Yeah good point - maybe it's worth putting that under its own heading? It just seems a bit different because the rest of the checklist is stuff that you'd want to check before merging the PR but this one you can only do after.

PR #4686, opened 2025-03-23
Show every.org section first for users from USA

### LINKED ISSUE #4666: Prioritize every.org donation link for US users
On donate page, show every.org section before the Stripe form if the user is coming from the US. We can use the existing way of detecting a user's country.

Nice to have: switch the buttons primary vs secondary to prioritize every.org in this case.

rakyi:
Tested with a VPN that the every.org section is first when connecting from the USA.

image

Resolves #4666


danyx23:
Hey @rakyi - I only had a look at the screenshot and this looks good! One thing that came to my mind: I think for the US audience we need a sentence or so to explain the stripe option below - otherwise it's a bit unclear why there is so much stuff below the call to action. Something like "You can also donate via Stripe by using the form below" (ask in the project channel for a better copy :) ).

rakyi:
Good point! Added the explanation after confirmin with Charlie.

marcelgerber (approved):
Nice! This works well.

Another enhancement would be to correctly pre-select the currency for Stripe based on the country code (e.g. USD in the US, Euro in Eurozone countries).
but that's just an idea - for now this awesome already :)

rakyi:
Good idea, writing it down for the future!

PR #4688, opened 2025-03-24
:hammer: Move explorers to adminSiteClient

Marigold:
## Context

This is the first part of our efforts to migrate explorers from `owid-content` repository to MySQL. This PR is harmless and only moves a couple of files as a preparation for full migration.

### Changes
- Move `ExplorerCreatePage` and `ExplorersListPage` to `adminSiteClient` to get access to `context` and knex transaction
- Remove `TemplatesComponent` (I couldn't figure out how was it used)
- Remove outdated `devTools/explorerTools`
- Remove outdated tests

## Testing guidance

- [x] Check if basic explorers functions work on the staging server


adminSiteClient/ExplorerCreatePage.tsx
```diff
@@ -20,17 +20,17 @@ import {
...
     ExplorerProgram,
-    EXPLORER_FILE_SUFFIX,
     makeFullPath,
     isEmpty,
 } from "@ourworldindata/explorer"
 import { GitCmsClient } from "../gitCms/GitCmsClient.js"
-import { GitCmsFile, GIT_CMS_BASE_ROUTE } from "../gitCms/GitCmsConstants.js"
-import { AdminManager } from "./AdminManager.js"
+import { GIT_CMS_BASE_ROUTE } from "../gitCms/GitCmsConstants.js"
+import { AdminManager } from "../explorerAdminClient/AdminManager.js"
```
rakyi, line 28: We can probably move everything else in `explorerAdminClient` too and just get rid of the directory, right?

rakyi (commented):
## Templates

> Remove `TemplatesComponent` (I couldn't figure out how was it used)

The template buttons are visible only after I uncheck the "Show Preview" checkbox first (even if I check it back again). That seems to be a bug, but otherwise the templates work. They are defined in `owid-content`, e.g. `explorers/grapher-template.explorer.tsv`.

We can either:
* Move the templates into `owid-grapher` and fix the bug(?) where the buttons don't show initially
* Get rid of them completely, but confirm first that nobody would miss them

image

## Scripts

@marcel you seem to be the main user of `extractGrapherIds` and `rewriteAllExplorers` scripts. Can you confirm that we can get rid of them without having a replacement as we move `owid-content` into the DB?

adminSiteClient/ExplorerCreatePage.tsx
Marigold: Done! I had it in "future work", but it wasn't as tricky as I expected.

Marigold:
I asked and no one is using templates (moreover, they haven't been updated in 4 years).

adminSiteClient/ExplorersListPage.tsx
```diff
@@ -32,9 +32,13 @@ import {
     GIT_CMS_REPO_URL,
 } from "../gitCms/GitCmsConstants.js"
 import { BAKED_BASE_URL } from "../settings/clientSettings.js"
-import { AdminManager } from "../explorerAdminClient/AdminManager.js"
+import { AdminManager } from "./AdminManager.js"
 import { AdminAppContext, AdminAppContextType } from "./AdminAppContext.js"
 
+// export interface AdminManager {
+//     loadingIndicatorSetting?: "loading" | "off" | "default"
+// }
```
rakyi, line 40: Delete the comment?

rakyi (approved):
Nice! I don't know if Marcel cares about the scripts. Probably fine to merge and restore them if necessary later.

PR #4691, opened 2025-03-24
🐛 fix indexing scheduled gdoc posts

ikesau:
Fixes `indexIndividualGdocPost` indexing posts that were scheduled for the future, which was causing not-live posts to show up in search


marcelgerber (approved):
nice!

PR #4694, opened 2025-03-25
Switch Stripe to dynamic payment methods

rakyi:
When we don't specify `payment_method_types` Stripe will pick and order them dynamically for us based on the allowed payment methods configured in the dashboard, customer location, currency etc.

https://docs.stripe.com/payments/payment-methods/dynamic-payment-methods

You can test this with a local dev Stripe setup and a VPN.

### DIFF
--- a/functions/donation/_utils/checkout.ts
+++ b/functions/donation/_utils/checkout.ts
@@ -14,23 +14,6 @@ const messageByInterval = {
     annual: "You will be charged annually and can cancel any time by writing to us at donate@ourworldindata.org.",
 } as const
 
-function getPaymentMethodTypes(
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
nice!

Before merging this, we should check that we have all relevant payment methods enabled in the Stripe dashboard.

In theory, I don't see a good reason why we wouldn't just enable almost every available payment method, unless [its fees](https://stripe.com/gb/pricing/local-payment-methods) are prohitively high. But probably good to double-check this with Este, also.

rakyi:
Thanks, for the review. Yep, I'm coordinating with Este on this. Didn't know about the nice pricing page!

Issue #4670, opened 2025-03-18
Cycle 2025.2: MIM Admin


ikesau:
Adding a view in the admin to specify Most Important Metrics (MIMs)

## Motivation

MIMs are the way we will emphasize particular views of our data that we think are the most important to understand each of our topics.

## Proposed scope

Create a new admin page that will allow authors to designate graphers, explorer views, or MDIMs as MIMs for a given topic.

**Must do**

- [ ] TBD, still establishing data model

PR #4711, opened 2025-03-27
🎉 MIMs admin UI

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/4670

This pull request introduces a new feature to manage "Most Important Metrics" (MIMs) in the admin. 

image

https://github.com/user-attachments/assets/36856737-4721-43ae-86b2-6561d51cc5b4

## Changes
- New Admin page
- New DB table - `mims`
- New function to validate explorer and grapher URLs
  - Opted not to validate query params, as we'll be able to do that during algolia indexing, much more easily

## To test
1. Go to http:staging-site-mim-admin/admin/mims
2. Paste various explorer and grapher URLs to various tags under various income groups
3. Rerank them
4. Reload the page, confirm everything is still in the right order


mlbrgl (approved):
Looks good to me! I like how you kept the level of engineering in check and aligned with the anticipated usage of fm.

A couple of nice to haves we could consider:
- given that our indexing refresh cycles are weekly still, a way to preview the FM per IG with thumbnails (e.g. borrowing from the data catalog ribbons)
- checking for deleted charts and marking orphaned FMs upon admin page load (I'm guessing you'll be checking for this during baking, and we don't delete charts that often, esp. those FM charts, so probably not a big deal)

ikesau:
Hey Matthieu - I've refactored the "all" behaviour to work as the originally intended "default" now. Would you mind having a look through the last 2 commits on this branch? 🙂 

This is the behaviour now:
image

baker/algolia/utils/shared.ts
```diff
@@ -61,3 +83,210 @@ export function scaleRecordScores<T extends { score: number }>(
...
+                    isIncomeGroupSpecificFM: false,
+                })
+            }
+        }
+    }
+
+    // Drop the defaults. We only want income-group copies
+    return expanded.filter(
+        (fm) => fm.incomeGroup !== FeaturedMetricIncomeGroup.Default
+    )
```
mlbrgl, line 231: it took me a moment to wrap my head around the "what happens if all income groups are filled in? How is the default case represented?" case, but it eventually made sense looking at the addition of isIncomeGroupSpecificFM to canonic chart and mdim views.

If at least one income group is blank, the the no country case will use the derived set of fm from that group. Otherwise, it will use the canonical view. For the user, there are no visible differences in using either records (thanks to originalAvailableEntities), and only one instance is being returned thanks to Algolia's deduplication.

Makes sense 👍🙂

mlbrgl (approved):
so nice that you found a solid way to handle this use case! 👏

PR #4713, opened 2025-03-28
Fix margin of Button without icon

rakyi:
## Context

When we recently changed how icons in `Button` work, we introduced this additional margin when there is no icon.

image


ikesau (approved):
🤦 

thanks! tested all 3 permutations and it works correctly, finally 😅

PR #4725, opened 2025-04-02
Fix Buildkite output groups

rakyi:
## Context

Buildkite uses special syntax where --- marks a start of the group in the output of a build. We were using it to mark an end of the group, making the rendered HTML collapsible sections confusing in the Builkite UI.

## Testing guidance

See the output of the Buildkite build for this PR.

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -159,18 +159,6 @@ function getProgressBarTotal(bakeSteps: BakeStepConfig): number {
     let total = minimum + bakeSteps.size
     // Redirects has two progress bar ticks
     if (bakeSteps.has("redirects")) total++
-    // Add a tick for the validation step that occurs when these two steps run
-    if (bakeSteps.has("dods") && bakeSteps.has("charts")) total++
-    // Add ticks for prefetching attachments, which will only run if any of these steps are enabled
-    if (
-        bakeSteps.has("gdocPosts") ||
-        bakeSteps.has("gdocTombstones") ||
-        bakeSteps.has("dataInsights") ||
-        bakeSteps.has("authors") ||
-        bakeSteps.has("multiDimPages")
-    ) {
-        total += 9
-    }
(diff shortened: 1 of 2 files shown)


baker/SiteBaker.tsx
```diff
@@ -159,18 +159,6 @@ function getProgressBarTotal(bakeSteps: BakeStepConfig): number {
...
-    // Add ticks for prefetching attachments, which will only run if any of these steps are enabled
-    if (
-        bakeSteps.has("gdocPosts") ||
-        bakeSteps.has("gdocTombstones") ||
-        bakeSteps.has("dataInsights") ||
-        bakeSteps.has("authors") ||
-        bakeSteps.has("multiDimPages")
-    ) {
-        total += 9
-    }
```
rakyi, line 173: This doesn't work well when we log the line at the start of the step instead of at the end of the step, because the prefetching group steps appear in the middle of a group that first needs prefetching, which is confusing. `console.log` should be good enough for prefetching instead.

ikesau (approved):
Wow, so sane. Funny how long we lived with this stupid thing.

I'm noticing for the first time, the pattern of "sub-progress bars" for certain steps (e.g. in the `bakeAllChangedGrapherPagesAndDeleteRemovedGraphers` step)

If you feel like being extra awesome, there are maybe a few more steps we could add those to. For instance, I'm seeing that on this branch, baking the blog index took 6 minutes (!) and I'm curious if each page took `360s/numberOfPages` time, or if there was one long step at the start.

Same goes for the prefetching attachments step. We can add a sub-progress bar there that doesn't have `---` at the beginning of its template string, to make it easier to debug build times from the logs.

rakyi:
You can enable timestamps in Buildkite to more or less get that additional information. This doesn't solve the problem when you run the bake in a terminal, but I think it's good enough. Feel free to open a new PR with the sub-progress bars!

image

PR #4728, opened 2025-04-03
:bug: Fix stuck deploy queue on staging servers

Marigold:
A lot of staging servers were failing in `getEmailSlackMentionsMap`, causing their deploy queues to run indefinitely. This happened because we used `SLACK_BOT_OAUTH_TOKEN` to detect staging servers—but we recently added that token to staging servers as well.

Fortunately, our staging servers now finally use `ENV=staging`, so we can clean up these hacks for good. 🎉

(I’ve deleted all `.pending` files on staging servers to unstuck the queues.)


rakyi (approved):
Thanks for the fix!

PR #4729, opened 2025-04-03
Wrangle less data

rakyi:
## Problem

1. We currently serialize the whole gdoc object to the HTML to the `window._OWID_GDOC_PROPS `, where it's used for hydration. It contains more data than we need, unnecessarily increasing the size of the page
   - For example, after the change the HTML of http://localhost:3030/foreign-aid-donations-increase becomes 16 kB smaller
3. We currently fetch the whole gdoc objects for a sitemap, even though we only need the `slug` and `updatedAt`. Not a big deal, but it was quick to improve

## Testing guidance

- Check that the sitemap looks the same after the change
- Check that all kinds of gdoc posts work without issues

I already tested the above locally.


db/db.ts
```diff
@@ -410,43 +408,6 @@ export const getImageMetadataByFilenames = async (
     return keyBy(rows, "filename")
 }
 
-export const getPublishedGdocPosts = async (
```
rakyi, line 413: This was only used in sitemap, so I moved and modified it there.

site/gdocs/OwidGdocPage.tsx
```diff
@@ -138,7 +139,30 @@ export default function OwidGdocPage({
...
+                                "linkedAuthors",
+                                "linkedDocuments",
+                                "linkedCharts",
+                                "linkedIndicators",
+                                "imageMetadata",
+                                "relatedCharts",
+                                "tags",
+                                "donors",
+                                "relatedCharts",
+                            ])
```
mlbrgl, line 165: good optimization idea! Maybe we could use something similar to what's being done in `checkIsLightningUpdate()` so that we get some kind of typechecking and documentation about what's included/ommitted and why. https://github.com/owid/owid-grapher/blob/8bb950b3e78662cc9fc59acfb99b9d22000858b2/adminSiteClient/gdocsDeploy.ts#L25

mlbrgl (approved):
LGTM!

baker/sitemap.ts
```diff
@@ -59,7 +90,7 @@ export const makeSitemap = async (
         undefined,
         (postrow) => !alreadyPublishedViaGdocsSlugsSet.has(postrow.slug)
     )
-    const gdocPosts = await db.getPublishedGdocPosts(knex)
+    const gdocPosts = await getGdocPosts(knex)
```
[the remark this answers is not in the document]
rakyi: Renamed.

rakyi:
Could you check again? I now also fixed the types.

mlbrgl (commented):
I'm not sure pick's typing is getting us the kind of confidence we are after, out of the box. If I misspell a key ("contant" instead of "content") or add a non-existent key in OwidGdoc (e.g. "xyz"), I get no type error. What's your reading?

mlbrgl:
I should have started with "thanks for the updates" btw 😅

So thanks for the updates!

rakyi:
It seems to work correctly for me. Is it possible your code wasn't up-to-date? Happy to jump on a call to investigate together.

SCR-20250411-jvpw
SCR-20250411-jwbh

mlbrgl:
* **#4757** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4757?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4729** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4729?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4729?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

mlbrgl:
@rakyi right, that makes sense. What I meant was that if you pick "contant" (with a typo), you get no error at the type level, and the return type of pick still advertises "content" for downstream code (a false positive of sorts).

Screenshot 2025-04-14 at 10 16 43


Here is a suggestion to make pick more type-aware: #4757

rakyi:
Thanks, Matthieu! I used `pick` from remeda, and we discussed and fixed a few other typing issues.

PR #4730, opened 2025-04-03
🎉 add OWID income groups to regions.ts

ikesau:
## Context

For featured metric indexing, we need a way to tell which income group a country belongs to. This PR changes the `regionsUpdater` script to include these regions, and adds code to omit them from the entity selector in graphers.

Income groups now appear in explorers' entity selectors, as I'm not sure if we want them there or not.

image

image

## Testing guidance
- Go to a grapher page and verify that an income group doesn't appear in the entity selector as a "local entity"


sophiamersmann (approved):
Looks good! One less thing I have to do this cycle hehe 

Having the income groups sorted to the top in explorers is fine, I think :)

PR #4738, opened 2025-04-07
Use Oxford comma to format list of authors

rakyi:
## Description

* Remove a duplicate function
* Split Bibtex formatting to a separate function
* Add more tests

## Context

To comply with our style guide, as [requested by Max](https://owid.slack.com/archives/C46U9LXRR/p1743841022366709).

## Testing guidance

Pls check if I could have missed a place where we format authors.


ikesau (commented):
image

This seems to contradict the tests you've written, but there shouldn't be a comma here
http://staging-site-fix-author-formatting/air-pollution-sources

rakyi:
Yeah, I actually realized independently that I missed this case and fixed it just now. 😄 It doesn't contradict the tests, though.

ikesau (approved):
Ah right, that was being accomplished with tsx.

All g now!

PR #4746, opened 2025-04-10
Fix copy for data insights

edomt:
## Context

Starting next week, we'll send our data insights three times per week. I've adapted the copy throughout the site to remove the word "daily" and replace "every weekday" with "every few days".

## Testing guidance

I _think_ I've identified all instances, but it's probably worth double-checking whether I've missed some or maybe other relevant words/sentences.


edomt:
@ikesau Would you have time to look at this early this week? Tomorrow will be the first day without a DI, so ideally we'd have this online by then (let me know if another reviewer makes more sense though)

PR #4779, opened 2025-04-17
Add support for external multi-dim embeds

### LINKED ISSUE #4775: Spike MDIM external embed
Timebox to two day.

rakyi:
## Context

This is a prototype of an external mdim embed. The final design might be different.

Resolves #4775.

`<iframe>` of the same mdim in a blank HTML page, with and without controls:

image

## Testing guidance

Test that the iframe behaves correctly. And since I also changed the mdim data page implementation, testing that would be appreciated too.

### Before merging

- [X] Google Analytics events were adapted to fit the changes in this PR

### DIFF
--- a/packages/@ourworldindata/components/src/Checkbox.tsx
+++ b/packages/@ourworldindata/components/src/Checkbox.tsx
@@ -4,6 +4,7 @@ import { faCheck } from "@fortawesome/free-solid-svg-icons"
 import cx from "classnames"
 
 export class Checkbox extends React.Component<{
+    className?: string
     checked: boolean
     onChange: React.ChangeEventHandler<HTMLInputElement>
     label: React.ReactNode
@@ -12,11 +13,17 @@ export class Checkbox extends React.Component<{
     "data-test"?: string
 }> {
     render(): React.ReactElement {
-        const { checked, onChange, label, disabled, id } = this.props
+        const { className, checked, onChange, label, disabled, id } = this.props
         const testHook = this.props["data-test"]
 
         return (
-            <div className={cx("checkbox", { "checkbox--disabled": disabled })}>
+            <div
+                className={cx(
+                    "checkbox",
+                    { "checkbox--disabled": disabled },
+                    className
+                )}
+            >
                 <label>
                     <input
                         id={id}
--- a/packages/@ourworldindata/components/src/RadioButton.tsx
+++ b/packages/@ourworldindata/components/src/RadioButton.tsx
@@ -1,6 +1,8 @@
 import * as React from "react"
+import cx from "classnames"
 
 export class RadioButton extends React.Component<{
+    className?: string
     checked: boolean
     onChange: React.ChangeEventHandler<HTMLInputElement>
     label: React.ReactNode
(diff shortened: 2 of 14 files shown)


packages/@ourworldindata/components/src/Checkbox.tsx
```diff
@@ -4,6 +4,7 @@ import { faCheck } from "@fortawesome/free-solid-svg-icons"
 import cx from "classnames"
 
 export class Checkbox extends React.Component<{
+    className?: string
```
rakyi, line 7: We need this to add margin in the embed modal.

packages/@ourworldindata/components/src/RadioButton.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from "react"
+import cx from "classnames"
 
 export class RadioButton extends React.Component<{
+    className?: string
```
rakyi, line 5: This was the easiest way to fix a type error after I added the `className` to `Checkbox`.

packages/@ourworldindata/explorer/src/Explorer.tsx
rakyi, line 1: I don't think the `hideControls` logic belongs here. Passing a JSX element deep into the component tree so it can be eventually rendered later is an antipattern. That's why I moved it to Grapher, where it is controlled with the new `canHideExternalControlsInEmbed` prop.

packages/@ourworldindata/grapher/src/core/relatedQuestion.ts
rakyi, line 1: Moved this here to allow fast refresh of the Grapher component. Fast refresh requires that the file with the component doesn't export other functions.

site/multiDim/MultiDim.tsx
```diff
@@ -55,6 +70,10 @@ export default function MultiDim({
         [config]
     )
 
+    useEffect(() => {
+        if (slug) analytics.logGrapherView(slug, { view: settings })
+    }, [slug, settings])
```
rakyi, line 75: Analytics was missing here.

site/multiDim/MultiDimDataPageContent.tsx
```diff
@@ -132,144 +132,148 @@ export type MultiDimDataPageContentProps = {
...
-        if (!grapher) return
-        const queryParams = Object.fromEntries(searchParams.entries())
-        updateGrapher(grapher, settings, queryParams)
-        // NOTE (Martin): This is the only way I was able to set the initial
-        // state on page load. Reconsider after the Grapher state refactor, i.e.
-        // when we decouple Grapher state from the Grapher components. Adding
-        // deps properly to the dep array leads to an infinite loop.
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [])
-
```
rakyi, line 272: I was able to get rid of this hack after I realized that duplicating the `settings` state in `useState` and URL params is better that doing this.

rakyi: Welp. I missed an edge case, this won't work. Keeping the hack for now.

marcelgerber (approved):
Nice, this works well!

I've tested this in an iframe (with controls shown/hidden), and also on the normal Mdim page.

marcelgerber:
You got some merge conflicts because I've also worked on the embed modal - but they should be relatively straightforward to resolve.

packages/@ourworldindata/grapher/src/core/Grapher.tsx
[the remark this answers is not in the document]
rakyi: Found the problem. It's unrelated to these changes and will be fixed separately.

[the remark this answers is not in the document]
rakyi: Fixed in https://github.com/owid/owid-grapher/pull/4801.

PR #4803, opened 2025-04-23
Add JSON-LD to gdoc posts

rakyi:
## Description 

Adding [JSON-LD](https://developers.google.com/search/docs/appearance/structured-data/article) might help with our click-through rates ([as claimed by Google](https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data)). It should make the Google search results prettier, with more context.

After we verify that it works and it is what we want, we can add more [types of JSON-LD](https://developers.google.com/search/docs/appearance/structured-data/search-gallery).

## Testing

[Rich Results Test](https://search.google.com/test/rich-results/result/r%2Farticles?id=8ytCZbzmjvGVoU5OnemjXA). The crawl failure is unrelated. It happens because Cloudflare Pages previews set `X-Robots-Tag: noindex`.

### DIFF
--- a/site/gdocs/OwidGdocPage.tsx
+++ b/site/gdocs/OwidGdocPage.tsx
@@ -19,6 +19,7 @@ import { match, P } from "ts-pattern"
 import {
     ARCHVED_THUMBNAIL_FILENAME,
     EnrichedBlockText,
+    OwidGdocPostInterface,
 } from "@ourworldindata/types"
 import { DATA_INSIGHT_ATOM_FEED_PROPS } from "../SiteConstants.js"
 import { Html } from "../Html.js"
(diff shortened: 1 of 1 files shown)


site/gdocs/OwidGdocPage.tsx
```diff
@@ -76,6 +77,81 @@ function getPageDesc(gdoc: OwidGdocUnionType): string | undefined {
...
+        <script
+            type="application/ld+json"
+            dangerouslySetInnerHTML={{
+                __html: JSON.stringify(data),
+            }}
+        />
+    )
+}
+
+function isPostPredicate(
```
ikesau, line 145: You might want to colocate and differentiate this from [checkIsGdocPostExcludingFragments](https://github.com/owid/owid-grapher/blob/master/packages/%40ourworldindata/utils/src/Util.ts#L1881)

A shame that we need 3 of these now, but oh well! 🙃

ikesau (approved):
Looks good! tested with articles and topic pages with various author permutations.

site/gdocs/OwidGdocPage.tsx
rakyi: Yeah... well the function param is a different type, and it's not clear we'll need the predicate anywhere else, so I'd keep it here for now.

PR #4828, opened 2025-05-01
✨ switch to meaningful time range for slopes

### LINKED ISSUE #4812: Choose appropriate time range when switching from line to slope charts
## Description

- For some of our charts, one example is [Child mortality rate](https://ourworldindata.org/grapher/child-mortality), only very few entities stretch the whole length of the time range
- After switching from a line to a slope chart, most countries say 'No data', which isn't helpful

## Expected behaviour

Grapher should try to automatically find a time range where all (or most) currently selected countries have data when the slope tab is clicked

## Steps to reproduce

1. Enable the slope tab for https://ourworldindata.org/grapher/child-mortality
2. Click on Slope
3. Note that all countries but Sweden don't have data

## Screenshots

Image

## Additional context

Has been described as a 'major limitation', so we should probably address this soon

sophiamersmann:
Resolves #4812

Example chart: [live](https://ourworldindata.org/grapher/child-mortality-igme?tab=chart) / [staging](http://staging-site-switch-to-slope-time/grapher/child-mortality-igme?tab=chart)

### Problem

- For some of our charts, only very few entities stretch the whole length of the time range
- After switching from a line to a slope chart, most countries say 'No data', which isn't helpful
- This has prevented authors to enable line/slope chart type switching for charts like [Child mortality rate](https://ourworldindata.org/grapher/child-mortality)

### Solution

- On switching to the slope chart tab, Grapher tries to find a time range for which most selected countries have data
- There are two concerns to balance:
    - We want to honour the user's selection
    - We want to show a meaningful slope chart
- The chosen compromise is to keep the end time fixed and only vary the start time (this also has a nice side effect of keeping the search space relatively small)
- It would be nicer if the time selection snaps back to the full range when the line tab is selected, but that's tricky to do because usually we want to keep the user selection as is

### Alternatives

- We might allow configs-per-chart-type at some point, which would also solve this problem, but we're not there yet (and we might never be)

### DIFF
--- a/packages/@ourworldindata/grapher/src/chart/ChartUtils.test.ts
+++ b/packages/@ourworldindata/grapher/src/chart/ChartUtils.test.ts
@@ -1,0 +1,98 @@
+import { expect, it, describe } from "vitest"
+import { OwidTable } from "@ourworldindata/core-table"
+import { findStartTimeForSlopeChart } from "./ChartUtils"
+
+describe(findStartTimeForSlopeChart, () => {
+    it("returns the original start time when all entities have data for it", () => {
+        const csv = `
+            entityName,gdp,time
+            USA,100,2000
+            USA,110,2002
+            USA,120,2005
+            Germany,180,2000
+            Germany,200,2002
+            Germany,220,2005
+            France,150,2000
+            France,160,2002
+            France,170,2005
+        `
+        const table = new OwidTable(csv)
+
+        const startTime = findStartTimeForSlopeChart(table, ["gdp"], 2000, 2005)
+        expect(startTime).toBe(2000)
+    })
+
+    it("finds a time for which all entities have data", () => {
+        const csv = `
+            entityName,gdp,time
+            USA,100,2000
+            USA,110,2002
+            USA,120,2005
+            Germany,200,2002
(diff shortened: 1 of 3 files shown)


ikesau (approved):
Nicely done! Clean implementation and good tests :)

In my attempt to build up my own repository of edge-case charts I found this one: ourworldindata.org/grapher/share-of-single-parent-households

The slope chart kind of breaks with or without your change, maybe we should remove it?

ikesau:
I do worry maybe a little that users might not notice that their time selection has been shifted when they navigate tabs, but if Marwa/Joe are okay with it, then that's fine.

sophiamersmann:
> In my attempt to build up my own repository of edge-case charts I found this one: https://ourworldindata.org/grapher/share-of-single-parent-households

Nice find! I'll ask the authors about it

> I do worry maybe a little that users might not notice that their time selection has been shifted when they navigate tabs, but if Marwa/Joe are okay with it, then that's fine.

I share the concern and we're usually more conservative but authors have avoided line/slope switching because of this particular issue, so we should to act.

I disabled the behaviour if the user has changed the time handles – that should make it a bit less disruptive.

PR #4832, opened 2025-05-02
🎉 support slope as default for line+slope charts

### LINKED ISSUE #4573: Add support for line/slope chart where the slope is the default view
## Core problem

For Graphers that have multiple chart type tabs, like line/slope charts, it's currently not possible to make the slope chart the default view.

## Proposed solution

The config field `tab` currently supports `table`, `map` and `chart` (where `chart` defaults to the first chart type). Add support for `line` and `slope` as well.

## Alternatives

We could also allow to configure slope/line charts (with the slope being the first chart type) and then keep the `tab` config field as is. But I think we should aim for consistent ordering

sophiamersmann:
Resolves #4573

Adds support for slope charts being the default in line/slope chart combinations.

More concretely, adds support for 'line' and 'slope' options to the `tab` config field, in addition to the currently available options, 'table', 'map' and 'chart'. If `tab` is 'chart' and there are more than one chart type in Grapher, the first chart type is shown by default.

### Technical details

- Tabs are handled on two levels in Grapher:
    - `tab` is 'table', 'map' or 'chart' (these options map to the config's tab field)
    - if `tab` is 'chart', then `chartTab` determines the currently active chart type
- `toObject()` serialises Grapher into a config object
    - if the active tab is different from the first chart type, it's stored in the config
- `updateFromObject()` constructs Grapher from a config object
    - if `tab` refers to a chart type, then `tab` and `chartTab` are set accordingly 

### Testing

- Go to http://staging-site-slope-as-default/admin/charts/712/edit
- Select the slope chart
- Refresh
- Note that the slope chart tab is shown by default

---
This is **part 1 of 7 in a stack** made with GitButler:
- <kbd>&nbsp;7&nbsp;</kbd> #4843 
- <kbd>&nbsp;6&nbsp;</kbd> #4842 
- <kbd>&nbsp;5&nbsp;</kbd> #4841 
- <kbd>&nbsp;4&nbsp;</kbd> #4839 
- <kbd>&nbsp;3&nbsp;</kbd> #4838 
- <kbd>&nbsp;2&nbsp;</kbd> #4833 
- <kbd>&nbsp;1&nbsp;</kbd> #4832 👈

### DIFF
--- a/packages/@ourworldindata/grapher/src/chart/ChartUtils.tsx
+++ b/packages/@ourworldindata/grapher/src/chart/ChartUtils.tsx
@@ -13,6 +13,8 @@ import {
     GrapherChartType,
     GRAPHER_CHART_TYPES,
     GRAPHER_TAB_QUERY_PARAMS,
+    GrapherTabOption,
+    GRAPHER_TAB_OPTIONS,
     InteractionState,
     SeriesName,
     GrapherInterface,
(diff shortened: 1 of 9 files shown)


ikesau (approved):
Nice! 

Tested _very_ rigorously 😅 

image

PR #4844, opened 2025-05-06
Add testimonials to donor list

rakyi:
## Description

Add testimonials interspersed in the list of donors, as requested by Charlie and Hassan. For the first version, we hard-code the list of testimonials, since it's likely we won't need to change it often. If/when the need to change them more often arises, we should be able to add them to the `{.donors}` Archie component.

## Context

* [Slack](https://owid.slack.com/archives/C07RNNQQD9V/p1746119872683459?thread_ts=1744211984.890759&cid=C07RNNQQD9V)
* [Figma](https://www.figma.com/design/kgwh14s9hheiWp7ajWUolJ/About-Pages?node-id=403-6802&m=dev)

image

## Testing guidance

A quick visual check should be enough.

- [X] Does the staging experience have sign-off from product stakeholders?

### Before merging

### DIFF
--- a/site/gdocs/components/Donors.tsx
+++ b/site/gdocs/components/Donors.tsx
@@ -1,29 +1,82 @@
-import { groupBy, removeDiacritics } from "@ourworldindata/utils"
+import * as _ from "lodash-es"
+import { removeDiacritics } from "@ourworldindata/utils"
 import { useDonors } from "../utils.js"
+
+function* generateTestimonials() {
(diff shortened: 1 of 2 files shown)


site/gdocs/components/Donors.tsx
```diff
@@ -1,29 +1,82 @@
-import { groupBy, removeDiacritics } from "@ourworldindata/utils"
+import * as _ from "lodash-es"
```
rakyi, line 1: In the long run, we should move to using Lodash functions from Lodash directly.

PR #4847, opened 2025-05-06
Switch internal non-leaf packages to just-in-time

rakyi:
## Description

Currently, all our internal packages in the `packages` dir are what Turborepo calls [compiled packages](https://turborepo.com/docs/core-concepts/internal-packages#compiled-packages). Here we switch packages other than `grapher` and `explorer` - which are bigger, and we intend to publish them separately - to [just-in-time packages](https://turborepo.com/docs/core-concepts/internal-packages#just-in-time-packages).

## Trade-offs

Compiling the packages separately means we don't have to compile the code again if something else changes. Switching to JIT packages means we'll lose that and running `tsc -b` will take longer. However, we avoid a chain of:

`utils` changes
-> `utils` re-compiles
-> `core-table` and `components` re-compile
-> `grapher` re-compiles
-> `explorer` re-compiles, 

and as the packages are changing, the `yarn startAdminDevServer` keeps restarting.

After the change, we shorten the chain to:

`utils` changes
-> `grapher` re-compiles
-> `explorer` re-compiles,

and the dev server restarts fewer times.

| Command | Before | After |
|--------|---: |---: |
| `yarn lerna run build --skip-nx-cache` | 28.19 s | 18.12 s |
| `yarn lerna run build` (using cache) after adding `console.log` to `comafyNumber` in `utils` | 26.46 s | 19.90 s |
| `yarn tsc -b -f` |  23.16 s | 40.52 s |
| `yarn buildVite` | 29.08 s | 31.36 s | 

We build Lerna packages much more often during development and run `tsc -b` rarely, therefore the trade-off is worth it.

### DIFF
--- a/packages/@ourworldindata/components/package.json
+++ b/packages/@ourworldindata/components/package.json
@@ -2,14 +2,10 @@
     "name": "@ourworldindata/components",
     "version": "0.0.0",
     "description": "This package contains shared UI components and styles for the Our World in Data site and the Grapher tool.",
-    "main": "dist/index.js",
+    "main": "src/index.ts",
+    "types": "src/index.ts",
     "directories": {
         "lib": "src"
-    },
-    "types": "dist",
-    "scripts": {
-        "build": "rm -rf ./dist && tsc --build",
-        "watch": "tsc --watch"
     },
     "keywords": [],
     "author": "",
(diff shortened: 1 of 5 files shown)


rakyi:
I think this is probably worth it after all. You both worked on our build setup in the past, so I'm curious what you think.

rakyi:
I recommend hiding whitespace to see the actual changes.

image

rakyi:
Another consideration here is that AI agents run tsc, and we don't want that to be slow, so I think we'll drop it here.

PR #4849, opened 2025-05-07
Fix narrative chart update lagging on prod

rakyi:
## Description

We must trigger a build if the updated narrative chart is used in an article.

## Context

[Slack discussion](https://owid.slack.com/archives/C46U9LXRR/p1745482169811779).

## Testing guidance

Check that a build is triggered when the narrative chart is used in an article.

### DIFF
--- a/adminSiteServer/apiRoutes/chartViews.ts
+++ b/adminSiteServer/apiRoutes/chartViews.ts
@@ -12,9 +12,9 @@ import {
     parseChartConfig,
     DbInsertChartView,
     ChartViewsTableName,
-    Base64String,
     ChartConfigsTableName,
     OwidGdocLinkType,
+    DbPlainUser,
     DbRawPostGdoc,
     DbRawImage,
     OwidGdocDataInsightContent,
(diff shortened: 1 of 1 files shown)


adminSiteServer/apiRoutes/chartViews.ts
```diff
@@ -239,16 +240,14 @@ export async function updateChartView(
...
     if (!rawConfig) {
         throw new JsonError("Invalid request", 400)
     }
 
-    const existingRow: Pick<
-        DbPlainChartView,
-        "chartConfigId" | "parentChartId"
-    > = await trx(ChartViewsTableName)
-        .select("parentChartId", "chartConfigId")
+    const existingRow = await trx<DbPlainChartView>(ChartViewsTableName)
```
rakyi, line 249: Knex does the `Pick`ing for use if we pass the type argument to `trx`.

adminSiteServer/apiRoutes/chartViews.ts
```diff
@@ -265,7 +264,7 @@ export async function updateChartView(
 
     await updateChartConfigInDbAndR2(
         trx,
-        existingRow.chartConfigId as Base64String,
```
rakyi, line 268: `chartConfigId` is in fact not a base64 string.

sophiamersmann (approved):

PR #4854, opened 2025-05-08
✨ add bar chart tab for all line charts

sophiamersmann:


marcelgerber (approved):
Awesome!

A few small issues I've noticed:
- I've somehow gotten into [this state](http://staging-site-bar-chart-tab/grapher/share-of-the-population-with-access-to-electricity?tab=discrete-bar&time=earliest..1990) (note the `earliest..1990`), and now can't switch to the line chart at all
- When playing an animation from the "Line" tab, it switches to the chart tab. We should make it so that you either get a line race or a bar race, depending on the tab you're on.

PR #4860, opened 2025-05-09
🐛 fix explorer featured metric indexing

ikesau:
`findMatchingRecordByPathnameAndQueryParams` was broken because I used the string `"explorer"` instead of the correct `"explorers"`

This PR fixes that by using the correct constant and creating one for Grapher as well, for consistency's sake. 

I'm holding off on using it everywhere because it competes with `BAKED_GRAPHER_URL`. I've added a point to next week's meeting's agenda to discuss.


rakyi (approved):
Looks good as a solution to the problem, but we should probably move these constants to `settings`, since we want to treat `grapher` (and `explorer`?) as reusable packages, so they should not contain constants only relevant for OWID.

PR #4861, opened 2025-05-09
Rename chart views to narrative charts

rakyi:
## Description

Before, we used "narrative charts" in the UI and "chart views" in most of the code. That's confusing, because we refer to the different configurations of explorer and mdim dimensions as chart views. Therefore we decided (with Daniel) to harmonize the naming to narrative charts.

## Testing guidance

Quick check that all narrative chart functionality still works the same.

## Checklist

### Before merging

- [X] Google Analytics events were adapted to fit the changes in this PR
  - [ ] Will let Bobbie know about the change

If DB migrations exists:

- [X] If columns have been added/deleted, all necessary views were recreated
- ~~The DB types in the ETL have been updated~~
- [X] If tables/views were added/removed, the Datasette export has been updated to take this into account

### After merging

- ~~If a table was touched that is synced to R2, the sync script to update R2 has been run~~


adminSiteClient/NarrativeChartIndexPage.tsx
```diff
@@ -147,17 +155,19 @@ export function ChartViewIndexPage() {
...
 
-        void getChartViews().then((res) => setChartViews(res.chartViews))
+        void getNarrativeCharts().then((res) =>
+            setNarrativeCharts(res.narrativeCharts)
+        )
     }, [admin])
 
     return (
         <AdminLayout title="Narrative charts">
-            <main className="ChartViewIndexPage">
```
rakyi, line 160: This class wasn't defined anywhere.

rakyi:
Thanks for the review and good catch!

No worries, it wasn't a clear-cut decision that renaming was necessary before, but it made sense to do it now.

PR #4863, opened 2025-05-12
Add text-wrap: balance to short article elements

rakyi:
## Description

Add the [`text-wrap: balance`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap#balance) to:

* Headings
* Image captions
* Pull quotes

Not adding it to blockquote, since we often use it to cite longer paragraphs of text, and `balance` only works for text spanning a limited number of lines, so it can easily lead to paragraphs of different length looking inconsistent next to each other.

## Context

[Slack discussion](https://owid.slack.com/archives/C03SK89695E/p1747028552252099).

Example of the difference it makes for headings:

| Before | After |
|--------|--------|
| image | image | 

## Testing guidance

I already tested all the components in existing articles that this modifies. You could double-check that. For that, the `posts_gdocs_components` table is useful.

### DIFF
--- a/site/gdocs/components/centered-article.scss
+++ b/site/gdocs/components/centered-article.scss
@@ -328,6 +328,10 @@
     }
 }
 
+.article-block__heading {
+    text-wrap: balance;
+}
+
 h1.article-block__heading {
     a.deep-link {
         height: 16px;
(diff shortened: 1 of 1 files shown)


mlbrgl (approved):
makes sense, thanks! 👍

PR #4877, opened 2025-05-15
🎉 support search for data tables

sophiamersmann:
---
⧓ Review in [Butler Review `#0H7M2quJT`](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/0H7M2quJT)

**data-table-search**

1 commit series (version 5)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 1/1 | [🎉 support search for data tables](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/0H7M2quJT/commit/bb68ad30-fc3a-4a24-838c-19e871b40fa6) | ⏳ |  |

_Please leave review feedback in the [Butler Review](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/0H7M2quJT)_

PR #4875, opened 2025-05-15
✨ (data table) drop show-selection-only toggle in favour of dropdown

sophiamersmann:
---
⧓ Review in [Butler Review `#7exHRsJKn`](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn)

**data-table-selection**

6 commit series (version 5)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 6/6 | [🔨 refactor controls row](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn/commit/441c745d-60b3-4abd-a620-071ad2ca7cb7) | ⏳ |  |
| 5/6 | [🔨 drop width calculation in controls row](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn/commit/fc78c24f-3c77-4ea7-958a-b57a1956a90e) | ⏳ |  |
| 4/6 | [🐛 (data table) default to the 'All' filter](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn/commit/0b1ddf14-3596-4796-a28b-ff4ce3e8422e) | ⏳ |  |
| 3/6 | [🐛 apply table selection only if the selection is not empty](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn/commit/2824f990-5962-4c02-819b-292a38f3e24f) | ⏳ |  |
| 2/6 | [✨ add filter dropdown for the data table](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn/commit/a76d2bbb-8b47-450e-847a-f87362646966) | ⏳ |  |
| 1/6 | [🔨 drop country/other grouping](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn/commit/22c86813-12f3-4f4c-82ee-03ab694ac085) | ⏳ |  |

_Please leave review feedback in the [Butler Review](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/7exHRsJKn)_

PR #4745, opened 2025-04-09
🎉 highlight regions on the map / TAS-975

sophiamersmann:
---
⧓ Review in [Butler Review `#6FcQRkrD9`](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/6FcQRkrD9)

**better-maps-regions**

1 commit series (version 5)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 1/1 | [🎉 highlight continents/regions/income groups on the map](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/6FcQRkrD9/commit/6d594241-9bc6-4f15-8d5e-1519c825977b) | ⏳ |  |

_Please leave review feedback in the [Butler Review](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/6FcQRkrD9)_

PR #4744, opened 2025-04-09
🎉 (entity selector) filter entities by type / TAS-923

sophiamersmann:
Resolves #3599 

Adds a 'Filter by type' dropdown to the entity selector.

### Notes

- I initially thought I'd use the newly imported `definedBy` field of the regions file, but it turns out that non-owid regions aren't really standardised which is why I resorted to string matching '... (SOURCE)' instead
- In addition to `WB`, `UNSD` and `WHO` defined in the regions file, I also added support for more sources; those are hard-coded

---
This is **part 4 of 16 in a stack** made with GitButler:
- <kbd>&nbsp;16&nbsp;</kbd> #4914 
- <kbd>&nbsp;15&nbsp;</kbd> #4911 
- <kbd>&nbsp;14&nbsp;</kbd> #4907 
- <kbd>&nbsp;13&nbsp;</kbd> #4883 
- <kbd>&nbsp;12&nbsp;</kbd> #4877 
- <kbd>&nbsp;11&nbsp;</kbd> #4876 
- <kbd>&nbsp;10&nbsp;</kbd> #4875 
- <kbd>&nbsp;9&nbsp;</kbd> #4864 
- <kbd>&nbsp;8&nbsp;</kbd> #4809 
- <kbd>&nbsp;7&nbsp;</kbd> #4808 
- <kbd>&nbsp;6&nbsp;</kbd> #4770 
- <kbd>&nbsp;5&nbsp;</kbd> #4745 
- <kbd>&nbsp;4&nbsp;</kbd> #4744 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #4743 
- <kbd>&nbsp;2&nbsp;</kbd> #4719 
- <kbd>&nbsp;1&nbsp;</kbd> #4708 

---
⧓ Review in [Butler Review `#1EhYge0Eu`](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/1EhYge0Eu)

**entity-selector-filter-dropdown**

13 commit series (version 5)

| Series | Commit Title | Status | Reviewers | 
| --- | --- | --- | --- |
| 13/13 | [🔨 default to 'all' if no filter options are available](https://gitbutler.com/sophiamersmann/owid-grapher/reviews/1EhYge0Eu/commit/a05949fd-e626-4da2-8e82-4c2ac4301982) | ⏳ |  |
[...]

PR #4883, opened 2025-05-16
🐛 avoid nonsensical stacking in stacked area/bar charts

sophiamersmann:
---
This is **part 13 of 13 in a stack** made with GitButler:
- <kbd>&nbsp;13&nbsp;</kbd> #4883 👈 
- <kbd>&nbsp;12&nbsp;</kbd> #4877 
- <kbd>&nbsp;11&nbsp;</kbd> #4876 
- <kbd>&nbsp;10&nbsp;</kbd> #4875 
- <kbd>&nbsp;9&nbsp;</kbd> #4864 
- <kbd>&nbsp;8&nbsp;</kbd> #4809 
- <kbd>&nbsp;7&nbsp;</kbd> #4808 
- <kbd>&nbsp;6&nbsp;</kbd> #4770 
- <kbd>&nbsp;5&nbsp;</kbd> #4745 
- <kbd>&nbsp;4&nbsp;</kbd> #4744 
- <kbd>&nbsp;3&nbsp;</kbd> #4743 
- <kbd>&nbsp;2&nbsp;</kbd> #4719 
- <kbd>&nbsp;1&nbsp;</kbd> #4708


ikesau:
I guess this is because we don't have member data for `Europe and Central Asia (PIP)`? (according to [WB](https://data.worldbank.org/?locations=Z7-AD), it _should_ include France (if that random page on their site is a valid source to use)), and we have `Europe and Central Asia (WB)` in our regions file. Might need to check with Pablo A to see if we can update it from `(PIP)` to (WB)

image

ikesau (approved):
Okay this took me a little bit of poking around in the code to understand how faceting strategies work, but it's making sense now. I went through a couple of pages of Stacked Area Charts on /admin/test and couldn't find any issues (except with the PIP region thing. I'm not sure which approach is the right one to take for fixing that. You probably have a better sense 🙂 )

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
```diff
@@ -459,3 +482,54 @@ export class AbstractStackedChart
...
+        )
+    }
+    return true
+}
+
+function getCountryNames(region: Region): string[] {
+    return checkHasMembers(region)
+        ? getCountryNamesForRegion(region)
+        : [region.name]
+}
```
marcelgerber, line 535: You don't need this because `getCountryNamesForRegion` does precisely that.

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
```diff
@@ -373,20 +379,37 @@ export class AbstractStackedChart
...
+        if (
+            // No facet strategy makes sense if columns are stacked and a single entity is selected
+            (!this.isEntitySeries && !areMultipleEntitiesSelected) ||
+            // No facet strategy makes sense if entities are stacked and we have a single column
+            (this.isEntitySeries &&
+                !hasMultipleYColumns &&
+                // The stacking must be sensible
+                checkIsStackingEntitiesSensible(selectedEntityNames))
+        )
             strategies.push(FacetStrategy.none)
```
marcelgerber, line 394: One thing I'm wondering here is if we can ever run into the scenario where we end up returning an empty array.

And we rely on this fact here: https://github.com/owid/owid-grapher/blob/aee8d4a17d5fe520d00fbfa42a52408811232a70/packages/%40ourworldindata/grapher/src/core/Grapher.tsx#L2944

In the past, this was either highly unlikely or completely impossible.

Now, in theory, it would be possible:
E.g. when selecting "World" and "Europe", and the chart has multiple units.

Probably, we don't have any charts where that's possible, but it would at least be good to make us aware of such issues. Could you throw an error in that case, and send that error to Sentry?

marcelgerber (approved):

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
sophiamersmann: It does, but the input type to `getCountryNamesForRegion` is narrowed to a `Region` that is not a `Country`. I needed it here but I thought it was confusing in general to implement `getCountryNamesForRegion` for countries.

packages/@ourworldindata/grapher/src/stackedCharts/AbstractStackedChart.tsx
sophiamersmann: Yeah, that's a good point! I'll add error handling.

sophiamersmann:
> I guess this is because we don't have member data for PIP regions? (according to [WB](https://data.worldbank.org/?locations=Z7-AD), Europe and Central Asia (PIP) should include France (if that random page on their site is a valid source to use))

Yup, there is nothing we can do if we don't have member data. We probably have tens of charts with PIP data, so adding PIP (and other sources) to the regions file would be best. I'll ask the data folks about it.

PR #4885, opened 2025-05-16
🐛 fix dod migration for prod DB

ikesau:
The [dod admin migration](https://github.com/owid/owid-grapher/pull/4880) assumed that the DB has an admin user with ID 1. This isn't the case! 

This PR fixes the failing migration.

The empty tables will need to be deleted from the prod DB before this is merged so that the migration doesn't crash trying to create tables that already exist.


Marigold (approved):
Looks good!

PR #4898, opened 2025-05-20
✨ (admin) add help text for tolerance fields

sophiamersmann:
Adds a note to both tolerance fields

### DIFF
--- a/adminSiteClient/DimensionCard.tsx
+++ b/adminSiteClient/DimensionCard.tsx
@@ -249,6 +249,9 @@ export class DimensionCard<
                             store={dimension.display}
                             auto={column.tolerance}
                             onBlur={this.onChange}
+                            helpText={`Tolerance for charts and maps. The tolerance
+                                set in the Map tab, if defined, overrides the map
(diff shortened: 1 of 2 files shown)


adminSiteClient/DimensionCard.tsx
```diff
@@ -249,6 +249,7 @@ export class DimensionCard<
                             store={dimension.display}
                             auto={column.tolerance}
                             onBlur={this.onChange}
+                            helpText="May be overridden by the map's tolerance setting"
```
pabloarosado, line 252: Maybe it's better to clarify: "Tolerance for charts (bar, marimekko, [others?]) and maps. The tolerance set in the Map tab, if defined, overrides the map tolerance defined here." (feel free to rephrase or shorten if you think this is too much detail).

pabloarosado (approved):
Thank you for adding the explanations! I've suggested some further clarifications, but I leave it up to you if you think your current proposal is enough detail.

sophiamersmann:
Thank you, @pabloarosado!

PR #4899, opened 2025-05-20
✨ add option to exclude sub-areas from parentTagArrays

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/4675

Allows us to filter out non-topic, non-area tags from the tag graph and does so for mdim, chart, and explorer indexing.


mlbrgl (approved):
makes sense to me! 👍

PR #4917, opened 2025-05-23
Fix incorrect imports from package dist dir

rakyi:


ikesau (approved):
🙈

ikesau:
Oops! Let me see if I can prevent my editor from finding/suggesting these.

rakyi:
Nice, thanks for sharing!

PR #4924, opened 2025-05-26
Allow previewing unpublished mdims in gdocs

rakyi:
## Description

We identify both normal charts and mdims by slug in the Archie component, but we can preview only unpublisehd mdims in the gdoc because slug is not unique for unpublished charts.

We still show a validation error in the admin and prevent publishing a gdoc with unpublished mdim.

## Context

Requested by Joe. Required for our work on new topic pages (and/or articles) with mdims.

image

## Testing guidance

Check that it's possible to preview unpublished mdim in a gdoc, but it's not possible to publish such gdoc.

### DIFF
--- a/adminSiteClient/GdocsSettingsForms.tsx
+++ b/adminSiteClient/GdocsSettingsForms.tsx
@@ -32,12 +32,12 @@ const GdocCommonErrors = ({
     )
     return (
         <>
-            {errorsToShowInDrawer.map((error) => (
+            {errorsToShowInDrawer.map((error, index) => (
                 <Alert
+                    key={index}
+                    className="GdocsSettingsForm__alert"
                     message={error.message}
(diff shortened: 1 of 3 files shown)


adminSiteClient/GdocsSettingsForms.tsx
```diff
@@ -32,12 +32,12 @@ const GdocCommonErrors = ({
...
     return (
         <>
-            {errorsToShowInDrawer.map((error) => (
+            {errorsToShowInDrawer.map((error, index) => (
                 <Alert
+                    key={index}
+                    className="GdocsSettingsForm__alert"
                     message={error.message}
                     type={error.type}
-                    key={error.message}
```
rakyi, line 39: We can get multiple errors with the same message, so we can't use the message as the key.

PR #4925, opened 2025-05-26
Add changes introduced with the new maps to MultiDim

rakyi:
Mirrors changes to `MultiDimDataPageContent` made in 20f01500183a26468a05ae9c07ecb3e15d699157.

The `MultiDim` component is used both for gdoc and iframe embeds. I assume we need the same logic there and you just didn't know the component exists. Is that correct?


rakyi:
Is the failing svg-tested a false positive? Something weird seems to be happening there.

sophiamersmann:
yes, those are false positives. if you want to be sure, you could rebase on master, but that's not necessary in my opinion.

PR #4931, opened 2025-05-27
🔨 throw if catalog paths can't be resolved to indicator ids

sophiamersmann:
Throws an error during baking if a catalog path can't be resolved to an indicator id for an explorer. This is in response to this [incident](https://owid.slack.com/archives/C46U9LXRR/p1748355592800989).

Baking fails, which is a sign that the code works 🌝 I think Lucas has updated the catalog paths by now, so probably should be ok to merge.

Is throwing an error what we usually do if something is off during baking? I remember that we used to have something like log-the-error-and-report-it-but-continue-execution.

### DIFF
--- a/baker/ExplorerBaker.tsx
+++ b/baker/ExplorerBaker.tsx
@@ -17,6 +17,7 @@ import {
     isNotErrorValueOrEmptyCell,
 } from "@ourworldindata/core-table"
 import { ColumnTypeNames } from "@ourworldindata/types"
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
Thank you!

Yes, I believe the log-error-but-continue-baking approach works a bit better here.
Otherwise, baking will be interrupted for a few hours or so while the error is present, which is probably worse than having a broken explorer around for a few hours.

PR #4934, opened 2025-05-28
Remove unused suggested_chart_revisions table

Marigold:
## Summary
• Removes the `suggested_chart_revisions` table which appears to be unused in the application code
• Analysis shows this table is only referenced in database migrations, with no TypeScript/JavaScript usage

## Test plan
- [x] Verify migration runs successfully
- [x] Confirm no application functionality is affected
- [x] Check that all existing migrations still work correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/db/migration/1748446282000-DropSuggestedChartRevisionsTable.ts
+++ b/db/migration/1748446282000-DropSuggestedChartRevisionsTable.ts
@@ -1,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class DropSuggestedChartRevisionsTable1748446282000
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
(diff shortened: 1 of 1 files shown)


db/migration/1748446282000-DropSuggestedChartRevisionsTable.ts
```diff
@@ -0,0 +1,15 @@
...
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            "DROP TABLE IF EXISTS `suggested_chart_revisions`"
+        )
+    }
+
+    public async down(_queryRunner: QueryRunner): Promise<void> {
+        // No rollback - table removal is permanent
```
rakyi, line 13: I just noticed this, and it's hopefully not really an issue, but as a reference for the future: proper migrations should always have a `down` migration, which is easy in this case: create the table.

PR #4943, opened 2025-06-02
Mdim dropdowns layout improvements

rakyi:
## Context

[Slack discussion](https://owid.slack.com/archives/C071F90MD0X/p1748420509861499).

## Testing guidance

Check that the dropdowns are fixed width on small screens. And that dropdowns row wraps under the heading when there isn't enough space.

### Before merging

### DIFF
--- a/site/multiDim/MultiDimDataPageSettingsPanel.scss
+++ b/site/multiDim/MultiDimDataPageSettingsPanel.scss
@@ -5,6 +5,7 @@ $toggle-full-width-breakpoint: 520px;
 
 .md-settings-row {
     display: flex;
+    flex-wrap: wrap;
     gap: 16px;
 
     .md-settings__configure-data {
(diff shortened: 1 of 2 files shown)


site/multiDim/MultiDimDataPageSettingsPanel.scss
```diff
@@ -1,6 +1,7 @@
 $indent: 16px;
 
 $toggle-height: 40px;
+$toggle-full-width-breakpoint: 520px;
```
rakyi, line 4: This is a bit arbitrary. It's not easy to calculate it dynamically, and this works well.

marcelgerber (approved):
This is definitely an improvement over what we have now, but do you think we can improve this even further using grid styles?

For example, this here doesn't look too great, with all the blank space on the right:
CleanShot 2025-06-11 at 10 47 58

rakyi:
Yeah, that didn't look great. I changed it so we jump from the variable size to full width directly without going through this fixed size.

I'm going to implement a more significant change for the dropdowns next anyway, since we talked about it more with Joe, Daniel, and Marwa, so this PR should be good enough for now.

PR #4945, opened 2025-06-02
Fix Grapher.isAdmin for gdoc previews in prod

rakyi:


packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -1292,7 +1307,7 @@ export class Grapher
         let variablesDataMap: MultipleOwidVariableDataDimensionsMap
 
         const startMark = performance.now()
-        if (this.isAdmin) {
+        if (this.isAdminObjectAvailable) {
```
rakyi, line 1310: Here we care about the `window.admin` object, which we then use in the nested calls rather than about whether we are in the admin, which is currently not the same thing.

PR #4949, opened 2025-06-02
🚧 [WIP] A/B testing mvp

bnjmacdonald:
The implementation is still buggy and missing some key features.

Key components of the PR:
- the A/B tests to run are hard-coded in ab-tests.json;
- the abTest.ts middleware assigns cookies for these tests (if the visitor isn't already assigned a cookie for the test);
- the addABTestClasses function in grapherTools.ts assigns CSS classes to the document body, reflecting the A/B testing cookies that are set;
- most of the remaining code deals with the actual CSS/HTML of the tests.

In this draft, I've hard-coded 2 tests just as part of the proof of concept: (1) showing/hiding the data page header; and (2) showing/hiding a button beneath the grapher that says "View insights about this data".


functions/_common/ab-tests.json
```diff
@@ -0,0 +1,18 @@
+[
+    {
+        "id": "ab-data-page-header",
+        "expires": "2025-06-30T00:00:00.000Z",
+        "arms": [
+            { "id": "ctl", "size": 0.5 },
+            { "id": "trt", "size": 0.5 }
```
rakyi: I'm not familiar with these abbreviations(?). Could we use full words to increase the readability? Unless those are standard terms like HTML.

functions/tsconfig.json
```diff
@@ -5,6 +5,7 @@
         "moduleResolution": "NodeNext",
         "lib": ["esnext"],
         "types": ["@cloudflare/workers-types"],
-        "esModuleInterop": true
+        "esModuleInterop": true,
+        "resolveJsonModule": true
```
rakyi: We don't really need to use a JSON file, do we? JSON doesn't allow comments or types and needs other kinds of special treatment. I think we can define the tests in `abTest.ts` instead.

site/DataPageContent.scss
rakyi, line 1: I don't think we need any of the `visibility` properties here. Using only `display` is fine.

Nesting a SCSS class in itself is not a good practice, not even sure if it's valid. The approach we had in mind with Ike was:

1. Add A/B test class to the `body`
2. Nest any related rules under it, likely without any other new classes

Example:

1. Add `.ab-data-page-insights-1-trt` to `body`
2. In `DataPageContent.scss`:

    ```scss
    .DataPageContent {
        .insights-btn {
            display: none;
        }
    }
    
    .ab-data-page-insights-1-trt {
        .DataPageContent {
            .insights-btn {
                display: block;
            }
        }
    }
    ```

So no `.ab-show` or `.ab-hide`.

functions/_common/abTest.ts
```diff
@@ -0,0 +1,51 @@
+import tests from "./ab-tests.json" with { type: "json" }
+// todo: send GA paramaters with all events, listing all experimental arms/conditions
+
+export const abTest = async (context) => {
+    const originalResponse = await context.next()
+    const cookie = context.request.headers.get("cookie")
+
+    if (tests && tests.length) {
+        let anyTrt = false
+        tests.map((test) => {
```
rakyi: Here, and also below, `map` creates a new array, but we ignore it, so we should either use `forEach` or `for...of`. (`for...of` is the most readable, IMO.)

functions/_common/abTest.ts
```diff
@@ -0,0 +1,51 @@
...
+                let assignedArm
+                for (const arm of test["arms"]) {
+                    const assignToArm =
+                        arm["cumulSize"] - arm["size"] <= percentage &&
+                        percentage < arm["cumulSize"]
+                    if (assignToArm) {
+                        assignedArm = arm["id"]
+                        break
+                    }
+                }
```
rakyi: It looks like we are generalizing the algorithm for arbitrary number of arms, even though we only have two. I suggest we start with the simplest implementation and assume only two arms, which can be described with a single float. We can make it more complex when we'll need it.

bnjmacdonald:
thanks a lot @rakyi! All your suggestions make sense.

site/DataPageContent.scss
bnjmacdonald: ✅ I've removed the `visibility` prop and gotten rid of `ab-show`

But I've kept the awkward logic of nesting `ab-data-page-insights-1-trt` in itself and using `ab-hide` (now `exp-hide`). So the rough logic looks like:

```scss
body {
    // by default, hide all elements part of the experiment
    .exp-data-page-insight-buttons-basic--t1__block {
        display: none;

        // add .exp-hide to elements that are shown by default but which you want
        // to hide for an experimental arm (must also include experimental arm in
        // the element's class names, e.g. "exp-data-page-insight-buttons-basic--t0__block exp-hide")
        &.exp-hide {
            display: block;
        }
    }

    // show elements in the arm the user is assigned to
    &.exp-data-page-insight-buttons-basic--t1
        .exp-data-page-insight-buttons-basic--t1__block {
        display: block;

        &.exp-hide {
            display: none !important;
        }

    }

    // styling for elements in {arm}
    &.exp-data-page-insight-buttons-basic--t1 {
        .DataPageContent {
            .chart-key-info {
                padding-bottom: 0;
                @include sm-up {
                    padding-bottom: 0;
                }
            }
            ...
        }
    }
}
```

Reasoning:

Regarding `.exp-hide`... this is necessary when we want to hide an element that is otherwise shown by default.
- e.g. in the `exp-data-page-insight-buttons-basic` experiment some conditions involve moving the "what you should know about this data" block down the page. Which I implement by adding a duplicate `<AboutThisData ... />` component lower down and then hiding the original if user is in a treatment condition.

Regarding nesting `exp-data-page-insight-buttons-basic` in itself... I did change the class names so that it's not literally the same class nested in itself (i.e. it's now `&.{experimentName}--t1 .{experimentName--t1__block {...}` instead of `&.{experimentName}--t1 .{experimentName--t1 {...}`.

But I've had to keep this more complicated approach rather than the simple one you laid out b/c of >2 arms per experiment. 

e.g. consider this block of code, which renders a button with `className="insights-link" in each of arm "t0" and "t1". Assume the user is assigned to "t0" (represented by the class in the wrapping `<body>`):

```jsx
<body class="exp-data-page-insight-buttons-basic--t0">
// button for arm t0
<Button
    className="insights-link"
    href="#about-the-data"
    text="Data sources & measurement"
    theme="solid-blue"
/>
// button for arm t1
<Button
    className="insights-link"
    href={insightsHref}
    text="View insights about this data"
    theme="solid-blue"
/>
</body>
```

If our css was just 

```scss
.DataPageContent {
    .insights-btn {
        display: none;
    }
}

.exp-data-page-insight-buttons-basic--t0,
.exp-data-page-insight-buttons-basic--t1 {
    .DataPageContent {
        .insights-btn {
            display: block;
        }
    }
}
```

then both buttons would be selected and rendered with `display: block` (but we only want to show one). So we need to add a class like `exp-data-page-insight-buttons-basic--t0__block` to each button and use css selectors that only set `display: block` when the body has `.exp-data-page-insight-buttons-basic--t0`and the child element has `.insights-btn` AND itself or a parent has `.exp-data-page-insight-buttons-basic--t0__block`.

Maybe I'm still not seeing a simpler way here, but I *think* this is necessary to deal with experiments that have >2 arms.

functions/_common/abTest.ts
bnjmacdonald: it's turned out that supporting experiments with >2 arms is pretty important right out of the gate, so I've kept this (and added other logic elsewhere to support >2 arms).

bnjmacdonald:
> I suggest using the [cookie](https://www.npmjs.com/package/cookie) package for parsing and serializing cookies to avoid potential gotchas and bugs, and to make the code more readable.
> 
> You might want to coordinate with Ike, since he is [adding GA to Cloudflare functions](https://github.com/owid/owid-grapher/pull/4977).

Edit: done ✅.

I haven't done this yet – added to the todo list above. Doesn't look like Ike used any cookie parsing packages in the `analyticsMiddleware` he added.

functions/_common/ab-tests.json
bnjmacdonald: Edit: I've renamed "c" to "control", "t1" to "treat1", etc.

As things currently stand, I've actually moved even more in the direction of shorthand, using arm ids like "c" for control and "t0", "t1", etc for treatment arms. 

I'm a little torn about whether to use "control" and "treat", but also don't feel strongly about it at all. Part of me wants to be a little opaque here so that the small minority of users who look at the html don't see a css class such as "data-page-insights-treat1" and maybe infer they are part of some big experiment (whereas in reality it's just a tiny experiment that renders a single button). "data-page-insights-t1" is much more ambiguous.

Probably not a very realistic concern though. And obviously we like to be transparent, and don't mind users being aware that we are doing some A/B testing (as most websites do). So I'll rename to "treat" and "control" (added to todo list).

site/runSiteFooterScripts.tsx
```diff
@@ -53,13 +54,44 @@ function hydrateSearchPage() {
...
+                props.totalPageCount = Math.ceil(
+                    dataInsights.length / DATA_INSIGHTS_INDEX_PAGE_SIZE
+                )
+                props.topic = { name: tag.name, slug: tag.slug }
+                props.dataInsights = dataInsights.slice(
+                    props.pageNumber * DATA_INSIGHTS_INDEX_PAGE_SIZE,
+                    (props.pageNumber + 1) * DATA_INSIGHTS_INDEX_PAGE_SIZE
+                )
+            }
+        }
```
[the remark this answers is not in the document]
bnjmacdonald: ignoring this b/c the current fallback on error (or on an invalid topicName) is to just render all data insights. Would be good in the future to provide an informative error message to the user, but the current behavior is fine for now b/c this is just an experimental/temporary feature.

functions/tsconfig.json
rakyi, line 1: Let's not commit empty line removals.

functions/experiments/types.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (!isFraction(n)) {
+        throw new Error(`Invalid Fraction: ${n} is not between 0 and 1`)
+    }
+    return n as Fraction
+}
+
+type ISODateString = string & { __brand: "ISODateString" }
+
+function isISODateString(value: string): value is ISODateString {
+    return !isNaN(Date.parse(value))
```
rakyi: `Date.parse()` works on other strings than ISO date strings. IMO, we don't have to go so much into type safety and could drop the branded types, but no pressure.

functions/experiments/Experiment.ts
```diff
@@ -0,0 +1,151 @@
...
+ * @property {Arm[]} arms - The list of arms (variants) in the experiment, each with its own fraction.
+ * @property {CookiePath[]} paths - The list of cookie paths where the experiment applies.
+ *
+ * @method getArmById - Retrieves an arm by its unique identifier.
+ * @method isExpired - Determines if the experiment has expired.
+ *
+ * @throws {Error} If the sum of arm fractions does not equal 1, if arm IDs are not unique, or if
+ * any `${experimentId}-${armId}` exceeds 100 characters.
+ */
+export class Experiment implements ExperimentInterface {
```
rakyi: We don't really need the separate interface, TS classes can be used directly as types.

functions/experiments/middleware.ts
```diff
@@ -0,0 +1,75 @@
+import { Experiment, validateUniqueExperimentIds } from "./Experiment.js"
+import { Arm, ServerCookie } from "./types.js"
+import { requestIsInIframe, isStaticAsset } from "./utils.js"
+import { experiments } from "./config.js"
+import * as cookie from "cookie"
+
+export const experimentsMiddleware = (context) => {
+    if (
+        isStaticAsset(context.request.url) ||
+        requestIsInIframe(context.request)
```
rakyi: For our first experiment this is useful, but we might want to experiment with the contents of iframes as well in the future. I think it would be better if we didn't treat iframe requests as special and filter iframes events out when doing the analysis.

functions/experiments/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+ * such as JavaScript, CSS, image, font, JSON, icon, or source map files.
+ *
+ * @param url - The URL string to check.
+ * @returns `true` if the URL points to a static asset, `false` otherwise.
+ *
+ * @example
+ * isStaticAsset("https://example.com/styles/main.css") // true
+ * isStaticAsset("https://example.com/data") // false
+ */
+export function isStaticAsset(url: string): boolean {
```
rakyi: Not every request that ends with a file extension is static. For example, our thumbnails, README files, CSVs, and other files are dynamic and could be part of an A/B test. But if we want to keep this logic, we could rename the function to something like `shouldSkipExperiments`.

functions/experiments/config.ts
```diff
@@ -0,0 +1,37 @@
...
+/*
+ * Hard-coded active experiments.
+ */
+export const experiments: Experiment[] = [
+    new Experiment({
+        /*
+         * Experiment: data-page-insight-butons-basic
+         *
+         * Control arm: no changes
+         * Control arm control1: "what you should know about this data" block is moved down the page (also the case in arms treat0 and treat1)
```
rakyi: I'm surprised we call this a control, despite changing something. Shouldn't it be the first treatment arm?

functions/experiments/middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (activeExperiments && activeExperiments.length) {
+        if (!validateUniqueExperimentIds(activeExperiments)) {
+            throw new Error(`Experiment IDs are not unique`)
+        }
+
+        for (const exp of activeExperiments) {
+            if (
+                !cookies ||
+                !Object.prototype.hasOwnProperty.call(cookies, exp.id)
+            ) {
```
rakyi: This should be sufficient:

```suggestion
           if (!(exp.id in cookies)) {
```

functions/experiments/middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+                // Check if the current request URL matches any of the experiment paths.
+                // If so, then assign the same experimental arm to all paths in the experiment
+                const requestPath = new URL(context.request.url).pathname
+                const isOnExperimentPath = exp.isUrlInPaths(requestPath)
+
+                if (isOnExperimentPath) {
+                    const assignedArm = assignToArm(exp)
+                    for (const path of exp.paths) {
+                        const value = `arm:${assignedArm.id}&sentrySampleRate:${assignedArm.replaysSessionSampleRate}`
+                        cookiesToSet.push({
```
rakyi: We don't really need a separate cookie for each path, do we? We should only need the path to decide whether we do something server-side, such as add a CSS class to the HTML. And for that we only need the experiment + arm and can look up the paths in the config.

site/DataInsightsIndexPage.tsx
```diff
@@ -20,6 +20,12 @@ export interface DataInsightsIndexPageProps {
     pageNumber: number
     totalPageCount: number
     isPreviewing?: boolean
+    topic?: TopicTag
```
rakyi: I find it useful to not use two different names for the same thing. Could this be:


```suggestion
    topicTag?: TopicTag
```

site/gdocs/pages/DataInsight.scss
```diff
@@ -252,13 +254,52 @@
...
+            }
+            .icon {
+                color: $blue-60;
+                font-size: 12px;
+            }
+            .close {
+                color: $blue-60;
+                font-size: 11px;
+                line-height: 0;
+            }
```
rakyi: These names are too generic. Our convention is to (mostly) use BEM (with fully spelled out names), so e.g. `filter-pill__name`.

site/DataPageContent.scss
rakyi: I think it's worth to talk about this on a call. 🙂

site/SentryUtils.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    try {
+        if (isConsentGranted) {
+            if (sessionStorage.getItem(SENTRY_SHOULD_RECORD_KEY) === "1") {
+                replay.start()
+            } else {
+                const isReplayRecording = !!replay.getReplayId()
+                if (!isReplayRecording) {
+                    const sampledRate = getSessionSampledRate()
+                    const p = computeSessionSampleRate(isConsentGranted)
```
rakyi: Please use a more descriptive variable name.

rakyi (commented):
Cool work! I appreciate all the thinking that went into this, that you were able to figure out all of this by yourself, and the ingenuity of implementing the first experiment.

However, I'm not sure the experiment makes sense when we degrade the DI page UX so much. I'd expect users to be less likely to click the link to DIs again if they experience a wonky flash of content. I believe we'd be better off linking to a topic page instead, since implementing the DI filtering by topic properly would be likely a lot of work.

adminSiteServer/mockSiteRouter.ts
```diff
@@ -304,28 +304,28 @@ getPlainRouteWithROTransaction(
...
+            await db.getPublishedDataInsightCount(trx, topicTag?.slug)
         )
 
-        if (topic && topic.slug !== undefined) {
+        if (topicTag && topicTag.slug !== undefined) {
             // if topic slug is not a valid topic, return NotFound page
             const validTopicSlugs = await db.getAllTopicSlugs(trx)
-            if (!validTopicSlugs.includes(topic.slug)) {
+            if (!validTopicSlugs.includes(topicTag.slug)) {
                 return res.status(404).send(renderNotFoundPage())
```
rakyi: Tried this on staging, and it doesn't seem to work. I get unfiltered data insights instead of a 404. But is 404 even the behavior we get in prod? It should ideally match, but not a big deal.

baker/GrapherBakingUtils.ts
```diff
@@ -58,6 +63,52 @@ export async function getTagToSlugMap(
     return tagsByIdAndName
 }
 
+/**
+ * Returns a map that can resolve Tag names and Tag IDs to whether the tag has any published data insights.
+ * e.g.
+ *   "Women's Rights" -> true
+ *   123 -> true
+ */
+export async function getTagToHasDataInsightsMap(
```
rakyi: Not a big deal, but we could use the native `Set` here, which would contain only tags that have data insights.

baker/GrapherBakingUtils.ts
```diff
@@ -58,6 +63,52 @@ export async function getTagToSlugMap(
...
+        WHERE t.slug IS NOT NULL
+        `
+    )
+
+    const map: Record<string | number, boolean> = {}
+    for (const row of rows) {
+        const hasDataInsight = !!row.hasDataInsight
+        if (row.slug) {
+            map[row.name] = hasDataInsight
+            map[row.id] = hasDataInsight
```
rakyi: We are only using this with names, so let's not put IDs there.

site/DataPageV2Content.tsx
```diff
@@ -100,18 +101,59 @@ export const DataPageV2Content = ({
...
+
+        const currentPath = window.location.pathname
+        const inExperiment = experimentIds.some((expId: string) => {
+            const experiment = experimentMap.get(expId)
+            return experiment?.isUrlInPaths(currentPath) ?? false
+        })
+
+        setIsPageInExperiment(inExperiment)
+    }, [assignedExperiments, experimentMap])
+
```
rakyi: There is nothing in this section that would intrinsically be a component state, therefore the code should live outside the component.

We can have essentially the same logic without the React hooks in a helper function(s) (e.g. in the utils package, close to other experiments helpers). And then run it once on the top level (which would execute when the JS file gets loaded) and assign the result to variable(s) right below the function definition. Then we could just import that variable in this file and use it directly without worrying about React state.

site/instrument.ts
rakyi, line 1: When this file is used, it must be imported first, so that the Sentry gets initialized as soon as possible. We should ideally not import anything in this file itself, therefore the contents of the `initializeSentry` should just be defined here inline. Also, all the helpers and anything else it needs should live here (and none of the code it doesn't need). We currently can't avoid importing some things, of course, e.g. the `clientSettings`.

rakyi (commented):
We are getting there. 🙂 I found a couple small issues and a few nitpicks/ideas for improvements which are not really necessary to fix (up to you). The important things are:

* Don't put the logic to check for experiments in the React component
* Put the Sentry init code into `instrument.ts`

site/instrument.ts
bnjmacdonald: I've refactored instrument.ts and SentryUtils.ts so that `Sentry.init` is now in instrument.ts. But `maybeSampleSession()` and all its dependencies are still getting imported from SentryUtils.ts (`maybeSampleSession()` must be called before Sentry.init, and is also called in CookiePreferencesManager.tsx on consent change).

site/DataPageV2Content.tsx
bnjmacdonald: I've *mostly* done this now, moving most of the logic to utils. 

If I've understood correctly, I tried what you suggested in [this commit](https://github.com/owid/owid-grapher/pull/4949/commits/929132d5ba76ebbf405884d3152149a00b7d0d4c), but ran into a problem that the client wasn't rendering components using the correct `experimentState`.

Specifically, clicking either of these buttons in arm `treat0` should scroll the user down to AboutThisData:  
Screenshot 2025-08-27 at 1 54 08 AM

But I couldn't get that to work without reintroducing `useEffect` to update experimentState on the client. Seems to be the only way I can get things to work properly, but perhaps I'm missing something here. 

Either way, at least most of the logic now lives outside the data page component.

site/instrument.ts
rakyi: OK, that's already an improvement.

site/DataPageV2Content.tsx
rakyi: I'm surprised by that, but not worth digging into further.

PR #4953, opened 2025-06-03
🐛 show error message in the admin

sophiamersmann:
I'm literally so confused by our error handling in the admin api router. `admin.requestJSON` expects `error` to have a `message` and `status`, so I added those for the charts endpoints. Not sure if we need to do this elsewhere as well?

Try updating [this chart on staging](http://staging-site-show-errors-in-admin/admin/charts/349/edit) (unless Bastian has fixed the problem by now)

### DIFF
--- a/adminSiteServer/apiRoutes/charts.ts
+++ b/adminSiteServer/apiRoutes/charts.ts
@@ -717,7 +717,7 @@ export async function createChart(
 
         return { success: true, chartId: chartId }
     } catch (err) {
-        return { success: false, error: String(err) }
+        return { success: false, error: { message: String(err), status: 500 } }
     }
 }
 
@@ -763,7 +763,7 @@ export async function updateChart(
     } catch (err) {
         return {
             success: false,
-            error: String(err),
+            error: { message: String(err), status: 500 },
         }
     }
 }


adminSiteServer/apiRoutes/charts.ts
```diff
@@ -717,7 +717,7 @@ export async function createChart(
 
         return { success: true, chartId: chartId }
     } catch (err) {
-        return { success: false, error: String(err) }
+        return { success: false, error: { message: String(err), status: 500 } }
```
ikesau, line 720: Might be better to do:

```suggestion
        return { success: false, error: new JsonError(err) }
```

It seems like an anti-pattern to me, but this is the way the admin was built 😅

ikesau (approved):

adminSiteServer/apiRoutes/charts.ts
sophiamersmann: ah, that makes much more sense!

sophiamersmann: The error message didn't make it through when I used JsonError. I just merged as is for now :)

PR #4955, opened 2025-06-04
Fix logo position in them main nav on a small screen

rakyi:
The logo currently isn't centered on small screens.

The `column-gap` polyfill doesn't work correctly when the last child is `display: hidden` which became the case at some point.

Went with this fix to be on the safe side, instead of switching to the `column-gap` property, which wouldn't work correctly for all widths by itself anyway.

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/site/SiteLogos.scss
+++ b/site/SiteLogos.scss
@@ -2,8 +2,15 @@
     display: flex;
     align-items: center;
     justify-content: space-between;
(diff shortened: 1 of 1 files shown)


mlbrgl (approved):
yes that makes sense, good catch!

Issue #4957, opened 2025-06-04
Zooming into the 2D view of a map on mobile and then clicking a country can trap the view


danyx23:
## Description

When you are on our website on mobile and zoom with a pinch gesture into our 2D map (zooming the entire website), click a country, then we switch you into the 3D view. The problem is that the globe now captures the pinch gesture to zoom only the globe in and out. This means you can't zoom out of the full website anymore. You are, in effect, trapped in this super zoomed in view and have to reload the website to get out again.

## Expected behaviour

Ideally, doing a pinch gesture on a 2D map on mobile would transition you into the 3D globe where zoom and pan is well defined, and not zoom the website as all

## Steps to reproduce

1. Go to our website on a mobile phone
2. Scroll to the first chart
3. Pinch zoom in quite a bit
4. Click a country
5. Try to get out again

## Screenshots

![Image](https://github.com/user-attachments/assets/c007df3b-c55b-4c38-85d9-2ae4c5134ae3)

## Environment

Mobile phones

ikesau:
A bandaid solution: you can read `window.visualViewport.scale` to see if the user has pinch-zoomed at all, and then prevent the switch to 3D view if so.

PR #4958, opened 2025-06-04
🐛 fix grapher 3D zoom trap

ikesau:
## Context

A "dumb" fix for https://github.com/owid/owid-grapher/issues/4957

## Video

https://github.com/user-attachments/assets/7d9a9982-120c-4cc9-b52f-0f2e418860b0

## Testing guidance

You can use either the MacOS simulator app, or your browser's phone simulator and zoom in via the track pad.

### DIFF
--- a/packages/@ourworldindata/grapher/src/mapCharts/ChoroplethMap.tsx
+++ b/packages/@ourworldindata/grapher/src/mapCharts/ChoroplethMap.tsx
@@ -329,8 +329,9 @@ export class ChoroplethMap extends React.Component<{
         if (isMapSelectionEnabled) {
             // select/deselect the country if allowed
             selection.toggleSelection(feature.id)
-        } else {
-            // rotate to the selected country on the globe on mobile
+        } else if (window?.visualViewport?.scale === 1) {
+            // rotate to the selected country on the globe
+            // but not if the user is zoomed in, otherwise they can get stuck in 3D mode
             globeController?.focusOnCountry(feature.id)
         }
     }


packages/@ourworldindata/grapher/src/mapCharts/ChoroplethGlobe.tsx
```diff
@@ -464,7 +464,11 @@ export class ChoroplethGlobe extends React.Component<{
         }
 
         // rotate to the selected country on mobile
-        if (!this.manager.isMapSelectionEnabled) {
+        if (
+            !this.manager.isMapSelectionEnabled &&
+            // don't switch to 3D if the user is zoomed in, otherwise they can get stuck in 3D mode
+            window?.visualViewport?.scale === 1
+        ) {
```
sophiamersmann: No need to do this here since we're already on the globe and clicking on a country only rotates to it

sophiamersmann (approved):
Thanks a lot, Ike! 🙏🏻

Really glad there is such a straightforward thing to do, thanks for researching!

packages/@ourworldindata/grapher/src/mapCharts/ChoroplethGlobe.tsx
ikesau: Oh yeah, you're right. I was thinking that this would help prevent people getting stuck in this case:
1. Enter globe mode
2. Scroll down page
3. Pinch zoom
4. Scroll back up
5. Click on globe

But I don't think it makes a difference. If they're zoomed in enough, they're stuck regardless 😬 

Should be pretty rare at least!

PR #4974, opened 2025-06-10
Fetch mdim FAQs from the DB only

rakyi:
Fetching metadata files for this is much slower and we can accomplish the same with using only the DB.

## Context

[Slack](https://owid.slack.com/archives/C46U9LXRR/p1749474993209939?thread_ts=1749206193.408539&cid=C46U9LXRR)

### DIFF
--- a/baker/MultiDimBaker.tsx
+++ b/baker/MultiDimBaker.tsx
@@ -8,13 +8,10 @@ import {
     MultiDimDataPageProps,
     FaqEntryKeyedByGdocIdAndFragmentId,
     MultiDimDataPageConfigEnriched,
+    PostsGdocsVariablesFaqsTableName,
+    DbPlainPostGdocVariableFaq,
 } from "@ourworldindata/types"
-import {
-    MultiDimDataPageConfig,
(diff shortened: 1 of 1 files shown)


baker/MultiDimBaker.tsx
```diff
@@ -65,61 +57,40 @@ export async function getRelevantVariableMetadata(
...
+    variableIds: Iterable<number>
+): Promise<FaqEntryKeyedByGdocIdAndFragmentId> => {
+    const faqRelations = await getFaqRelationsFromDb(knex, [...variableIds])
+    const faqGdocIds = uniq(faqRelations.map((rel) => rel.gdocId))
+    const faqGdocs = await fetchAndParseFaqs(knex, faqGdocIds, {
         isPreviewing: false,
     })
 
-    Object.values(variableMetadataDict).forEach((metadata) => {
-        const { errors: faqResolveErrors } = resolveFaqsForVariable(
```
rakyi, line 86: We don't have to do this anymore, because the consistency of the data should be guaranteed by the DB, since we are using FKs.

Issue #4781, opened 2025-04-18
Add analytics tracking of data downloads (CF function invocations)


danyx23:
## Core problem

We currently don't know how much our data is downloaded and from where. It would be good to have analytics for our (server side) data downloads. One possible approach could be to manually invoke an event in GA in our CF download code but there might be other, better solutions.

## Proposed solution

Analytics are registered as users download data in a way that is easy for us to understand. Research available options, e.g. https://developers.cloudflare.com/analytics/analytics-engine/

We also need to integrate the data into our analytics pipeline. Coordinate with Bobbie/Mojmir.

## Alternatives

Continue not to have analytics for data downloads.

## Context

One sub-question related to this is how much of our data download traffic comes from AI companies (either user sessions about data or web crawling of our downloads).

A similar concern was tracked before in https://github.com/owid/analytics/issues/176

danyx23:
Here are [some options that Google Gemini came up with](https://docs.google.com/document/d/1a7dXaUFiD1z4ll6IUgykBVm2g3R07kHgQRfWhNINZqs/edit?tab=t.0) and which all seem like reasonable leads to evaluate.

PR #4977, opened 2025-06-10
✨ add google analytics to pages functions

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/4781

## Testing guidance

For local testing, update your local `.dev.vars` file with variables from the the 1Password "Cloudflare Functions GA4 env variables" note.

Run `make up.full`

Requests to http://localhost:8788/grapher/life-expectancy.readme.md should send an event to Google Analytics with the following shape:

```js
{
  client_id: "01975bac-3447-779c-ba96-d3be88a3d2ed"
  event: {
    name: 'cf_function_invocation',
    params: {
      page_location: 'http://localhost:8788/grapher/life-expectancy.readme.md',
      page_path: '/grapher/life-expectancy.readme.md',
      referrer: null
      user_agent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:140.0) Gecko/20100101 Firefox/140.0',
      method: 'GET',
      country: null,
      status_code: 200
    }
  }
}
```

Still need to set the vars in staging so this can be tested there 👍 


functions/_common/analytics.ts
```diff
@@ -0,0 +1,133 @@
...
+export async function sendEventToGA4(
+    event: { name: string; params: Record<string, string | number> },
+    request: Request,
+    env: Env
+) {
+    if (!validateAnalyticsEnvVariables(env)) {
+        return
+    }
+    const cookies = parseCookies(request)
+    const client_id =
```
ikesau: I'm not sure this will work correctly, yet. Need to test it with real GTM cookies.

functions/_common/analytics.ts
```diff
@@ -0,0 +1,133 @@
...
+            headers: {
+                "Content-Type": "application/json",
+            },
+            body: JSON.stringify({
+                client_id,
+                events: [event],
+            }),
+        })
+        if (!response.ok) {
+            console.error(
```
rakyi: Should we capture the error also in Sentry in this case? Maybe not when fetch throws, since we probably can't do anything about a network error, but we can probably fix e.g. a 403 here?

functions/_common/analytics.ts
```diff
@@ -0,0 +1,133 @@
...
+        console.error("Error sending event to GA4 MP:", error)
+    }
+}
+
+export function getCommonEventParams(request: Request, pathname: string) {
+    const url = new URL(request.url)
+    return {
+        page_location: url.href,
+        page_path: pathname,
+        referrer: request.headers.get("referer") || "(direct)",
```
rakyi: Do we have an established convention of using `(direct)` when we don't have a referrer? Otherwise, we could just omit it or use null or empty string would also work. Same for the other missing values. Otherwise the consumer of those events needs to know the magic string we use for every missing value.

functions/_common/analytics.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+function shouldSample(env: Env): boolean {
+    return (
+        Math.random() < parseInt(env.CLOUDFLARE_GOOGLE_ANALYTICS_SAMPLING_RATE)
+    )
+}
+
+// Helper function to detect download format
+function getDownloadFormat(pathname: string): string | null {
```
rakyi: I'm not sure if we want to treat our json configs as data download. It should be mostly us, who fetch them. Have you confirmed this with Daniel/Bobbie? They are not included in the bundle when you download data from the data download. But if we want them, the we should probably treat images as "downloads" too.

And more generally, I thought we'll be logging every request (when sampled), not just downloads. Did I misunderstand?

functions/_common/analytics.ts
ikesau: Good idea! I'll just track everything and we can narrow it down if it's too noisy.

functions/_common/analytics.ts
ikesau: True, I'll just set them to null instead.

functions/_common/analytics.ts
ikesau: After chatting with Daniel, we're just gonna run analytics sampling on every route, not just downloads (so I've refactored the commit to be more agnostic)

functions/_common/cookieTools.ts
rakyi, line 1: We could use the [cookie](https://www.npmjs.com/package/cookie) package, which has tests to make sure it's RFC compliant and exposes a nice API.

rakyi (approved):
LGTM, just one more idea for a potential improvement.

functions/_common/cookieTools.ts
ikesau: Yeah, idk, I thought about alternatives - I had hoped we could use the other cookie parsing library we have, but it's specifically an express middleware.

Our usecase here is so simple, I decided to not bring in another dependency just for this.

rakyi: Ok. Btw, approve from me means it's good to merge. This was optional. 🙂

PR #5001, opened 2025-06-17
Fix spacing of logos in archive on small screen

rakyi:
There was a missing margin between the OWID logo and the affiliate logos on archive site on a small screen introduced in 0324820c2.

We can simplify the code, but the change introduces smaller margin between the OWID logo and the affiliate logos on live site on a single size, which probably wasn't even intentional, since the margin is smaller on both smaller and larger screen sizes.

| Before | After |
|--------|--------|
|  | SCR-20250617-lefo |
| SCR-20250617-leip | SCR-20250617-lelp |

### DIFF
--- a/site/SiteLogos.scss
+++ b/site/SiteLogos.scss
@@ -4,14 +4,6 @@
     justify-content: space-between;
     flex-shrink: 0;
 
-    .logo-owid {
-        margin-right: 24px;
-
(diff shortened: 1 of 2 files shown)


rakyi:
That 24 px is what we used for years. I'd rather not change it now, since it's unrelated.

PR #5016, opened 2025-06-20
Move some variable metadata queries to DB

rakyi:
We can't move all call sites to fetch the metadata from the DB, because for data pages and Algolia indexing we need the dimension entities, which we currently don't have in the DB.

Unfortunately, it doesn't seem to make a noticeable change in bake times in staging, but presumably we hit the R2 less, which is good in itself, and the code is a bit simpler.

### DIFF
--- a/db/model/Variable.ts
+++ b/db/model/Variable.ts
@@ -3,7 +3,6 @@ import { Writable } from "stream"
 import * as db from "../db.js"
 import {
     retryPromise,
-    isEmpty,
     omitUndefinedValues,
     mergeGrapherConfigs,
     diffGrapherConfigs,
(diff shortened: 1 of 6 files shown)


db/model/Variable.ts
```diff
@@ -570,6 +569,57 @@ export async function getAllChartsForIndicator(
     }))
 }
 
+export async function getDatapageIndicatorId(
+    knex: db.KnexReadonlyTransaction,
+    grapher: GrapherInterface
+): Promise<number | undefined> {
+    // If we have a single Y variable and that one has a schema version >= 2,
```
ikesau: I think this comment could be rephrased a bit in JSDoc style (so that it appears on mouseover) to explain that it will return a number ID when a grapher is datapage-valid, and undefined if the grapher isn't a datapage valid grapher.

Basically, "then we show the datapage based on this information." is a bit indirect when it comes to explaining what this function is doing.

ikesau (approved):
Looks good! Code makes sense and the pages I tested after a local bake all worked 🙂

db/model/Variable.ts
rakyi: Good point, I didn't really read the existing comments. Changed!

PR #5026, opened 2025-06-23
✨ enable comparison lines for all chart types using DualAxis

sophiamersmann:
Comparison lines used to be supported for LineCharts and ScatterPlots only. Now they're also supported for StackedArea, StackedBar and Marimekkos charts, and whenever DualAxis is used.


ikesau (approved):
Nice! Tested on all the relevant charts and they work, including for static exports 🙂 (though with some aforementioned placement issues)

PR #5027, opened 2025-06-23
🎉 add vertical comparison lines

sophiamersmann:


packages/@ourworldindata/grapher/src/comparisonLine/CustomComparisonLine.tsx
```diff
@@ -0,0 +1,149 @@
...
+                </Halo>
+            </text>
+        )
+    }
+
+    render(): React.ReactElement | null {
+        if (!this.linePath) return null
+
+        return (
+            <g id={makeIdForHumanConsumption("horizontal-comparison-line")}>
```
ikesau: I think this will violate [the HTML spec](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute) if we have 2 horizontal comparison lines. 

Personally, I'm not sure if we/grapher cares about that (i.e. if this is just for figma) but if you know that it's something we mostly manage to avoid, we should try do the same here.

packages/@ourworldindata/grapher/src/schema/grapher-schema.008.yaml
```diff
@@ -217,17 +217,26 @@ properties:
...
+                      yEquals:
+                          type: string
+                  additionalProperties: false
+                - type: object
+                  properties:
+                      label:
+                          type: string
+                      xEquals:
+                          type: number
+                  required: [xEquals]
```
ikesau, line 240: ooc, what's `required`doing here? And why does the `yEquals` object not have the same?

ikesau (approved):
Nice!

I tested the side-by-side clipping logic before even seeing that you'd written code to handle all the cases and was really impressed at how you'd handled it :)

packages/@ourworldindata/grapher/src/schema/grapher-schema.008.yaml
sophiamersmann: `yEquals` defaults to `x` if not provided (not my choice, that’s just how it works right now). For `xEquals`, defaulting to zero (or another constant) felt a bit arbitrary, so I figured it’s better to make it required. That said, I get how it might be confusing since `yEquals` behaves differently. Maybe I should add a note in the description at the very least.

packages/@ourworldindata/grapher/src/comparisonLine/CustomComparisonLine.tsx
sophiamersmann: True! I'll add the label to it so it collides less often.

These IDs are just so Figma shows a human-readable name. Unfortunately, there’s no better way to do it... so far, it hasn't caused any problems, but it's really not ideal..

PR #5031, opened 2025-06-23
replace "and" with "&" in data labels

LeeBreisacher:
## Context

Issue https://github.com/owid/owid-grapher/issues/4669. Standardization of tag names with "and" and "&" 

image

## Testing guidance

Go to https://ourworldindata.org/data and observe that ampersand (&) is used everywhere instead of the word "and".

### DIFF
--- a/site/search/DataCatalogRibbon.tsx
+++ b/site/search/DataCatalogRibbon.tsx
@@ -18,7 +18,7 @@ export const DataCatalogRibbon = ({
     selectedCountries: Region[]
 }) => {
     if (result.nbHits === 0) return null
-    const titleLabel = result.title.replace(" and ", " & ")
+    const titleLabel = result.title.replaceAll(" and ", " & ")
     const handleAddTopicClick = (e: React.MouseEvent) => {
         e.preventDefault()
         addTopic(result.title)
--- a/site/search/DataCatalogRibbon.tsx
+++ b/site/search/DataCatalogRibbon.tsx
@@ -21,7 +21,7 @@ export const DataCatalogRibbon = ({
     const titleLabel = result.title.replace(" and ", " & ")
     const handleAddTopicClick = (e: React.MouseEvent) => {
         e.preventDefault()
-        addTopic(titleLabel)
+        addTopic(result.title)
         window.scrollTo({
             top: 0,
         })
@@ -31,7 +31,7 @@ export const DataCatalogRibbon = ({
         <div className="data-catalog-ribbon col-start-2 span-cols-12 col-sm-start-2 span-sm-cols-13">
             <button
                 className="data-catalog-ribbon__header-button"
-                aria-label={`Add topic ${titleLabel} to filters`}
+                aria-label={`Add topic ${result.title} to filters`}
                 onClick={handleAddTopicClick}
             >
                 <div className="data-catalog-ribbon__header">
@@ -57,7 +57,7 @@ export const DataCatalogRibbon = ({
                                         hit,
                                         i + 1,
                                         "ribbon",
-                                        titleLabel
+                                        result.title
                                     )
                                 }}
                                 searchQueryRegionsMatches={selectedCountries}
--- a/site/search/DataCatalogRibbon.tsx
+++ b/site/search/DataCatalogRibbon.tsx
@@ -18,9 +18,10 @@ export const DataCatalogRibbon = ({
     selectedCountries: Region[]
 }) => {
     if (result.nbHits === 0) return null
+    const titleLabel = result.title.replace(" and ", " & ")
     const handleAddTopicClick = (e: React.MouseEvent) => {
         e.preventDefault()
-        addTopic(result.title)
+        addTopic(titleLabel)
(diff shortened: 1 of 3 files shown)


site/search/DataCatalogRibbon.tsx
```diff
@@ -18,9 +18,10 @@ export const DataCatalogRibbon = ({
     selectedCountries: Region[]
 }) => {
     if (result.nbHits === 0) return null
+    const titleLabel = result.title.replace(" and ", " & ")
     const handleAddTopicClick = (e: React.MouseEvent) => {
         e.preventDefault()
-        addTopic(result.title)
+        addTopic(titleLabel)
```
ikesau: This isn't quite right. This means that we'll add `X & Y` as the tag filter that we send to Algolia, when the data we've indexed to Algolia is still tagged with `X and Y`

If you test locally using the credentials Marcel provided, you can see how this leads to unintended behaviour:

https://github.com/user-attachments/assets/c3f5de41-05f9-4b0e-b5de-58918073a5a6

site/search/DataCatalogRibbon.tsx
```diff
@@ -30,11 +31,11 @@ export const DataCatalogRibbon = ({
         <div className="data-catalog-ribbon col-start-2 span-cols-12 col-sm-start-2 span-sm-cols-13">
             <button
                 className="data-catalog-ribbon__header-button"
-                aria-label={`Add topic ${result.title} to filters`}
+                aria-label={`Add topic ${titleLabel} to filters`}
```
ikesau: I think screenreaders pronounce it the same way in either case, so we don't need to update these aria-labels.

site/search/DataCatalogRibbon.tsx
```diff
@@ -56,7 +57,7 @@ export const DataCatalogRibbon = ({
                                         hit,
                                         i + 1,
                                         "ribbon",
-                                        result.title
+                                        titleLabel
```
ikesau: This is for analytics, so we should probably just keep the title the same for data continuity purposes.

site/search/TopicsRefinementList.tsx
```diff
@@ -44,16 +44,17 @@ export const TopicsRefinementList = ({
...
             >
                 {entries.map(([facetName, count], i) => {
+                    const facetNameLabel = facetName.replace(" and ", " & ")
                     const isLast = i === entries.length - 1
                     return (
                         <React.Fragment key={i}>
                             <li className="data-catalog-refinement-list__list-item">
                                 <button
-                                    aria-label={`Filter by ${facetName}`}
+                                    aria-label={`Filter by ${facetNameLabel}`}
```
ikesau: Another aria-label that we don't really need to update 🙂

site/search/DataCatalogRibbon.tsx
LeeBreisacher: Ah, right. Thanks for the good review -- I thought maybe I was going overboard a bit. I will take another shot at it. 
> Let me know if you need any help setting up your local so that you can query Algolia
Yes, I don't think I know what "query Algolia" means (you're aware I'm still quite new to all this). Marcel gave me these entries for my .env:
ALGOLIA_ID=ASCB5XMYF2
ALGOLIA_SEARCH_KEY=bafe9c4659e5657bf750a38fbee5c269
Is that what you're referring to?

site/search/DataCatalogRibbon.tsx
```diff
@@ -67,7 +68,7 @@ export const DataCatalogRibbon = ({
             </div>
             <button
                 className="data-catalog-ribbon__see-all-button"
-                aria-label={`Add ${result.title} to filters`}
+                aria-label={`Add ${titleLabel} to filters`}
```
ikesau: We also don't need this change here 🙂

site/search/DataCatalogRibbon.tsx
ikesau: Yeah, so you should have a .env file in the root of the grapher repo. If you paste those 2 lines into it and run `make up`, you should be able to go to `http://localhost:3030/data` and use the data catalog from your local instance (which will have your changes so that you can test there aren't any regressions)

site/search/DataCatalogRibbon.tsx
LeeBreisacher: arg. thanks! sorry...

site/search/DataCatalogRibbon.tsx
```diff
@@ -18,6 +18,7 @@ export const DataCatalogRibbon = ({
     selectedCountries: Region[]
 }) => {
     if (result.nbHits === 0) return null
+    const titleLabel = result.title.replace(" and ", " & ")
```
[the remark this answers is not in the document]
ikesau: I think this is unlikely to be relevant, but it's not wrong. @LeeBreisacher can you make this change where applicable, also? 🙂

ikesau (approved):
Great, thanks @LeeBreisacher ! (Sorry for the delay, I had taken some time off)

PR #5033, opened 2025-06-23
Import lodash directly

rakyi:
- Don't re-export lodash in @ourworldindata/utils
- Use `import * as _ from "lodash-es"` consistently

This change is mostly for improving readability by removing unnecessary indirection of the re-exports and inconsistency in lodash imports, and making it easier to see which functions are from lodash.

I used Claude Code to create a few scripts for automating the migration with [jscodeshift](https://github.com/facebook/jscodeshift) and removing extra newlines. Then I manually fixed the remaining edge cases.


rakyi:
The above BundleMon report seems to be a false positive. Rebase on master fixed it.

marcelgerber (approved):
Nice! Feel free to go ahead with this, before it causes too many painful conflicts.

PR #5051, opened 2025-06-25
Add yarn typecheckLerna

rakyi:
We need to run tsc separately on packages built with tsup, which only strips types. We can use Lerna to check types everywhere in parallel.


marcelgerber (approved):
Good catch! And the fact that CI is failing right now is a good sign 😁

rakyi:
Yeah, I wanted to see it fail. 😄

PR #5052, opened 2025-06-25
Always show scrollbar when DimensionDropdown overflows

rakyi:
[Slack](https://owid.slack.com/archives/C071F90MD0X/p1750847557886759)


rakyi:
Yep, thanks for the additional tests!

PR #5054, opened 2025-06-26
Use nocache query param in admin mdim preview

rakyi:
## Description

- When data managers work on mdims they should see their changes ASAP on the staging server
- Also add missing archival context params that got lost with grapher refactor 

Part of #5002 to fix the biggest pain point ASAP.

## Testing guidance

Check that all data, metadata, and configs are loaded with `nocache` query param on mdim data page admin preview.


danyx23 (approved):
Looks good! I have only read through the code and given it a very brief test - if you want a more in-depth review please ask someone else to add theirs but it seems like this isn't needed. Thanks for picking this up!

PR #5066, opened 2025-07-01
Add autoFocus to EntitySelector search input

rakyi:
As requested on [Slack](https://owid.slack.com/archives/C0149NKLZAM/p1751020560265779). Hope this naive solution is good enough.


packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -1330,6 +1330,7 @@ export class EntitySelector extends React.Component<{
                     placeholder={`Search for ${a(this.searchPlaceholderEntityType)}`}
                     trackNote="entity_selector_search"
                     onKeyDown={this.onSearchKeyDown}
+                    autoFocus
```
sophiamersmann: The `EntitySelector` component has an `autoFocus` prop that probably got lost in a recent refactor. Can you use that here so that `autoFocus` is set only when it should be?

sophiamersmann (approved):

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
rakyi: Yep, I think I got it now.

PR #5067, opened 2025-07-01
Add date to ResearchAndWritingLink by default

### LINKED ISSUE #4042: Feature request: Datestamp for article thumbnails
## Core problem
Adding a date stamp below the article thumbnail on topic pages. @mrwbkrm has a design ready for this.

For example, on [the upcoming Covid page](https://admin.owid.io/admin/gdocs/1DW5SzHglTI1FQstFTa-q6RibahuDzmKLvYOvUzFRJvU/preview), we have several articles written about updates we made, but it’s not clear when they were updated unless people click each one. 

Image

Similarly, I think it could be helpful on other topics, for people to tell which articles are newer or older.

[Discussed on Slack [here](https://owid.slack.com/archives/C0149NKLZAM/p1728382980606889)]

## Proposed solution

Marwa's design:

Image

rakyi:
Fixes #4042

I tested it locally. We agreed with Joe and Marwa that dates should be shown by default.


rakyi:
Fixed the above.

PR #5145, opened 2025-07-17
Document-database-edit

danyx23:
Edit database documentation texts

### DIFF
--- a/db/CLAUDE.md
+++ b/db/CLAUDE.md
@@ -1,0 +1,7 @@
+This folder contains the database access code for the OWID admin. The database used is MySQL 8. The docs subfolder a README.md file describing the most important tables and overall notes, and then one TABLE-NAME.md file per table describing the table in more detail. Use these docs before creating new migrations or writing queries.
+
+The migration subfolder contains the pre-migrations-schema.sql file that describes the original Mysql database schema ca 2018. Building on this state, the various \*.ts files describe the migrations since then (the filenames are prefixed by unix timestamp so can be listed alphabetically to see the correct order). To create new migrations, run the `yarn createDbMigration` which will create a new migration file.
+
+The model folder contains helper functions to query most tables (not all tables are used by the Grapher admin - some are used by auxiliary systems). The most complex models are those related to our GDocs based CMS, represented in the posts_gdocs table. The related model classes are in the model/GDoc subfolder and contain substantial business logic and a class hierarchy.
+
+The tests folder contains database tests. These can be run via `make dbtest` which spins up a test mysql database, runs the migrations and then runs the database tests.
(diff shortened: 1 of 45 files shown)


db/CLAUDE.md
```diff
@@ -0,0 +1,7 @@
+This folder contains the database access code for the OWID admin. The database used is MySQL 8. The docs subfolder a README.md file describing the most important tables and overall notes, and then one TABLE-NAME.md file per table describing the table in more detail. Use these docs before creating new migrations or writing queries.
```
lucasrodes:
```suggestion
This folder contains the database access code for the OWID admin. The database used is MySQL 8. The docs subfolder contains a README.md file describing the most important tables and overall notes, and then one TABLE-NAME.yml file per table describing the table in more detail. Use these docs before creating new migrations or writing queries.
```

db/docs/housekeeper_reviews.yml
```diff
@@ -1,12 +1,11 @@
 metadata:
-    description: "Housekeeper provides suggestions for content maintenance and quality assurance. This table tracks suggested reviews for various objects (charts, datasets, etc.) that may require maintenance, updates, or quality checks. Part of the automated housekeeping system for content management."
-
+    description: Housekeeper provides suggestions for content maintenance and quality assurance. This table tracks suggested reviews for various objects (charts, datasets, etc.) that may require maintenance, updates, or quality checks. Part of the automated housekeeping system for content management during ETL updates.
```
lucasrodes: Actually we use it independently of updates. Instead we have a bot that checks things & shared them on slack. The tool is managed in ETL.

```suggestion
    description: Housekeeper provides suggestions for content maintenance and quality assurance. This table tracks suggested reviews for various objects (charts, datasets, etc.) that may require maintenance, updates, or quality checks. Part of the automated housekeeping system for content management in ETL.
```

db/docs/narrative_charts.yml
```diff
@@ -1,21 +1,21 @@
 metadata:
-    description: "Named chart views with specific configurations for use in narrative content. These are chart variations that have been customized for specific storytelling contexts, often with particular country selections, time ranges, or visual settings that support the narrative flow of articles."
+    description: Named chart views with specific configurations for use in a narrative context like an article. These are chart variations that have been customized for specific storytelling contexts, often with particular country selections, time ranges, or visual settings that support the narrative flow of articles. Narrative charts can be based on either standalone charts or multi-dim chart views (and, probably from Q4 2025, explorer views)
```
lucasrodes, line 2: should we maybe avoid making future references? seems a bit risky to me

db/docs/posts_gdocs_links.yml
```diff
@@ -1,19 +1,19 @@
 metadata:
-    description: "Link tracking system for Google Docs posts that enables internal analytics and link health monitoring. Extracts and categorizes all links from Google Docs content to track internal references, external citations, and embedded charts/explorers."
+    description: Link tracking system for Google Docs posts. All outgoing links and references are tracked here. The target can be a things like grapher charts, explorers etc which can be joined via slug (be careful to take redirects into account when doing that!)
```
lucasrodes: minor typo

```suggestion
    description: Link tracking system for Google Docs posts. All outgoing links and references are tracked here. The target can be things like grapher charts, explorers etc which can be joined via slug (be careful to take redirects into account when doing that!)
```

Also, maybe good to clarify what the warning (last sentence in parents) actually means

db/docs/narrative_charts.yml
danyx23: I think it's fine - we're very likely to add this soon and even if not, the worst that could happen would be a query that fails once.

danyx23:
Thank you @lucasrodes! I accepted your suggestions and clarified the role of chart_configs and chart_revisions.

The foreign key annotations were indeed done with the script that I added :)

PR #5167, opened 2025-07-22
:lipstick: create yarn commands to run only changed files through prettier and lint

danyx23:
When working with coding agents, waiting for `yarn prettierAll` and/or `yarn testLint` takes quite a long time. This PR adds commands that only run these for changed files (uncommited files)


Marigold:
(Nice, we have the same in the [makefile](https://github.com/owid/etl/blob/master/default.mk#L46-L62) as `make check`)

PR #5169, opened 2025-07-22
Materialize explorer views in the DB

danyx23:
## Context

For a while now, we've been storing grapher chart configs in the chart_configs table in MySQL. The standalone Grapher charts that live in the charts table reference those. Similarly, the MDIMs reference the chart_configs table.

The one thing that was still missing was explorer views. This PR adds materializing Grapher configs for every published explorer view into the chartconfig table.

Most of the core logic of this PR was developed by @sophiamersmann, resurrected from #4836

## Testing guidance

- Run `make migrate` to apply the migrations
- Run `yarn refreshExplorerViews --stats` to insert explorer views for all existing explorers
- Inspect the db to see that things are as they should be
- Changes to explorers (either via the api or the admin) should update the views correctly

## Checklist

### Before merging

If DB migrations exists:

### After merging

- [ ] run the `yarn refreshExplorerViews` script to fill explorer_views
- [x] If a table was touched that is synced to R2, the sync script to update R2 has been run


danyx23:
@sophiamersmann I think this should be ready now. This does not upload the explorer view configs to R2 yet. I'm a bit torn on whether it should. On the one hand, we are not using the constructed chart views for rendering anywhere, so this is an arguement against pushing to R2. On the other hand, some tools might benefit form having it in R2 in the future, and it's semantically quite close to mdim chart configs that we have there as well - in a way, it may be less mental overhead to just also push explorer view chart configs to R2 as well. What do you think?

danyx23:
I also added some tooling for working with claude to this repo and made the tests less chatty

db/model/ExplorerViewsHelpers.ts
```diff
@@ -0,0 +1,279 @@
...
+
+function columnDefToDimensionObject(
+    colDef?: OwidColumnDef
+): Partial<OwidChartDimensionInterface> {
+    if (!colDef) return {}
+    return omitUndefinedValues({
+        variableId: colDef.owidVariableId,
+        slug: colDef.slug,
+        // pipe all fields of the explorer's column grammar into the display field
+        // this is not necessarilly correct; we need to add a translation step here
```
sophiamersmann: @danyx23, my implementation wasn't complete and this is one of the more major pieces missing. `OwidColumnDef` doesn't correspond to the `display` field one to one, so we need a translation step here.

The other big piece I also haven’t gotten around to is adding support for the `duplicate` transform column.

More generally, my implementation was done before the Grapher state refactor, and I ended up abandoning it because I was hoping there’d be a way to use GrapherState that would save us from having to rebuild the configs from scratch like I did here. Do you think that's worth exploring?

danyx23:
@sophiamersmann I removed most of your original code now and implemented this by instantiating an explorer instead and iterating over the views. I added a bunch of optimizations, e.g. a way to let grapher fetch only metadata since this is all we need for this case - this could also be done by supplying a custom loader - let me know if you think that would be better!

db/model/ExplorerViews.ts
```diff
@@ -0,0 +1,483 @@
...
+
+            // Extract the generated config from the explorer's grapher state
+            const config = explorer.grapherState.toObject(false)
+
+            // Grapher uses an internal fallback chain for some important properties.
+            // For explorer views we want to have a config that materializes as much
+            // as possible of what is shown, even if by default grapher would fall back
+            // to defaults (e.g. from the first Y indicator metadata), so we fill
+            // the important properties with the fallback defaults here
+            config.title = config.title ?? explorer.grapherState.defaultTitle
```
danyx23, line 205: @sophiamersmann These properties are set explicitly so that we get the value after all the fallback chains have kicked in. Are there others where this should also be done?

danyx23:
I wrote about this [briefly on slack](https://owid.slack.com/archives/CQQUA2C2U/p1753947920580319) - the explorer views upserting can take quite a while for some of the big explorers like the food explorer with 2000 views. This is then taking several minutes, and having an HTTP handler "hang" for so long is not great. I looked a bit into how best to avoid this but all of the solutions are pretty labour intensive (just having a fire and forget promise that is not awaited for filling the explorer views is problematic with the transaction scope) - so I would try and see if we can get away with this for now. In the longer term, explorers will hopefully become less important anyway. 

But please let me know if you disagree or see other issues with this!

danyx23:
On the stagings server, the views as already generated, so you can run sql queries there if you are interested in what the grapher configs look like:
```bash
ssh owid@staging-site-explore-views-in-chart-confi
yarn query "select ev.explorerSlug, explorerView, cc.full ->> '$.title' as title from explorer_views ev join chart_configs cc on ev.chartConfigId = cc.id limit 10"
```

danyx23:
* **#5169** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5239** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5239?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#5240](https://github.com/owid/owid-grapher/pull/5240) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5240?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

danyx23:
I've squashed the commits of this PR and extracted the changes to the claude instructions into a base PR

sophiamersmann:
@danyx23, it'd be convenient if you pushed to constructed configs to R2 since I'd like to try using them in search to render the data table.

db/migration/1753194645809-MigrateExplorerViewsToUseChartConfigs.ts
```diff
@@ -0,0 +1,126 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+// cSpell:ignore uuidv7
```
rakyi: Could you add the word to your local dictionary instead? Since I assume cSpell is not a tool most of us use. The comment appears multiple times in this file.

db/model/ExplorerViews.ts
```diff
@@ -0,0 +1,483 @@
...
+                WHERE v.id IN (?)
+            `,
+            [requiredVariableIds]
+        )
+
+        // check if all required variable IDs exist in the database
+        const missingIds = requiredVariableIds.filter(
+            (id: number) =>
+                !partialGrapherConfigRows.find((row) => row.id === id)
+        )
```
rakyi: This could be [_.difference](https://lodash.com/docs/4.17.15#difference), which should also be faster.

db/model/ExplorerViews.ts
```diff
@@ -0,0 +1,483 @@
...
+                      config: row.grapherConfigAdmin as string,
+                  })
+                : {}
+            const etlConfig = row.grapherConfigETL
+                ? parseGrapherConfigFromRow({
+                      id: row.id,
+                      config: row.grapherConfigETL as string,
+                  })
+                : {}
+            const mergedConfig = mergeGrapherConfigs(etlConfig, adminConfig)
```
rakyi, line 134: Is this logic correct? We pick either/or, in `getMergedGrapherConfigForVariable`.

db/model/ExplorerViews.ts
```diff
@@ -0,0 +1,483 @@
...
+            })
+        }
+    }
+
+    return generatedViews
+}
+
+// Helper function to create deterministic JSON serialization
+// by sorting object keys to ensure consistent string comparison
+function deterministicStringify(obj: any): string {
```
rakyi: We are already using the `safe-stable-stringify` package in multiple places, which we could also use here instead of this helper.

db/model/ExplorerViews.ts
```diff
@@ -0,0 +1,483 @@
...
+
+    for (const view of existingViews) {
+        try {
+            const parsedView = JSON.parse(view.explorerView)
+            const deterministicKey = deterministicStringify(parsedView)
+            existingViewsMap.set(deterministicKey, view)
+        } catch {
+            // Skip views with invalid JSON
+            console.warn(
+                `Skipping view with invalid JSON: ${view.explorerView}`
```
rakyi: Why is this a warning instead of an error? It seems that storing invalid JSON in that column should never happen/be a big deal.

packages/@ourworldindata/types/src/dbTypes/ExplorerViews.ts
```diff
@@ -0,0 +1,39 @@
+import { JsonString } from "../domainTypes/Various"
+
+export const ExplorerViewsTableName = "explorer_views"
+export interface DbInsertExplorerView {
+    id?: number
+    explorerSlug: string
+    explorerView: JsonString
```
rakyi: I think we could improve the naming. Currently, it's possible to have `explorerView.explorerView`, which is confusing. Could we call this field `explorerDimensions` or maybe just `dimensions`? We use the word "dimensions" in mdim config.

packages/@ourworldindata/types/src/dbTypes/ExplorerViews.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export type DbRawExplorerView = Required<
+    Omit<DbInsertExplorerView, "chartConfigId" | "error">
+> & {
+    chartConfigId: string | null
+    error: string | null
+}
+
+export type DbEnrichedExplorerView = Omit<DbRawExplorerView, "explorerView"> & {
+    explorerView: Record<string, string>
```
rakyi: This could be null when there is error, right?

db/model/ExplorerViews.ts
```diff
@@ -0,0 +1,483 @@
...
+        await knex("explorer_views")
+            .where("explorerSlug", slug)
+            .whereIn("explorerView", removedViewStrings)
+            .delete()
+    }
+
+    // Update existing views with changed configs
+    for (const { existing, generated } of updatedViews) {
+        if (generated.error) {
+            // Update to error state, remove chart config reference
```
rakyi, line 432: And delete the chart config itself as well?

devTools/refreshExplorerViews.ts
```diff
@@ -0,0 +1,296 @@
...
+    return db.knexRaw(knex, query)
+}
+
+async function prepareGrapherConfigsForExplorerViews(
+    knex: db.KnexReadWriteTransaction,
+    slugWildcard?: string
+): Promise<ExplorerProcessingStats[]> {
+    // Fetch published explorers and existing explorer slugs
+    const publishedExplorers = await fetchPublishedExplorers(knex, slugWildcard)
+    const existingExplorerRows = await db.knexRaw<{ explorerSlug: string }>(
```
rakyi: Could this be `existingExplorerViewRows` - and similarly also the variable names that derive from this? I find it quite confusing that we call it just explorer rows.

devTools/refreshExplorerViews.ts
```diff
@@ -0,0 +1,296 @@
...
+}
+
+function printStatsTable(stats: ExplorerProcessingStats[]): void {
+    const newExplorers = stats.filter((s) => s.isNew)
+    const existingExplorers = stats.filter((s) => !s.isNew)
+
+    console.log("\n" + "=".repeat(100))
+    console.log("Explorer Processing Statistics (Optimized)")
+    console.log("=".repeat(100))
+    console.log(
```
rakyi: Maybe we could use [`console.table()`](https://developer.mozilla.org/en-US/docs/Web/API/console/table_static) here (and below)?

devTools/refreshExplorerViews.ts
```diff
@@ -0,0 +1,296 @@
...
+
+Examples:
+  yarn tsx devTools/refreshExplorerViews.ts
+  yarn tsx devTools/refreshExplorerViews.ts --stats
+  yarn tsx devTools/refreshExplorerViews.ts --slug="energy%" --stats
+  yarn tsx devTools/refreshExplorerViews.ts --slug="energy-mix"
+`)
+}
+
+const main = async (): Promise<void> => {
```
rakyi: Not sure if it's worth rewriting now, but we are using yargs in other scripts, see e.g. `baker/archival/archiveChangedPages.ts`.

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -477,7 +485,7 @@ export class Explorer
 
         const previousTab = this.grapherState.activeTab
 
-        this.updateGrapherFromExplorer()
+        void this.updateGrapherFromExplorer()
```
rakyi: I'm surprised we continue with updating the grapher state below without awaiting this call. Are you sure it is safe? Also other uses of `void`.

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -531,17 +539,14 @@ export class Explorer
         this.explorerProgram.constructTable(slug)
     )
 
-    @action.bound updateGrapherFromExplorer() {
+    @action.bound async updateGrapherFromExplorer() {
```
rakyi, line 546: This was [discussed before](https://github.com/owid/owid-grapher/pull/5203#discussion_r2242428718), and it seemed risky to make it async. Any new insights since the previous discussion, @sophiamersmann?

packages/@ourworldindata/explorer/src/GrapherGrammar.ts
```diff
@@ -306,28 +338,43 @@ export const GrapherGrammar: Grammar<GrapherCellDef> = {
...
 } as const
+
+const omitEmptyStringValues = <T extends Record<string, any>>(obj: T) =>
+    R.omitBy(obj, (parsedValue) => parsedValue === "")
+
+const omitEmptyObjectValues = <T extends Record<string, any>>(obj: T) =>
+    R.omitBy(
+        obj,
+        (parsedValue) => R.isPlainObject(parsedValue) && R.isEmpty(parsedValue)
+    )
```
rakyi: Tests for these helpers would be nice.

adminSiteServer/app.test.ts
```diff
@@ -1338,3 +1339,598 @@ describe("OwidAdminApp: validateChartSlug", { timeout: 10000 }, async () => {
...
+        })
+        expect(nutrientAdequateView).toBeTruthy()
+
+        // Chart configs should have been created for any new successful views
+        const finalChartConfigsCount = await getChartConfigsCount()
+        // Should have original 3 charts plus configs for successful views
+        expect(finalChartConfigsCount).toBeGreaterThanOrEqual(3)
+    })
+
+    it("should handle foreign key relationships correctly", async () => {
```
rakyi: I'm not sure this test adds anything beyond what we test in other tests. Does it?

adminSiteServer/app.test.ts
rakyi, line 1: Nice tests!

packages/@ourworldindata/explorer/src/Explorer.tsx
sophiamersmann: It's on my to do list but I haven't looked into it yet, so no new insights from my ~site~ side.

db/model/ExplorerViews.ts
danyx23: I thought so - but @sophiamersmann I was surprised to see that `getMergedGrapherConfigForVariable` does this differently. Is there a reason for this? I thougth we want to merge the layers and use etl first, then add admin on top?

sophiamersmann: Depends if you're dealing with patch or full configs. In `getMergedGrapherConfigForVariable`, we take either or because the admin's full (!) config already includes the etl's config. If I see correctly, here you're dealing with patch configs, so merging them is what we want.

danyx23: That makes sense - thanks for clarifying, Sophia!

db/model/ExplorerViews.ts
danyx23: Yeah it basically should not happen. I've changed this to logging via sentry and throwing again so we stop processing at this point. Maybe at some point in the future we'll want to relax this again but better to fail hard for now 🙌

packages/@ourworldindata/explorer/src/Explorer.tsx
danyx23: I changed a bunch of things in how explorers await things in this PR. Let's try to get this right on this PR here :).

packages/@ourworldindata/explorer/src/Explorer.tsx
danyx23: Yeah this one should be awaited. The remaining voids are all in tail position and should be fine

db/model/ExplorerViews.ts
```diff
@@ -17,8 +17,9 @@ import {
 import { transformExplorerProgramToResolveCatalogPaths } from "./ExplorerCatalogResolver.js"
 import { insertChartConfig, updateExistingConfigPair } from "./ChartConfigs.js"
 import { uuidv7 } from "uuidv7"
-import { isEqual } from "lodash-es"
+import { isEqual, difference } from "lodash-es"
```
rakyi: Noticed this only now: I think it's better if we stick to namespaced imports (since we also use Remeda, that has the same/similarly named functions) and do `import * as _ from "lodash-es"`.

db/model/ExplorerViews.ts
```diff
@@ -292,14 +290,21 @@ export async function refreshExplorerViewsForSlug(
 
     for (const view of existingViews) {
         try {
-            const parsedView = JSON.parse(view.explorerView)
-            const deterministicKey = deterministicStringify(parsedView)
+            const parsedView = JSON.parse(view.dimensions)
+            const deterministicKey = stringify(parsedView as object)
```
rakyi: Here and everywhere else where we use `JSON.parse` with `strigify` it would be nicer to use the actual type:

```suggestion
            const parsedView: Record<string, string> = JSON.parse(
                view.dimensions
            )
            const deterministicKey = stringify(parsedView)
```

db/model/ExplorerViews.ts
```diff
@@ -292,14 +290,21 @@ export async function refreshExplorerViewsForSlug(
...
+            void logErrorAndMaybeCaptureInSentry(
+                new Error(
+                    `Explorer view contains invalid JSON for explorer ${slug}: ${view.dimensions}`
+                )
             )
+            if (ex instanceof Error) {
+                throw new Error(
+                    `Failed to parse explorer view dimensions for explorer ${slug}: ${ex.message}`
+                )
+            }
```
rakyi: We should always re-throw here, not only when it's an Error. And since we already log the error above, it should be fine to just:

```suggestion
            throw ex
```

db/model/ExplorerViews.ts
```diff
@@ -292,14 +290,21 @@ export async function refreshExplorerViewsForSlug(
...
             existingViewsMap.set(deterministicKey, view)
-        } catch {
-            // Skip views with invalid JSON
-            console.warn(
-                `Skipping view with invalid JSON: ${view.explorerView}`
+        } catch (ex) {
+            // Skip views with invalid JSON - this indicates a data integrity issue
+            void logErrorAndMaybeCaptureInSentry(
+                new Error(
+                    `Explorer view contains invalid JSON for explorer ${slug}: ${view.dimensions}`
```
rakyi:
```suggestion
                    `Explorer view contains invalid JSON for explorer ${slug}: ${view.dimensions}`,
                    { cause: ex }
```

functions/explorers/[slug].ts
```diff
@@ -78,7 +78,7 @@ async function handleThumbnailRequest(
             bounds
         )
         const explorer = new Explorer(explorerProps)
-        explorer.updateGrapherFromExplorer()
+        await explorer.updateGrapherFromExplorer()
         while (!explorer.grapherState.isReady) {
             await new Promise((resolve) => setTimeout(resolve, 100))
         }
```
Comment (rakyi) [inline functions/explorers/[slug].ts]: Can we remove the while loop now that we `await` the update? Or is it not guaranteed to be ready afterwards?

cspell.json
rakyi, line 1: I meant that you'd add it to your local dictionary (supposedly global) without committing it into the repo.

danyx23: I know that some of use this, which is why I think a shared file makes sense - but I moved it now to the .vscode subfolder where it is less in the way. I hope that is ok with you!

db/model/ExplorerViews.ts
danyx23: Yeah nice, this is better

cspell.json
rakyi: Sure, that makes sense!

Issue #4927, opened 2025-05-26
A component that can control Grapher state via text links


ikesau:
We want to make a new component (name TBD) that will allow authors to write about an embedded grapher chart with inline CTAs that update the Grapher.

e.g. "And as we can see, **global energy consumption** has increased steadily over the last 40 years"

This would probably take the form of something like

```
{.name-tbd}
default-view: https://ourworldindata.org/grapher/per-capita-energy-use

[.+content]
Energy use throughout the world remains highly uneven, though the [global trend](https://ourworldindata.org/grapher/per-capita-energy-use?tab=chart&country=~OWID_WRL) is increasing.
[]
{}
```

(except written with gdocs wysiwyg, not markdown)

ikesau:
https://github.com/owid/owid-grapher/pull/5183

PR #5183, opened 2025-07-24
🎉 Guided Chart component

ikesau:
## Context

https://github.com/owid/owid-grapher/issues/4927

## Testing guidance

[You can use this testing gdoc](https://docs.google.com/document/d/16bKYpPJdB_BP1Y3-qpoSmbQiRTdhWjTAt59xtvyWSNU/edit?tab=t.0)

### Before merging

If DB migrations exists:

### After merging


site/gdocs/components/SpanElement.tsx
```diff
@@ -11,6 +15,12 @@ export default function SpanElement({
     span: Span
     shouldRenderLinks?: boolean
 }): React.ReactElement {
+    const handleGuidedChartLinkClick = useGuidedChartLinkHandler()
+    const [hasHydrated, setHasHydrated] = React.useState(false)
```
ikesau, line 19: This is so that in the no-JS case, these will show and work as regular links

packages/@ourworldindata/grapher/src/chart/GuidedChartUtils.tsx
```diff
@@ -0,0 +1,48 @@
...
+    React.createContext<GuidedChartContextValue | null>(null)
+
+/**
+ * If called within a `GuidedChartContext`, sets the context's `grapherStateRef`
+ * to a new `GrapherState` instance initialized with the provided config.
+ * If no context is available, returns a local ref initialized with the config.
+ * This is so the `GrapherState` can be controlled from a GuidedChart,
+ * but also allows for local usage when not
+ */
+export function useMaybeGlobalGrapherStateRef(
```
ikesau, line 27: This is used in `site/MultiDim.tsx` and `Grapher.tsx`

Instead of calls to `useRef`, they now call this, which will instantiate and attach the `GrapherState` to the `GuidedChart` ref instead of their local one, if the `GuidedChartContext` exists.

Otherwise, it's same as before.

site/AriaAnnouncer.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react"
```
ikesau, line 1: This was originally implemented as an ancillary feature in the `GuidedChart.tsx` component, but it messed up some of the grid margin override rules to have a random div in the main `ArticleBlocks` node, so I lifted it up into its own thing that we can reuse.

site/gdocs/components/layout.ts
```diff
@@ -97,31 +98,19 @@ const layouts: { [key in Container]: Layouts} = {
         ["socials"]: "span-cols-3 col-sm-start-2 span-sm-cols-12",
     },
     ["sticky-right-left-column"]: {
-        ["chart"]: "span-cols-5 col-start-1 span-md-cols-10 col-md-start-2 span-sm-cols-12 col-sm-start-1",
```
ikesau, line 100: As best as I could tell, these weren't doing anything. I reduced them all to a simplified default because on sm and md we want everything to stack vertically anyway.

site/css/grid-margin-overrides.scss
rakyi, line 1: I didn't know this file existed! I might have put some stuff that should have been here elsewhere. Will try to keep it in mind. 🙂

packages/@ourworldindata/grapher/src/chart/GuidedChartUtils.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react"
```
rakyi: React docs say ES module style imports are preferred:


```suggestion
iimport * as React from "react"
```

site/gdocs/components/GuidedChart.tsx
```diff
@@ -0,0 +1,98 @@
...
+    d,
+    containerType = "default",
+}: {
+    d: EnrichedBlockGuidedChart
+    containerType?: Container
+    className?: string
+}) {
+    const stateRef = useRef<GrapherState | null>(null)
+    const chartRef = useRef<HTMLDivElement | null>(null)
+    const multiDimRef = useRef<{
```
rakyi: Does this really need to be a ref instead of state? It seems like it doesn't.

packages/@ourworldindata/grapher/src/chart/GuidedChartUtils.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { MultiDimDataPageConfig } from "@ourworldindata/utils"
+import { MultiDimDimensionChoices } from "@ourworldindata/types"
+
+export interface GuidedChartContextValue {
+    grapherStateRef: React.RefObject<GrapherState>
+    chartRef?: React.RefObject<HTMLDivElement>
+    onGuidedChartLinkClick?: (href: string) => void
+    registerMultiDim?: (registrationData: {
+        config: MultiDimDataPageConfig
+        updater: (newSettings: MultiDimDimensionChoices) => void
```
rakyi: A convention-following callback name would be `onSettingsChange`. Could we use that?

site/multiDim/MultiDim.tsx
```diff
@@ -75,6 +82,19 @@ export default function MultiDim({
...
     })
 
+    // Register with GuidedChartContext for guided chart link support
+    const guidedChartContext = useContext(GuidedChartContext)
+    const hasRegistered = useRef(false)
+    useEffect(() => {
+        if (guidedChartContext?.registerMultiDim && !hasRegistered.current) {
+            guidedChartContext.registerMultiDim({
+                config,
+                updater: setSettings,
```
rakyi: It seems natural to pass `handleSettingsChange`. Could we do that? Then we wouldn't need to duplicate the `filterToAvailableChoices` logic. And it seems natural that the logic would live in this component.

rakyi (approved):
I had a few ideas for improvements, but nothing blocking. Nice work! 🙂

PR #5188, opened 2025-07-25
Fix mdim data page share menu

rakyi:
[Slack](https://owid.slack.com/archives/C01GAN73R5W/p1753434769858419)

This doesn't fix the gdoc embeds, will continue with that next.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/Grapher.tsx
+++ b/packages/@ourworldindata/grapher/src/core/Grapher.tsx
@@ -1148,10 +1148,12 @@ export class GrapherState {
     }
 
     @computed get editUrl(): string | undefined {
-        if (this.showAdminControls && this.adminBaseUrl) {
-            return `${this.adminBaseUrl}/admin/${
-                this.manager?.adminEditPath ?? `charts/${this.id}/edit`
-            }`
+        let editPath = this.manager?.adminEditPath
+        if (!editPath && this.id) {
+            editPath = `charts/${this.id}/edit`
+        }
+        if (this.showAdminControls && this.adminBaseUrl && editPath) {
+            return `${this.adminBaseUrl}/admin/${editPath}`
         }
         return undefined
     }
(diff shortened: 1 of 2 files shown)


site/multiDim/MultiDimDataPageContent.tsx
```diff
@@ -217,9 +220,6 @@ export function DataPageContent({
                             ...grapherConfig.value,
                             ...baseGrapherConfig,
                         }
-                        if (config.manager)
-                            config.manager.adminEditPath = adminEditPath
-                        else config.manager = { adminEditPath }
```
rakyi, line 222: We were overriding the manager here. I'm guessing this must be a conflict resolution artifact? I think neither the config we fetch nor the base config will ever have a manager set, so it should be safe to ignore it. Could you double-check that, pls?

packages/@ourworldindata/grapher/src/core/Grapher.tsx
```diff
@@ -1148,10 +1148,12 @@ export class GrapherState {
...
-        if (this.showAdminControls && this.adminBaseUrl) {
-            return `${this.adminBaseUrl}/admin/${
-                this.manager?.adminEditPath ?? `charts/${this.id}/edit`
-            }`
+        let editPath = this.manager?.adminEditPath
+        if (!editPath && this.id) {
+            editPath = `charts/${this.id}/edit`
+        }
+        if (this.showAdminControls && this.adminBaseUrl && editPath) {
+            return `${this.adminBaseUrl}/admin/${editPath}`
```
rakyi, line 1156: Fixes a bug where we might end up with `undefined` in the edit URL when an mdim view has more than one indicator.

rakyi:
I just discovered the grapher in admin hits an infinite loop when you click "Edit" on an mdim view. Will fix that too in a follow-up.

site/multiDim/MultiDimDataPageContent.tsx
danyx23: Yes, I think this must be an artifact, I can't think of a case where this would be intended behaviour

danyx23 (approved):
This looks good, but I just saw that the iframe embed code that is shown for the non-archive version and the share link url that is copied is missing the view parameters in the query string. Doesn't have to be in this PR but probably good to also fix soon.

rakyi:
A barrage of bugs. 😢 All shall be fixed!

PR #5196, opened 2025-07-28
Fix hydration error in Table

rakyi:
Browser adds `<tbody>` implicitly when it's missing, which breaks the hydration, and React raises error in dev. It's possible React didn't do the validation in older versions.


ikesau (approved):
Changes look good, tested with all table types and it works with no hydration error 👍

PR #5198, opened 2025-07-28
medium chart hit component

sophiamersmann:
## Context

Links to issues, Figma, Slack, and a technical introduction to the work.

Add if relevant, i.e. might not be necessary when there are no UI changes.

## Testing guidance

## Checklist

### Before merging

If DB migrations exists:

### After merging


sophiamersmann:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5198?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5247** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5247?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5246** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5246?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5245** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5245?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5259** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5204** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5204?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5203** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5198** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4999** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4999?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 39 other dependent PRs ([#5071](https://github.com/owid/owid-grapher/pull/5071) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5071?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5074](https://github.com/owid/owid-grapher/pull/5074) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5074?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5083](https://github.com/owid/owid-grapher/pull/5083) <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5083?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 36 others)
* **#4867** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/4867?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5005** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5005?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

functions/_common/downloadFunctions.ts
```diff
@@ -245,124 +237,7 @@ export async function fetchDataValuesForGrapher(
...
-    const result = omitUndefinedValues({
-        entityName,
-        startTime: grapherState.startTime,
-        endTime: grapherState.endTime,
-        columns: makeColumnInfoForRelevantSlugs(grapherState),
-        startTimeValues: makeDimensionValuesForTime(grapherState, startTime),
-        endTimeValues: makeDimensionValuesForTime(grapherState, endTime),
-        source: grapherState.sourcesLine,
-    })
+    const result = constructGrapherValuesJson(grapherState, entityName)
```
sophiamersmann, line 240: Moved elsewhere so I can reuse the logic on the client

packages/@ourworldindata/grapher/src/GrapherValuesJsonHelpers.ts
```diff
@@ -0,0 +1,173 @@
+import * as _ from "lodash-es"
```
sophiamersmann, line 1: Moved the prep work for the `.values.json` endpoint here because I needed the `GrapherState` type

site/search/SearchChartHitDataDisplay.tsx
```diff
@@ -20,12 +22,10 @@ export function SearchChartHitDataDisplay({
     startValue,
     trend,
     showLocationIcon,
+    className,
 }: SearchChartHitDataDisplayProps): React.ReactElement | null {
-    // Remove parentheses from the beginning and end of the unit
-    const strippedUnit = unit?.replace(/(^\(|\)$)/g, "")
```
sophiamersmann, line 25: Done elsewhere

site/search/SearchChartHitMedium.tsx
```diff
@@ -1,219 +1,705 @@
...
 
-    const fullQueryParams =
-        isExplorerView || isMultiDimView
-            ? hit.queryParams! + entityQueryStr.replace("?", "&")
-            : entityQueryStr
+    // The grapherState contains MobX observables. When observable values are modified,
+    // MobX automatically triggers reactions and recomputes derived values. To improve
+    // performance, we wrap all state modifications in runInAction() so that MobX
+    // batches all changes and only runs expensive computations (like transformTable)
+    // once at the end, rather than after each individual state change.
```
sophiamersmann, line 138: This is what made switching between Data and Writing slow

functions/grapher/[slug].ts
Comment (rakyi) [inline functions/grapher/[slug].ts:1]: I understand why you made this change, and maybe you already thought about this, but here's an idea.

Every time we fetch mdim view config from the client only using the slug + query params (as we do for the thumbnails), we have to make three requests in the CF function (for each view, so 3N requests):

1. Get a 404 for normal grapher config
2. Fetch mdim config
3. Fetch mdim view config

A different approach is possible with N + 1 requests per N views in the results (+ 1 is the mdim config):

1. Fetch the mdim config by slug on the client directly (since we know it's an mdim, contrary to the thumbnail request) - we have a function for it already (`functions/multi-dim/[slug].json.ts`) - and cache it (using react-query). That's only one request, even when we have many views of that mdim in the results.
2. Fetch the mdim view config by UUID we get from the mdim config.

We do this for normal mdim page and embeds.

Now, I didn't measure anything and I don't know how big of a deal it is in practice, but it seems it could be both faster and make us pay less for R2 requests. And unless you really want to, we don't have to change it in this PR. But I wanted to write it up for future reference.

An important question is e.g. how many views per mdim do we (usually) show in the results?

packages/@ourworldindata/grapher/src/chart/ChartTabs.ts
```diff
@@ -101,13 +101,19 @@ export function mapChartTypeNameToQueryParam(
...
         .with(GRAPHER_TAB_NAMES.Table, () => GRAPHER_TAB_CONFIG_OPTIONS.table)
         .with(GRAPHER_TAB_NAMES.WorldMap, () => GRAPHER_TAB_CONFIG_OPTIONS.map)
-        .otherwise((tabName) => mapChartTypeNameToQueryParam(tabName))
+        .otherwise((tabName) => mapChartTypeNameToTabConfigOption(tabName))
+}
+
+export function mapGrapherTabNameToQueryParam(
+    tabName: GrapherTabName
+): GrapherTabQueryParam {
+    return mapGrapherTabNameToConfigOption(tabName)
```
rakyi: Do we really need the alias? If so, it could be:

```suggestion
export const mapGrapherTabNameToQueryParam = mapGrapherTabNameToConfigOption
```

site/search/SearchChartHitDataTable.tsx
```diff
@@ -0,0 +1,102 @@
...
+        time !== undefined && rows.every((row) => row.time === time)
+
+    // Hide the time in each row if it's shown in the title
+    const displayRows = shouldShowTimeInTitle
+        ? rows.map((item) => R.omit(item, ["time"]))
+        : rows
+
+    return (
+        <div className="search-chart-hit-table">
+            <div className="search-chart-hit-table-content">
```
rakyi, line 35: I didn't check with VoiceOver, but this is likely less accessible than a real table. Consider switching to a real table - maybe in a follow-up PR?

site/search/SearchChartHitMedium.scss
```diff
@@ -1,175 +1,127 @@
...
+    .search-chart-hit-table {
+        height: 100%;
+        width: 100%;
+        background: #fff;
+        padding: 6px 12px;
 
-        &::after {
-            content: ", ";
+        &.single-slot {
+            aspect-ratio: calc(
```
rakyi, line 42: I had a use case for aspect-ratio as well, but it's not supported in Safari 14.1, which [we say we support](https://github.com/owid/owid-grapher/blob/f07218e3bf2ecec9dcd38c4c747c9750e17f2c22/docs/browser-support.md#L1). So I used [an older technique](https://github.com/owid/owid-grapher/blob/78af6f56f5b51fe542481aa22d0a6c14fa48a260/site/search/SearchDataInsightHit.scss#L14-L20). We should use that here (and below) too, I think.

site/search/queries.ts
```diff
@@ -64,6 +64,8 @@ export const searchQueryKeys = {
 export const chartHitQueryKeys = {
     chartInfo: (slug: string, entities: string[]) =>
         ["chart-info", slug, entities] as const,
+    chartConfig: (slug: string) => ["chart-config", slug] as const,
+    chartData: (slug: string) => ["chart-data", slug] as const,
```
rakyi: So chart info and config are really specific to a slug, right? Data and metadata are reused between charts. So ideally we'd cache on the level of indicator and used indicator ID in the key. Do you have a sense of how much could the indicator requests be repeated in the search results? It's not too bad anyway, since they are already cached in the browser, but it would be even nicer if we cached them on the app level as well.

Doesn't have to be changed in this PR, but another thing to consider. Maybe add a code comment explaining the trade-off.

site/search/SearchChartHitMedium.tsx
```diff
@@ -1,219 +1,705 @@
...
-                                    )}
-                                </div>
-                                <div className="search-chart-hit-medium__thumbnail-label">
-                                    {chartType.label}
-                                </div>
-                            </a>
+                    // Single-time line charts are rendered as bar charts
+                    // by Grapher. Adjusting the time param makes sure
+                    // Grapher actually shows a line chart. This is important
+                    // since we offer separate links for going to the line
```
rakyi, line 329:
```suggestion
                    // since we often separate links for going to the line
```

site/search/SearchChartHitMedium.tsx
```diff
@@ -1,219 +1,705 @@
...
+        placedTabs,
+        dataTableProps,
+    }: {
+        placedTabs?: { tab: GrapherTabName; slot: GridSlot }[]
+        dataTableProps?: SearchChartHitDataTableProps
+    }
+) {
+    if (!placedTabs || !dataTableProps) return
+
+    // Check how much many rows are available for the table and update
```
rakyi:
```suggestion
    // Check how many rows are available for the table and update
```

rakyi (commented):
Nice work!

I didn't fully understand all the Grapher parts, since I'm missing a lot of related context (e.g. how faceting works), but the new code looks good. You could ask Marcel to review it as well, but I think it's probably fine. 🙂

I left a bunch of comments, but the only issue that really needs addressing is the `aspect-ratio` thing.

functions/grapher/[slug].ts
Comment (sophiamersmann) [reply functions/grapher/[slug].ts:1]: That makes sense to me!

Should I revert the changes in this file then?

packages/@ourworldindata/grapher/src/chart/ChartTabs.ts
sophiamersmann: Done

site/search/queries.ts
sophiamersmann: I just realised the chartConfig should include view params in case of mdims.

sophiamersmann: I think color and size indicators (like population and continents, etc.) are probably repeated quite a lot.

site/search/SearchChartHitDataTable.tsx
sophiamersmann: I considered using a table but then switched to grid/flex because it made it much easier to make the layout work.

site/search/SearchChartHitMedium.scss
sophiamersmann: Thanks for the pointer! I noticed it wasn't working but didn't know how to fix it.

site/search/SearchChartHitMedium.tsx
sophiamersmann: This is meant to be 'offer'

sophiamersmann:
Thanks @rakyi! I added smaller corrections to this PR. The rest I'll address in follow-up PRs (including the aspect-ratio issue)

functions/grapher/[slug].ts
Comment (rakyi) [reply functions/grapher/[slug].ts:1]: If we are going with the other approach I suggested (fetch the mdim config on the client), then yes, I think we should revert them.

site/search/SearchChartHitMedium.tsx
rakyi: Haha, I somehow didn't parse it as a sentence with "offer". 😄 Fair enough.

site/search/SearchChartHitMedium.scss
rakyi: Curious to hear how are you testing old browsers.

sophiamersmann:
Folded into `search-tpl`

PR #5268, opened 2025-08-12
Add sortable narrative and references count columns to admin charts list

danyx23:
Introduce new sortable columns for narrative charts and references count in the admin charts list

Adds new indices to make the references joins fast enough. 

The references excludes wordpress pages since they don't really matter anymore and make the query slow.


db/migration/1755013960275-AddPostsGdocsLinksTargetIndexes.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class AddPostsGdocsLinksTargetIndexes1755013960275
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        // Add index on target column (with prefix length for TEXT column) for improved JOIN performance
+        await queryRunner.query(`
+            CREATE INDEX idx_posts_gdocs_links_target 
+            ON posts_gdocs_links(target(100))
+        `)
```
marcelgerber: Personally, I would only add the second (composite) index below, and not add this one here.

The composite index can always also be used for queries that benefit from an index on `target(100)`.

marcelgerber (changes requested):
Hm, this PR doesn't work at all for the dataset page's chart list (e.g. [this one](http://staging-site-better-charts-admin-table/admin/datasets/6835#:~:text=Republish%20all%20charts-,Charts,-Showing%2010%20of)), which is what Pablo initially requested.

The code for this request is somewhere else: https://github.com/owid/owid-grapher/blob/070dcf0b7636953e428ce30f1f278d87221847b3/adminSiteServer/apiRoutes/datasets.ts#L191-L208

db/migration/1755013960275-AddPostsGdocsLinksTargetIndexes.ts
danyx23: Excellent point!

danyx23:
Ah, great catch! I always forget about these secondary views. I added it now to all relevant places. Thanks!

marcelgerber (approved):
Nice! this works well now.

The only thing I'm wondering now is that we now have a highly complex, mostly-duplicated query in 4 different places.
Could a MySQL view be a good alternative here? That we can then rely on in all of these 4 places, and just combine them with the relevant filters/joins for each use case?

danyx23:
@marcelgerber claude says yes :)

danyx23:
* **#5268** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5268?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

db/migration/1755647662664-CreateChartReferencesView.ts
```diff
@@ -0,0 +1,55 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class CreateChartReferencesView1755647662664 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            CREATE VIEW chart_references_view AS
```
marcelgerber, line 8: not entirely sure about our MySQL naming conventions: should this end in `_view`?

danyx23: We haven't done this much in the past but I think it's helpful to be aware of potential performance implications.

PR #5271, opened 2025-08-13
✨ Sort user location to top in entity selector

### LINKED ISSUE #5180: Entity selector: Always sort current location to the top
## Core problem

The current location shows up at the top when sorting by name, but not when sorting by value. Now that sorting by value is the default, the current location isn’t as prominent as it used to be.

## Proposed solution

Always pin the current location to the top.

danyx23:
## Summary

Implements user location sorting to the top in Grapher's entity selector for **all** sorting modes and entity filters, not just when sorting by name with countries filter.

Closes #5180

### Problem
The entity selector already detected user location and displayed it with a location icon, but only sorted it to the top when:
- ✅ Sorting by name + Countries filter  
- ❌ Sorting by name + Other filters (continents, income groups, etc.)
- ❌ Sorting by numerical columns (population, GDP per capita, etc.)

### Solution
Updated `EntitySelector.tsx` to **always** prioritize user location:

1. **Universal local sorting**: Modified filtering logic to enable `sortLocalsAndWorldToTop: true` for all entity filters
2. **Enhanced numerical sorting**: Extended numerical column sorting to respect local entity prioritization while maintaining proper numerical ordering within each group

### Technical Implementation
- **Location**: `packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx`
- **Key changes**:
  - Line 1043: Always enable local sorting for countries filter  
  - Line 1031: Enable local sorting for "all" entities filter
  - Line 1096-1108: Enhanced numerical sorting with local entity prioritization
- **Preserved behavior**: 
  - Entities without data values still sorted alphabetically at the end
  - Local entity ordering respects user's geographic hierarchy (country, then regions)
  - Location icon continues to display for user's entities

## Test Plan

- [x] All existing EntitySelector tests pass
- [x] All Grapher package tests pass (478/478)  
- [x] TypeScript compilation successful

## Expected User Experience

Users will now see their location consistently sorted to the top in the entity selector regardless of:
- **Sort column**: Name, Population, GDP per capita, or any other indicator
- **Entity filter**: All entities, Countries only, Continents, Income groups, etc.

The user's country/region will always appear first with the familiar location arrow icon (📍), providing a consistent and intuitive experience.


packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -1049,69 +1052,71 @@ export class EntitySelector extends React.Component<EntitySelectorProps> {
...
+        // Value sorting: missing values go last
+        const [withValues, withoutValues] = _.partition(entities, (e) =>
+            isFiniteWithGuard(byValue(e))
+        )
 
-            const [localEntities, otherEntities] = _.partition(
-                entitiesWithoutWorld,
-                (entity: SearchableEntity) => entity.isLocal
+        let sortedWithValues: SearchableEntity[]
+        if (options.sortLocalsAndWorldToTop) {
```
sophiamersmann: Did you skip World intentionally in this branch?

pr-plan.md
```diff
@@ -0,0 +1,44 @@
+# PR project report
```
sophiamersmann: You probably forgot to delete this file?

danyx23:
* **#5271** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5271?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5271?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

pr-plan.md
danyx23: Indeed, thanks!

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
danyx23: I'm a bit torn on this. I added it back in now and I'm not sure it's better. E.g. on[ life-expectancy](https://sort-location-to-top-5180.owid.pages.dev/grapher/child-mortality), there are now 4 entities that are auto-added to the top, three of which have the location icon and World that doesn't. I think people might not understand that the list is still sorted below the 4 entities? Ideally we'd have some other way to communicate that these 4 are special, but that would make this PR a bigger thing.

Do you think we should ship with World being sorted up or without?

sophiamersmann: I agree it's a bit confusing if it's sorted up without any indication. In that case, could you add a comment that the behaviour is intentional and maybe refactor/rename `sortLocalsAndWorldToTop` such that the method does what the name suggest?

Issue #5231, opened 2025-08-05
Set Sentry user to Google Analytics ID for session replays


ikesau:
We should use the `_ga` Google ID cookie as a user ID, so that we're able to understand a bit more about our users' journeys on the site. Currently every replay session is by an "Anonymous User"

If `analyticsConsent` is true, then the cookie must exist, so it should be a simple matter of adding it to the config when we init the replay integration.

PR #5274, opened 2025-08-14
Set Sentry user ID to Google Analytics client ID for session replays

danyx23:
Implements #5231 

This pull request enhances the integration between Google Analytics and Sentry for session replay tracking. The main improvement is that, when analytics consent is given and the app is not running in an iframe, the Google Analytics client ID is now set as the Sentry user ID. This is achieved using a robust polling mechanism to wait for the GA cookie, accommodating users who accept cookies later in their session.

**Analytics and Sentry integration:**

* Added logic to poll for the Google Analytics (`_ga`) cookie using `retryPromise` with exponential backoff, allowing up to 10 retries and a total wait time of about 17 minutes. This ensures the client ID can be captured even if cookies are accepted after initial page load. (`site/instrument.ts`, [site/instrument.tsR38-R72](diffhunk://#diff-d2a6a44e94d4517ec3f1b3b115d6ad48e4444735fa92867060824114138fe1d7R38-R72))
* Once the GA cookie is found, extracts the client ID and sets it as the Sentry user ID for improved session replay correlation. (`site/instrument.ts`, [site/instrument.tsR38-R72](diffhunk://#diff-d2a6a44e94d4517ec3f1b3b115d6ad48e4444735fa92867060824114138fe1d7R38-R72))

**Dependency updates:**

* Imported `retryPromise` from `@ourworldindata/utils` and `Cookies` from `js-cookie` to support the new polling logic. (`site/instrument.ts`, [site/instrument.tsL2-R3](diffhunk://#diff-d2a6a44e94d4517ec3f1b3b115d6ad48e4444735fa92867060824114138fe1d7L2-R3))

## Notes
- The way we extract the GA id is a bit unorthodox but widely used. I tried to do this another way but ran into several obstacles. If this were to break in the future it wouldn't be particularly tragic, so I think it's ok to run with this.

### DIFF
--- a/site/instrument.ts
+++ b/site/instrument.ts
@@ -1,5 +1,6 @@
 import * as Sentry from "@sentry/react"
-import { isInIFrame } from "@ourworldindata/utils"
+import { isInIFrame, retryPromise } from "@ourworldindata/utils"
+import Cookies from "js-cookie"
 import {
     COMMIT_SHA,
     ENV,
@@ -7,18 +8,6 @@ import {
     SENTRY_DSN,
 } from "../settings/clientSettings.js"
 import { getPreferenceValue, PreferenceType } from "./cookiePreferences.js"
-
-// Type definition for gtag function
-declare global {
-    interface Window {
-        gtag?: (
-            command: string,
-            targetId: string,
-            config: string,
-            callback: (value: string) => void
-        ) => void
-    }
-}
 
 if (LOAD_SENTRY) {
     const analyticsConsent = getPreferenceValue(PreferenceType.Analytics)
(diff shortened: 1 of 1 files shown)


site/instrument.ts
```diff
@@ -34,4 +35,39 @@ if (LOAD_SENTRY) {
         release: COMMIT_SHA,
         ...sentryOpts,
     })
+
+    // Set Google Analytics client ID as Sentry user ID for session replays
+    if (analyticsConsent && !isInIFrame()) {
```
rakyi, line 40: In the description, you say:

> This is achieved using a robust polling mechanism to wait for the GA cookie, accommodating users who accept cookies later in their session.

Won't `analyticsConsent` be false if the user didn't accept the analytics yet? Unless I'm missing something, we'd only start the polling after the consent was already given, in which case we might not need to poll at all.

If I'm wrong about this, please add a comment to the code that explains it.

bnjmacdonald:
closing as duplicate of #5355

Issue #5290, opened 2025-08-20
Deploy queue occasionally gets stuck


Marigold:
## Problem

Occasionally, [deploy queue](https://admin.owid.io/admin/deploys) gets stuck and `pm2 log deploy-queue` is repeating "Deploying site..." over and over again.

```
567|deploy-queue  | 2025-08-20T12:57:03: Deploying site...
567|deploy-queue  | 2025-08-20T12:57:03: ---
567|deploy-queue  | 2025-08-20T12:57:03: 📆 2025-08-20T12:57:03.600Z
567|deploy-queue  | 2025-08-20T12:57:03:
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart number-of-clinical-trials (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart trials-by-intervention (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart number-of-clinical-trials-by-sponsor (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart number-of-clinical-trials-by-purpose (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart number-of-clinical-trials-by-study-type (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart number-of-clinical-trials-by-status (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart average-study-length-by-phase (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Publishing chart number-of-clinical-trials-with-results-posted (by Tuna Acisu)
567|deploy-queue  | 2025-08-20T12:57:03: Dod early-phase-one-trial created (by Saloni Dattani)
567|deploy-queue  | 2025-08-20T12:57:03: Dod phase-one-trial created (by Saloni Dattani)
567|deploy-queue  | 2025-08-20T12:57:03: Dod phase-two-trial created (by Saloni Dattani)
567|deploy-queue  | 2025-08-20T12:57:03: Dod phase-three-trial created (by Saloni Dattani)
567|deploy-queue  | 2025-08-20T12:57:03: Dod phase-four-trial created (by Saloni Dattani)
567|deploy-queue  | 2025-08-20T12:57:03: Dod cber-fda created (by Saloni Dattani)
[...]

PR #5292, opened 2025-08-20
♻️ Simplify deploy queue server implementation

Marigold:
Fix for https://github.com/owid/owid-grapher/issues/5290

## Human summary

Deploy queue logic was outdated and unnecessarily complex. Retries in case of errors are managed by `pm2` process manager, so we don't need the machinery in code. It wasn't sending errors to Sentry (we had a frozen queue a few times already).

## Summary
• Replace file watching with simple polling loop in startDeployQueueServer.ts (every 5 seconds)
• Remove retry logic (MAX_SUCCESSIVE_FAILURES) and while loop from deployIfQueueIsNotEmpty
• Replace error swallowing with proper error propagation using logErrorAndMaybeCaptureInSentry
• Let PM2 handle process restarts instead of masking errors with retry logic

## Test plan
- [ ] Verify deploy queue server still processes deployments correctly
- [ ] Confirm errors are properly logged to Sentry and cause process restart
- [ ] Test that polling interval works as expected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/baker/startDeployQueueServer.ts
+++ b/baker/startDeployQueueServer.ts
@@ -26,15 +26,25 @@ const main = async () => {
         process.exit(1)
     }
 
-    // Poll for changes every 5 seconds
-    while (true) {
+    console.log(`Watching for changes to: ${DEPLOY_QUEUE_FILE_PATH}`)
+
+    // Watch for file changes and run deploy function when file is modified
+    fs.watchFile(DEPLOY_QUEUE_FILE_PATH, async () => {
         try {
+            console.log(
+                "Deploy queue file changed, checking for deployments..."
+            )
             await runDeployIfQueueIsNotEmpty()
         } catch (error) {
             await logErrorAndMaybeCaptureInSentry(error)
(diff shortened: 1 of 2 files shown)


baker/startDeployQueueServer.ts
```diff
@@ -25,14 +26,16 @@ const main = async () => {
...
 
-    // Listen for file changes
-    fs.watchFile(DEPLOY_QUEUE_FILE_PATH, () => {
-        // Start deploy after 10 seconds in order to avoid the quick successive
-        // deploys triggered by Wordpress.
-        setTimeout(runDeployIfQueueIsNotEmpty, 10 * 1000)
-    })
-
-    void runDeployIfQueueIsNotEmpty()
+    // Poll for changes every 5 seconds
```
rakyi: Previously, we wouldn't call `runDeployIfQueueIsNotEmpty` unless the file changed. Now we call it every 5 seconds even when it doesn't change, which seems more risky in case something goes wrong in `runDeployIfQueueIsNotEmpty`. Could we keep the file watch?

Issue #5082, opened 2025-07-03
Manual line breaks in a gdoc cause rendering errors


rakyi:
## Description

When the user inserts manual line breaks in a gdoc with <kbd>Shift</kbd> + <kbd>Enter</kbd> it results in whitespace control characters being present in the gdoc content, which cause at least these known issues:

1. The content renders incorrectly, as if there was no whitespace:
   
  <img width="593" alt="Image" src="https://github.com/user-attachments/assets/3eb4207e-0ec2-4b40-9017-e46d19c40830" />

2. It makes the data insights RSS feed invalid:

   ```
   XML Parsing Error: not well-formed
   Location: https://ourworldindata.org/atom-data-insights.xml
   Line Number 55, Column 2136:
   ```

  This is a problem, because we use the feed to get DIs into Mailchimp to send the DI newsletter. See [Slack](https://owid.slack.com/archives/C46U9LXRR/p1751535337201879).

This is how the charcters look in gdoc, when you enable View → Show non-printing characters:

<img width="692" alt="Image" src="https://github.com/user-attachments/assets/234d22f9-3c14-4d5f-aff2-ab6b02e15e30" />

## Expected behaviour

It should not be possible to publish a gdoc with these characters. The user should encouter a validation error in the admin.

ikesau:
There are 2 ways to fix this:
1. Find and replace all of these characters with newlines in the `archieToEnriched` step
2. Add validation to `parseText` to catch these characters to make the authors fix the gdoc.

I think option 2 is better, if it can work 🙂

PR #5298, opened 2025-08-20
🐛 fix whitespace in gdocs

ikesau:
## Context
Resolves https://github.com/owid/owid-grapher/issues/5082

Fixes an issue with line tabulation characters making it into our JSON and rendering on mobile and prevents such characters from getting published again

image

## Testing guidance

[You can use this gdoc](https://docs.google.com/document/d/19Qkw0KWlNFziB4t-ePazy-7If1PQaCfbiA-aUtakHJo/edit)

Confirm a vertical tab is present via `View` > `Show non-printing characters`

image

Preview the article and confirm there's an error:
image

Confirm that the suggested RegEx will catch the offending character:
image

Replace it and confirm that there's no error

Check other articles and make sure they can be (re)published as expected.

### DIFF
--- a/db/migration/1755709232592-RemoveTabsFromGdocContent.ts
+++ b/db/migration/1755709232592-RemoveTabsFromGdocContent.ts
@@ -1,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class RemoveTabsFromGdocContent1755709232592
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`-- sql
+            UPDATE posts_gdocs
+            SET content = REPLACE(content, '\\\\u000b', ' ')
+            WHERE content LIKE '%\\\\u000b%'
+        `)
+    }
+
+    public async down(_: QueryRunner): Promise<void> {
+        // no-op
+    }
+}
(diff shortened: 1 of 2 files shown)


db/migration/1755709232592-RemoveTabsFromGdocContent.ts
rakyi, line 1: I'm not sure what exactly are we doing here. Are we replacing the whole control character with just a whitespace? Shouldn't it be a newline, since it shows up as a newline in gdocs?

rakyi (approved):
Nice work! No whitespace characters shall escape on our watch, unless we escape them ourselves!

db/migration/1755709232592-RemoveTabsFromGdocContent.ts
ikesau: Yeah, good catch, but not exactly. The correct thing to do would be to reconstruct the entire JSON body. Newlines are meant to signify the creation of a new `{ "type": "text" }` block; newlines in the `text` field of one of these blocks don't render.

I didn't do that here because a) it would be even more work for an issue that already overran my time estimate, b) none of these missing newlines were severe enough to be caught in editorial review ergo they must be relatively inconsequential, and c) replacing them with spaces _is_ an improvement to the cases where currently there's no space at all.

PR #5313, opened 2025-08-25
Improve ExplorerViews performance with ID-based operations and index

danyx23:
Fixes poor performance of updating explorer views. The `refreshExplorerViewsForSlug` function now queries the existing view ids and then uses those to delete/update instead of using an inefficient where query.

Also adds an index on `explorer_views.explorerSlug` to make those queries not need a full table scan each time.

### DIFF
--- a/db/model/ExplorerViews.ts
+++ b/db/model/ExplorerViews.ts
@@ -273,13 +273,19 @@ export async function refreshExplorerViewsForSlug(
     // Fetch existing explorer views with their chart configs
     type ExistingView = Pick<
         DbRawExplorerView,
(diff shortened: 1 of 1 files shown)


db/migration/1756128544832-AddExplorerViewsExplorerSlugIndex.ts
rakyi, line 1: I don't think we need the separate index, since slug is already a FK and MySQL creates index on a FK automatically.

danyx23: Right! Good catch. This was my fault for a change :)

PR #5314, opened 2025-08-25
Improve multi-dim preview link in admin

### LINKED ISSUE #5249: MDIMs: need to wait before preview link is available in admin
## Core problem

The “preview” button for MDIMs always takes some extra time to show, so we have to wait a few seconds before clicking on it.

It seems that the chart thumbnail is downloaded/created first.

## Proposed solution
The link to the preview should be visible without the need for the thumbnails to be downloaded.
Maybe a "Preview" link or something next to it.

## Demo
https://github.com/user-attachments/assets/2c2b2e82-8748-4c20-abf7-76c6bca1d5fa

rakyi:
Show a clickable fallback when the image has not loaded yet or there was an error loading the image.

Resolves #5249

image

### DIFF
--- a/adminSiteClient/MultiDimIndexPage.tsx
+++ b/adminSiteClient/MultiDimIndexPage.tsx
@@ -41,6 +41,67 @@ type MultiDim = Omit<ApiMultiDim, "updatedAt"> & {
     updatedAt: Date
 }
 
+function PreviewLink({
+    slug,
+    catalogPath,
+}: {
+    slug: string | null
+    catalogPath: string | null
+}) {
+    const [status, setStatus] = useState<"loading" | "success" | "error">(
+        "loading"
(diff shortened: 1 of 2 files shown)


rakyi:
How would aspect-ratio help? We don't know the width of the image before it loads. I could hard code the width, but then we'd need to fix it manually when the thumbnail dimensions change, so I didn't do it - but then we wouldn't need aspect-ratio, so I'm probably missing something.

ikesau:
Yeah you could do it with explicit width/height too, of course. And yeah, it would be via hard-coding the aspect ratio of our mdim thumbnails, which I thought was a fairly stable constant?

rakyi:
OK, I think I was a bit too conservative here. 🙂 Added the width.

PR #5320, opened 2025-08-25
🐛 fix resource panel with many small paragraphs

ikesau:
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/site/css/grid.scss
+++ b/site/css/grid.scss
@@ -29,7 +29,7 @@ $grid-max-colums: 14;
 $grid-col-spans: (2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14);
 $grid-col-start: (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14);
 $grid-col-end: (2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14);
-$grid-row-spans: (1, 2, 3);
+$grid-row-spans: (1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
 
 // Remapping our breakpoint variables that we prefix with $ so that we can use them in class names
 // e.g. .grid-sm-cols-12
@@ -97,7 +97,7 @@ $grid-responsive: (lg, md, sm);
 // Spans Rows
 @each $span in $grid-row-spans {
     .span-rows-#{$span} {
-        grid-row: span #{$span} / span #{$span};
+        grid-row: span #{$span};
     }
 }
 
--- a/site/gdocs/components/layout.ts
+++ b/site/gdocs/components/layout.ts
@@ -55,7 +55,7 @@ const layouts: { [key in Container]: Layouts} = {
         ["pull-quote--left-center"]: "col-start-5 span-cols-6 col-md-start-3 span-md-cols-10 span-sm-cols-12 col-sm-start-2",
         ["pull-quote--right-center"]: "col-start-5 span-cols-6 col-md-start-3 span-md-cols-10 span-sm-cols-12 col-sm-start-2",
         ["pull-quote--right"]: "span-cols-14 grid grid-cols-12-full-width",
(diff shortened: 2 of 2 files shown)


site/css/grid.scss
```diff
@@ -97,7 +97,7 @@ $grid-responsive: (lg, md, sm);
 // Spans Rows
 @each $span in $grid-row-spans {
     .span-rows-#{$span} {
-        grid-row: span #{$span} / span #{$span};
+        grid-row: span #{$span};
```
ikesau, line 100: This was invalid syntax

site/gdocs/components/ResourcePanel.scss
```diff
@@ -65,3 +65,12 @@
         }
     }
 }
+
+// Specifically when the resource panel is used in LTPs, position it explicitly
+// in the first row beneath the header. This means there can only be one
+// resource panel per LTP, but that's currently our only expected usecase.
+.topic-page-header ~ .resource-panel {
```
ikesau, line 72: This allows us to do this for LTPs:
https://github.com/user-attachments/assets/42c34e6f-92bf-4845-ac43-c62f314f8df5


And this for regular articles (if authors ever want to)
https://github.com/user-attachments/assets/2e8c943c-af88-4fe5-b1be-ea351ff22306

mlbrgl: given that this only applies to LTP, it looks like we could merge this with https://github.com/owid/owid-grapher/blob/fix-resource-panel/site/gdocs/components/topic-page.scss#L285-L290

(seeing how regular articles not being affected by any of these rules, and getting their positioning from utility classes in layout.ts)

It would also help keep the discrepancy in grid-row-end values at bay: for the same purpose, it is sometimes 5 (layout.ts), 10 (here) or 6 (topic-page.scss).

site/css/grid.scss
mlbrgl: this one is a bit tricky. It is used by tailwind ([row](https://tailwindcss.com/docs/grid-row), [col](https://tailwindcss.com/docs/grid-column)):

Screenshot 2025-09-02 at 10 31 32

and the [spec makes some provision for this syntax](https://drafts.csswg.org/css-grid-2/#grid-placement-errors)

Screenshot 2025-09-02 at 10 32 28

What I understood is that tailwind possibly uses this tolerance to make their code more maintainable. However in our simplified case, this doesn't look warranted since we only use one syntax for `grid-row` and `grid-col`.

So I'm fine (or even in favour) of simplifying, and I would apply the same simplification throughout the file.

mlbrgl (approved):
I might have mentioned this before but I'm a bit ambivalent about enforcing the positioning of this block. It seems a bit arbitrary that we have so many loosely enforced editorial guidelines, while this one is strongly enforced.

Ideally, we would surface this as a warning in the admin, but I guess the effort is not justified at this point.

JoeHasell:
In case it's helpful to report, I'm encountering the same issue [here](https://admin.owid.io/admin/gdocs/1kbDZmVXI_DWIXL1SR7wD-_c67KUOV1670wRUhlHNz-s/preview) — with not-so-small paragraphs:

Screenshot 2025-09-03 at 16 14 42

ikesau:
Thanks @JoeHasell  - you just need to put it below whichever paragraph you want it to appear at on mobile.

e.g.
```
paragraph-1

paragraph-2

resource-panel

paragraph-3
```

JoeHasell:
Ah thanks Ike! I added to the documentation to make it clear:

> **Resource Panel**
> A sidebar CTA that can link out to charts and, if the article is tagged, will include a link to the data catalog pre-filtered to the tag.
> 
> This component is intended to be used in the introduction section of linear topic pages. 
>
>In the GDoc, the component should be placed at least after the first paragraph of text. 
> 
> On desktop, it will always appear at the top of the section (on the right), irrespective of how it is ordered in the GDoc. On mobile, it will appear within the text according to where you place it in the GDoc. For example, if you place it between the 3rd and 4th paragraphs, this is where it will show on mobile — but on desktop, it will still show aligned with the top of the 1st paragraph.

PR #5332, opened 2025-08-27
🐛 fix thumbnails for csv based explorers

danyx23:
Makes sure that csv based explorers properly await the loading of their data

### DIFF
--- a/packages/@ourworldindata/explorer/src/Explorer.tsx
+++ b/packages/@ourworldindata/explorer/src/Explorer.tsx
@@ -534,11 +534,6 @@ export class Explorer
         return !this.isNarrow
     }
 
-    private futureGrapherTable = new PromiseSwitcher<OwidTable>({
-        onResolve: (table) => this.setGrapherTable(table),
-        onReject: (error) => this.grapher?.setError(error),
-    })
-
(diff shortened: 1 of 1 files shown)


packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -859,7 +854,13 @@ export class Explorer
         this.setGrapherTable(
             BlankOwidTable(tableSlug, `Loading table '${tableSlug}'`)
         )
-        void this.futureGrapherTable.set(this.tableLoader.get(tableSlug))
```
rakyi, line 862: Does using `await` here instead of `void` not solve it? I find the `PromiseSwitcher` abstraction mind-bending, but it seems useful, and we lose that functionality (handling [out of order arrival of network requests](https://react.dev/reference/react/useEffect#fetching-data-with-effects) when switching views).

Not sure how to get the same functionality without it in the current implementation, but we should solve it as it is a common problem when fetching data.

danyx23: Very good point. I now did what you suggested in #5334, where I then unified how we handle this problem across different data loading code paths

PR #5336, opened 2025-08-27
🐛 fix explorer view configs

sophiamersmann:
Daniel recently added a table for explorer view configs, and for search I added a new endpoint that returns the grapher config for a specific explorer view ([example](http://staging-site-search-tpl/explorers/conflict-data.config.json?Conflict+type=Intrastate+conflicts&Measure=Conflict+deaths&Conflict+sub-type=All+sub-types&Data+source=Uppsala+Conflict+Data+Program&Sub-measure=Country+and+regional+data)). Both spin up an Explorer, pick a specific view, then serialise the grapher state with `grapherState.toObject()`.

The tricky part is that explorers only partially rely on that config. So when those exported configs are used to render a chart, there are visible differences compared to the explorer itself. Here’s what I noticed:
- Columns defined by a duplicate transform are added to the dimensions array as a `table` property, which means they’re not used as chart dimensions
- Color scale properties aren’t serialised because they’re set directly on the column def
- Display properties also aren’t serialised for the same reason

This PR kind of snowballed into a refactor of the code handling of indicator-based explorers. I dropped support for two explorer features to make the code a bit simpler:
- Multi-step transforms for indicator-based explorers (not used anywhere)
- The `where` transform (dropped for convenience, see my code comments below)

I would have wanted to do a bit more testing before I open this up for review and there are still a few things I'd want to improve and/or double-check, but I still think it's useful to hand this over to you before my holiday. 

One thing that I might want to rewrite is how the display properties are handled. I think it might be nicer to drop the special handling via `isDisplayProperty` and just serialise whatever is set on the column defs, similar to how it’s done for the color scales. 

The easiest way to test is to go into the console and look at `grapherState.object`.

To do:
- [x] Broken: http://staging-site-fix-explorer-dimension-prope/explorers/conflict-data-source?Data+source=Uppsala+Conflict+Data+Program&Conflict+type=Extrastate+conflicts&Measure=Conflict+deaths&Conflict+sub-type=&Sub-measure=Country+and+region+data

### DIFF
--- a/db/model/ExplorerViews.ts
+++ b/db/model/ExplorerViews.ts
@@ -195,7 +195,7 @@ async function iterateExplorerViews(
             await explorer.updateGrapherFromExplorer()
 
             // Extract the generated config from the explorer's grapher state
-            const config = explorer.grapherState.toObject(false)
+            const config = explorer.grapherState.toObject()
 
             // Grapher uses an internal fallback chain for some important properties.
             // For explorer views we want to have a config that materializes as much
--- a/packages/@ourworldindata/core-table/src/Transforms.ts
+++ b/packages/@ourworldindata/core-table/src/Transforms.ts
@@ -16,18 +16,19 @@ import {
 } from "./ErrorValues.js"
 import * as R from "remeda"
 
-enum TransformParamType {
-    TimeSlug = "TimeSlug", // column with time
-    EntitySlug = "EntitySlug", // column with entity
-    DataSlug = "DataSlug", // column with data
-    ColumnSlug = "ColumnSlug", // any column
+export enum TransformParamType {
+    /** Column with time */
+    TimeSlug = "TimeSlug",
+    /** Column with entity */
+    EntitySlug = "EntitySlug",
+    /** Column with data */
+    DataSlug = "DataSlug",
     Number = "Number",
     String = "String",
 }
 
 interface TransformParam {
     type: TransformParamType
-    spread?: boolean
 }
 
 interface Transform {
@@ -267,67 +268,6 @@ const subtract: Transform = {
     },
 }
 
-enum WhereOperators {
-    is = "is",
-    isNot = "isNot",
-    isGreaterThan = "isGreaterThan",
-    isGreaterThanOrEqual = "isGreaterThanOrEqual",
-    isLessThan = "isLessThan",
-    isLessThanOrEqual = "isLessThanOrEqual",
-}
-// Todo: add tests/expand capabilities/remove?
-// Currently this just supports `columnSlug where someColumnSlug (isNot|is) this or that or this`
-const where: Transform = {
-    params: [
-        { type: TransformParamType.DataSlug },
-        {
-            type: TransformParamType.ColumnSlug,
-        },
-        {
-            type: TransformParamType.String,
-            spread: true,
-        },
-    ],
-    fn: (
-        columnStore: CoreColumnStore,
-        columnSlug: ColumnSlug,
-        conditionSlug: ColumnSlug,
-        ...condition: string[]
-    ): CoreValueType[] => {
-        const values = columnStore[columnSlug]
-        const conditionValues = columnStore[conditionSlug]
-        const operator = condition.shift()
-        let passes: (value: any) => boolean = () => true
-        if (
-            operator === WhereOperators.isNot ||
-            operator === WhereOperators.is
-        ) {
-            const result = operator === "isNot" ? false : true
-            const list = condition.join(" ").split(" or ")
-            const set = new Set(list)
-            passes = (value: any): boolean =>
-                set.has(value) ? result : !result
-        } else if (operator === WhereOperators.isGreaterThan)
-            passes = (value: any): boolean =>
-                value > parseFloat(condition.join(""))
-        else if (operator === WhereOperators.isGreaterThanOrEqual)
-            passes = (value: any): boolean =>
-                value >= parseFloat(condition.join(""))
-        else if (operator === WhereOperators.isLessThan)
-            passes = (value: any): boolean =>
-                value < parseFloat(condition.join(""))
-        else if (operator === WhereOperators.isLessThanOrEqual)
-            passes = (value: any): boolean =>
-                value <= parseFloat(condition.join(""))
-
-        return values.map((value, index) =>
-            passes(conditionValues[index])
-                ? value
-                : ErrorValueTypes.FilteredValue
-        )
-    },
-}
-
 // Assumptions: data is sorted by entity, then time, and time is a continous integer with a row for each time step.
 // todo: move tests over from CE
 const percentChange: Transform = {
@@ -339,7 +279,7 @@ const percentChange: Transform = {
     ],
     fn: (
         columnStore: CoreColumnStore,
-        timeSlug: ColumnSlug,
+        _timeSlug: ColumnSlug,
         entitySlug: ColumnSlug,
         columnSlug: ColumnSlug,
         windowSize: number
@@ -400,7 +340,7 @@ const duplicate: Transform = {
     ): CoreValueType[] => _.cloneDeep(columnStore[columnSlug]),
 }
 
-const availableTransforms: Record<string, Transform> = {
+const availableTransforms = {
     asPercentageOf,
     timeSinceEntityExceededThreshold,
     divideBy,
@@ -408,24 +348,56 @@ const availableTransforms: Record<string, Transform> = {
     percentChange,
     multiplyBy,
     subtract,
-    where,
     duplicate,
 } as const
 
-export const AvailableTransforms = Object.keys(availableTransforms)
+type TransformName = keyof typeof availableTransforms
+export const availableTransformNames = Object.keys(
+    availableTransforms
+) as TransformName[]
 
-export const extractTransformNameAndParams = (
+const isTransformName = (word: string): word is TransformName =>
+    availableTransformNames.includes(word as any)
+
+/**
+ * Extracts the transform name and params from a transform string.
+ *
+ * For example, 'rollingAverage time entity population 5' is parsed as
+ * {
+ *     transformName: 'rollingAverage',
+ *     params: [
+ *        { type: TransformParamType.TimeSlug, value: 'time' },
+ *        { type: TransformParamType.EntitySlug, value: 'entity' },
+ *        { type: TransformParamType.DataSlug, value: 'population' },
+ *        { type: TransformParamType.Number, value: '5' },
+ *    ]
+ * }
+ */
+export const parseTransformString = (
     transform: string
-): { transformName: string; params: string[] } | undefined => {
+):
+    | {
+          transformName: TransformName
+          params: { type: TransformParamType; value: string }[]
+      }
+    | undefined => {
     const words = transform.split(" ")
-    const transformName = words.find(
-        (word) => availableTransforms[word] !== undefined
-    )
+
+    const transformName = words.find((word) => isTransformName(word))
+
     if (!transformName) {
         console.warn(`Warning: transform '${transformName}' not found`)
         return
     }
-    const params = words.filter((word) => word !== transformName)
+
+    const values = words.filter((word) => word !== transformName)
+
+    // Extract data slugs from transforms (i.e. columns with data, not time or entity)
+    const paramDefs = availableTransforms[transformName].params
+    const params = R.zip(paramDefs, values).map(([paramDef, value]) => {
+        return { type: paramDef.type, value }
+    })
+
     return { transformName, params }
 }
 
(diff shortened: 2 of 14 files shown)


sophiamersmann:
* **#5337** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5337?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5336** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5336?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5336?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -17,10 +17,12 @@ import {
 import * as R from "remeda"
 
 enum TransformParamType {
-    TimeSlug = "TimeSlug", // column with time
-    EntitySlug = "EntitySlug", // column with entity
-    DataSlug = "DataSlug", // column with data
-    ColumnSlug = "ColumnSlug", // any column
```
sophiamersmann, line 23: I got rid of the `ColumnSlug` because it was a bit annoying that it can be anything (time, entity or data)

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -267,67 +269,6 @@ const subtract: Transform = {
...
-    isLessThan = "isLessThan",
-    isLessThanOrEqual = "isLessThanOrEqual",
-}
-// Todo: add tests/expand capabilities/remove?
-// Currently this just supports `columnSlug where someColumnSlug (isNot|is) this or that or this`
-const where: Transform = {
-    params: [
-        { type: TransformParamType.DataSlug },
-        {
-            type: TransformParamType.ColumnSlug,
```
sophiamersmann, line 284: The only reason I got rid of the `where` transform is so that I can get rid of `TransformParamType.ColumnSlug`. Not a strong reason, but I also don't believe the where transform is particularly useful to us now or in the future (it's def not currently in use)

packages/@ourworldindata/core-table/src/Transforms.ts
```diff
@@ -458,26 +398,44 @@ export const applyTransforms = (
...
+ * Extracts column slugs that represent data columns from a transform string.
+ *
+ * Parses a transform string to identify parameters that are data column slugs
+ * (as opposed to time, entity, number, or string parameters). This is useful
+ * for understanding which data columns a transform depends on.
+ *
+ * For example, returns ['population'] for the transform string
+ * 'rollingAverage time entity population 5'.
+ */
+export const extractDataSlugsFromTransformString = (
```
sophiamersmann: I haven't changed the logic. Just added comments and simplified

packages/@ourworldindata/explorer/src/ColumnGrammar.ts
```diff
@@ -38,6 +38,7 @@ export const ColumnGrammar: Grammar<ColumnCellDef> = {
         keyword: "name",
         description:
             "This is the name that may appear on the y or x axis of a chart",
+        isDisplayProperty: true,
```
sophiamersmann, line 41: Adding `isDisplayProperty` here makes it so that `name` is added to the display object of a dimension. It looks benign, but this _could_ lead to a change in the `displayName` fallback chain, which might lead to visible differences. This is something I'd like to look in more after my holiday.

packages/@ourworldindata/explorer/src/ExplorerProgram.ts
```diff
@@ -315,6 +335,16 @@ export class ExplorerProgram extends GridProgram {
         return this.columnDefsByTableSlug.get(undefined) ?? []
     }
 
+    get columnDefsWithoutTableSlugByIdOrSlug(): Record<
```
sophiamersmann: Copy-pasted from the Explorer component

packages/@ourworldindata/explorer/src/ExplorerProgram.ts
```diff
@@ -366,30 +396,40 @@ export class ExplorerProgram extends GridProgram {
         return clone
     }
 
-    constructExplorerGrapherConfig(grapherRow: any): ExplorerGrapherInterface {
+    /**
+     * Grapher config for the currently selected row including global settings.
+     * Includes all columns that are part of the GrapherGrammar.
+     */
+    get explorerGrapherConfig(): ExplorerGrapherInterface {
```
sophiamersmann, line 409: No real change. Just got rid of the `constructExplorerGrapherConfig` indirection

packages/@ourworldindata/explorer/src/ExplorerProgram.ts
```diff
@@ -366,30 +396,40 @@ export class ExplorerProgram extends GridProgram {
...
-    constructGrapherConfig(
-        explorerGrapherConfig: ExplorerGrapherInterface
-    ): GrapherInterface {
+    /**
+     * Grapher config for the currently selected row, with explorer-specific
+     * fields translated to valid GrapherInterface fields.
+     *
+     * For example, `yAxisMin` is translated to `{yAxis: {min: ... }}`
+     */
+    get grapherConfig(): GrapherInterface {
```
sophiamersmann, line 436: Same here. No real change. Just got rid of the `constructGrapherConfig` indirection

packages/@ourworldindata/explorer/src/ExplorerProgram.ts
```diff
@@ -547,19 +661,23 @@ const parseColumnDefs = (block: string[][]): OwidColumnDef[] => {
...
-            if (cellDef?.display) {
-                // move field into 'display' object
+            if (
+                cellDef?.isDisplayProperty &&
+                isNotErrorValueOrEmptyCell(row[field]) &&
+                row[field] !== ""
+            ) {
                 row.display = row.display || {}
                 row.display[field] = row[field]
-                delete row[field]
```
sophiamersmann, line 562: Not deleting the field from the top level feels a little safer to me, but it might be ok to delete it as well. Not sure. That's something I'd want to look into after my holiday

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -563,39 +564,19 @@ export class Explorer
...
-    private getBaseColumnsForColumnWithTransform(slug: string): string[] {
+    /**
+     * Extracts the data slugs (variable IDs or column slugs) that a transformed
+     * column depends on.
+     *
+     * For example, if a column has the transform 'divideBy 170775 other_slug', this method
+     * would return ['170775', 'other_slug'] as these are the columns that the
+     * transformed column needs to perform its calculation.
+     */
+    private getDataSlugsForTransformedColumn(slug: string): string[] {
```
sophiamersmann: This used to be more complicated because it supported multi-step transforms

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -686,128 +666,111 @@ export class Explorer
...
+                            if (index === 0) {
+                                // If the transform depends on a single variable
+                                // (e.g. duplicate transform), then add that
+                                // variable as proper chart dimension. If the
+                                // transform depends on multiple variables, then
+                                // the first variable id is treated as the main
+                                // variable for the dimension.
+                                dimensions.unshift({
+                                    variableId: variableId,
+                                    property: dimension.property,
```
sophiamersmann: This is an important change. Previously, all dimensions needed for a transformed columns were added as `table` property

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -686,128 +666,111 @@ export class Explorer
...
-                mergedDef.display ??= {}
-                if (manuallyProvidedDef.name)
-                    mergedDef.display.name = manuallyProvidedDef.name
-                if (manuallyProvidedDef.unit)
-                    mergedDef.display.unit = manuallyProvidedDef.unit
-                if (manuallyProvidedDef.shortUnit)
-                    mergedDef.display.shortUnit = manuallyProvidedDef.shortUnit
-
-                return mergedDef
-            })
```
sophiamersmann, line 810: _I think_ this can be safely dropped now that the display properties are marked as such and added to the display object when parsed

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -993,10 +956,6 @@ export class Explorer
         )
     }
 
-    @computed private get downloadDataLink(): string | undefined {
```
sophiamersmann, line 1013: Unused

packages/@ourworldindata/grapher/src/chart/ChartDimension.ts
```diff
@@ -86,17 +87,20 @@ export class ChartDimension
...
-                },
-                new ChartDimensionDefaults()
-            )
+        const keysToSerialize = [
+            "variableId",
+            "property",
+            "display",
+            "targetYear",
+        ]
+        const obj: OwidChartDimensionInterface = objectWithPersistablesToObject(
```
sophiamersmann, line 96: display is a persistable, so I think we should call `objectWithPersistablesToObject`

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -686,128 +666,118 @@ export class Explorer
...
-                const mergedDef = { ...def, ...manuallyProvidedDef }
-
-                // update display properties
-                mergedDef.display ??= {}
-                if (manuallyProvidedDef.name)
-                    mergedDef.display.name = manuallyProvidedDef.name
-                if (manuallyProvidedDef.unit)
-                    mergedDef.display.unit = manuallyProvidedDef.unit
-                if (manuallyProvidedDef.shortUnit)
-                    mergedDef.display.shortUnit = manuallyProvidedDef.shortUnit
```
sophiamersmann, line 821: I think this can be safely dropped now that the display properties are marked as such and added to the display object when parsed

db/model/ExplorerViews.ts
```diff
@@ -195,7 +195,7 @@ async function iterateExplorerViews(
             await explorer.updateGrapherFromExplorer()
 
             // Extract the generated config from the explorer's grapher state
-            const config = explorer.grapherState.toObject(false)
+            const config = explorer.grapherState.toObject()
```
sophiamersmann, line 198: Probably debatable but I don't think we need to write the defaults into the database since we don't do this for any other config?

packages/@ourworldindata/explorer/src/Explorer.tsx
marcelgerber: What is the purpose here?
I understand that having it as `table` will mean it doesn't show up if we just render the chart based on the `chart_config` alone; but if we have it like this, and have a column defined as `multiplyBy [varId] 2`, then it will show up wrong when rendered based on the `chart_config`, no?
Or is the purpose simply to get more correct metadata to show up?

packages/@ourworldindata/explorer/src/Explorer.tsx
```diff
@@ -686,128 +666,118 @@ export class Explorer
...
+                                    variableId: variableId,
+                                    property: dimension.property,
+                                    display: dimension.display,
+                                })
+                            } else {
+                                // If a transformed column depends on multiple
+                                // variables, we use DimensionProperty.table to
+                                // prevent intermediate variables from appearing
+                                // as chart dimensions
+                                dimensions.unshift({
```
marcelgerber: Using `unshift` will prepend them in the reverse order if there's multiple dimensions, no? Don't know whether that can cause any issues, if they are in a different order (?) from `ySlugs`?

packages/@ourworldindata/explorer/src/ExplorerProgram.ts
```diff
@@ -414,22 +454,96 @@ export class ExplorerProgram extends GridProgram {
...
-     * Grapher config for the currently selected row, with explorer-specific
-     * fields translated to valid GrapherInterface fields.
+     * All dimensions (x, y, color, size) for the currently selected Grapher row.
      *
-     * For example, `yAxisMin` is translated to `{yAxis: {min: ... }}`
+     * Dimensions are either specified by a slug or variable id.
      */
-    get grapherConfig(): GrapherInterface {
-        return this.constructGrapherConfig(this.explorerGrapherConfig)
+    get dimensionsOfSelectedRow(): ExplorerChartDimensionInterface[] {
```
marcelgerber, line 467: Could add a test for this one :)

marcelgerber (approved):

packages/@ourworldindata/explorer/src/Explorer.tsx
sophiamersmann: Yes, it shows up incorrectly for the multiply (and other) transforms, but it works as expected for the `duplicate` transform which is the only one actually used for indicator-based explorers right now. At one point, I had a special case for the duplicate transform where the actual property was used, while all other transforms just added it as `table` dimensions. Now that I think about it again, that's probably the way to go. Better to show nothing than to risk displaying something that's just wrong.

packages/@ourworldindata/explorer/src/Explorer.tsx
sophiamersmann: True... I think I’ll just sort the y-dimensions after the fact to match the order in `ySlugs`. It’s not a huge deal, but the order can matter in stacked area charts for example. Plus, for search, I rely on both the thumbnail (which is rendered server-side) and the table (built from the config on the client) having the same dimension order.

PR #5348, opened 2025-08-29
:sparkles: Update lastEditedAt if tags are updated

Marigold:
We want to start syncing charts with just tags changes in ETL's chart-diff. For that, it'd be very useful to know the time of an update. Can we bump `lastEditedAt` when we make edits to tags? It'd show up at the top of charts view in Admin. Are there any risks from doing this?


ikesau (approved):
Yeah, this seems good 👍

PR #5372, opened 2025-09-05
🚧  [WIP] a/b: add data pages experiment 2

bnjmacdonald:
## Context

Forthcoming...

...

### DIFF
--- a/site/DataPageContent.scss
+++ b/site/DataPageContent.scss
@@ -234,6 +234,18 @@ $experiments: (
         treat0,
         treat1,
     ),
+    exp-data-page-insight-btns-full: (
+        control1,
+        treat000,
+        treat010,
+        treat100,
+        treat110,
+        treat101,
+        treat200,
+        treat210,
+        treat201,
+        treat211,
+    ),
 );
 
 $experimentId: exp-data-page-insight-btns-basic;
--- a/site/DataPageV2Content.tsx
+++ b/site/DataPageV2Content.tsx
@@ -35,6 +35,9 @@ import { GrapherWithFallback } from "./GrapherWithFallback.js"
 import { AttachmentsContext } from "./gdocs/AttachmentsContext.js"
 import { DocumentContext } from "./gdocs/DocumentContext.js"
 import { faArrowRight, faArrowDown } from "@fortawesome/free-solid-svg-icons"
+import Image from "./gdocs/components/Image.js"
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
+import { DataInsightLink } from "@ourworldindata/types"
 
 declare global {
     interface Window {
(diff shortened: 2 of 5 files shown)


site/DataPageV2Content.tsx
```diff
@@ -112,6 +114,11 @@ export const DataPageV2Content = ({
         }
     }
 
+    // insight links for when multiple links are shown
+    const insightLinks = datapageData.dataInsights?.length
+        ? datapageData.dataInsights?.slice(0, 3)
```
rakyi: We already slice them during baking, so there should be no more than three here, right?

site/DataPageV2Content.tsx
```diff
@@ -300,67 +318,178 @@ export const DataPageV2Content = ({
...
+    dataTrackNote?: string
+}) => {
+    const aProps = {
+        href: `/data-insights/${insightLink.slug}`,
+        className:
+            "grid grid-cols-12 span-cols-4 span-lg-cols-6 span-sm-cols-12 owid-btn owid-btn--solid-blue",
+        "data-track-note": dataTrackNote,
+    }
+    return (
+        <a {...aProps}>
```
rakyi: This is unnecessary indirection. Could you move the `aProps` inline?

site/DataPageV2Content.tsx
```diff
@@ -300,67 +318,178 @@ export const DataPageV2Content = ({
...
+            <div className="span-cols-12">
+                <p className="item__type">Data insight</p>
+                <span className="item__title">
+                    {insightLink.title}
+                    <FontAwesomeIcon
+                        className={cx("icon", {
+                            "owid-btn--icon-right": insightLink.title,
+                        })}
```
rakyi: Please use less generic class name than `icon` to avoid conflicts and also more specific name than `owid-btn--icon-right` which we are stealing from the generic `Button` component.

site/DataPageContent.scss
```diff
@@ -302,25 +291,37 @@ $armIds: control1, treat0, treat1;
...
 
-            .insight__title {
-                margin: 0;
+                .item__title {
+                    font-size: 16px;
+                    margin: 0;
 
-                @include subtitle-2-bold;
+                    .icon {
+                        vertical-align: middle;
```
rakyi: Font awesome icons are generally well vertically aligned with the text when you don't override the defaults. It's possible we are doing something wrong if we need to specify this again here, but probably not important.

bnjmacdonald: ah right, `vertical-align: middle` made a small difference when I had `@include subtitle-2-bold;`, but I deleted the latter (which rendered the former unnecessary). Removing `vertical-align: middle` ✅

PR #5383, opened 2025-09-09
🐝 remove deprecated econ tag

ikesau:
Per @bertharc 's request. We're merging `Labor Supply & Employment` into `Labor Force Participation & Employment`

See https://github.com/owid/etl/pull/4988 also

## Context

It's a little tricky to migrate all foreign keys, because you risk violating composite unique constraints.

e.g. if I have the table

```
tagId, chartSlug, createdBy
1, population, 1
2, population, 4
```

I can't update row 1 to be `2, population, 1` because there's a unique constraint on `tagId, chartSlug`

So I had to write each of these tables' unique constraints as of today's date to check whether or not it was safe to migrate the old ID: if there's no violation, then migrate. If there is, then delete the row with the old ID.


rakyi (approved):
I don't see any problems with this approach, but I didn't full grasp why we do this.

In your example from the PR description, you seem to be saying we can't change the old tag to the new tag if the new tag is already used. In that case, wouldn't it be possible to skip the change and only remove the old tag?

rakyi:
(My confusion was resolved in a huddle. 🙂)

PR #5385, opened 2025-09-10
:bug: Add MySQL lock when updating `country_latest_data` to avoid deadlocks

Marigold:
Concurrent calls to update charts using Admin API endpoint `/charts/[chart_id]` sometimes fail with deadlock. Adding a MySQL lock prevents them.

(Note that `country_latest_data` is scheduled for deprecation anyway, so this should be low risk)


Marigold:
Sorry, I shouldn't have complicated it with updated logic (that could be different, as you say). I reverted to the original and added MySQL lock, which should do the trick.

rakyi (approved):

PR #5392, opened 2025-09-11
Switch to mdim view config ID for analytics

### LINKED ISSUE #5365: Include view ID for MDim analytics events
## Core problem

Google Analytics has a 100 character limit in the field of each GA4 event param, so we can't include a serialization of the query params to know which Mdim view a user is acting on.

## Proposed solution

We can include the `chartConfigId` UUID from the `multi_dim_x_chart_configs` table in MDim pages/embeds, then include them in the GA event, so that we can resolve user actions to specific views.

rakyi:
Resolves #5365

## Testing guidance

Check that the selected view config ID is sent to analytics.

### Before merging

### After merging

- [x] Let Bobbie know about the change

### DIFF
--- a/site/multiDim/hooks.ts
+++ b/site/multiDim/hooks.ts
@@ -1,11 +1,19 @@
-import { useMemo } from "react"
+import * as _ from "lodash-es"
+import { useEffect, useMemo, useRef } from "react"
 
-import { GrapherProgrammaticInterface } from "@ourworldindata/grapher"
+import {
+    GrapherAnalytics,
+    GrapherProgrammaticInterface,
+} from "@ourworldindata/grapher"
+import { MultiDimDimensionChoices } from "@ourworldindata/types"
+import { MultiDimDataPageConfig } from "@ourworldindata/utils"
 import {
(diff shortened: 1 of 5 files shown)


site/multiDim/hooks.ts
```diff
@@ -24,3 +31,21 @@ export const useBaseGrapherConfig = (
...
 }
+
+export function useMultiDimAnalytics(
+    slug: string | null,
+    config: MultiDimDataPageConfig,
+    settings: MultiDimDimensionChoices
+) {
+    useEffect(() => {
+        // Log analytics event on page load and when the settings change.
+        if (slug) {
```
ikesau: This is how it worked before, too, but this logs an view every time the Grapher changes, even if the settings don't change (e.g. updating the time selection logs another grapher view)

And while this is true in a sense, I'm not sure if that's helpful? Like maybe we should only log one of these events when the `viewConfigId` changes?

rakyi: Good catch, I totally missed that. Fixing.

PR #5423, opened 2025-09-19
🐛🤖 Ensure explorer views refresh when dependencies change

### LINKED ISSUE #5396: Explorer views are not updated when some of their dependencies change
### Current Job Creation Pattern
Explorer view refresh jobs (`refresh_explorer_views`) are only created in **one place**: when an **explorer configuration itself** is updated via the admin API in `adminSiteServer/apiRoutes/explorer.ts`.

### What Doesn't Trigger Explorer Jobs

I examined the grapher config update endpoints and **none of them** create explorer refresh jobs:

1. **Direct chart updates** (`adminSiteServer/apiRoutes/charts.ts` - `saveGrapher` function)
2. **Variable-level config updates** (`adminSiteServer/apiRoutes/variables.ts` - ETL/admin config functions)  
3. **Bulk chart updates** (`adminSiteServer/apiRoutes/bulkUpdates.ts`)

These operations only:

- Update the chart configurations in the database
- Sync configs to R2 storage
- Trigger static builds if the chart is published
- But **do not** refresh explorer views

### The Implication
This means that if a grapher config that's referenced by an explorer gets updated, the explorer views will become stale until the explorer configuration itself is next updated. The explorer views contain pre-generated chart configurations, and these won't reflect the updated grapher config until someone manually updates the explorer or a manual refresh is triggered.

## Expected behaviour

Explorer views get updated when any of their dependencies change.

## Steps to reproduce

1. Run the async worker:

   ```
   npx tsx --tsconfig tsconfig.tsx.json devTools/worker/explorerJobsWorker.ts
   ```
3. Go to http://localhost:3030/admin/charts/5796/edit
4. Update the title of the chart and save it
5. Observe that no job runs, even though [this explorer](http://localhost:3030/explorers/democracy?Dataset=Polity&Metric=%C2%AD%C2%AD%C2%ADPolitical+regime&Sub-metric=Age+of+democracy&country=ARG~AUS~BWA~CHN~OWID_WRL) uses the chart

## Additional context

[Slack](https://owid.slack.com/archives/CQQUA2C2U/p1757608608272249)

danyx23:
Fixes #5396

- add helper to enqueue refresh jobs from chart and variable updates
- wire helper into admin routes so dependent explorers get queued
- extend queue suite and db test harness to cover refresh triggers without flakes
- pin dbtest mysql image to 8.4.5 to avoid upstream hang

### DIFF
--- a/adminSiteServer/apiRoutes/variables.ts
+++ b/adminSiteServer/apiRoutes/variables.ts
@@ -26,6 +26,7 @@ import {
     updateGrapherConfigAdminOfVariable,
     updateGrapherConfigETLOfVariable,
 } from "../../db/model/Variable.js"
+import { enqueueExplorerRefreshJobsForDependencies } from "../../db/model/Explorer.js"
 import { DATA_API_URL } from "../../settings/clientSettings.js"
 import * as db from "../../db/db.js"
 import {
@@ -341,6 +342,17 @@ export async function putVariablesVariableIdGrapherConfigETL(
         await updateGrapherConfigETLOfVariable(trx, variable, validConfig)
 
     await updateGrapherConfigsInR2(trx, updatedCharts, updatedMultiDimViews)
+    const chartIdsForRefresh = Array.from(
+        new Set(
+            updatedCharts
+                .map((chart) => chart.chartId)
+                .filter((id): id is number => typeof id === "number")
+        )
+    )
+    await enqueueExplorerRefreshJobsForDependencies(trx, {
+        chartIds: chartIdsForRefresh,
+        variableIds: [variableId],
+    })
     const allUpdatedConfigs = [...updatedCharts, ...updatedMultiDimViews]
 
     if (allUpdatedConfigs.some(({ isPublished }) => isPublished)) {
@@ -416,6 +428,17 @@ export async function deleteVariablesVariableIdGrapherConfigETL(
             updates
         )
     await updateGrapherConfigsInR2(trx, updatedCharts, updatedMultiDimViews)
+    const chartIdsForRefresh = Array.from(
+        new Set(
+            updatedCharts
+                .map((chart) => chart.chartId)
+                .filter((id): id is number => typeof id === "number")
+        )
+    )
+    await enqueueExplorerRefreshJobsForDependencies(trx, {
+        chartIds: chartIdsForRefresh,
+        variableIds: [variableId],
+    })
     const allUpdatedConfigs = [...updatedCharts, ...updatedMultiDimViews]
 
     if (allUpdatedConfigs.some(({ isPublished }) => isPublished)) {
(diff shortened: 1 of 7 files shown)


danyx23:
@codex please review

db/model/Explorer.ts
rakyi, line 1: Not a big deal, but I noticed we changed several SQL queries from the Knex query builder to raw SQL, which makes it less type safe. Not worth changing back, but I think it's good to use the query builder when it's not too annoying.

adminSiteServer/apiRoutes/variables.ts
```diff
@@ -341,6 +342,17 @@ export async function putVariablesVariableIdGrapherConfigETL(
         await updateGrapherConfigETLOfVariable(trx, variable, validConfig)
 
     await updateGrapherConfigsInR2(trx, updatedCharts, updatedMultiDimViews)
+    const chartIdsForRefresh = Array.from(
+        new Set(
+            updatedCharts
+                .map((chart) => chart.chartId)
+                .filter((id): id is number => typeof id === "number")
```
rakyi, line 349: We repeat this filter several times in this file, but `chartId` can't be anything else than a number, can it?

adminSiteServer/tests/explorer-views.queue.test.ts
rakyi, line 252: Seems like this comment is outdated now.

adminSiteServer/tests/explorer-views.queue.test.ts
```diff
@@ -415,4 +423,109 @@ describe("Explorer queue semantics", { timeout: 20000 }, () => {
...
+        })
+
+        const explorerTsv = `explorerTitle\tVariable Explorer\nexplorerSubtitle\tTest variable based explorer\nisPublished\ttrue\ngraphers\n\tyVariableIds\ttitle\n\t${variableId}\tIndicator view`
+
+        await env.request({
+            method: "PUT",
+            path: `/explorers/${slug}`,
+            body: JSON.stringify({ tsv: explorerTsv, commitMessage: "init" }),
+        })
+        await env.request({
```
rakyi, line 492: This second request doesn't really do anything, does it? We pass the same params except a different `commitMessage`.

rakyi:
Thanks for fixing this!

adminSiteServer/tests/explorer-views.queue.test.ts
danyx23: No this is actually required - the explorer api is a bit weird, in that the first push to it updates the explorer but does not publish it and then the second push publishes it. We should really change this, it's a super weird quirk.

adminSiteServer/apiRoutes/variables.ts
danyx23: Yes true, this is needlessly defensive

db/model/Explorer.ts
danyx23: Ah yes, two trivial queries were converted to raw SQL. In general I think that for complex queries I prefer SQL and in the  [Database access guidelines ](https://www.notion.so/owid/Database-access-in-Grapher-d9db343c2bfb4ae0b14b3dec72f686c6?source=copy_link) we say that anything involving a join should be done as raw SQL

PR #5451, opened 2025-10-01
:sparkles: Add JSON-LD specs to grapher pages

Marigold:
Add JSON-LD (https://schema.org/Dataset) to grapher & data pages. We hope this could improve how LLMs navigate grapher page and fetch data. I did a couple of tests to see if it helps LLMs at inference time, but it didn't seem to do anything (LLM knew about JSON-LD in HTML, but weren't using download links or metadata in any way).

### DIFF
--- a/site/JsonLdDataset.tsx
+++ b/site/JsonLdDataset.tsx
@@ -1,4 +1,9 @@
-import type { Dataset, WithContext, SearchAction, PropertyValue } from "schema-dts"
+import type {
+    Dataset,
+    WithContext,
+    SearchAction,
+    PropertyValue,
+} from "schema-dts"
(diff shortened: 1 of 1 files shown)


Marigold:
Closing as JSON-LD didn't make much difference. It's unclear whether LLMs are using it or not (maybe in training phase). @rakyi what do you think, should we give it a try despite not knowing its effect? (If yes, we'd have to do this properly) If not, I'll close it.

rakyi:
I'd close it now, and we might return to it in the future.

But I think there might have been a misunderstanding — I didn't intend to use JSON-LD for instructing LLMs on how to modify the query params. Only for [structured data about dataset(s)](https://developers.google.com/search/docs/appearance/structured-data/dataset).

PR #5469, opened 2025-10-07
Tell agent to check types less often

rakyi:
Should we try going with this for a while and tweak it further if necessary?

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,7 +8,7 @@
 - make migrate: apply migrations
 - make dbtest: run database and api tests
 
-When you have completed implementing a set of changes, ALWAYS run `yarn typecheck` and fix any errors you have.
+When you have completed implementing a big set of changes, run `yarn typecheck` and fix any errors you have.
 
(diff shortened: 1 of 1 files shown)


danyx23:
Sure! One thing we could try is to just tell it to run the typechecker once it thinks it is done with the things the user asked. This is sort of the point where you'd want it to do this, right?

rakyi:
I'd prefer to not run it most of the time, so that's why I went with "a big set of changes". Usually it doesn't find any type errors and I don't want to wait for/cancel the type checking, even if it runs only once per prompt. So for small changes, I'd like it to not run at all.

Does it seem too little to you? Too bad overriding LLM "config" is not as easy as JSON. 😄

danyx23:
You can add a user CLAUDE.md/AGENTS.md file for most agents at ~/.claude/CLAUDE.md or something like this, so if we have diverging personal preferences that would also be a place to deal with them. 

Happy to run with your version and see what the results will be!

rakyi:
Hmm, I never tried overriding these files, I wonder how they determine precedence and how reliable it is.

OK, I'll merge this now and if it's not good in practice, I'll either iterate or revert.

PR #5474, opened 2025-10-08
📜 slight tweaks to agents instructions

danyx23:
Adds a few minor tweaks to CLAUDE.md files:
- tell AI not to use inline styles
- tell AI what to pay attention to when adding migrations


danyx23:
The symlink looks correct when I check it out in a new worktree, so I think this is fine. Thanks for the review!

PR #5488, opened 2025-10-10
Add SSG for multi-dim data pages

rakyi:
| Before | After |
|--------|--------|
| image | image |

A clear improvement, but we unfortunately still miss the metadata. We could add that in the future, but that will be much more work.


marcelgerber (approved):
nice!

PR #5499, opened 2025-10-14
Switch SiteRedirectsIndexPage to antd components

rakyi:
## Context

This page was among the first things I made after joining OWID and I didn't even realize I could have used Ant Design there.

I wanted to get rid of the `react-hook-form` (it's great, but not really necessary) and this kind of rewrite is cheap using AI.

| Before | After |
|--------|--------|
| image | image |

## Testing guidance

You could double-check if everything still works (I already checked), if you want.

### DIFF
--- a/adminSiteClient/SiteRedirectsIndexPage.tsx
+++ b/adminSiteClient/SiteRedirectsIndexPage.tsx
@@ -1,11 +1,23 @@
-import cx from "classnames"
-import { useContext, useEffect, useState } from "react"
-import { useForm } from "react-hook-form"
-
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query"
+import { useContext, useEffect, useMemo, useState } from "react"
+import {
+    Alert,
+    Button,
+    Flex,
+    Form,
+    Input,
+    Popconfirm,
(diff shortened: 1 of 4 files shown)


adminSiteClient/SiteRedirectsIndexPage.tsx
```diff
@@ -20,213 +32,344 @@ type Redirect = {
...
+                                    redirects for the site. Don't use this page
+                                    to create redirects for charts, use the{" "}
+                                    <Link to="/redirects">
+                                        chart redirects page
+                                    </Link>{" "}
+                                    instead.
+                                </Typography.Paragraph>
+                                <Typography.Paragraph>
+                                    The source has to start with a slash and any
+                                    query parameters (/war-and-peace
```
ikesau, line 231: It hasn't been an issue thus far, and the validation ensures we don't run into problems from misunderstanding this, but for maximal clarity you could split this into 2 sentences. `...with a slash. Any query parameters...`

rakyi: Yeah, that's better. Done.

PR #5500, opened 2025-10-14
🐛 (tooltip) default to 0 if no min value is given

### LINKED ISSUE #5284: MDIMs should have y-axis min set to 0 as default
## Core problem

This was previously a commonly reported issue with line charts (see #3797), and it would be useful to implement it in MDIMs as well. In D&R we pretty much would always want the y-axis minimum to default to 0. 

## Proposed solution

In the MDIM config, this should be the default value: `"yAxis": {"min": 0}`.

## Context

In MDIMs, the main manifestation of the issue that I've noticed is that it impacts the map hover view, e.g., it looks like this when the ymin is not set, exagerating small variations in mostly flat trends: 

Image

sophiamersmann:
Fixes #5284

### DIFF
--- a/packages/@ourworldindata/grapher/src/mapCharts/MapSparkline.tsx
+++ b/packages/@ourworldindata/grapher/src/mapCharts/MapSparkline.tsx
@@ -138,7 +138,7 @@ export class MapSparkline extends React.Component<MapSparklineProps> {
                 },
                 // Copy min/max from top-level Grapher config if Y column == Map column
                 min: this.manager.mapAndYColumnAreTheSame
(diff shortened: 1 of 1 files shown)


sophiamersmann:
* **#5500** <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/owid/owid-grapher/5500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #5508, opened 2025-10-14
✨ Minimal CTA-only Announcements

ikesau:
## Context

After chatting with Joe, I realized that we'd like announcements to be able to link directly to charts/articles from the homepage. Thus the "cta-only" announcement was born.

It simply adds a `{.cta}` front-matter property that can be specified for gdocs with type `announcement`.

When present, it renders the announcement using the Announcement's excerpt as its text content, plus the cta as a link.

During my implementation of this, I realized that I hadn't implemented linkedCharts and linkedDocuments for announcements yet, so this PR also fixes that.

Unfortunately, this required some extra work to get working on the homepage's latest announcements section, because the traditional loadLinkedCharts / loadLinkedDocuments flow assumes that the links are present on the gdoc's static JSON, whereas latestAnnouncements is loaded as an attachment. So `GdocHomepage` needs to also resolve these links in `_loadSubclassAttachments` after the latest announcements have been fetched.

| Homepage | Latest Page |
|--------|--------|
| image | image | 

## Testing guidance

```
title: We have updated our many of our charts on income inequality
type: announcement
authors: Joe Hassel
excerpt: We’ve updated data from the Luxembourg Income Study, allowing us to compare inequality before and after tax.
kicker: Data update
{.cta}
url: https://ourworldindata.org/grapher/economic-inequality-gini-index
text: See the new data
{}
[+body]
[]
```
or
```
title: We've just updated part of a topic page
type: announcement
authors: Max Roser
excerpt: Whoa! An update to a topic page, perhaps an anchor link
kicker: Update
{.cta}
url: https://docs.google.com/document/u/0/d/published-id
text: Read the article
{}
[+body]
[]
```

### DIFF
--- a/site/NewsletterSubscription.tsx
+++ b/site/NewsletterSubscription.tsx
@@ -13,7 +13,7 @@ const analytics = new SiteAnalytics()
 export const NewsletterSubscription = ({
     context,
 }: {
-    context?:
+    context:
         | NewsletterSubscriptionContext.Homepage
         | NewsletterSubscriptionContext.Floating
 }) => {
@@ -76,7 +76,7 @@ export const NewsletterSubscriptionForm = ({
     context,
     className = "",
 }: {
-    context?: NewsletterSubscriptionContext
+    context: NewsletterSubscriptionContext
     className?: string
 }) => {
     const DATA_INSIGHTS = "16"
@@ -118,6 +118,7 @@ export const NewsletterSubscriptionForm = ({
                 Receive our latest work by email.
             </span>
             <img
+                alt=""
                 className="newsletter-subscription-form__checkbox-image"
                 src="/images/biweekly-newsletter.webp"
                 width={1200}
@@ -154,6 +155,7 @@ export const NewsletterSubscriptionForm = ({
                 </a>
             </div>
             <img
+                alt=""
                 className="newsletter-subscription-form__checkbox-image"
                 src="/images/data-insights.webp"
                 width={1200}
--- a/site/OwidSocials.tsx
+++ b/site/OwidSocials.tsx
@@ -1,5 +1,4 @@
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
-import * as React from "react"
 import { RSS_FEEDS, SOCIALS } from "./SiteConstants.js"
 import { NewsletterSubscriptionContext } from "./newsletter.js"
 


db/model/Gdoc/GdocBase.ts
```diff
@@ -115,6 +115,75 @@ export async function getLinkedIndicatorsForCharts(
     })
 }
 
+/**
+ * Helper function to load linked charts for given grapher and explorer slugs.
+ * This is shared between GdocBase.loadLinkedCharts and GdocHomepage.loadAnnouncementLinks
+ */
+export async function loadLinkedChartsForSlugs(
```
ikesau, line 121: This is a cut-paste from `GdocBase`

db/model/Gdoc/GdocBase.ts
```diff
@@ -115,6 +115,75 @@ export async function getLinkedIndicatorsForCharts(
     })
 }
 
+/**
+ * Helper function to load linked charts for given grapher and explorer slugs.
+ * This is shared between GdocBase.loadLinkedCharts and GdocHomepage.loadAnnouncementLinks
```
rakyi: I'd remove this line. The usage can be inferred from code, and the comment can easily become outdated.

db/model/Gdoc/GdocPost.ts
```diff
@@ -250,46 +260,67 @@ export const enrichLatestPageItems = async (
...
+    ) as Record<string, LinkedChart>
+
+    const linkedDocuments = Object.assign(
+        {},
+        ...announcementModels.map((m) => m.linkedDocuments)
+    ) as Record<string, OwidGdocMinimalPostInterface>
+
+    // Extract announcements' linked documents' featured images
+    // (in case they're needed for prominent links)
+    const linkedDocumentFeaturedImageFilenames = pipe(
```
rakyi: Since Lodash and Remeda share some function names and also since they often read similar to our own functions, most of their usage is now in the form of `R.pipe()`, with `import * as R from "remeda"`.

Would you mind changing imports to that style? If it's too much work, then we can keep it as is, but I think it's worth using that style going into the future.

site/gdocs/components/HomepageIntro.tsx
```diff
@@ -116,6 +118,87 @@ function FeaturedWorkTile({
...
+    const { linkedDocument } = useLinkedDocument(announcement.cta?.url || "")
+    // If there's not enough space to show all of the third announcement, we link to the latest page, no matter what
+    // If there is enough space, it behaves as normal:
+    // - cta: link directly to the resource
+    // - no cta: link to the latest page
+    const announcementRef = useRef<HTMLLIElement>(null)
+    const isOverflowing = useIsOverflowing(
+        announcementRef as RefObject<HTMLElement>
+    )
+    const latestPageLink = `/latest#${announcement.slug}`
```
rakyi, line 163: We don't list all announcements on `/latest`, right? Too bad we can't have a stable link that would always work, but I guess that would be a lot harder to do.

site/gdocs/components/HomepageIntro.tsx
```diff
@@ -116,6 +118,87 @@ function FeaturedWorkTile({
...
+          linkedDocument?.url ||
+          announcement.cta?.url ||
+          latestPageLink
+
+    return (
+        <li
+            className={cx("homepage-intro__announcement", {
+                "homepage-intro__announcement--is-overflowing": isOverflowing,
+            })}
+            key={announcement.id}
```
rakyi: There should be no key here, and instead it should be `announcement.id` in the `HomepageAnnouncements` where we use this component.

site/gdocs/components/HomepageIntro.tsx
ikesau: We could, but Joe/Ed want it to go to the feed page so that it's situated in the context of the rest of our work and minimizes the editorial significance/permanence of announcements.

PR #5513, opened 2025-10-15
Switch Grapher to React Aria select

rakyi:
## Context

We don't want to ship two different packages for our select components to reduce bundle size.

Unfortunately, I didn't go all the way — we still use `react-select` in `GlobalEntitySelector`, which is hard to switch and only used in country pages, which should be soon removed/reworked. So the hope is we'll get rid of `react-select` completely at that point in `site`.

It's still used in admin, but bundle size there doesn't really matter. We'll probably switch it to antd components there eventually.

## Testing guidance

Please check that we get equivalent functionality in grapher and explorer with after the change.

### Before merging

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/grapher.scss
+++ b/packages/@ourworldindata/grapher/src/core/grapher.scss
@@ -58,7 +58,6 @@ $zindex-controls-drawer: 150;
     @import "../popover/Popover.scss";
     @import "../controls/SettingsMenu.scss";
     @import "../controls/MapRegionDropdown.scss";
-    @import "../controls/MapCountryDropdown.scss";
     @import "../captionedChart/CaptionedChart.scss";
     @import "../timeline/TimelineComponent.scss";
     @import "../controls/ContentSwitchers.scss";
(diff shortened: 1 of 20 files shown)


site/multiDim/DimensionDropdown.tsx
```diff
@@ -69,7 +69,7 @@ export default function DimensionDropdown({
             isDisabled={isDisabled}
             isOpen={isOpen}
             onOpenChange={setIsOpen}
-            selectedKey={value}
```
rakyi: `selectedKey` is now deprecated, and the new prop is called `value`.

packages/@ourworldindata/grapher/src/core/grapher.scss
```diff
@@ -40,6 +40,13 @@ $zindex-controls-backdrop: 130;
 $zindex-controls-popup: 140;
 $zindex-controls-drawer: 150;
 
+// Importing here so the variables above are accessible. Must be imported
+// outside of .GrapherComponent, because the dropdown options get rendered
+// outside of it.
+@import "../controls/Dropdown.scss";
+@import "../controls/SearchDropdown.scss";
+@import "../controls/MapCountryDropdown.scss";
```
sophiamersmann: Is there a reason you moved these here instead of below where the rest of the unscoped imports are?

rakyi: I think it was probably just an oversight. Moving it below.

rakyi:
Good catch. Fixed that regression now as well.

PR #5516, opened 2025-10-15
:sparkles: Bump OpenAI version from gpt-4o to gpt-5

Marigold:
Automatic alt text generation is apparently [not great](https://owid.slack.com/archives/C03NA1B2P1U/p1760529280770049). Bumping OpenAI version from `gpt-4o-mini` to `gpt-5-mini` might help. We could even go for `gpt-5` if we're fine with slower generation.

EDIT: replies by Charlie and Lucas

> ike has looked into this a bunch, including using different models — we had a team hour recently about it and saw various trade-offs. it also wasn’t clear that 5 is better, but happy to experiment and see (accuracy should probs be the top priority)

> my sense is that `gpt5 > gpt-4o, gpt-5-mini >>> gpt-4-mini`

### DIFF
--- a/adminSiteServer/imagesHelpers.ts
+++ b/adminSiteServer/imagesHelpers.ts
@@ -123,7 +123,11 @@ export async function fetchGptGeneratedAltText(url: string) {
                 content: `Generate alt text for this image, describing it for a vision impaired person using a screen reader.
 Do not say "alt text:".
 Do not say "The image...".
+Do not mention colors unless they are important.
 Do not use markdown in the text.
+If the image has a title, mention it first.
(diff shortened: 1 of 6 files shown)


ikesau:
Hey Mojmir 🙂 

[I also looked into this](https://github.com/owid/owid-grapher/compare/gpt-5-mini), and while it's hard to judge, we found the tone of gpt-5-mini to be slightly worse?

I think it hallucinates figures far less often, though, so for that reason we should probably just make the switch.

Feel free to bring in my changes to the prompt from my branch into this one, or merge it instead.

PR #5525, opened 2025-10-15
Use prefetched attachments during gdoc baking

Marigold:
## Summary
Optimizes gdoc baking performance by using prefetched attachment data instead of making redundant database queries.

## Problem
During site baking, we prefetch all gdoc attachments (authors, charts, images, etc.) to avoid N+1 queries. However, the `validate()` method was still calling `mapSlugsToIds()` and `getMinimalAuthorsByNames()` hundreds of times during baking, leading to redundant database queries.

## Solution
- Added optional `prefetched` parameters to `loadState()`, `validate()`, `loadLinkedAuthors()`, and `loadLinkedCharts()`
- Built and cached a slug-to-id mapping during chart prefetching in `SiteBaker`
- Updated `SiteBaker` to pass prefetched data through the baking pipeline
- Maintained backward compatibility (all new parameters are optional, so non-baking code paths continue to work as before)

## Performance Impact
**~25% faster baking** (60s → 45s for `--steps charts gdocPosts`)

## Testing
- ✅ `yarn typecheck` passes
- ✅ `yarn testLintChanged` passes
- ✅ `yarn fixPrettierChanged` passes
- ✅ Local bake completes successfully

## Changes
- **db/model/Gdoc/GdocBase.ts**: Modified `loadState()`, `validate()`, `loadLinkedAuthors()`, and `loadLinkedCharts()` to accept and use prefetched data
- **baker/SiteBaker.tsx**: Built slug-to-id map during prefetching and passed it to `validate()`

Co-Authored-By: Claude <noreply@anthropic.com>


db/model/Gdoc/GdocBase.ts
```diff
@@ -741,8 +750,11 @@ export class GdocBase implements OwidGdocBaseInterface {
         }
     }
 
-    async loadLinkedCharts(knex: db.KnexReadonlyTransaction): Promise<void> {
```
marcelgerber, line 744: We should also make sure that the `getPublishedExplorersBySlug` call in here is also cached, it's also happening thousands of times.

Marigold: Yeah, that's a big bottleneck. I've added it in [this PR](https://github.com/owid/owid-grapher/pull/5535). It'd make sense to first merge that one and then revisit whether this PR helps at all.

Marigold:
Closing, build times were optimized recently and this wasn't a bottleneck.

PR #5527, opened 2025-10-16
✨ Use KnexReadonly instead of KnexReadonlyTransaction for baking

Marigold:
## Motivation

Baking isn't writing to MySQL (it used to because of GDocs images, but it's not doing it any more), but it's using a transaction for the whole baking process. This can sometimes result in deadlocks and issues when cloning staging servers from snapshot of `staging-site-master` (this has been partly resolved). Still, maybe we don't need one huge transaction?

## Summary

- Changes baking code to use `KnexReadonly` (regular Knex instance) instead of `KnexReadonlyTransaction`
- Replaces `knexReadonlyTransaction()` with `knexReadonly()` for baking operations
- Baking doesn't need transactions since it's just reading data to generate static files

## Test plan

- [ ] Verify local bake still works: `yarn buildLocalBake`
- [ ] Run type checks: `yarn typecheck`

### DIFF
--- a/baker/DeployUtils.ts
+++ b/baker/DeployUtils.ts
@@ -167,9 +167,7 @@ const getSlackMentionByEmail = _.memoize(
  * the end of the current one, as long as there are changes in the queue.
  * If there are no changes in the queue, a deploy won't be initiated.
  */
-export const deployIfQueueIsNotEmpty = async (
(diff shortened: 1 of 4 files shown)


Marigold:
Closing, it's not causing as many headaches after build times optimization.

PR #5535, opened 2025-10-16
:zap: optimize chart baking performance

Marigold:
# 🧑‍🦱 Summary

I let Claude profile our baking process and find the biggest bottlenecks. There were some easy wins which decreased runtime from 15 min to about 9 min.
- Cache `mapSlugsToIds`, `getDods`, `narrativeChartNames` and `publishedExplorersBySlug`
- Don't load state of gdocs in `blogIndex` step
  - The biggest bottleneck was `getLinkedIndicatorsForCharts`, but I couldn't optimize it any more
  - Perhaps there's a better way than passing `loadState: false`?
- Use `chart_dimensions` in `getRelatedChartsForVariable` 

Cache is saved in a transaction and not in module, because that could cause problems when reusing those functions elsewhere.

The biggest bottleneck is chart baking now. That comes primarily from loading data & metadata JSON files from R2. Simply increasing parallelization didn't help.

# 🤖 Summary

This PR optimizes the performance of chart baking, data insights baking, blog index baking, and gdoc posts baking by addressing N+1 query problems, optimizing SQL queries, and skipping unnecessary operations.

## Chart Baking Optimization

**Optimize SQL query** in getRelatedChartsForVariable to use indexed chart_dimensions table join instead of JSON_CONTAINS()
- Query time reduced from ~28ms to ~0.8ms per call (97% faster)
- Function execution time reduced from 64s to 1.8s total (for 2288 calls)
- **Impact**: Chart baking improved from **~75s to ~63s** for 4522 charts **(16% faster, saving ~12 seconds)**

This optimization accounts for 88% of the chart baking performance improvement.

## Data Insights Baking Optimizations

- **Memoize getPublishedExplorersBySlug()** to avoid fetching all explorers 355 times
  - **Biggest optimization**: saves ~46 seconds (244% improvement)
- **Memoize mapSlugsToIds()** to avoid fetching all chart slugs 355 times
  - Saves ~7 seconds (38% improvement)

**Impact**: Data insights baking improved from **~6 minutes to ~19 seconds** for 355 insights **(95% faster, saving ~5.7 minutes)**

## Blog Index Baking Optimization

**Skip expensive loadState() operations** when loading blog index posts
- Add optional `loadState` parameter to `loadGdocFromGdocBase()` and `getAndLoadListedGdocPosts()` (defaults to true)
- Use `loadState: false` in `getBlogIndex()` since blog index only needs basic metadata
- **Impact**: Blog index baking improved from **~14.2s to ~270ms** for 399 posts **(98% faster, saving ~14 seconds)**

### Why this works
The blog index only displays post cards with basic information (title, slug, excerpt, authors, date, thumbnail), so it doesn't need the expensive operations that `loadState()` performs:
- `loadLinkedCharts`: 2,932s across 399 posts (7.3s avg per post) - loads full chart configs
- `validate`: 1,413s (3.5s avg) - validates all content and links
- `loadImageMetadataFromDB`: 201s (504ms avg) - loads all image metadata
- Other operations: 240s

When 399 posts load in parallel, these operations create massive database contention. Skipping them for blog index is safe and dramatically improves performance.

## GDoc Posts Baking Optimization

**Memoize validation data queries** to avoid fetching the same data 601 times during gdoc posts baking
- Memoize `getDods()` - was taking 3.2s (5.3ms × 601 calls), now 21ms
- Memoize `getAllNarrativeChartNames()` - was taking 462ms (0.8ms × 601 calls), now 1ms
- **Impact**: GDoc posts baking improved from **~13.8s to ~9.5s** for 601 posts **(31% faster, saving ~4.3 seconds)**
- **Impact**: Validate function improved from **~4.4s to ~1.0s** **(77% faster, saving ~3.4 seconds)**

### Why this works
The `validate()` function in GdocBase fetches validation data (chart slugs, explorers, narrative chart names, DODs) for every gdoc post to validate links and content. This data is the same for all posts, so fetching it once and caching it saves significant time.

### Chart Baking
The main bottleneck was the SQL query in getRelatedChartsForVariable using JSON_CONTAINS() which cannot utilize indexes. By rewriting it to use an INNER JOIN on the chart_dimensions table (which has an indexed variableId column), query performance improved by 97%.

### Data Insights Baking  
The main issue was N+1 query problems where the same data was being fetched repeatedly for each insight being baked:
- getPublishedExplorersBySlug() was querying all explorers 355 times
- mapSlugsToIds() was querying all chart slugs 355 times

The memoization ensures these queries are only executed once and cached for subsequent calls within the same baking session.

### Transaction-Scoped Caching Implementation
All memoization uses **transaction-scoped caching** rather than module-level caching:
- Cache is stored on the knex transaction object: `(knex as any).__queryCache`
- Cache is automatically cleared when the transaction ends
- Safe for both long-running baking transactions and short web app requests
- Created `cachedInTransaction<T>()` helper function to eliminate code duplication

This approach ensures cached data never becomes stale and works correctly in both baking and web app contexts.

### Blog Index Baking
The optimization adds a `loadState` option that allows callers to skip expensive operations when they only need basic gdoc metadata. This is implemented by:
1. Adding optional `loadState` parameter to `loadGdocFromGdocBase()` (defaults to true for backward compatibility)
2. Passing the option through `getAndLoadListedGdocPosts()`
3. Using `loadState: false` in `getBlogIndex()` which only needs basic post information

### GDoc Posts Baking
The validate() function was calling getDods() and getAllNarrativeChartNames() for every gdoc post, even though the data is identical for all posts. Memoizing these functions ensures they only query the database once, with subsequent calls returning the cached result.

## Testing

Each optimization was tested individually by removing it and measuring the performance impact. Only the optimizations with significant impact were kept in the final PR.

For the blog index optimization, testing with `loadState: true` showed that batch query optimization for `loadLinkedCharts` didn't help (actually ~8% slower) because 601 parallel posts still create database contention. The skip-loadState approach is much more effective.

### DIFF
--- a/db/model/Post.ts
+++ b/db/model/Post.ts
@@ -153,7 +153,7 @@ const getFullPost = async (
 
 export const getBlogIndex = _.memoize(
     async (knex: db.KnexReadonlyTransaction): Promise<IndexPost[]> => {
-        // TEMPORARY: Using loadState: false to skip expensive operations for blog index
+        // Using loadState: false to skip expensive operations for blog index
         const gdocPosts = await getAndLoadListedGdocPosts(knex, {
             loadState: false,
         })


db/model/Dod.ts
```diff
@@ -12,17 +12,19 @@ import {
...
-        `-- sql
-        SELECT * FROM dods
-        ORDER BY updatedAt DESC`
-    )
+    return cachedInTransaction(knex, "dods", async () => {
+        return knexRaw<DbPlainDod>(
+            knex,
+            `-- sql
+            SELECT * FROM dods
+            ORDER BY updatedAt DESC`
```
marcelgerber, line 25: This is a total aside, and also not really important, but I just noticed that we don't index `dods` by `updatedAt`.
And also, the admin is the only place that benefits from DoDs being ordered by `updatedAt`, all the other ones run `indexBy` on it either way.

marcelgerber (approved):
Caching them on the level of the knex transaction is galaxy brain!

Issue #2842, opened 2023-10-24
Gdoc italics inside a link get parsed in the wrong order


ikesau:
This source in Gdoc:

<img width="630" alt="image" src="https://github.com/owid/owid-grapher/assets/11844404/e18b918b-c57a-45a7-82ca-723a56a97209">

Results in this HTML being rendered: (3 separate `a` tags):

<img width="526" alt="image" src="https://github.com/owid/owid-grapher/assets/11844404/277610fd-65f1-4d88-b694-23185b0a34c8">

i.e.

```
<a>differences</a>
<em><a>within</a></em>
<a>countries</a>
```

It should be:
```
<a>differences <em>within</em> countries</a>
```

ikesau:
Ideally, https://github.com/owid/owid-grapher/issues/2483 should also be fixed as part of this

PR #5536, opened 2025-10-16
🐛 Fix subscript parsing

ikesau:
## Context

Fixes https://github.com/owid/owid-grapher/issues/2962 and https://github.com/owid/owid-grapher/issues/2842

Our code that converts the Google Docs AST to an ArchieML-ish text format had a bug where it would create new spans any time a new element began, nesting them inside a new `SpanLink` if that was present on the element.

e.g.  the Google Docs equivalent of [This is a link with **formatting** inside it](https://en.wikipedia.org/wiki/Main_Page) would get converted to [This is a link with ](https://en.wikipedia.org/wiki/Main_Page)[**formatting**](https://en.wikipedia.org/wiki/Main_Page)[ inside it](https://en.wikipedia.org/wiki/Main_Page) (check the source of this text to see what I mean)

This PR fixes the issue by finding such cases and concatenating all the children of the first SpanLink into one child array.

I also wrote a script to identify posts where this has happened, so we can republish these documents.

It says the only afflicted posts are `war-and-peace`, `sdgs/economic-growth`, `population-projections`, `better-learning`, and `corruption-development-relationship` - this seems like a low number, but I can at least confirm that it's correct in each case (verified with the following CSS on these pages: `a + a { background: lime; color: red; }`)

| Before | After |
|--------|--------|
| image | image |


db/model/Gdoc/gdocToArchie.ts
```diff
@@ -9,11 +9,118 @@ import {
...
+                if (!el.textRun) return null // Should not happen in this logic block
+
+                const baseSpan: SpanSimpleText = {
+                    spanType: "span-simple-text",
+                    text: el.textRun.content || "",
+                }
+
+                return applyTextStyles(el.textRun, baseSpan)
+            })
+            .filter((span): span is Span => span !== null)
```
rakyi: I don't think we need the `: span is Span` part.

rakyi (approved):
Nice! I'm glad we are finally fixing this.

Could you add tests for more cases, e.g. parsing two different consecutive links with different styles, e.g. bold, italic etc.?

ikesau:
I republished the affected articles ✅

PR #5556, opened 2025-10-20
chore: exclude react-aria locales from build

marcelgerber:
As suggested here: https://react-spectrum.adobe.com/react-aria/internationalization.html#vite


marcelgerber:
@codex review

PR #5567, opened 2025-10-21
✨ Expand analytics to all routes and capture full user-agents

Marigold:
## 🧔‍♂️ Summary

I'm trying to analyze traffic from LLM bots. This can be done by inspecting the User-Agent, but unfortunately, LLMs don't run JavaScript. As a result, Google Tag Manager doesn't detect them and doesn't send their data to GA4 (and BigQuery). Luckily, we already have a Cloudflare middleware that sends `cf_function_invocation` events via the Measurement Protocol!

We should start tracking all our pages in the Cloudflare function, not just `/grapher/*` and similar paths. I’ve excluded paths like `/assets/*` and `/fonts/*`, but there might be more we want to exclude. We can refine the exclusions after collecting some initial data.

As a second step, if the User-Agent length is greater than 100 (which is common for LLMs), store it as `user_agent_next` so we can concatenate them later in BigQuery.

## 🤖 Summary
- Enable analytics tracking on all routes (not just grapher/donation/explorers)
- Capture full user-agent strings by adding `user_agent_next` parameter for chars 100-200
- Exclude static assets to avoid unnecessary function invocations

## Motivation
We want to track LLM bot traffic (ChatGPT, Claude, Perplexity, etc.) across the entire site. These bots don't execute JavaScript, so we need server-side tracking. Bot identifiers in user-agent strings often appear after character 100, which was previously truncated due to GA4's 100-char limit.

## Changes
1. **`_routes.json`**: Changed from specific paths to `/*` with exclusions for static assets
2. **`functions/_common/analytics.ts`**: Added `user_agent_next` field to capture chars 100-200 of user-agent when it exceeds 100 chars

## Example
For user-agent: `Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko); compatible; ChatGPT-User/1.0; +https://openai.com/chatgpt`
- `user_agent`: First 100 chars (truncated)
- `user_agent_next`: Chars 100-200 (contains "ChatGPT-User/1.0...")

## Testing
- TypeScript: ✅ `yarn typecheck` passed
- Linting: ✅ `yarn testLintChanged` passed
- Formatting: ✅ `yarn fixPrettierChanged` passed

### DIFF
--- a/_routes.json
+++ b/_routes.json
@@ -8,9 +8,9 @@
         "/*.css",
         "/*.js",
         "/images/*",
-        "*.woff2",
-        "*.woff",
-        "*.ttf",
-        "*.ico"
+        "/*.woff2",
+        "/*.woff",
+        "/*.ttf",
+        "/*.ico"
     ]
 }


functions/_common/analytics.ts
```diff
@@ -48,16 +48,22 @@ export function getCommonEventParams(request: Request, pathname: string) {
...
         pathname: pathname,
         referrer,
         user_agent,
         method: request.method,
         country: request.headers.get("cf-ipcountry") || "",
     }
 
+    // If user-agent is longer than 100 chars, capture the next 100 chars
+    if (fullUserAgent.length > 100) {
+        params.user_agent_next = fullUserAgent.slice(100, 200)
```
ikesau, line 64: Hopefully there aren't UA's larger than 200, but should we maybe make this more dynamic?

_routes.json
```diff
@@ -1,11 +1,16 @@
...
-    "include": [
-        "/grapher/*",
-        "/deleted/*",
-        "/donation/*",
-        "/explorers/*",
-        "/multi-dim/*"
-    ],
-    "exclude": ["/grapher/_grapherRedirects.json"]
+    "include": ["/*"],
+    "exclude": [
```
ikesau, line 4: Is there a way you generated this list? Otherwise I'll build my own and double-check there's nothing missing

Marigold: I checked a few of our pages and used my best judgment to create this list. I’m still a bit of a noob, so I’ve probably missed some. If you have other suggestions, please speak up!

(I plan to check random events in BigQuery after we merge this to see if I missed anything.)

functions/_common/analytics.ts
Marigold: User agents > 200 would be very rare. I'm afraid of adding more parameters dynamically, I think that GA has some limits on number of parameters and I won't want to mess with that. UAs longer than 200 would be bots anyway.

PR #5572, opened 2025-10-22
Add archived videos to archived articles

rakyi:
## Context

Part of #5366

## Testing guidance

Check that the videos are properly archived and used in archived articles. Example article with a video:

https://ourworldindata.org/new-feature-embed-archived-charts

You can archive it with:

```
yarn buildArchive --latestDir --type posts --postSlugs new-feature-embed-archived-charts
```

PR #5575, opened 2025-10-23
migrate(db): add computed column & index on `posts_gdocs_components.type`

marcelgerber:
should aid us a bit with #5572, and any other future needs for this table.


rakyi (approved):
Good idea, we should have done it a while ago!

PR #5578, opened 2025-10-24
🎉 subscribe-bar component

ikesau:
Adds an inline box to link to the subscribe page

## Screenshots / Videos / Diagrams

image

## Testing guidance

You can use the following archie:
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque eget feugiat purus. In elementum nisl orci, id viverra lacus molestie nec. Integer lacinia lorem id magna pretium blandit. Suspendisse nibh turpis, ornare vel elementum eget, aliquam vel massa. Quisque justo dui, eleifend aliquam cursus eu, hendrerit quis nisl. Pellentesque id lacinia dui. Suspendisse tempus id augue id scelerisque. Quisque feugiat non erat a fermentum. Sed vel magna est. Aenean aliquam imperdiet eros, eu tempor nisi convallis a. Nunc feugiat, arcu sit amet pellentesque sagittis, ipsum elit tincidunt arcu, quis condimentum justo leo at mauris. Donec id leo blandit, porttitor diam a, congue metus. Fusce vel nulla nulla. Ut laoreet diam mauris, non maximus dui pulvinar in.


{.subscribe-bar}
align: center
{}

Ut tincidunt tempus lorem, sed tempor orci tincidunt eget. Curabitur fermentum eros quis augue lobortis, nec rutrum tellus aliquet. Integer suscipit aliquam efficitur. Donec porttitor nulla velit, a ornare dui congue quis. Etiam nibh velit, cursus non auctor vitae, venenatis sollicitudin justo. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean consequat scelerisque arcu, nec viverra est molestie eget.

Praesent ut commodo felis, pellentesque vehicula nisi. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Pellentesque placerat orci sed dolor faucibus, facilisis varius ligula fermentum. Vestibulum eget aliquam tortor, et porta mauris. Fusce eget nunc ex. Praesent vehicula ex quis congue facilisis. Vestibulum finibus eu enim elementum tempus. Donec eu dolor felis.



{.subscribe-bar}
align: left
{}

[...]

PR #5589, opened 2025-10-28
🎉 Automatic subscribe bar

ikesau:
## Context

A follow-up to https://github.com/owid/owid-grapher/pull/5578 that adds the subscribe-bar by default to all articles and linear topic pages.

Can be overridden by writing `hide-subscribe-banner: true` in a post's front-matter.

Also renames subscribe-bar to subscribe-banner, which I think is a more natural name that the rest of the team already seems to prefer using.

image

## Testing guidance

Look at various articles and LTPs on staging


site/gdocs/components/gdocComponentUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+    for (let index = blocks.length - 1; index >= 0; index--) {
+        const block = blocks[index]
+        if (block.type === "heading") {
+            const headingBlock = block as EnrichedBlockHeading
+            if (headingBlock.level === 1) {
+                lastHeadingIndex = index
+                break
+            }
+        }
+    }
```
marcelgerber: This might be better done using [`blocks.findLastIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex).

PR #5590, opened 2025-10-28
🔨 allow queryStr override in renderSingleGrapherOnGrapherPage

Marigold:
## Summary
- Add optional `queryStr` parameter to `renderSingleGrapherOnGrapherPage` function
- Allows programmatic override of URL parameters when rendering grapher pages
- Maintains backward compatibility by defaulting to `window.location.search` when not provided

## Use case
This enables calling code to supply custom URL parameters instead of always reading from `window.location.search`, useful for programmatic rendering scenarios.

## Test plan
- [x] TypeScript compilation passes
- [x] Backward compatible - existing call sites work without changes
- [ ] Manual testing with custom `queryStr` parameter


Marigold:
Good idea, that's more elegant.

PR #5592, opened 2025-10-28
🎉 Country profiles

ikesau:
[Figma](https://www.figma.com/design/50j5XbYEx9YuFcD7GZcd6F/Country-Profiles?node-id=1428-3117&p=f&t=XZByWkxTXQfKCSFq-0)

Country profiles are a new archie gdoc type.

The idea is you write a gdoc with `type: profile` and `scope: countries`

This gdoc is a template featuring generic content and placeholder tokens (e.g.`$entityName` when we need to refer to the country by name.)

That template then gets instantiated with every country in its scope.

image

[Staging example](http://staging-site-country-profiles/admin/gdocs/1pue4IXf8zCU-qz3Hqqd6UGP2-A5LqAGZ1RbO_OSRNk8/preview) 

(You can change the previewed entity in the top of the preview settings drawer)

image

In a [follow-up PR](https://github.com/owid/owid-grapher/pull/5639), a component is added which allows us to show/hide sections based on which country is instantiated.


ikesau:
@copilot review

ikesau:
@codex review

adminSiteServer/mockSiteRouter.ts
```diff
@@ -389,7 +393,12 @@ getPlainRouteWithROTransaction(
 
         try {
             return res.send(
-                await renderGdocsPageBySlug(trx, pageNumberOrSlug, true)
+                await renderGdocsPageBySlug(
+                    trx,
+                    pageNumberOrSlug,
+                    true,
+                    OwidGdocType.DataInsight
```
ikesau: We have to add this namespace param now because slugs are no longer unique

packages/@ourworldindata/utils/src/regions.json
```diff
@@ -35,7 +35,8 @@
         "name": "Aland Islands",
         "slug": "aland-islands",
         "regionType": "country",
-        "isUnlisted": true
+        "isUnlisted": true,
+        "article": "the"
```
sophiamersmann, line 39: The `regions.json` file is automatically generated by [this script](https://github.com/owid/owid-grapher/tree/master/devTools/regionsUpdater), so your manual changes will be overwritten on the next update. You could either ask a data scientist to add a `article` column to the [ETL regions file](https://github.com/owid/etl/tree/master/etl/steps/data/garden/regions) or you could hardcode the list of countries that need an article and update the regionsUpdater script

packages/@ourworldindata/utils/src/regions.ts
```diff
@@ -355,3 +363,50 @@ export const getRegionAlternativeNames = (
     }
     return _regionAlternativeNames.get(regionName)!
 }
+
+const getRegionArticle = (region?: Region): string | undefined => {
```
sophiamersmann: Is there a reason for being so defensive? This could just be `region?.article`, no?

packages/@ourworldindata/utils/src/regions.ts
```diff
@@ -355,3 +363,50 @@ export const getRegionAlternativeNames = (
...
+        const article = (region as { article?: string }).article
+        if (typeof article === "string") {
+            const trimmed = article.trim()
+            if (trimmed.length > 0) return trimmed
+        }
+    }
+    return undefined
+}
+
+const getArticleForEntity = (entity: string): string | undefined => {
```
sophiamersmann: Same here. Let's simplify and drop the by code and by slug cases? Btw, if you're passing around entity names, you could use the EntityName type. It's of course just a string, but a bit more explicit

packages/@ourworldindata/utils/src/profiles.ts
```diff
@@ -0,0 +1,179 @@
...
+            scopeValue === "regions" ||
+            scopeValue === "all"
+        ) {
+            continue
+        }
+
+        const withoutParenthetical = removeTrailingParenthetical(scopeValue)
+        const matchingRegion =
+            getRegionByNameOrVariantName(scopeValue) ??
+            getRegionByNameOrVariantName(withoutParenthetical) ??
```
sophiamersmann: I don't understand why we do this since `Africa` and `Africa (WHO)` for example are distinct entities?

packages/@ourworldindata/utils/src/profiles.ts
```diff
@@ -0,0 +1,179 @@
...
+        if (!region?.code || !region?.name) return
+        if (!entitiesByCode.has(region.code)) {
+            entitiesByCode.set(region.code, {
+                name: region.name,
+                code: region.code,
+            })
+        }
+    }
+
+    const nonCountryRegions = regions.filter(
```
sophiamersmann: Probably only becomes relevant when the regions scope is used for the first time, but I wonder if the scope makes sense in its current form since it includes multiple 'versions' of each continent, plus other more niche aggregates like Polynesia

packages/@ourworldindata/types/src/gdocTypes/Gdoc.ts
```diff
@@ -237,6 +238,33 @@ export interface OwidGdocAnnouncementInterface extends OwidGdocBaseInterface {
     content: OwidGdocAnnouncementContent
 }
 
+export interface OwidGdocProfileContent {
+    type: OwidGdocType.Profile
+    title: string
+    authors: string[]
+    scope: string // e.g. ["countries", "regions"]
```
sophiamersmann: The comment confused me for a moment, `e.g. "countries, regions"` makes more sense here, right?

packages/@ourworldindata/utils/src/profiles.ts
```diff
@@ -0,0 +1,179 @@
...
+    const entityNameWithArticle = articulateEntity(entity.name)
+    const replacements: ReplacementPair[] = [
+        ["$entityName", entityNameWithArticle],
+        ["$entityCode", entity.code],
+        ["$noArticleEntityName", entity.name],
+    ]
+
+    const clonedContent = _.cloneDeep(content)
+
+    clonedContent.title = replaceTokens(clonedContent.title, replacements)
```
sophiamersmann: Out of curiosity, could you also just run `replaceStringsInObject(clonedContent)`?

baker/countryIndexes.tsx
```diff
@@ -80,6 +80,7 @@ export const countryIndicatorVariables = async (
         return rows.map(parseVariablesRow)
     })
 
+// TODO: what is this? check with matthieu/sophia
```
sophiamersmann: This is the first time I'm reading this code. Maybe Matthieu knows more

baker/siteRenderers.tsx
```diff
@@ -488,85 +483,10 @@ ${dataInsights
...
-    return renderToHtmlPage(
-        <LongFormPage
-            withCitation={true}
-            post={formattedCountryProfile}
-            overrides={overrides}
-            formattingOptions={formattingOptions}
-            baseUrl={BAKED_BASE_URL}
-        />
-    )
+export const flushCache = () => {
```
sophiamersmann: Do we need this although it's empty?

adminSiteServer/mockSiteRouter.ts
```diff
@@ -572,7 +564,8 @@ getPlainRouteNonIdempotentWithRWTransaction(
             const page = await renderGdocsPageBySlug(
                 trx,
                 req.params.authorSlug,
-                true
+                true,
+                OwidGdocType.Author
```
sophiamersmann: This makes the comment above unnecessary, or maybe I'm misunderstanding?

db/model/Gdoc/GdocFactory.ts
```diff
@@ -297,7 +303,8 @@ export async function getMinimalGdocBaseObjects(
 export async function getPublishedGdocBaseObjectBySlug(
     knex: KnexReadonlyTransaction,
     slug: string,
-    fetchLinkedTags: boolean
+    fetchLinkedTags: boolean,
+    type?: OwidGdocType
```
sophiamersmann: Maybe it's a bit safer to make this non-optional if slugs are not unique anymore?

adminSiteClient/GdocsSettingsForms.tsx
```diff
@@ -368,3 +375,80 @@ export const GdocAboutPageSettings = ({
...
+    selectedEntity?: string
+    setSelectedEntity: Dispatch<SetStateAction<string | undefined>>
+    entitiesInScope: { value: string; label: string }[]
+}) => {
+    return (
+        <form className="GdocsSettingsForm">
+            <div
+                className="form-group"
+                style={{
+                    background: "#f9f9f9",
```
sophiamersmann, line 399: Did you want these styles to be inlined?

packages/@ourworldindata/utils/src/profiles.ts
sophiamersmann: We also keep adding more aggregates to the regions file. If the regions scope is used, does that mean that anytime we update the regions file, we get more profiles? Not sure that's expected (not really an issue for countries because that's a relatively stable list)

packages/@ourworldindata/utils/src/profiles.ts
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    const nonCountryRegions = regions.filter(
+        (region) => !checkIsCountry(region)
+    )
+
+    for (const scopeValue of scopeValues) {
+        if (scopeValue === "countries" || scopeValue === "all") {
+            for (const country of countries) addEntity(country)
```
sophiamersmann, line 167: We also have a special "Other" category that contains unrecognised countries like Somaliland. Not saying it should be included, just wanted to make you aware!

packages/@ourworldindata/utils/src/profiles.ts
ikesau: Yeah actually this is all unnecessarily coupled given the function's so generic 😓

PR #5604, opened 2025-10-30
cleanup(gdocs): remove old chart overrides

marcelgerber:
I've converted all outstanding charts to proper narrative charts now, so we can now remove this code.

This PR is split up into commits, and it's probably easier to review commit-by-commit.

TODO:
- [x] Check all deployed content for references to `data-grapher-config`, there ideally shouldn't be any left at this point
	- I've checked this using `rg -l data-grapher-config -- ~/dist` on the prod machine
- [x] Check on staging server that archive context is properly passed along to the embedded chart


marcelgerber:
@sophiamersmann can you think of any other code that we can now remove? for reference, the PR that added all of this code is https://github.com/owid/owid-grapher/pull/2258/files?w=1.

Or is there any of these settings that you'd like to keep around? I've kept `hasTableTab` around for now, because I'm sure it'll be useful in the future even if it's not currently exposed, but the others I've removed now.

marcelgerber:
@codex review

sophiamersmann (approved):
No, I can't think of anything. I believe you've covered everything! I would remove as many hide directives as possible, which I think you have done. Keeping `hasTableTab` is also reasonable. There is some code in the ControlsRow component that annoyed me while working on small multiple maps, so I might revisit this after you merge, but there's nothing for you to address in this PR.

Thank you very much for doing this! I was annoyed more than once by having to maintain this code. I really appreciate it!

PR #5605, opened 2025-10-30
fix(mdim): properly pass archive information, such that it can be shown in an embedded mdim

marcelgerber:
The before/after here is in https://ourworldindata.org/vaccination.

| Before | After |
|--------|--------|
| CleanShot 2025-10-30 at 18 28 30 | CleanShot 2025-10-30 at 18 28 57 | 

## What's going on here:

This is the code that's rendering the `<MultiDimEmbed>`, it has `chartConfig = { archiveContext }`:
https://github.com/owid/owid-grapher/blob/8fa135a0e5b9c4bacf060e7bb05a21964575836d/site/gdocs/components/Chart.tsx#L160-L165

https://github.com/owid/owid-grapher/blob/8fa135a0e5b9c4bacf060e7bb05a21964575836d/site/gdocs/components/Chart.tsx#L114-L121

Then, `<MultiDimEmbed>` just passes these along to `<MultiDim>` as-is:
https://github.com/owid/owid-grapher/blob/8fa135a0e5b9c4bacf060e7bb05a21964575836d/site/MultiDimEmbed.tsx#L90-L97

`<MultiDim>` always sets `baseGrapherConfig.archiveContext`, even if `archiveContext` is undefined, thereby overriding whatever is in `localGrapherConfig`:
https://github.com/owid/owid-grapher/blob/8fa135a0e5b9c4bacf060e7bb05a21964575836d/site/multiDim/MultiDim.tsx#L69-L74

---

I'm not entirely sure whether there's other props in `baseGrapherConfig` that should always override what's in `localGrapherConfig`. It's your call, @rakyi, and there are also a bunch of other solution we can alternatively consider! (such as explicitly passing `archiveContext` to the `<MultiDim>` component in this case).

### DIFF
--- a/site/multiDim/MultiDim.tsx
+++ b/site/multiDim/MultiDim.tsx
@@ -165,8 +165,8 @@ export default function MultiDim({
                 if (ignoreFetchedData) return
                 const grapherConfig = {
                     ...viewGrapherConfig,
-                    ...localGrapherConfig,
                     ...baseGrapherConfig,
+                    ...localGrapherConfig,
                 }
                 if (slug) {
                     grapherConfig.slug = slug // Needed for the URL used for sharing.


rakyi (approved):
Ouch, this is tricky. Thanks for the fix! I think it's good for now.

PR #5606, opened 2025-10-30
✨ unique post slug+type when published

ikesau:
## Context

Enforces uniqueness on type+slug for published rows in posts_gdocs.

This came up because I was thinking about how to handle slugs for country profiles and in my research was surprised to see that we didn't actually have a uniqueness constraint set in the DB (we've instead been enforcing it at the app level)

I think this is a nice promise to make, and it helps clarify what the correct solution for country profiles is (i.e. only storing the `energy` part of the slug, if the full URL is `https://ourworldindata.org/profile/energy/austria`)

https://github.com/user-attachments/assets/ac4987b9-ed73-4a63-9d90-32c2299da934

## Testing guidance

Try to save an article/data-insight with a duplicated slug and confirm that it's possible
Try to publish an article/data-insight with a duplicated slug and confirm that an error throws

## Checklist
- [x] Update the documentation in db/docs (not applicable)


ikesau:
Codex raises a good point. I don't think this is worth it if we need to keep around the canonical URL check regardless.

PR #5607, opened 2025-10-31
✨ Update heading on newsletter signup popup

ikesau:
One more spot to hide this message per Charlie's request

| Before | After |
|--------|--------|
| image | image |


ikesau:
Actually I think at this point I should refactor (again) how we render this form

See https://github.com/owid/owid-grapher/issues/5608

PR #5637, opened 2025-11-06
✨ use short entity names for more chart types

### LINKED ISSUE #5596: Implement short names for all chart types
## Core problem

Long entities don't look good in crowded charts, like line charts or stacked area and bar charts. We have a short_name solution for this, but it's currently implemented for Marimekkos.

## Proposed solution

Implement short names for all chart types, especially when there are several entities selected

## Context

[Slack conversation here](https://owid.slack.com/archives/C0149NKLZAM/p1761577972777329)

There is now a `Middle East, North Africa, Afghanistan and Pakistan (WB)` region from the World Bank which could be shortened in some contexts as `MENA, Afghanistan and Pakistan (WB)`:

| Stacked area | Stacked bar |
|--------|--------|
| Image | Image |

sophiamersmann:
Fixes #5596 

Uses short names for all chart types except for the following:

* Stacked bar charts (since entity names only appear in the legend)
* Map charts (since labels only appear in the tooltip)

We used to display the short version of an entity name only if the actual name was longer than a certain threshold. But I think we can assume that if a short name is available, we want to use it in the chart area. The longer version will still be used in the tooltip, entity selector, and data table.

---
- <kbd>&nbsp;2&nbsp;</kbd> #5641 
- <kbd>&nbsp;1&nbsp;</kbd> #5637 👈

### DIFF
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
@@ -39,8 +39,10 @@ import {
 } from "../horizontalColorLegend/HorizontalColorLegends"
 import { DiscreteBarChartState } from "./DiscreteBarChartState"
 import { ChartComponentProps } from "../chart/ChartTypeMap.js"
-import { makeProjectedDataPatternId } from "./DiscreteBarChartHelpers"
-import { TextWrap } from "@ourworldindata/components"
+import {
+    fitLabelToBarHeight,
+    makeProjectedDataPatternId,
+} from "./DiscreteBarChartHelpers"
 import { OwidTable } from "@ourworldindata/core-table"
 import { HorizontalAxis } from "../axis/Axis"
 import { GRAPHER_DARK_TEXT } from "../color/ColorConstants"
@@ -53,9 +55,6 @@ const GAP__ENTITY_LABEL__BAR = 5
 
 /** The gap between the the entity label and negative value label */
 const GAP__ENTITY_LABEL__VALUE_LABEL = 10
-
-/** If an entity name exceeds this width, we use the short name instead (if available) */
-const SOFT_MAX_LABEL_WIDTH = 90
 
 export interface Label {
     valueString: string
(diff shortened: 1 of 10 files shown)


sophiamersmann:
@codex review

packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChart.tsx
```diff
@@ -136,48 +135,16 @@ export class DiscreteBarChart
...
-            if (tooLong && series.shortEntityName) {
-                label = new TextWrap({
-                    text: series.shortEntityName,
-                    maxWidth: label.maxWidth,
-                    ...this.entityLabelStyle,
-                })
-            }
-
-            return { ...series, label }
+            return { ...series, label: labelWrap }
```
sophiamersmann, line 147: No need to review. I rewrote this in the next PR

packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartHelpers.ts
```diff
@@ -1,6 +1,54 @@
...
+
+        // Labels shouldn't exceed this width
+        if (currMaxWidth > maxWidth) break
+
+        labelWrap = makeTextWrap(currMaxWidth)
+        step += 1
+    }
+
+    return labelWrap
+}
```
sophiamersmann, line 54: No need to review. I rewrote this in the next PR

packages/@ourworldindata/grapher/src/stackedCharts/StackedDiscreteBars.tsx
```diff
@@ -248,45 +245,16 @@ export class StackedDiscreteBars
...
-            if (tooLong && item.shortEntityName) {
-                label = new TextWrap({
-                    text: item.shortEntityName,
-                    maxWidth: label.maxWidth,
-                    ...this.labelStyle,
-                })
-            }
-
-            return { ...item, label }
+            return { ...item, label: labelWrap }
```
sophiamersmann, line 257: No need to review. I rewrote this in the next PR

packages/@ourworldindata/grapher/src/stackedCharts/MarimekkoChart.tsx
```diff
@@ -579,16 +576,9 @@ export class MarimekkoChart
...
-        let label = item.entityName
-        let labelBounds = Bounds.forText(label, {
-            fontSize,
-        })
-        if (labelBounds.width > SOFT_MAX_LABEL_WIDTH && item.shortEntityName) {
-            label = item.shortEntityName
-            labelBounds = Bounds.forText(label, {
-                fontSize,
-            })
-        }
```
sophiamersmann, line 591: We used to measure the original label first, but I think it's ok to just use the shorter version in the chart area if available

ikesau (approved):

PR #5638, opened 2025-11-06
🎉 add Swap axes button for single X/Y charts

### LINKED ISSUE #4496: Add an indicator swapper for scatterplots in Grapher
## Core problem

When creating a scatterplot, it is often the case we want to explore how the view looks with the indicators swapped between axes. To try that, we need to manually search again for the indicators for each axis, even if the indicators are already in the view. This takes some time and it's prone to error when indicators have complex names.

## Proposed solution
To include a "Indicator swapper" button that swaps the indicators shown in the x and y axes.

## Context

[Post in Slack](https://owid.slack.com/archives/C0149NKLZAM/p1738084196885549)

sophiamersmann:
Resolves #4496


sophiamersmann:
@codex review

PR #5641, opened 2025-11-07
🐛 fix overlapping discrete bar chart labels

### LINKED ISSUE #5461: Discrete bar chart labels overlap
## Description

In [this chart](https://ourworldindata.org/grapher/rate-of-violent-deaths-non-state-societies), the discrete bar chart labels overlap on smaller screens.

## Expected behaviour

The labels should shrink and not overlap. The labels also shouldn't take up that much space if possible.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/rate-of-violent-deaths-non-state-societies
2. Make screen smaller (or check it on your phone)
3. Note that labels overlap for some screen sizes

## Screenshots

Image

sophiamersmann:
Fixes #5461

Improves the labelling algorithm for discrete bar charts in cases with limited space (see screenshots).

It works in three steps: First, we find the minimum width that produces at most the available number of lines. If that doesn't work (i.e., we're at maximum width, but the text wraps into more lines than available), we reduce the font size until the text fits. If the font size becomes too small to read, we truncate the label as a last resort.

I'm not convinced that truncating lines is the right approach because it hides content, but using a comically small font size, which is difficult to read, also hides content in a way. So I went with the ellipsis solution hoping that we will only need this in very few cases.

The new strategy introduces line breaks a bit more frequently than the previous one. In some cases, that's ok, in other cases it would be better to avoid line breaks or, ideally, to implement them in a balanced manner, ensuring that all lines have approximately the same length. Idk, but for now, I think this is a net positive.

Examples:
- http://staging-site-fix-overlapping-labels/grapher/rate-of-violent-deaths-non-state-societies
- http://staging-site-fix-overlapping-labels/grapher/grocery-bag-environmental-impact

| Before | After |
|--------|--------|
| Screenshot 2025-11-07 at 17 15 04 | Screenshot 2025-11-10 at 11 23 11 | 
| Screenshot 2025-11-07 at 17 14 28 | Screenshot 2025-11-10 at 11 23 25 | 

---
- <kbd>&nbsp;2&nbsp;</kbd> #5641 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5637

### DIFF
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartHelpers.ts
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartHelpers.ts
@@ -1,4 +1,6 @@
-import { TextWrap } from "@ourworldindata/components"
+import { TextWrap, shortenWithEllipsis } from "@ourworldindata/components"
+import { EntityName } from "@ourworldindata/types"
+import { max } from "lodash-es"
 
 interface FontSettings {
     fontSize: number
@@ -20,8 +22,9 @@ export function makeProjectedDataPatternId(color: string): string {
  * 1. Calculate the maximum number of lines that can fit in the bar height
  * 2. Use binary search to find the minimum width that produces at most maxLines
  * 3. If the label doesn't fit even at maxWidth, reduce font size until it fits
+ * 4. If the label still doesn't fit, truncate the label with ellipsis
  */
-export function wrapLabelForHeight({
+function wrapLabelForHeight({
     label,
     availableHeight,
     minWidth,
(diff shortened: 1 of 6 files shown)


packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartHelpers.ts
```diff
@@ -1,54 +1,250 @@
...
+            low = mid + step
+        }
+    }
+
+    return bestWidth
+}
+
+/**
+ * Decrease font size until the label fits within availableHeight at maxWidth.
+ * Stops at minFontSize (typically 8px).
```
ikesau: One thing I found a little confusing here is that the minimum of `minFontSize` is **6**px, but it's actually always 8px because of `originalFontSize` 😅

sophiamersmann: I think the comment is out of date as well

PR #5642, opened 2025-11-07
🐛 use debounced change handlers for form fields

sophiamersmann:
Fixes https://github.com/owid/owid-grapher/issues/4887


adminSiteClient/Forms.tsx
```diff
@@ -70,12 +81,18 @@ export class TextField extends React.Component<TextFieldProps> {
     }
 
     @bind onBlur(e: React.FocusEvent<HTMLInputElement>) {
+        this.debouncedOnValue.cancel()
```
marcelgerber: Don't think we should do this: This means that when you quickly enter something and then tab away, the new value is never applied.

PR #5649, opened 2025-11-10
🐛 drop entities without plottable data

### LINKED ISSUE #3492: Marimekko: Entities offered for selection that have no data for any year
## Description

~[This chart](https://ourworldindata.org/grapher/share-of-people-who-say-that-looking-after-the-environment-is-important-to-them?time=latest&country=~OWID_GFR) offers "West Germany" for selection but "West Germany" doesn't have any data~ Unfortunately, the link is broken. [Here](https://ourworldindata.org/grapher/share-of-children-reaching-sufficient-reading-comprehension-by-the-end-of-lower-secondary-age?time=2017&country=~OWID_WRL) is another example.

## Expected behaviour

We should remove all entities from the selector that don't have data. This includes:
- Entities that are filtered because they're not in any color group (if the setting is enabled)
- Entities the are excluded, i.e. specified in `excludedEntities`
- If the timeline is disabled, we should remove entities that don't have data for the current year ([example chart](https://ourworldindata.org/grapher/multidimensional-poverty-index-mpi?tab=chart&country=OWID_YPR~YEM~OWID_YAR))

## Steps to reproduce

1. Go to this https://ourworldindata.org/grapher/share-of-children-reaching-sufficient-reading-comprehension-by-the-end-of-lower-secondary-age?time=2017&country=~OWID_WRL
2. Select "World"
3. Look at every year
4. Note that "World" doesn't have data for any of the years

## Additional context

Note that World does have data in the input table but isn't rendered because it doesn't belong in any of the color groups.

See #3455 for an attempt to fix and why it wasn't successful.

sophiamersmann:
Fixes #3492

Some entities only have data for the colour dimension, so they technically exist in the data (which is why they appear in the entity selector), but they're not plotted because they don't have any actual (plottable) data.

The SVG tester reports color differences because after dropping entities, we might miss some color categories and colors are applied in order. For the continents, it would be nice if we ensured that semantic colors are consistent across all charts (e.g. Europe is always denim, etc). This is fixed in the follow up PR

Example: Search for 'Monaco' in https://ourworldindata.org/grapher/gdp-per-capita-worldbank?tab=line&time=1990..2024&country=IND~CHN~MCO~LUX

---
This is **part 1 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #5672 
- <kbd>&nbsp;3&nbsp;</kbd> #5671 
- <kbd>&nbsp;2&nbsp;</kbd> #5675 
- <kbd>&nbsp;1&nbsp;</kbd> #5649 👈

### DIFF
--- a/packages/@ourworldindata/core-table/src/OwidTable.test.ts
+++ b/packages/@ourworldindata/core-table/src/OwidTable.test.ts
@@ -677,3 +677,55 @@ gdp	annotation`
         (table.get("gdp").def as OwidColumnDef).annotationsColumnSlug
     ).toEqual("annotation")
 })
+
+describe("dropEntitiesThatHaveNoDataInAllColumns", () => {
+    it("drops only entities missing data in every specified column", () => {
+        const csv = `year,entityName,gdp,population,lifeExpectancy
+2020,USA,1000,330,78
+2020,China,2000,1400,76
+2020,OnlyGDP,500,,
+2020,OnlyPopulation,,100,
+2020,OnlyLifeExpectancy,,,80
+2020,NoDataAtAll,,,`
+        const table = new OwidTable(csv, [
+            { slug: "gdp", type: ColumnTypeNames.Numeric },
+            { slug: "population", type: ColumnTypeNames.Numeric },
+            { slug: "lifeExpectancy", type: ColumnTypeNames.Numeric },
+            { slug: "year", type: ColumnTypeNames.Year },
+        ])
+
+        const result = table.dropEntitiesThatHaveNoDataInAllColumns([
+            "gdp",
+            "population",
+            "lifeExpectancy",
+        ])
+
+        // Should keep all entities except NoDataAtAll
+        expect(result.availableEntityNames).toEqual([
+            "USA",
+            "China",
+            "OnlyGDP",
+            "OnlyPopulation",
+            "OnlyLifeExpectancy",
+        ])
+    })
+
+    it("filters correctly when checking a subset of columns", () => {
+        const csv = `year,entityName,gdp,population,continent
+2020,USA,1000,330,North America
+2020,China,,1400,Asia
+2020,India,,500,Asia`
+        const table = new OwidTable(csv, [
+            { slug: "gdp", type: ColumnTypeNames.Numeric },
+            { slug: "population", type: ColumnTypeNames.Numeric },
+            { slug: "year", type: ColumnTypeNames.Year },
+            { slug: "continent", type: ColumnTypeNames.Continent },
+        ])
+
+        const result = table.dropEntitiesThatHaveNoDataInAllColumns(["gdp"])
+
+        // Only USA should remain when checking just GDP
+        expect(result.availableEntityNames).toEqual(["USA"])
+        expect(result.numRows).toEqual(1)
+    })
+})
--- a/packages/@ourworldindata/core-table/src/OwidTable.ts
+++ b/packages/@ourworldindata/core-table/src/OwidTable.ts
@@ -337,6 +337,64 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
         )
     }
 
+    /**
+     * Drop _all rows_ for an entity if the entity has no data in ALL of the specified columns.
+     * An entity is kept if it has at least one valid value in ANY of the columns.
+     */
+    dropEntitiesThatHaveNoDataInAllColumns(columnSlugs: ColumnSlug[]): this {
+        const indexesByEntityName = this.rowIndicesByEntityName
+
+        // Start with all entities and remove those that have no data in ALL columns
+        const entityNamesToKeep = new Set(indexesByEntityName.keys())
+
+        for (const entityName of indexesByEntityName.keys()) {
+            const indicesForEntityName = indexesByEntityName.get(entityName)
+            if (!indicesForEntityName)
+                throw new Error("Unexpected: entity not found in index map")
+
+            // Check if the entity has at least one valid value in ANY column
+            let hasValidValueInAnyColumn = false
+
+            for (const slug of columnSlugs) {
+                const col = this.get(slug)
+
+                const hasSomeValidValueForEntityInCol =
+                    indicesForEntityName.some((index) =>
+                        isNotErrorValue(col.valuesIncludingErrorValues[index])
+                    )
+
+                if (hasSomeValidValueForEntityInCol) {
+                    hasValidValueInAnyColumn = true
+                    break // No need to check other columns for this entity
+                }
+            }
+
+            // If the entity has no valid value in any column, drop it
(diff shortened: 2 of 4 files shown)


packages/@ourworldindata/core-table/src/OwidTable.ts
```diff
@@ -319,6 +319,64 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
         )
     }
 
+    /**
+     * Drop _all rows_ for an entity if the entity has no data in ALL of the specified columns.
+     * An entity is kept if it has at least one valid value in ANY of the columns.
+     */
+    dropEntitiesThatHaveNoDataInAllColumns(columnSlugs: ColumnSlug[]): this {
```
marcelgerber, line 344: Leaving my thought process here: I was wondering if we could instead simply use `dropRowsWithErrorValuesForAllColumns`, which does quite a similar thing (the only difference being that it drops _all_ rows that have error values in all specified cols, and not "just" entities where all rows are empty in these cols).

But thinking about it a bit more, that's probably a tad too much, and might negatively impact some charts when it's applied to all chart types directly in `tableAfterAuthorTimelineAndEntityFilter`.
Although I'm not quite sure now, is there any chart type that does something with empty table rows? Probably not, right?

packages/@ourworldindata/core-table/src/OwidTable.ts
```diff
@@ -319,6 +319,64 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
...
+        const indexesByEntityName = this.rowIndicesByEntityName
```
marcelgerber, line 345: If we opt to keep this around and not go with `dropRowsWithErrorValuesForAllColumns`, then I believe we can implement it differently to make it more efficient:
1. Build the union of all of the columns' `validRowIndices`
2. Then, compute the entity name set found at these indices
3. Then, filter by these entities (just like `entityNamesToKeep` here)

Or actually, another idea would be to refactor `dropRowsWithErrorValuesForAllColumns` (and likewise `...ForAnyColumn`) to be split in two, one method that just _finds_ all the row indices that would be dropped (or kept), and then another method that can then apply additional logic on top of that (like here: finding the entity names affected, and filtering by these).
I see already that you've done some refactoring and improvements in #5671 and #5672, but think this would still be a valuable improvement on top of this.

packages/@ourworldindata/core-table/src/OwidTable.test.ts
```diff
@@ -646,3 +646,55 @@ gdp	annotation`
         (table.get("gdp").def as OwidColumnDef).annotationsColumnSlug
     ).toEqual("annotation")
 })
+
+describe("dropEntitiesThatHaveNoDataInAllColumns", () => {
```
marcelgerber, line 681: These tests, as they are now, also pass with `dropRowsWithErrorValuesForAllColumns`. It would be great if you can rewrite them so they capture the differences between the two.

packages/@ourworldindata/core-table/src/OwidTable.ts
sophiamersmann: Using `dropRowsWithErrorValuesForAllColumns` was my first attempt but this results in color/size values not being defined in some cases (although infinite tolerance is used), and I wasn't sure how to best fix it, so I went with the `dropEntitiesThatHaveNoDataInAllColumns` approach. 

Basically, when using `dropRows..`, tolerance interpolation for the x/y dimensions can recreate rows that were previously dropped (e.g. Canada has data for 2001, 2011, 2016, 2021, and y-tolerance adds a row for 2023). But those newly created rows don't have color/size values because color/size tolerance already happened earlier in the pipeline. The only fix I could think of is re-applying color/size tolerance again after x/y tolerance has been applied. I don't know, maybe that's better than `dropEntitiesThatHaveNoDataInAllColumns` or maybe there's another solution all together, not sure.

marcelgerber: Hm. So, in your case, `dropEntitiesThatHaveNoDataInAllColumns` was being applied then _before_ applying tolerance, correct? What if we instead applied it after running tolerance, that should work, no? (obviously that's also a bit annoying, because we probably need to also run it for `tableForSelection`).

But maybe this all is also getting quite complicated now, and it's just easier - and easier to reason about - the way you do it here, with `dropEntitiesThatHaveNoDataInAllColumns`.

The only other thing that'd lean me towards `dropRowsWithErrorValuesForAllColumns` is if it also can resolve #5768 - which I think it might be.

sophiamersmann: It's applied as you see here, after color tolerance and before x/y tolerance. Moving `dropRows...` after x/y tolerance should probably work. I haven't really considered it because I didn't want to meddle with `transformTableForSelection`.

If we keep `dropRows...` in GrapherState, then I think that should solve #5768, but if we move it after tolerance, then I don't think it does because at that point, tolerance will have filled the 'artificial' leading/trailing times with values.

Another way of doing this which I considered was moving color/size tolerance back to the transformTable functions, and fixing the issue that made us move color/size tolerance to the beginning of the pipeline some other way, but that also seemed complicated, so I dropped it

marcelgerber: Ah, makes sense! Do you recall now what the reason was for moving color/size tolerance out of transformTable? Is it to do with joining a mix of daily and yearly indicators?

sophiamersmann: This was in response to Lucas [updating the Continents indicator](https://github.com/owid/owid-grapher/issues/4176) which surfaced [this bug](https://github.com/owid/owid-grapher/issues/4284): All dots showed up in gray for some scatter plots because the new Continents variable has data for the year 2023, but the charts only has data up to 2021 for example, which meant that `tableAfterAuthorTimelineAndEntityFilter` drops all data for 2022 and 2023, including all color values. Moving the color tolerance earlier in the pipeline worked at the time, but now that we're running into problems here, maybe we can just do something more specific to solve that problem? Now that I'm writing this I also remember [this issue](https://github.com/owid/owid-grapher/issues/4872) you opened a while ago. If we added special treatment for color and size variables that way, then probably we can move color tolerance back to the transformTable function safely?

sophiamersmann:
* **#5649** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5649?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5649?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sophiamersmann:
Ah, I didn't realise there was more to do here. I don't have time now, so I'll just close it

PR #5664, opened 2025-11-13
🎉 Raycast snippets

ikesau:
Adds a script that generates [Raycast Snippets](https://manual.raycast.com/snippets/how-to-import-snippets) from our `exampleEnrichedBlocks.ts` file and adds a check in the test command to ensure none are missing.

We can also designate blocks that don't need snippets with the `EXEMPT_BLOCK_TYPES` constant.


mlbrgl (approved):
Such a good idea!

Imported and tested into a gdoc... it's like a dream finally coming true 😭🥹

So you're thinking of running this manually once in a while?

ikesau:
Yeah, we can experiment with maintenance and distribution, but at a minimum, you'll have to add a snippet each time you add a new archie component (or add it to the excludes list)

I think it'll also be nice to invite people to add more complex snippets also (`sticky-right-with-chart`/`sticky-left-with-chart` was my toy example of this)

Issue #5583, opened 2025-10-27
Chart references counted differently across Grapher admin


paarriagadap:
## Description

There are inconsistent counts of chart references in chart tables and the Refs tab in the Grapher admin.

## Expected behaviour

Reference counts should be the same everywhere to avoid confusion.

## Steps to reproduce

1. Go to [this chart table](https://admin.owid.io/admin/charts?chartSearch=extreme+poverty).
2. See that there are 17 references and 2 narrative charts
3. Open the very same chart (Edit button, or click [here](https://admin.owid.io/admin/charts/5111/edit))
4. The Refs tab shows 21 references

## Screenshots

See chart table detail for the chart described above

<img width="400" height="402" alt="Image" src="https://github.com/user-attachments/assets/bcb632cc-014c-44bf-92a6-c571c057d836" />

And Refs tab detail
<img width="216" height="124" alt="Image" src="https://github.com/user-attachments/assets/948571a4-969a-4625-a050-ff8b8a8b6d54" />

## Additional context

See [Slack conversation](https://owid.slack.com/archives/C46U9LXRR/p1760638028762849). @rakyi says

> At minimum I can see that in the list view we count only published explorers, while in the detail view we also include unpublished ones. There might be other differences.

PR #5665, opened 2025-11-13
🐛 Improve chart refs

ikesau:
Somewhat resolves https://github.com/owid/owid-grapher/issues/5583

We had a bit of ambiguity in our refs tab / chart index page because both counts are derived in different ways.

The chart index page uses the `chart_references_view` view, which only considers published gdocs, whereas the chart editor page also looks at unpublished data insights.

I think this is useful behaviour: we want to know if a data insight has already been written about a chart, even if it hasn't yet been published.

So this PR adds some language to clarify that for our authors.

It also wasn't counting data insights that linked to charts that in ways other than the `grapher-url` front-matter property, so I fixed that by including data insights in the `getGdocsPostReferencesByChartId` function.

Consequently, data insights then showed up in the old grapher page Related Writing section, so I had to add `type` to a few SQL statements so that we can call `getCanonicalUrl` and add the `/data-insight/` sub-path where necessary.

### Example 1 (Data insight is published, shown twice)
image

### Example 2 (Data insight is unpublished, shown once)
image

### Related Articles
https://github.com/user-attachments/assets/0d7d1e9b-b52b-4a1a-99e0-d30998b0d0af

### DIFF
--- a/db/model/Post.ts
+++ b/db/model/Post.ts
@@ -270,7 +270,8 @@ export const getWordpressPostReferencesByChartId = async (
                 p.title,
                 p.slug,
                 p.id,
-                CONCAT("${BAKED_BASE_URL}","/",p.slug) as url
+                CONCAT("${BAKED_BASE_URL}","/",p.slug) as url,
+                'article' as type
             FROM
                 posts p
                 JOIN posts_links pl ON p.id = pl.sourceId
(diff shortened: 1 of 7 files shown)


db/model/Post.ts
```diff
@@ -321,16 +322,16 @@ export const getGdocsPostReferencesByChartId = async (
...
             FROM
                 posts_gdocs pg
                 JOIN posts_gdocs_links pgl ON pg.id = pgl.sourceId
                 JOIN chart_slug_mapping csm ON pgl.target = csm.target_slug
             WHERE
                 pg.type NOT IN (
                     '${OwidGdocType.Fragment}',
-                    '${OwidGdocType.AboutPage}',
-                    '${OwidGdocType.DataInsight}'
+                    '${OwidGdocType.AboutPage}'
```
marcelgerber: is there a good reason why we're excluding about pages?

ikesau: Nope, no good reason! Changed it to only exclude fragments 👍

Issue #5077, opened 2025-07-02
Cycle 2025.4: Static viz admin


ikesau:
A new space in the Grapher admin to manage and add additional metadata to static viz graphics, as well as a new Archie component to specify and render them with.

## Motivation

Our static viz is often very high quality, but it's hard/impossible for our readers to understand, audit, recreate, or search for them. A richer set of metadata attached to our static viz will allow users to learn more about how each static viz is made, download its data and search for them on our site.

## Proposed scope

[Figma rough sketches](https://www.figma.com/design/AzeEMejpYQ624hDckcA4oY/Static-Viz-admin?node-id=0-1&p=f&t=CWcvWjvXXCcGy54u-0)

A `static_viz` is going to be a new first-class citizen in the DB that can refer to up to 2 rows from the `images` table and add metadata to them.

- [x] `static_viz` table in the DB (3h)
- [x] CRUD API for static viz (1d)
- [x] Admin table for viewing static viz (3d)
- [x] Admin page for creating/editing static viz (1w)
- [ ] ArchieML component for rendering static viz (design pending) (3d-1w)
- [x] GPT text extraction button (1d)

### Might do
- [ ] Algolia index creation (3d)
- [ ] Add way to upgrade existing images into static viz (1d)

danyx23:
@ikesau could you update this issue to indicate the current status please? Thanks!

PR #5667, opened 2025-11-14
Move `Modal.tsx` to `@ourworldindata/components`

ikesau:
Allows us to reuse the Modal component for the upcoming [Rich Static Viz component](https://github.com/owid/owid-grapher/issues/5077).

I removed the logic that purportedly allowed users to click on e.g. the settings drawer without dismissing the modal, as the `.modal-overlay` was preventing that from working anyway. Maybe I've missed a case where it was needed?

**Before**

https://github.com/user-attachments/assets/6ed13346-e326-4fee-9b0b-c63df2031e2d

**After**

https://github.com/user-attachments/assets/14e8eb1f-9515-4bdc-acef-8789cb88c8d6


ikesau:
Nvm, this is more trouble than it's worth

PR #5686, opened 2025-11-19
✨ download modal improvements

### LINKED ISSUE #5014: Improve the download modal
Bobbie identified a few papercuts in the new download modal (see [Figma](https://www.figma.com/board/BhbeWpdM1pWr2uZajFkbpo/Retrospective-on-redesigned-Grapher-download-modal?node-id=0-1&p=f&t=csY0Mqp2jv3QB5wu-0))

Here are the individual tasks we should consider doing as part of this:
- [ ] On the 2 download data buttons (full or displayed), show how many rows of data the csv file will have
- [ ] Move the data API options below the URL fields (some people seem to be confused by the radio buttons and seem to think they are click buttons)
- [ ] On the table tab, when opening the download modal, default to showing the "Data" tab and on the Viz tab change the text to say something like "Table can't be exported to image"
- [ ] When "show selection only" is enabled in the table tab then the "download only displayed data" button should only download this subset.

sophiamersmann:
Resolves #5014

A collection of download modal improvements, as outlined in #5014:

- On the 2 download data buttons (full or displayed), show how many rows of data the csv file will have
- Move the data API options below the URL fields (some people seem to be confused by the radio buttons and seem to think they are click buttons)
- On the table tab, when opening the download modal, default to showing the "Data" tab and on the Viz tab change the text to say something like "Table can't be exported to image"
- When "show selection only" is enabled in the table tab then the "download only displayed data" button should only download this subset.

It makes sense to review this by commit.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
+++ b/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
@@ -172,6 +172,7 @@ import {
     EntityRegionTypeGroup,
     groupEntityNamesByRegionType,
     EntityNamesByRegionType,
+    isEntityRegionType,
 } from "./EntitiesByRegionType.js"
 import {
     getEntityNamesParam,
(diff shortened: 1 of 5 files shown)


packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -957,6 +958,34 @@ export class GrapherState {
 
         return table
     }
+
+    @computed get filteredTableForDisplay(): OwidTable {
```
sophiamersmann, line 962: I'm not very happy with this new `filteredTableForDisplay` property, but couldn't think of another way

PR #5687, opened 2025-11-19
Update feedback modal suggestions

rakyi:
As [requested by Charlie](https://owid.slack.com/archives/C03NV9Z3YSV/p1762508260754189).

### DIFF
--- a/site/Feedback.tsx
+++ b/site/Feedback.tsx
@@ -52,15 +52,29 @@ class Feedback {
 }
 
 const vaccinationRegex = /vaccination|vaccine|doses|vaccinat/i
-const licensingRegex = /license|licensing|copyright|permission|permit/i
+const licensingRegex = /license|licence|licensing|copyright|permission|permit/i
 const citationRegex = /cite|citation|citing|reference/i
 const translateRegex = /translat/i
+const fundingRegex = /\b(fund|funds|funding|funded|funder)\b/i
+const reusingChartsRegex = /(use|reuse|using|reusing)\s+(chart|image|picture)/i
+const reusingDataRegex =
+    /(use|reuse|using|reusing|utilize|utilizing|utilise|utilising)\s+data/i
+const visualizationToolRegex =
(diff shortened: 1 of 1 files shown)


site/Feedback.tsx
```diff
@@ -55,12 +55,25 @@ const vaccinationRegex = /vaccination|vaccine|doses|vaccinat/i
 const licensingRegex = /license|licensing|copyright|permission|permit/i
 const citationRegex = /cite|citation|citing|reference/i
 const translateRegex = /translat/i
+const fundingRegex = /\b(fund|funds|funding|funded|funder)\b/i
+const reusingChartsRegex = /(use|reuse|using|reusing)\s+(chart|image|picture)/i
+const reusingDataRegex = /(use|reuse|using|reusing|utilize|utilizing)\s+data/i
```
ikesau: might want to add UK spelling here too (`utilise|utilising`)

site/Feedback.tsx
```diff
@@ -55,12 +55,25 @@ const vaccinationRegex = /vaccination|vaccine|doses|vaccinat/i
...
+const visualizationToolRegex =
+    /grapher|grapher\s+reusability|(use|reuse|using|reusing)\s+grapher|data\s+viz\s+tool|data\s+visualization\s+tool|visualization\s+software/i
```
ikesau: just checking, this only matches "use grapher" and not "use data visualization tool", right?

rakyi: It matches both.

site/Feedback.tsx
rakyi: Added various British spelling variants.

site/Feedback.tsx
ikesau: image

It doesn't, because a different regex eats it first.

No biggie though 🙂

rakyi: Ah, I didn't realize it was being matched by a different regex. Thought you meant it might not be matched at all. Fixed in master.

Issue #5207, opened 2025-07-30
Marimekko charts add 'fake years' of data by including regions in the tolerance


spoonerf:
## Description

In [this chart](https://ourworldindata.org/grapher/maternal-mortality?tab=marimekko&time=latest&country=) the data goes up to 2020, but the Marimekko view shows up to 2023, because this is the year the world regions indicator, which is used for colour, goes up to 2023. Because the tolerance is set to 5 it pushes the max year of the chart up to 2023. 

## Expected behaviour

The max year shown in a chart should be limited to the max time in the data being displayed (and not the max time of the variables used for colour or size). 

## Steps to reproduce

1. Go to this URL https://ourworldindata.org/grapher/maternal-mortality?tab=marimekko&time=latest&country=
2. Scroll the year to the latest value; it should be 2023.

Examples:

* https://ourworldindata.org/grapher/share-of-students-at-end-of-primary-education-achieving-minimum-math-proficiency
* https://ourworldindata.org/grapher/share-of-students-at-end-of-lower-secondary-education-achieving-minimum-math-proficiency
* https://ourworldindata.org/grapher/daily-income-or-consumption-of-the-richest-10-marimekko?tab=map

## Screenshots

Marimekko view:
<img width="1003" height="715" alt="Image" src="https://github.com/user-attachments/assets/045ac291-c406-4ea4-abfd-c7fc8aab0cf6" />

Raw data:
<img width="857" height="623" alt="Image" src="https://github.com/user-attachments/assets/eb87531f-63ad-40e7-992c-3b47601829f6" />

sophiamersmann:
@spoonerf, the chart no longer works as an example for this bug because its data has been updated. Do you know of another example chart with this bug?

spoonerf:
Yep sure, here are some: 

* https://ourworldindata.org/grapher/share-of-students-at-end-of-primary-education-achieving-minimum-math-proficiency
* https://ourworldindata.org/grapher/share-of-students-at-end-of-lower-secondary-education-achieving-minimum-math-proficiency
* https://ourworldindata.org/grapher/daily-income-or-consumption-of-the-richest-10-marimekko?tab=map

sophiamersmann:
Thanks a lot! Added them to the pr description

PR #5694, opened 2025-11-20
🐛 prevent applying tolerance twice

### LINKED ISSUE #4891: Marimekko: Tolerance range too wide
## Description

Tolerance is set to 3 for the y-indicator in [this chart](https://ourworldindata.org/grapher/share-of-children-reaching-sufficient-reading-comprehension-by-the-end-of-lower-secondary-age). Belarus has data for the year 2018. Tolerance should thus add data points for years 2015...2021, but currently years 2012...2023 show the 2018 value.

## Expected behaviour

Only years 2015...2021 should have interpolated data.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/share-of-children-reaching-sufficient-reading-comprehension-by-the-end-of-lower-secondary-age?country=~BLR
2. Drag the timeline handle and note that Belarus has interpolated data for years 2012...2023

sophiamersmann:
Resolves #4891

Example: https://ourworldindata.org/grapher/share-of-children-reaching-sufficient-reading-comprehension-by-the-end-of-lower-secondary-age?time=2012&country=~BLR

If a column has tolerance=3, and an entity (e.g. Belarus) has data at year 2018, that data should be visible for years 2015-2021 (exactly ±3 years from 2018). However, the bug caused the data to appear at years 2012-2024 (±6 years), effectively
applying the tolerance twice.

The data transformation pipeline has two stages where tolerance is applied:
1.	Interpolation (in transformTable):
	- Creates a complete table with all entity-year combinations
	- Fills missing years using tolerance (e.g., 2015-2021 get value from 2018)
	- Years outside tolerance (e.g., 2014) remain empty
2.	Filtering (in filterByTargetTimes):
	- For a target year (e.g., 2014), finds the closest available year
	- Originally used tolerance AGAIN to search (looking ±3 years from 2014)
	- This could find 2015, which is within tolerance of 2014
	- Result: 2018 data shown at 2014 (via 2015), which is 4 years away—outside the intended tolerance

Normally this bug is hidden because after interpolation (step 1), every year has a row. The bug only manifests when some rows are removed between interpolation and filtering, such as when showNoDataArea=false in Marimekko charts.

Fixed by dropping tolerance from `filterByTargetTimes`.

### SVG tester

This happens because the y-axis has quasi infinite tolerance (9999), but the x-axis doesn't have tolerance set, which is why then most rows are dropped. This can be fixed by assigning an appropriate tolerance value for the x-axis (which I did on prod and [staging](http://staging-site-fix-marimekko-tolerance/admin/charts/5349/edit))

Screenshot 2025-11-25 at 10 12 05

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherState.test.ts
+++ b/packages/@ourworldindata/grapher/src/core/GrapherState.test.ts
@@ -28,6 +28,7 @@ import {
     SampleColumnSlugs,
     SynthesizeGDPTable,
     OwidTable,
+    ErrorValueTypes,
 } from "@ourworldindata/core-table"
 import { legacyToCurrentGrapherQueryParams } from "./GrapherUrlMigrations"
 import { setSelectedEntityNamesParam } from "./EntityUrlBuilder"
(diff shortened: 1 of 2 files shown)


sophiamersmann:
@codex review

sophiamersmann:
I believe that's tracked in [this separate issue](https://github.com/owid/owid-grapher/issues/5207). It's on my list of things to do, but I think I'd like to merge this one as is.

PR #5706, opened 2025-11-24
Fix: Add typecheck:functions script and resolve type errors

Marigold:
This PR adds a new script `yarn typecheck:functions` to properly typecheck the `functions/` directory.

Changes:
- Added `typecheck:functions` script to `package.json`.
- Configured `functions/tsconfig.json` to use project references for local packages, ensuring correct type resolution.
- Added `functions/types.d.ts` to handle `.wasm` and `.bin` imports.
- Fixed various type errors in `functions/` code (mostly related to `caches.default`, `Response` types, and global `window` usage).

PR #5707, opened 2025-11-24
enhance: add typechecking for functions

marcelgerber:
Fixes all current type errors inside `functions`, and includes them in our `typecheck` step.

Supersedes #5706.

In contrast to #5706, this is not its own command but is instead integrated into the normal `yarn typecheck`.

## Why do we need to yarn patch `preact`?

This is super annoying. Preact's types includes a `/// <reference lib="dom" />` line, and that means that globally, everything downstream will also pull in the "dom" types (with type definitions like `caches`, `Response`, etc. that differ from CF's types).
Even worse, there is no "good" way to either ignore this line, explicitly exclude "dom", or at least explicitly mark `@cloudflare/workers-types` to be of higher priority. So, a patch it is.

### DIFF
--- a/functions/api/search/index.test.ts
+++ b/functions/api/search/index.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from "vitest"
+import { describe, it, expect, vi, beforeEach, assert } from "vitest"
 import { onRequestGet } from "./index.js"
 import * as searchApi from "./searchApi.js"
 import type { Env } from "../../_common/env.js"
@@ -88,6 +88,7 @@ describe("Search API endpoint", () => {
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            assert(typeof body === "object" && "error" in body)
             expect(body.error).toBe("Invalid type parameter")
         })
     })
@@ -158,6 +159,7 @@ describe("Search API endpoint", () => {
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            assert(typeof body === "object" && "error" in body)
             expect(body.error).toBe("Invalid page parameter")
         })
 
@@ -172,6 +174,7 @@ describe("Search API endpoint", () => {
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            assert(typeof body === "object" && "error" in body)
             expect(body.error).toBe("Invalid page parameter")
         })
 
@@ -186,6 +189,7 @@ describe("Search API endpoint", () => {
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            assert(typeof body === "object" && "error" in body)
             expect(body.error).toBe("Invalid hitsPerPage parameter")
         })
 
@@ -200,6 +204,7 @@ describe("Search API endpoint", () => {
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            assert(typeof body === "object" && "error" in body)
             expect(body.error).toBe("Invalid hitsPerPage parameter")
         })
     })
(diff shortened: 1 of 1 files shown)


rakyi (approved):
Cool that you found a workaround for the Preact issue!

However, it makes me wonder: wouldn't everyone who is running a Preact app on Cloudflare Workers hit this? I can't find anything about it on Google, so it makes me wonder if there is something specific to our setup that causes us to hit it.

I wouldn't invest much more time into it, though. And I assume the issue will go away when/if we switch to `@algolia/autocomplete-core` and thus get rid of the Preact dependency, right?

package.json
```diff
@@ -250,7 +250,9 @@
         "wrangler": "^4.45.4"
     },
     "resolutions": {
-        "whatwg-url": "15.0.0"
+        "whatwg-url": "15.0.0",
+        "preact@npm:^10.13.2": "patch:preact@npm%3A10.17.0#~/.yarn/patches/preact-npm-10.17.0-e831e619a5.patch",
+        "preact@npm:^10.10.0": "patch:preact@npm%3A10.17.0#~/.yarn/patches/preact-npm-10.17.0-e831e619a5.patch"
```
rakyi, line 255: These refer to the old Preact version, but you now changed it to `10.17.0` IIUIC. Is this still correct?

marcelgerber: yep, they've been added by Yarn automatically (`yarn patch preact`), and they do the right thing.
I haven't changed the version of Preact in any way, we were using 10.17.0 before and are still doing so now.

marcelgerber:
> However, it makes me wonder: wouldn't everyone who is running a Preact app on Cloudflare Workers hit this? I can't find anything about it on Google, so it makes me wonder if there is something specific to our setup that causes us to hit it.

Yes, that's correct. It's such a niche issue, though, that I assume you can't really find anything. But I've definitely found a good bunch of issues that stem from the usage of `<reference>` or importing "dom" and thereby polluting global types (and that there's not really any workaround for this 🥲), e.g. https://github.com/microsoft/TypeScript/issues/33901 and https://github.com/microsoft/TypeScript/issues/43990.

Also, I've created a standalone project (`npx wrangler init`), and have observed that the same issue happens there once you import Preact. So, it's likely not anything to do with our setup.

Incidentally, since TS 5.5, TS is kind-of cracking down on the usage of `<reference>` in libs, albeit for other reasons: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#simplified-reference-directive-declaration-emit
Alas, Preact currently uses an older version of TS still, and the issue is still present in the latest Preact release.
 
> And I assume the issue will go away when/if we switch to `@algolia/autocomplete-core` and thus get rid of the Preact dependency, right?

Not really, it's both `@algolia/autocomplete-js` and `instantsearch.js` that rely on Preact. It's probably not likely anytime soon that we get rid of both of them.

rakyi:
Thanks for the thorough explanation. I think we use instantsearch.js only for types, so Preact would get tree-shaken if that was the only usage.

marcelgerber:
> I think we use instantsearch.js only for types, so Preact would get tree-shaken if that was the only usage.

Nope, that also doesn't help, precisely because we _only_ care about the types here, and TS doesn't do any kind of tree shaking (it's just bundlers that do tree shaking, not TS).

In fact, it is actually `instantsearch.js` that is causing Preact types to be loaded.
I learned today that you can see this info by running `cd functions && yarn tsc --explainFiles`, which proved quite useful for this PR.

rakyi:
Aah, that makes sense, and good to know!

PR #5720, opened 2025-11-26
🔨 fix loading indicator in mdims

### LINKED ISSUE #5405: Loading indicator sometimes not visible in multi-dim
## Description

When you change a dropdown on a multi-dim data page on a slow connection, loading indicator is sometimes not shown. This might make the user think the page is broken, since they get no visible feedback and nothing changes.

I wasn't able to reproduce it consistently, but it seems to happen quite a bit when simulating a GPRS connection in the Firefox dev tools.

## Expected behaviour

Loading indicator should always be shown (when the loading takes a while) when switching between multi-dim views.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/vaccination-coverage-who-unicef
2. In dev tools enable network throttling, e.g. GPRS network (in Firefox)
3. Try changing dropdowns

## Additional context

[Slack](https://owid.slack.com/archives/C014SGT6W9X/p1753088266279909?thread_ts=1753085190.516389&cid=C014SGT6W9X)

sophiamersmann:
Fixes #5405

Adds a `isDataReady` flag to track when a chart has finished loading, similar to the existing `isConfigReady` flag.

When `isConfigReady` is true and `whatAreWeWaitingFor()` checks if we're ready to render, it may check outdated data (if the new data hasn't loaded yet). This results in the loading spinner not being displayed.

I tested by:
- checking that the loading spinner appears in mdims
- checking that the "No table loaded yet" message doesn't flash for embedded charts

Not sure if there's something else we need to pay attention to.

### DIFF
--- a/packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
+++ b/packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
@@ -219,7 +219,12 @@ abstract class AbstractCaptionedChart extends React.Component<CaptionedChartProp
     }
 
     private renderLoadingIndicator(): React.ReactElement {
-        return <LoadingIndicator title={this.manager.whatAreWeWaitingFor} />
+        return (
+            <LoadingIndicator
+                title={this.manager.whatAreWeWaitingFor}
+                bounds={this.bounds}
+            />
+        )
     }
 
     private renderTimeline(): React.ReactElement {


sophiamersmann:
@codex review

packages/@ourworldindata/grapher/src/captionedChart/CaptionedChart.tsx
```diff
@@ -219,7 +219,12 @@ abstract class AbstractCaptionedChart extends React.Component<CaptionedChartProp
     }
 
     private renderLoadingIndicator(): React.ReactElement {
-        return <LoadingIndicator title={this.manager.whatAreWeWaitingFor} />
+        return (
+            <LoadingIndicator
+                title={this.manager.whatAreWeWaitingFor}
+                bounds={this.bounds}
+            />
+        )
```
sophiamersmann, line 227: Unrelated; positions the loading indicator in the center of the chart area (not in the center of the entire frame, including the entity selector)

rakyi (changes requested):
Actually, I'm not sure if it's unrelated, let me test a bit more.

rakyi (approved):
OK, I can reliably reproduce it on master, so it should be unrelated.

PR #5725, opened 2025-11-26
✨ show loading state for data downloads

sophiamersmann:
Adds a loading state to the download buttons in Grapher's download modal

---
This is **part 3 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #5732 
- <kbd>&nbsp;3&nbsp;</kbd> #5725 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #5713

### DIFF
--- a/packages/@ourworldindata/grapher/src/modal/DownloadModal.tsx
+++ b/packages/@ourworldindata/grapher/src/modal/DownloadModal.tsx
@@ -3,6 +3,7 @@ import { useCallback, useMemo, useState } from "react"
 import * as React from "react"
 import { observable, computed, action, makeObservable } from "mobx"
 import { observer } from "mobx-react"
+import cx from "classnames"
 import {
     Bounds,
     canWriteToClipboard,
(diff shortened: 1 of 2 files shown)


packages/@ourworldindata/grapher/src/modal/DownloadModal.tsx
```diff
@@ -924,6 +926,14 @@ export const DownloadModalDataTab = (props: DownloadModalProps) => {
     const serverSideDownloadAvailable =
         props.manager.isServerSideDownloadAvailable
 
+    // Track which download button is currently loading
```
rakyi: This doesn't behave entirely correctly on a slow connection:

1. Click the first button -> shows spinner on the first button
2. Click the second button (while the first is loading) -> no feedback (although no additional request is made)

And even though the buttons become disabled, they still appear (and behave) as if clickable.

We could either also visually disable both buttons when either of them is loading or split the loading state to track each button separately (and still disable visually whichever one is loading).

sophiamersmann: I decided to split the the loading state and moved it into the Button components

PR #5728, opened 2025-11-27
Fix race condition in multi-dim data fetching

### LINKED ISSUE #5727: Multi-dim can show a wrong view on a slow connection
## Description

When the user switches between dropdowns quickly on a slow connection, the multi-dim can get into a state where the dropdowns don't correspond to the chart shown below.

## Expected behaviour

State of dropdowns and chart should never get out of sync.

## Steps to reproduce

1. Go to http://localhost:3030/grapher/vaccination-coverage-who-unicef (or equivalent prod URL)
2. Switch on network throttling (regular 2G) in the dev tools
3. Switch the first dropdown:
    1. To the second option
    2. To the third option
    3. Back to the second option
5. Wait for all network requests to complete
6. Observe that the chart title and dropdown value don't correspond

## Video

https://github.com/user-attachments/assets/15c29fce-877d-4ba1-b5c8-026fb390adcf

## Environment

- OS: macOS 15.7.2
- Browser: Firefox 145.0.2

## Additional context

Related to (but probably independent of) #5405.

rakyi:
Fixes #5727

One day, we'll rewrite the whole spaghetti multi-dim data loading and state synchronization, but today is not the day.

### DIFF
--- a/site/multiDim/MultiDimDataPageContent.tsx
+++ b/site/multiDim/MultiDimDataPageContent.tsx
@@ -110,6 +110,9 @@ export function DataPageContent({
     // within grapher, so we have to initialize it early with
     // a truthy value
     const managerRef = useRef<GrapherManager>({ adminEditPath: "" })
+    // Used to track the latest updateGrapher request to avoid race conditions
+    // when quickly switching between dimension dropdowns on slow connections.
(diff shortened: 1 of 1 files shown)


rakyi:
It seems with this fix we can end up in a perpetual loading state and I can't figure out why.

PR #5732, opened 2025-11-28
🔨🤖 add viewId column to the explorer_views table

sophiamersmann:
Adds a viewId column to the explorer_views table, similar to the mdim view ids we already have.

---
This is **part 4 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #5732 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #5725 
- <kbd>&nbsp;1&nbsp;</kbd> #5713

### DIFF
--- a/adminSiteServer/multiDim.ts
+++ b/adminSiteServer/multiDim.ts
@@ -27,7 +27,7 @@ import {
 import {
     mergeGrapherConfigs,
     MultiDimDataPageConfig,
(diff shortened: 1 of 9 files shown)


sophiamersmann:
I thought I wanted to use it for zip filenames, but then realised that that doesn't really makes sense. No other motivation, but I thought it's still sensible to have a viewId available in the db.

PR #5745, opened 2025-12-02
fix(docs): add missing query params to chart API and combine image endpoints

Marigold:
This PR updates the OpenAPI specification for the Chart API to include missing query parameters for various endpoints and combines the PNG and SVG endpoints into a single endpoint with a format parameter.

### DIFF
--- a/docs/chart-api.openapi.yaml
+++ b/docs/chart-api.openapi.yaml
@@ -150,8 +150,6 @@ paths:
                 - Data
             parameters:
                 - $ref: "#/components/parameters/slug"
-                - $ref: "#/components/parameters/time"
-                - $ref: "#/components/parameters/country"
                 - name: v
                   in: query
                   description: Version number (e.g., 1)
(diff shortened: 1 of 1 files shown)


docs/chart-api.openapi.yaml
sophiamersmann: `mapSelect` works like `country` (can be multiple entities as well, with a tilde as separator)

sophiamersmann: or 'earliest'

docs/chart-api.openapi.yaml
```diff
@@ -141,6 +150,8 @@ paths:
                 - Data
             parameters:
                 - $ref: "#/components/parameters/slug"
+                - $ref: "#/components/parameters/time"
+                - $ref: "#/components/parameters/country"
```
sophiamersmann: The csv endpoint shouldn't depend on time or country

sophiamersmann: It does depend on some of the query params (those that the downloaded tables depend on), but I don't think that needs to be reflected here...

docs/chart-api.openapi.yaml
```diff
@@ -411,6 +422,7 @@ paths:
             parameters:
                 - $ref: "#/components/parameters/slug"
                 - $ref: "#/components/parameters/country"
+                - $ref: "#/components/parameters/time"
```
sophiamersmann, line 425: All Grapher params apply here to. I'd either drop country and time or add all grapher query params here

sophiamersmann: Of course not all of them apply like the globe ones for example don't. I'm sure there's more, so maybe not worth getting into the details

PR #5746, opened 2025-12-02
✨ disable switching from line to bar chart via the timeline

sophiamersmann:
Resolves #5223, resolves #5568

Disables switching from a line/slope to a bar chart via the timeline, as discussed. End of an era!

The migration should take care of all charts where the discrete bar chart is currently the default implicitly. If you check the charts reported by the SVG tester on staging, they should look unchanged.

PR #5747, opened 2025-12-02
🎉 select a year by typing

sophiamersmann:
Adds the ability to select a year by typing it into an input field. Support for dates is added in the next PR. I thought review was easier this way.

This isn't implemented for touch devices, it only works on desktop. 

Not sure how much to say about the feature itself – I hope it should be fairly intuitive. The tooltips that appear on hover are clickable. When clicked, the tooltip turns into an input field and you can type a year, hit Enter, and then the handle snaps to the closest time point.

~It's a bit awkward that using arrows to increment or decrement years won't have any effect until a year is reached that is closer to a different existing year, at which point it snaps.~ Improved by jumping to the next available year on arrow up/down in the input field

Examples (these point to the last staging site that includes all changes from this stack):
- Years: http://staging-site-grapher-time-selection-misc/grapher/life-expectancy
- BCE: http://staging-site-grapher-time-selection-misc/grapher/population-long-run-with-projections
- Dates: http://staging-site-grapher-time-selection-misc/grapher/excess-deaths-cumulative-economist-single-entity

### DIFF
--- a/packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
+++ b/packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
@@ -68,7 +68,6 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
     private lastUpdatedTooltip?: MarkerType
 
     private editHandle?: Exclude<MarkerType, MarkerType.Hover>
-    private isEditableTimeTooltipHovered: boolean = false
 
     private slider?: Element | HTMLElement | null
     private playButton?: Element | HTMLElement | null
@@ -84,14 +83,12 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
             | "lastUpdatedTooltip"
             | "hoverTime"
             | "editHandle"
-            | "isEditableTimeTooltipHovered"
         >(this, {
             startTooltipVisible: observable,
             endTooltipVisible: observable,
             lastUpdatedTooltip: observable,
             hoverTime: observable,
             editHandle: observable,
-            isEditableTimeTooltipHovered: observable,
         })
     }
 
@@ -249,8 +246,7 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
             !this.manager.isSingleTimeSelectionActive &&
             !this.isDragging &&
             !this.areBothHandlesVisible &&
-            !this.editHandle &&
-            !this.isEditableTimeTooltipHovered
+            !this.editHandle
         )
     }
 
@@ -422,7 +418,6 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
 
     @action.bound private resetEditState(): void {
         this.editHandle = undefined
-        this.isEditableTimeTooltipHovered = false
     }
 
     @action.bound private onCompleteYear(year?: number): void {
@@ -544,7 +539,7 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
 
         return (
             <EditableYearTooltip
-                type={MarkerType.Start}
+                position="left"
                 isEditing={this.editHandle === MarkerType.Start}
                 currentTime={startTime}
                 formattedTime={formattedStartTime}
@@ -553,11 +548,7 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
                 onChange={this.onChangeYear}
                 getTimeForKey={this.getStartTimeForNavigationKey}
                 onMouseEnter={action(() => {
-                    this.isEditableTimeTooltipHovered = true
                     this.hoverTime = undefined
-                })}
-                onMouseLeave={action(() => {
-                    this.isEditableTimeTooltipHovered = false
                 })}
                 onBlur={action(() => {
                     this.startTooltipVisible = false
@@ -578,9 +569,16 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
             return <SimpleTimeTooltip formattedTime={formattedEndTime} />
         }
 
+        // Avoid overlap with the hover tooltip when hovering to the right of the handle
+        const position =
+            this.hoverTime !== undefined &&
+            this.hoverTime > this.controller.endTime
+                ? "left"
+                : "right"
+
         return (
             <EditableYearTooltip
-                type={MarkerType.End}
+                position={position}
                 isEditing={this.editHandle === MarkerType.End}
                 currentTime={endTime}
                 formattedTime={formattedEndTime}
@@ -589,11 +587,7 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
                 onChange={this.onChangeYear}
                 getTimeForKey={this.getEndTimeForNavigationKey}
                 onMouseEnter={action(() => {
-                    this.isEditableTimeTooltipHovered = true
                     this.hoverTime = undefined
-                })}
-                onMouseLeave={action(() => {
-                    this.isEditableTimeTooltipHovered = false
                 })}
                 onBlur={action(() => {
                     this.startTooltipVisible = false
@@ -613,17 +607,15 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
             return <SimpleTimeTooltip formattedTime={formattedHoverTime} />
         }
 
-        // Show to the left if hovering over start handle, right if hovering over end handle
-        const type =
-            hoverTime < this.controller.endTime
-                ? MarkerType.Start
-                : MarkerType.End
+        // Show on the left if hovering to the left of the existing handle,
+        // right if hovering to the right of the existing handle
+        const position = hoverTime < this.controller.endTime ? "left" : "right"
 
         // The hover tooltip is not editable (since it's impossible to click on
         // the tooltip), but it uses the same component for consistent styling
         return (
             <EditableYearTooltip
-                type={type}
+                position={position}
                 currentTime={hoverTime}
                 formattedTime={formattedHoverTime}
             />
@@ -869,7 +861,7 @@ function TimelineHandle({
 }
 
 function EditableYearTooltip({
-    type,
+    position,
     isEditing = false,
     currentTime,
     formattedTime,
@@ -881,7 +873,7 @@ function EditableYearTooltip({
     onMouseLeave,
     onBlur,
 }: {
-    type: Exclude<MarkerType, MarkerType.Hover>
+    position: "left" | "right"
     isEditing?: boolean
     currentTime: number
     formattedTime: string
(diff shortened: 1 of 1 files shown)


sophiamersmann:
* **#5864** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5790** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5790?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5761** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5761?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5747** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5747?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5746** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sophiamersmann:
@codex review

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -428,6 +428,26 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
         }
     }
 
+    @action.bound private handleYearKeyboardNavigation(
+        key: string,
+        currentTime: number
+    ): number {
```
marcelgerber: The keyboard behavior for when hitting ArrowUp and then reaching the endTime is a bit confusing, as you can see in the video ([chart link](http://staging-site-discrete-timeline/grapher/life-expectancy?time=earliest..1800)): You can reach the endTime, but then when you hit enter it won't let you do just that.

https://github.com/user-attachments/assets/7553079d-dfa7-4574-8ec4-9bcfeb59b1f5

I can see two possible ways to make that more intuitive:
- (1) disallow increasing to 1800 in the first place, since `allowHandlesOnSameTime` is disabled
- (2) allow increasing beyond 1800 (as it is very much possible when entering numbers), and then possibly also "skip" over 1800 since that's an impossible option

packages/@ourworldindata/grapher/src/timeline/TimelineController.test.ts
marcelgerber, line 1: Do you want to LLM-add test cases for the new methods like `setStartTimeFromInput` etc., too?

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -334,39 +396,56 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
...
+            return controller.getNextTime(currentTime)
+        } else if (key === "ArrowDown") {
+            return controller.getPrevTime(currentTime)
+        } else if (key === "PageUp") {
+            return controller.getLargeStepForward(currentTime)
+        } else if (key === "PageDown") {
+            return controller.getLargeStepBackward(currentTime)
+        } else if (key === "Home") {
+            return controller.minTime
+        } else {
```
marcelgerber: This should either be 
```suggestion
        } else if (key === "End") {
```
, or if you really wanna go even more safe then key could be typed with the proper possibilities of what it can be, and then you can use `p-match` to properly match it.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -334,39 +396,56 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
...
+        if (!this.areBothHandlesVisible) {
+            this.controller.setStartAndEndTimeFromInput(year)
+        } else if (this.editHandle === MarkerType.Start) {
+            this.controller.setStartTimeFromInput(year)
+        } else {
+            this.controller.setEndTimeFromInput(year)
+        }
+    }
+
+    @action.bound private handleYearKeyboardNavigation(
```
marcelgerber: Possibly (and not really important in any way), we could also allow for other, laptop-keyboard-friendly combos here, like Shift+Up/Down or Shift+Left/Right instead of PgUp, PgDown, Home, End.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -542,86 +725,225 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
...
+    onComplete?: (year?: number) => void
+    onChange?: (year: number) => void
+    onKeyboardNavigation?: (key: string, currentTime: number) => number
+    onMouseEnter?: () => void
+    onMouseLeave?: () => void
+    onBlur?: () => void
+}): React.ReactElement {
+    const [inputValue, setInputValue] = React.useState(currentTime.toString())
+
+    const isEditing = editHandle === type
```
marcelgerber: Maybe instead of passing `editHandle` and then computing `isEditing` here, we can just pass `isEditing` instead and leave `editHandle` to be an implementation detail of the caller.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -409,6 +488,107 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
...
+        return (
+            <EditableYearTooltip
+                type={MarkerType.Start}
+                editHandle={this.editHandle}
+                currentTime={startTime}
+                formattedTime={formattedStartTime}
+                onStartEditing={() => this.onStartEditing(MarkerType.Start)}
+                onComplete={this.onCompleteYear}
+                onChange={this.onChangeYear}
+                onKeyboardNavigation={this.handleYearKeyboardNavigation}
```
marcelgerber: Given that `handleYearKeyboardNavigation` isn't modifying state at all, but rather computing a new state (a new year), maybe we should rename it to reflect that (and possibly `onKeyboardNavigation`, too)?

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -43,6 +47,19 @@ interface TimelineComponentProps {
 export class TimelineComponent extends React.Component<TimelineComponentProps> {
     base = React.createRef<HTMLDivElement>()
 
+    private hoverTime?: Time
+
+    private startTooltipVisible: boolean = false
+    private endTooltipVisible: boolean = false
+    private lastUpdatedTooltip?: MarkerType
+
+    private editHandle?: Exclude<MarkerType, MarkerType.Hover>
```
marcelgerber, line 70: Now that I think about it some more, we don't really need this, do we?
We only need the information that one of the edit handles is currently actively being edited, but not which one - that info we can instead just pass along as part of the `onChangeYear` handler.
And then `isEditing` just becomes state of the `Editable...` component.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -542,86 +725,225 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
...
+    formattedTime: string
+    onStartEditing?: () => void
+    onComplete?: (year?: number) => void
+    onChange?: (year: number) => void
+    onKeyboardNavigation?: (key: string, currentTime: number) => number
+    onMouseEnter?: () => void
+    onMouseLeave?: () => void
+    onBlur?: () => void
+}): React.ReactElement {
+    const [inputValue, setInputValue] = React.useState(currentTime.toString())
```
marcelgerber, line 888: Should we make this a number instead? Given that we're using an `<input type='number'>`, and that's well-supported across browsers, we can just move the `parseIntOrUndefined` step to the actual `setInputValue` call site.

Advantage: This way, we can stop users from entering float numbers, like `2021.1`.
Disadvantage: We might need special handling for `-`, because `-` in and of itself is not a valid integer.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
marcelgerber: Hm okay, now that I thought about is even some more, I guess we need to avoid the case where both inputs are active at the same time, and using `editHandle` is a good way to do just that.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
sophiamersmann: I like (2) better, but (1) is much easier to implement, so I went with that for now

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
sophiamersmann: I think the current combos are used because they're listed here: https://www.w3.org/WAI/ARIA/apg/patterns/slider/

What happens on Shift+Up/Down or Shift+Left/Right?

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
sophiamersmann: It would be much nicer but I can't think of another way to make the '-' show up in the input field

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
marcelgerber: My thinking was that they would do the same thing as PgUp (Shift+Up), PgDown (Shift+Down), Home (Shift+Left), End (Shift+Right), but in the end it is also probably a bad idea and we shouldn't do it, especially because Shift+Left/Right is used for text selection.

I also didn't really realise that you can use Fn+ArrowKeys on a Macbook to type Home/End/PgUp/PgDown, so that's a better option actually.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
marcelgerber: Yeah, you're right on that.

One option would be to go with react-aria's [NumberField](https://react-spectrum.adobe.com/react-aria/NumberField.html), which is overall quite nice, but I don't know if it's worth using that when we have custom behavior ourselves for key presses and which values we want to snap to.

One simple option to prevent float numbers from being entered (but not from being pasted, e.g.) would be simply to do: `onKeyDown={(e) => { if (e.key === ".") e.preventDefault() }}`

sophiamersmann: I'll just call e.preventDefault() then, everything else seems overkill to me 👍🏻

sophiamersmann:
### Merge activity

* **Jan 6, 9:21 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5747).
* **Jan 6, 9:22 AM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5747) rebased this pull request as part of a merge.
* **Jan 6, 9:23 AM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5747).

PR #5754, opened 2025-12-02
✨ Configurable archie chart sizes

ikesau:
## Context

Adds a `size` attribute to our archie charts, to allow for narrow and wider charts.

Defaults to `wide` (which is the same as the current size)

https://github.com/user-attachments/assets/31238ed2-df63-484a-9dcd-0939d64fac41

## Testing guidance

[You can use this test gdoc](https://docs.google.com/document/d/1rfsNPpAsTerRHMgarGt8IKV7rANOPpxsoIRXGCcBpkM/edit?tab=t.0)

  - narrative charts don't render wide, but they're centered, at least 🤷

### DIFF
--- a/db/model/Gdoc/rawToArchie.ts
+++ b/db/model/Gdoc/rawToArchie.ts
@@ -26,7 +26,6 @@ import {
     RawBlockStickyLeftContainer,
     RawBlockStickyRightContainer,
     RawBlockText,
-    RawBlockUrl,
     RawBlockAdditionalCharts,
     RawBlockAllCharts,
     RawBlockCallout,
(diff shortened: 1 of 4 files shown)


db/model/Gdoc/exampleEnrichedBlocks.ts
```diff
@@ -46,6 +46,7 @@ const enrichedBlockText: EnrichedBlockText = {
 const enrichedChart: EnrichedBlockChart = {
     type: "chart",
     url: "https://ourworldindata.org/grapher/total-cases-covid-19",
+    size: BlockImageSize.Wide,
```
rakyi: Are we sure that image and chart sizes are gonna stay in sync, i.e. when we add a size it should apply to both etc.? If so, I think it would be good to rename the type to e.g. `BlockSize` or something else more generic.

If we are not sure that they should be the same, a new type for chart sizes would be better.

PR #5761, opened 2025-12-03
🎉 select a date by typing

sophiamersmann:
Follow up to the previous PR. Adds support for selecting a specific date by typing.

The date input field is implemented with React Aria, which adds quite a bit of JS to our bundle. Not sure there's anything we can do about that?

### DIFF
--- a/packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
+++ b/packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
@@ -555,7 +555,7 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
     }
 
     @computed private get startHandleTooltip(): React.ReactNode {
-        const { startTime, minTime, maxTime } = this.controller
+        const { startTime } = this.controller
 
         if (!this.startTooltipVisible || !this.areBothHandlesVisible) {
             return undefined
@@ -569,14 +569,20 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
 
         const isEditing = this.editHandle === MarkerType.Start
 
+        const isDateValid = (date: CalendarDate): boolean => {
+            const { minTime, endTime, allowHandlesOnSameTime } = this.controller
+            const time = calendarDateToDaysSinceEpoch(date)
+            if (time < minTime) return false
+            return allowHandlesOnSameTime ? time <= endTime : time < endTime
+        }
+
         return this.isDailyData ? (
             <EditableDateTooltip
                 position="left"
                 isEditing={isEditing}
                 formattedTime={formattedStartTime}
                 currentTime={startTime}
-                minTime={minTime}
-                maxTime={maxTime}
+                isDateValid={isDateValid}
                 onStartEditing={() => this.onStartEditing(MarkerType.Start)}
                 onComplete={this.onCompleteDate}
                 onChange={this.onChangeDate}
@@ -603,7 +609,7 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
     }
 
     @computed private get endHandleTooltip(): React.ReactNode {
-        const { endTime, minTime, maxTime } = this.controller
+        const { endTime } = this.controller
 
         if (!this.endTooltipVisible) return undefined
 
@@ -622,14 +628,21 @@ export class TimelineComponent extends React.Component<TimelineComponentProps> {
 
         const isEditing = this.editHandle === MarkerType.End
 
+        const isDateValid = (date: CalendarDate): boolean => {
+            const { maxTime, startTime, allowHandlesOnSameTime } =
+                this.controller
+            const time = calendarDateToDaysSinceEpoch(date)
+            if (time > maxTime) return false
+            return allowHandlesOnSameTime ? time >= startTime : time > startTime
+        }
+
         return this.isDailyData ? (
             <EditableDateTooltip
                 position={position}
                 isEditing={isEditing}
                 formattedTime={formattedEndTime}
                 currentTime={endTime}
-                minTime={minTime}
-                maxTime={maxTime}
+                isDateValid={isDateValid}
                 onStartEditing={() => this.onStartEditing(MarkerType.End)}
                 onComplete={this.onCompleteDate}
                 onChange={this.onChangeDate}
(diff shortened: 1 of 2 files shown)


sophiamersmann:
* **#5864** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5790** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5790?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5761** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5761?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5761?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#5747** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5747?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5746** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sophiamersmann:
@codex review

ikesau:
Hey Sophia, awesome work on this!

Maybe you've already tried to dial this in, but there's a perceived baseline shift when the new element comes in.

https://github.com/user-attachments/assets/0d79a982-cdbe-4023-a4d9-1278d6044fc8

sophiamersmann:
Should be fixed, thanks @ikesau!

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -988,6 +1177,27 @@ function SimpleTimeTooltip({
...
+        date.date()
+    )
+}
+
+/** Convert CalendarDate to days-since-epoch integer */
+export function calendarDateToDaysSinceEpoch(date: CalendarDate): number {
+    const year = date.year
+    const month = date.month.toString().padStart(2, "0")
+    const day = date.day.toString().padStart(2, "0")
+    const isoString = `${year}-${month}-${day}`
```
marcelgerber, line 1208:
```suggestion
    const isoString = date.toString()
```

this'll always return a ISO string, see https://react-aria.adobe.com/internationalized/date/CalendarDate#to-a-string

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
```diff
@@ -957,6 +1024,128 @@ function EditableYearTooltip({
...
+        React.useState(false)
+
+    const handleKeyDown = (e: React.KeyboardEvent): void => {
+        if (e.key === "Enter") {
+            onComplete(currentValue)
+        } else if (e.key === "Escape") {
+            onComplete()
+        } else if (e.key === "ArrowUp" || e.key === "ArrowDown") {
+            // Apply immediately on arrow key changes
+            setShouldApplyImmediately(true)
```
marcelgerber, line 1117: This, for me, only works after the second key press: The first time I press arrow down, it doesn't auto-apply. Makes me think the `handleChange` callback is called before this one?

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
marcelgerber, line 1: More so than for years, I find myself craving +/- or stepper buttons, because going from e.g. 2022-12-31 to 2023-01-01 requires interacting with all three segments.

Something like this could work? Very much optional, though!

Image

marcelgerber (approved):

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
sophiamersmann: Agent fluff that made it through... thanks!

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
sophiamersmann: Same... I looked into it briefly, but I don't think there's a good solution (other than adding more stuff as you suggest). Since we don't have that much daily data, I'm okay to go ahead and if we get feedback, we can improve later.

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
sophiamersmann: None of the fixes I tried work :( Not sure what to do. Do you have an idea? Otherwise I'd just leave this as is

marcelgerber:
@sophiamersmann I just found another odd case, when you have a chart that supports both single-year and multi-year (like scatter or map): when it is single-time, and you're hovering to the right of the set time (or over the timeline-end marker), the tooltip gets rendered atop the grey timeline, when it should probably just be to the left of the handle.
[chart url](http://staging-site-discrete-timeline/grapher/life-expectancy).

CleanShot 2025-12-18 at 10 13 22

packages/@ourworldindata/grapher/src/timeline/TimelineComponent.tsx
marcelgerber: I've also tried, but also couldn't find a good fix :(

sophiamersmann: Thanks for trying! I guess in that case we just wait for the bug being reported 😅

sophiamersmann:
Thanks, @marcelgerber, now fixed! I'll do one more round of testing and then probably ship later today

sophiamersmann:
### Merge activity

* **Jan 6, 9:21 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5761).
* **Jan 6, 9:25 AM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5761) rebased this pull request as part of a merge.
* **Jan 6, 9:25 AM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5761).

PR #5763, opened 2025-12-03
fix(analytics): stop sending along past values for GA variables

### LINKED ISSUE #5735: Analytics: unclear if chart downloads come from explorer views or grapher charts
## Description
Chart download data analytics appears to be off at times (affecting ~2.5% of the data, which translates to around 20,000 downloads in 2025). The issue is that for some records, it is not evident if the download is coming from a Grapher chart or an Explorer view.

## Context
I'm looking at the analytics data on chart downloads, which comes from `owid.grapher_click` events with `event_action` matching pattern `"chart_download_(full|filtered)--(client|server)"`.

[→ Slack discussion](https://owid.slack.com/archives/CQQUA2C2U/p1764257547478649)

This event has the field `explorer_view`, which, to my knowledge, should only contain a value when `grapher_path` is an Explorer. However, I'm also observing non-NULL values for Grapher charts.

Additionally, I've observed that there is the field `explorer_path` which aparrently shouldn't be registered (from Slack discussion).

## Query
The query below queries our Google Analytics raw data for November 26, 2025, and retrieves records where an `explorer_path` is registered, and this differs from `grapher_path`.

Be mindful if you change the line `owid-analytics.analytics_253393922.events_20251126` not to incur high costs.

```sql
with raw AS (
  SELECT
    (SELECT value.string_value FROM UNNEST(event_params) WHERE key = 'grapher_path') AS grapher_path,
    (SELECT value.string_value FROM UNNEST(event_params) WHERE key = 'explorer_path') AS explorer_path,
    (SELECT value.string_value FROM UNNEST(event_params) WHERE key = 'explorer_view') AS explorer_view,
    *,
  FROM `owid-analytics.analytics_253393922.events_20251126`
    WHERE event_name = 'owid.grapher_click'
    AND 
      (SELECT value.string_value FROM UNNEST(event_params) WHERE key = 'event_action') IN (
          "chart_download_full_data--server",
          "chart_download_filtered_data--server",
          "chart_download_full_data--client",
          "chart_download_filtered_data--client"
      )
)
SELECT * FROM raw
WHERE (explorer_view IS NOT NULL) AND (grapher_path like "/grapher%");
```

## Steps to reproduce

1. Run query above in BQ console
2. Inspect the results. You will see that there are rows with `explorer_view` values but grapher-like values in `grapher_path`.

## Screenshots

Image

marcelgerber:
Fixes #5735.

So, Ike and I found a major problem in our analytics stack.

Basically, whenever we set a custom variable ("Data Layer Variable version 2" in GTM parlance) from our code via `dataLayer.push()`, it would be set to that new value _from now on until it gets overridden_, i.e. every future event would send it along even if it isn't meant to do just that.

In particular for us, this is mostly relevant with `explorer_path` and `grapher_path` metadata. Imagine the following chain of events:
```ts
{event: "owid.grapher_view", grapherPath: "/grapher/life-expectancy"} // user first views the life expectancy chart embedded into that page
{event: "owid.explorer_view", explorerPath: "/explorers/population-and-demography"} // user first views the explorer embedded into the same page
```
... then the second event would result in this event being sent to GA:
```ts
{event: "owid.explorer_view", grapherPath: "/grapher/life-expectancy", explorerPath: "/explorers/population-and-demography"}
//                              ⤴️ shouldn't be here, but is 🤷‍♂️
```

This is because, by default, GTM's data layer merges variables and stores them in its own data structures, and they wouldn't get reset ever.
Putting the `_clear: true` on the object makes it disregard these.

## Which events are affected?
All `owid.*` events we've sent through GA4, ever. Other events, like `page_view` and `scroll`, are unaffected.

It's also important to note that all event properties that have been explicitly set for the event, are also unaffected. It's mostly about `explorer_path`, `grapher_path`, sometimes `event_target`, etc.

## Other ways to solve this
I found at least two other ways to solve this, also:
1. Using `dataLayer.push(function () { this.reset() })`, [see here](https://developers.google.com/tag-platform/tag-manager/datalayer#custom_data_layer_methods).
2. By explicitly setting all unset properties to `null` or `undefined` for each event.
3. By using `_clear: true` (this is what we're doing here).

No clue if there's one option that's better than the others.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherAnalytics.ts
+++ b/packages/@ourworldindata/grapher/src/core/GrapherAnalytics.ts
@@ -11,9 +11,11 @@ import {
 
 const DEBUG = false
 
+type GAEventWithClear = GAEvent & { _clear: boolean }
+
 // Add type information for dataLayer global provided by Google Tag Manager
 type WindowWithDataLayer = Window & {
-    dataLayer?: (GAEvent | GAConsent)[]
+    dataLayer?: (GAEventWithClear | GAConsent)[]
 }
 declare const window: WindowWithDataLayer
 
@@ -179,7 +181,14 @@ export class GrapherAnalytics {
             console.log("Analytics.logToGA", event)
             return
         }
-        if (typeof window !== "undefined") window.dataLayer?.push(event)
+        if (typeof window !== "undefined") {
+            // It's very important that we clear (_clear) the data layer whenever we push an event,
+            // otherwise it will retain all properties from previous events which can lead to
+            // confusing and incorrect events data being sent.
+            // See https://www.simoahava.com/analytics/two-simple-data-model-tricks/.
+            // see also https://github.com/google/data-layer-helper/blob/4a65b385db1fac710d33bf5d1345e598e3d117fc/README.md#preventing-default-recursive-merge
+            window.dataLayer?.push({ ...event, _clear: true })
+        }
     }
 
     updateGAConsentSettings(consent: GAConsentParams): void {


marcelgerber:
@codex Is there any one of these methods that is better than the others?

> I found at least two other ways to solve this, also:
> 
> Using dataLayer.push(function () { this.reset() }), [see here](https://developers.google.com/tag-platform/tag-manager/datalayer#custom_data_layer_methods).
> By explicitly setting all unset properties to null or undefined for each event.
> By using _clear: true (this is what we're doing here).

PR #5767, opened 2025-12-04
🐛 remove <AdditionalCharts /> from markdown generation

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/5668

### DIFF
--- a/db/model/Gdoc/enrichedToMarkdown.ts
+++ b/db/model/Gdoc/enrichedToMarkdown.ts
@@ -72,6 +72,12 @@ const CUSTOM_MARKDOWN_COMPONENTS = {
     StaticViz: "StaticViz",
 }
 
+// Assumes that the contents of these tags can be removed entirely
+const CUSTOM_MULTILINE_MARKDOWN_COMPONENTS = {
+    AdditionalCharts: "AdditionalCharts",
+    KeyIndicatorCollection: "KeyIndicatorCollection",
+}
+
 function markdownComponent(
     componentName: keyof typeof CUSTOM_MARKDOWN_COMPONENTS,
     attributes: Record<string, string | undefined>,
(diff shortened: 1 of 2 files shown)


db/model/Gdoc/enrichedToMarkdown.ts
```diff
@@ -96,6 +102,15 @@ export function stripCustomMarkdownComponents(content: string): string {
         const regex = new RegExp(`<${componentName}[^\n]*?/>`, "g")
         strippedContent = strippedContent.replace(regex, "")
     }
+    for (const componentName of Object.values(
+        CUSTOM_MULTILINE_MARKDOWN_COMPONENTS
+    )) {
+        const regex = new RegExp(
+            `<${componentName}[\\s\\S]*?<\\/${componentName}>`,
```
mlbrgl, line 110: it seems a bit arbitrary that multi-line components support their opening tag being spread over multiple lines when the self closing ones specifically forbid it. Not an issue, but the inconsistency gave me pause.

db/model/Gdoc/enrichedToMarkdown.ts
```diff
@@ -72,6 +72,11 @@ const CUSTOM_MARKDOWN_COMPONENTS = {
     StaticViz: "StaticViz",
 }
 
+const CUSTOM_MULTILINE_MARKDOWN_COMPONENTS = {
+    AdditionalCharts: "AdditionalCharts",
+    KeyIndicatorCollection: "KeyIndicatorCollection",
```
mlbrgl, line 78: we seem to have deemed the text of the key indicators not md worthy, but I wonder if baking the assumption that all multiline component's content is never included in the md will hold for all future multiline components.

Not necessarily to be addressed here but maybe a comment/warning might be nice?

mlbrgl (approved):
nice 👍 And thanks for the extra testing!

Issue #5285, opened 2025-08-18
Better image positioning on mobile


ikesau:
In the [updated IPL article](https://ourworldindata.org/new-international-poverty-line-3-dollars-per-day), the layout is optimized for desktop. On mobile, images are positioned very far away from where they're first mentioned.

| Desktop | Mobile |
|--------|--------|
| <img width="1560" height="601" alt="Image" src="https://github.com/user-attachments/assets/1ed9d0ba-d292-46af-a556-a056984a92cd" /> | <img width="568" height="1956" alt="Image" src="https://github.com/user-attachments/assets/3f7868eb-6bad-4845-b9cb-16c95fc9da2d" /> | 

This is a consequence of our grid system naively stacking columns atop one another on mobile. It would be nice if we could give more granular control to authors to specify where an image should show up on mobile.

2 approaches I can think of:
1. Add `desktopOnly` and `mobileOnly` flags to the image archie block, so authors could specify an inline image for mobile and an image in the other column on desktop **we're going with this one**
2. Different handling of sticky columns on mobile, maybe 9:1 column width so that you can swipe left/right as you're reading, to keep the image accessible.

PR #5771, opened 2025-12-04
✨ hide blocks on mobile/desktop

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/5285

Allows authors to specify whether an image/chart/video should only show on desktop/tablets, or mobile. 

This will allow us to handle the case where a two-column layout becomes hard to read on mobile because the image gets pushed to the very bottom of the column, while the text assumes that the reader can see the chart.

desktopOnly mobileOnly

## Testing guidance

[You can use this gdoc](https://docs.google.com/document/d/1DjArt6cC9bwFRnTUKjc3GaWm6WggSSqdb9HclkYZ4k4/edit?tab=t.0)

I've tested that mobileOnly charts don't hydrate until they're shown.

### Before merging

### DIFF
--- a/site/gdocs/components/AtomArticleBlocks.tsx
+++ b/site/gdocs/components/AtomArticleBlocks.tsx
@@ -50,13 +50,17 @@ function AtomArticleBlock({
         )
     }
     return match(block)
-        .with({ type: "image" }, (block) => (
-            <Image
-                filename={block.filename}
-                smallFilename={block.smallFilename}
-                alt={block.alt}
-            />
-        ))
+        .with({ type: "image" }, (block) => {
+            // Skip mobile-only blocks in Atom feeds (assuming a desktop image will also be specified)
+            if (block.visibility === "mobile") return null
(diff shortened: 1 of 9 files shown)


site/gdocs/components/AtomArticleBlocks.tsx
```diff
@@ -50,13 +50,17 @@ function AtomArticleBlock({
...
-        .with({ type: "image" }, (block) => (
-            <Image
-                filename={block.filename}
-                smallFilename={block.smallFilename}
-                alt={block.alt}
-            />
-        ))
+        .with({ type: "image" }, (block) => {
+            // Skip mobile-only blocks in Atom feeds
+            if (block.mobileOnly) return null
```
ikesau: This assumes that a mobileOnly image will only ever be used in concert with a desktopOnly image.

ikesau:
Good point! Thanks 🙂

PR #5783, opened 2025-12-08
Enable `strict: true` for TS in `functions/`

rakyi:
We only recently added `functions/` to `yarn typecheck`, and forgot to enable `strict: true` which meant many kinds of potential errors were not caught by TS, see e.g. #5780

### DIFF
--- a/functions/_common/search/constructSearchResultDataTableContent.ts
+++ b/functions/_common/search/constructSearchResultDataTableContent.ts
@@ -1,4 +1,5 @@
 import * as _ from "lodash-es"
+import * as R from "remeda"
 import { match } from "ts-pattern"
 import {
     GrapherState,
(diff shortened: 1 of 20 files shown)


functions/_common/search/constructSearchResultDataTableContent.ts
```diff
@@ -135,21 +136,21 @@ function buildDataTableContentForLineChart({
 
     // Group series by name to handle cases where multiple series share the same name,
     // which can happen when projections are included alongside historical data
-    const groupedSeries = _.groupBy(
+    const groupedSeries = R.groupBy(
```
rakyi, line 139: Remeda has a more strict return type indicating that the groups can't be empty, which helps below.

PR #5808, opened 2025-12-11
perf: faster tolerance interpolation with special handling for common cases

marcelgerber:
Changes tolerance in such a way that it uses `validIndices` to more quickly find the next value-full row.

Using `validIndices` also allows quickly detecting and handling common cases: The case when there's no valid value at all for an entity, and the case when there's only a single valid value for an entity, and the case when all values are already present before running tolerance.


marcelgerber:
@codex review

PR #5812, opened 2025-12-11
Add multi-dim redirects

### LINKED ISSUE #5211: Create a way to redirect chart slugs to MultiDim views
## Core problem

With the `chart_slug_redirects` table we have a convenient way to redirect chart slugs to other charts (where the target is referenced by chart id). We want to enable something similar for redirecting chart slugs to multi-dim views.

This is a bit more complex, because we need to redirect to a specific view of the Multi-dim. In the database, we'll need a new table that references the `multidim-x-chart-configs` table at a specific row, identified by the internal id. 

In addition, we should maybe store the json serialization or query param of the view choice - in case the structure of the multi-dim changes in the future? Or we might want to have a FK constrain that prevents deletions? The latter is probably more in line with what we do with narrative views (is this true?). This is annoying for structural updates in the ETL though. Needs some more thought.

rakyi:
## Overview

Add admin API and UI to create, read, and delete redirects and integrate them throughout the code base.

Source of a multi-dim redirect can be:

- grapher
- explorer
- multi-dim

Target of a multi-dim redirect can be:

- default view of a multi-dim
- specific view of a multi-dim, i.e. multi-dim URL with query params

Multi-dim redirects affect:

- URL resolution in Cloudflare Pages and Functions: data pages, configs, thumbnails, etc.
- URL resolution during baking: if an article includes a chart that redirects to a multi-dim, it should use the multi-dim instead
- Dependency tracking in the archive: archived page checksums should be calculated using the resolved redirect targets

## Context

Resolves #5211

The implementation diverges from the initial design mostly by usinng a single table for storage, instead of separate tables based on the type of the source (the source might no longer exist).

## Future work

- Index page in admin to list all multi-dim redirects, similar to chart and site redirects
- Support for mapping explorer view query params to multi-dim view query params

image

## Testing guidance

There is a ton of stuff that we should test, unfortunately. I already tested most of it manually, but double checking it would be great. Not sure if trying to list it from manually here is worthwile, see what places we touch in code.

### Before merging

If DB migrations exists:

### DIFF
--- a/adminSiteServer/apiRoutes/mdims.ts
+++ b/adminSiteServer/apiRoutes/mdims.ts
@@ -1,10 +1,21 @@
 import {
+    ChartSlugRedirectsTableName,
+    DbEnrichedMultiDimDataPage,
     DbPlainMultiDimDataPage,
+    DbInsertMultiDimRedirect,
+    DbPlainMultiDimRedirect,
     JsonError,
     MultiDimDataPageConfigRaw,
     MultiDimDataPagesTableName,
+    MultiDimRedirectsTableName,
 } from "@ourworldindata/types"
+import {
+    MultiDimDataPageConfig,
+    queryParamsToStr,
+    Url,
+} from "@ourworldindata/utils"
 import { getMultiDimDataPageById } from "../../db/model/MultiDimDataPage.js"
+import { redirectWithSourceExists } from "../../db/model/Redirect.js"
 import { expectInt } from "../../serverUtils/serverUtil.js"
 import {
     upsertMultiDim,
@@ -20,6 +31,166 @@ import {
     isValidCatalogPath,
 } from "../validation.js"
 import e from "express"
+import * as z from "zod"
+
+function buildRedirectTargetDescription(
+    multiDim: DbEnrichedMultiDimDataPage,
+    viewConfigId: string | null
+): string {
+    let targetDescription = `/grapher/${multiDim.slug}`
+    if (viewConfigId) {
+        const mdimConfig = MultiDimDataPageConfig.fromObject(multiDim.config)
+        const dimensions = mdimConfig.findViewDimensionsByConfigId(viewConfigId)
+        if (dimensions) {
+            targetDescription += queryParamsToStr(dimensions)
+        }
+    }
+    return targetDescription
+}
+
+async function validatePathIsNotRedirectSource(
+    trx: db.KnexReadonlyTransaction,
+    path: string
+): Promise<void> {
+    if (await redirectWithSourceExists(trx, path)) {
+        throw new JsonError(
+            `'${path}' is already a source of an existing site redirect`,
+            400
+        )
+    }
+    const existingMultiDimRedirect = await trx<{ id: number }>(
+        MultiDimRedirectsTableName
+    )
+        .select("id")
+        .where("source", path)
+        .first()
+    if (existingMultiDimRedirect) {
+        throw new JsonError(
+            `'${path}' is already a source of an existing multi-dim redirect`,
+            400
+        )
+    }
+    const slug = Url.fromURL(path).slug
+    const existingChartSlugRedirect = await trx<{ id: number }>(
+        ChartSlugRedirectsTableName
+    )
+        .select("id")
+        .where("slug", slug)
+        .first()
+    if (existingChartSlugRedirect) {
+        throw new JsonError(
+            `'${path}' is already a source of an existing chart slug redirect`,
+            400
+        )
+    }
+}
+
+async function validateMultiDimRedirect(
+    trx: db.KnexReadonlyTransaction,
+    source: string,
+    targetSlug: string
+): Promise<void> {
+    const targetPath = `/grapher/${targetSlug}`
+
+    // Check source is not already a redirect source
+    await validatePathIsNotRedirectSource(trx, source)
+
+    // Check source is not already a redirect target (would create chain: X -> source -> target)
+    const sourceIsTargetOfSiteRedirect = await trx<{ id: number }>("redirects")
+        .select("id")
+        .where("target", source)
+        .first()
+    if (sourceIsTargetOfSiteRedirect) {
+        throw new JsonError(
+            `Creating this redirect would form a redirect chain: '${source}' is already a target of an existing site redirect`,
+            400
+        )
+    }
+    const sourceSlug = Url.fromURL(source).slug
+    // Multi-dim targets are always at /grapher/ paths, so only check if source is a /grapher/ path
+    if (source.startsWith("/grapher/")) {
+        const sourceIsMultiDimTarget = await db.knexRaw<{ source: string }>(
+            trx,
+            `-- sql
+            SELECT mdr.source
+            FROM ${MultiDimRedirectsTableName} mdr
+            JOIN ${MultiDimDataPagesTableName} mdp ON mdr.multiDimId = mdp.id
+            WHERE mdp.slug = ?
+            LIMIT 1`,
+            [sourceSlug]
+        )
+        if (sourceIsMultiDimTarget.length > 0) {
+            throw new JsonError(
+                `Creating this redirect would form a redirect chain: '${source}' is already a target of redirect '${sourceIsMultiDimTarget[0].source}'`,
+                400
+            )
+        }
+    }
+    // Check if source matches a chart's current slug that has old slugs redirecting to it
+    const sourceIsChartSlugTarget = await db.knexRaw<{ slug: string }>(
+        trx,
+        `-- sql
+        SELECT csr.slug
+        FROM ${ChartSlugRedirectsTableName} csr
+        JOIN charts c ON csr.chart_id = c.id
+        JOIN chart_configs cc ON c.configId = cc.id
+        WHERE cc.slug = ?
+        LIMIT 1`,
+        [sourceSlug]
+    )
+    if (sourceIsChartSlugTarget.length > 0) {
+        throw new JsonError(
+            `Creating this redirect would form a redirect chain: '${source}' is already a target of chart slug redirect '${sourceIsChartSlugTarget[0].slug}'`,
+            400
+        )
+    }
+
+    // Check target is not already a redirect source (would create chain: source -> target -> Y)
+    if (await redirectWithSourceExists(trx, targetPath)) {
+        throw new JsonError(
+            `Creating this redirect would form a redirect chain: target '${targetPath}' is already a source of an existing site redirect`,
+            400
+        )
+    }
+    const targetInMultiDim = await trx<{ id: number }>(
+        MultiDimRedirectsTableName
+    )
+        .select("id")
+        .where("source", targetPath)
+        .first()
+    if (targetInMultiDim) {
+        throw new JsonError(
+            `Creating this redirect would form a redirect chain: target '${targetPath}' is already a source of an existing multi-dim redirect`,
+            400
+        )
+    }
+    const targetInChartSlug = await trx<{ id: number }>(
+        ChartSlugRedirectsTableName
+    )
+        .select("id")
+        .where("slug", targetSlug)
+        .first()
+    if (targetInChartSlug) {
+        throw new JsonError(
+            `Creating this redirect would form a redirect chain: target '${targetPath}' is already a source of an existing chart slug redirect`,
+            400
+        )
+    }
+}
+
+async function createSlugChangeRedirect(
+    trx: db.KnexReadWriteTransaction,
+    multiDim: DbEnrichedMultiDimDataPage,
+    previousSlug: string
+): Promise<void> {
+    const source = `/grapher/${previousSlug}`
+    await validateMultiDimRedirect(trx, source, multiDim.slug!)
+    await trx(MultiDimRedirectsTableName).insert({
+        source,
+        multiDimId: multiDim.id,
+        viewConfigId: null,
+    })
+}
 
 export async function handleGetMultiDims(
     _req: Request,
@@ -52,6 +223,27 @@ export async function handleGetMultiDims(
     } catch (error) {
         throw new JsonError("Failed to fetch multi-dims", 500, { cause: error })
     }
+}
+
+export async function handleGetMultiDim(
+    req: Request,
+    _res: e.Response<any, Record<string, any>>,
+    trx: db.KnexReadonlyTransaction
+) {
+    const id = expectInt(req.params.id)
+    const row = await trx<DbPlainMultiDimDataPage>(MultiDimDataPagesTableName)
+        .select("id", "catalogPath", "slug", "config", "updatedAt", "published")
+        .where("id", id)
+        .first()
+    if (!row) {
+        throw new JsonError("Multi-dimensional data page not found", 404)
+    }
+    const multiDim = {
+        ...row,
+        config: JSON.parse(row.config),
+        published: Boolean(row.published),
+    }
+    return { multiDim }
 }
 
 export async function handlePutMultiDim(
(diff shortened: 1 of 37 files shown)


rakyi:
* **#5812** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5812?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#5825](https://github.com/owid/owid-grapher/pull/5825) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5825?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5835](https://github.com/owid/owid-grapher/pull/5835) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5835?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5812?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

marcelgerber:
@codex review

site/gdocs/utils.ts
```diff
@@ -152,12 +153,22 @@ export const useLinkedChart = (
...
 
+    const resolvedUrl = (() => {
+        const parsedResolvedUrl = Url.fromURL(linkedChart.resolvedUrl)
+        if (!queryString) return parsedResolvedUrl.fullUrl
+
+        return parsedResolvedUrl.setQueryParams({
+            ...parsedResolvedUrl.queryParams,
+            ...parsedOriginalUrl.queryParams,
+        }).fullUrl
+    })()
```
marcelgerber: I get the slight benefit of using an IIFE here (namely, that you don't need to have mutable vars), but it's not a pattern we normally use.

I think it's fine to pull this out into "normal" variables, or alternatively into a scoped `resolveRedirectQueryParams`-or-so function.

rakyi:
@mlbrgl I think we can change the function Codex found to merge the query params, right? We currently have no redirects in the DB which have query params on the source.

functions/_common/redirectTools.ts
```diff
@@ -1,12 +1,18 @@
...
 export function createRedirectResponse(
-    redirSlug: string,
-    currentUrl: URL
+    currentUrl: URL,
+    targetPath: string,
+    targetParams?: URLSearchParams
 ): Response {
+    const mergedParams = new URLSearchParams(targetParams)
+    for (const [key, value] of currentUrl.searchParams.entries()) {
+        mergedParams.append(key, value)
```
marcelgerber:
```suggestion
        mergedParams.set(key, value)
```

I don't think `append` is correct here, as it can result in duplicated entries in the URL (like `?a=1&a=2`), which we never have.

functions/_common/redirectTools.ts
```diff
@@ -66,3 +82,77 @@ export async function getOptionalRedirectForSlug(
...
+            url,
+            `/explorers/${slug.toLowerCase()}${extension}`
+        )
+
+    console.log("Looking up explorer slug and extension", {
+        slug,
+        extension,
+    })
+
+    // Explorer redirects contain full target paths (e.g. /grapher/slug?params)
```
marcelgerber, line 114:
```suggestion
    // Explorer redirects contain full target paths (e.g. /explorers/slug?params)
```

db/model/MultiDimRedirects.ts
```diff
@@ -0,0 +1,144 @@
...
+            `No matching view found for viewConfigId ${viewConfigId} on multi-dim ${multiDimSlug}`
+        )
+    }
+
+    const params = new URLSearchParams()
+    const sortedDimensions = Object.entries(dimensions).sort(([keyA], [keyB]) =>
+        keyA.localeCompare(keyB)
+    )
+    for (const [dimension, choice] of sortedDimensions) {
+        params.append(dimension, choice)
```
marcelgerber: it doesn't really matter here, but:

```suggestion
        params.set(dimension, choice)
```

db/model/MultiDimRedirects.ts
```diff
@@ -0,0 +1,144 @@
...
+            AND ${whereClause}
+        `,
+        [sourcePrefix, ...params]
+    )
+
+    if (slugs) {
+        for (const slug of slugs) {
+            redirectMap.set(slug, { targetSlug: slug })
+        }
+    }
```
marcelgerber: Could you add a comment why we're doing this? I was a bit confused about (but understand it now, we want to fill `redirectMap` with a default of a non-redirect).

Also, I believe its a bit weird that the map "shape" is slightly different depending on whether `slugs` is passed along or not, maybe we can just change that.

baker/redirectsFromDb.ts
```diff
@@ -28,12 +37,23 @@ export async function getRecentGrapherRedirects(
...
     )
+
+    const grapherPrefix = "/grapher/"
+    const multiDimRedirects = await getRecentMultiDimRedirects(knex)
+
+    return [
+        ...recentGrapherRedirects.map((row) => ({
+            source: `${grapherPrefix}${row.source}`,
+            target: `${grapherPrefix}${row.target}`,
+        })),
```
marcelgerber: Shall we do the prepending in the mysql query above already, using `CONCAT()`?

baker/redirectsFromDb.ts
```diff
@@ -1,17 +1,26 @@
 import * as _ from "lodash-es"
 import * as db from "../db/db.js"
 import { getRedirectsFromDb } from "../db/model/Redirect.js"
+import {
+    getMultiDimRedirectTargets,
+    getRecentMultiDimRedirects,
+} from "../db/model/MultiDimRedirects.js"
 
-export async function getRecentGrapherRedirects(
+export interface RecentNonSiteRedirect {
```
mlbrgl: nit: doesn't need to be exported (the general convention being: "if it is exported, it should probably live in types")

db/model/archival/archivalDb.ts
```diff
@@ -792,124 +879,306 @@ export const getPostChecksumsFromDb = async (
...
-        Array.from(allExplorerSlugs)
+        originalSlugs,
+        "/explorers/"
     )
-    const explorerChecksumsMap = new Map(
-        explorerChecksums.map((e) => [e.explorerSlug, e])
+
+    const slugsRedirectingToMultiDim = new Set(
+        originalSlugs.filter((slug) => multiDimRedirects.get(slug) !== slug)
     )
```
marcelgerber, line 1061: I guess this would also filter out the case when `/explorers/covid` redirects to `/grapher/covid`?

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -119,3 +185,262 @@ export async function handlePatchMultiDim(
...
+        .leftJoin("charts", "charts.configId", "chart_configs.id")
+        .select(trx.raw("1"))
+        .where("chart_configs.slug", sourceSlug)
+        .whereNotNull("charts.publishedAt")
+        .first()
+
+    const publishedExplorerExists = await trx(ExplorersTableName)
+        .select(trx.raw("1"))
+        .where({ slug: sourceSlug, isPublished: true })
+        .first()
```
marcelgerber: Explorers can only cause a conflict in case `source.startsWith("/explorers")`.

marcelgerber (approved):
Whoa, what a bunch of code for this!

Overall this is good, but I also can't say I've fully grasped every single line of this all.
Let me know if there's a specific thing that I should test explicitly!

And, one more overall note: The resolution of "for this slug, what is the actual page (slug & type) behind this" logic we really should centralise. We need it for gdocs, for archiving, for the mock router, etc. But: we can of course also do this in a followup refactoring.

site/gdocs/utils.ts
rakyi: I agree, it slipped through my own review of agent generated code.

functions/_common/redirectTools.ts
rakyi: Good catch! Again, the AI agent did something sensible, but different than what we want, and I didn't notice.

functions/_common/redirectTools.ts
rakyi: The original is actually correct. `_grapherRedirects.json` contains only mapping of a slug to a slug, but in `_explorerRedirects.json` we map from an explorer slug to `/grapher/<targetSlug>`, i.e. we include the full path in the target instead of assuming it's at `/grapher/`.

db/model/MultiDimRedirects.ts
rakyi: What do you mean by a different shape? In both cases it should be a map from a string to:

```ts
{
    targetSlug: string
    queryStr?: string
}
```

rakyi: We can actually just remove the mapping to the same slug part.

marcelgerber: Yeah, shape is the wrong word to use, and I was really talking about getting rid of the "default" values (of assuming a redirect to itself)

baker/redirectsFromDb.ts
rakyi: Good idea, changing.

db/model/archival/archivalDb.ts
rakyi: Ooh, great find! I need to fix that in multiple places.

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -107,6 +159,20 @@ export async function handlePatchMultiDim(
...
+                .select("source")
+                .where("multiDimId", id)
+                .limit(1)
+
+            if (existingRedirects.length > 0) {
+                throw new JsonError(
+                    `Cannot unpublish multi-dim: redirects exist pointing to this multi-dim (e.g., '${existingRedirects[0].source}'). Please delete the redirects first.`,
+                    400
+                )
+            }
```
mlbrgl, line 323: this seems different from what is being done when unpublishing charts, where redirects to a published chart are being deleted upon unpublishing it: https://github.com/owid/owid-grapher/blob/74009cda2266a7bde27c3531839de3f815c11dbe/adminSiteServer/apiRoutes/charts.ts#L530-L535

adminSiteServer/apiRoutes/mdims.ts
```diff
@@ -119,3 +185,269 @@ export async function handlePatchMultiDim(
...
+    if (sourceSlug) {
+        const sourceMultiDim = await getMultiDimDataPageBySlug(
+            trx,
+            sourceSlug,
+            { onlyPublished: false }
+        )
+
+        if (sourceMultiDim && sourceMultiDim.id !== multiDimId) {
+            const existingTargetRedirect = await trx<{
+                source: string
```
mlbrgl: I don't get how we can ever get there. This requires both:
- the sourceSlug to be matching an unpublished multi-dim page (since matching a published multi-dim page is handled above)
- and the (unpublished) multi-dim page to be the target of a redirect, which is not possible in the current state of the UI (since a multi-dim redirect can only be created for published multi-dim - but maybe this is relaxed in the next PR?)

functions/explorers/[slug].ts
```diff
@@ -151,7 +154,20 @@ export const onRequest: PagesFunction<Env> = async (context) => {
             context
         )
         .catch(async (e) => {
-            console.log("Handling 404 for", url.pathname)
-            return error(500, e)
+            // Here we do a unified after the fact handling of 404s to check
```
Comment (mlbrgl) [inline functions/explorers/[slug].ts]: could this be rephrased?

site/gdocs/utils.ts
```diff
@@ -152,12 +153,20 @@ export const useLinkedChart = (
...
+        : parsedResolvedUrl.fullUrl
+
     return {
         linkedChart: {
             ...linkedChart,
             // linkedCharts doesn't store any querystring information, because it's indexed by slug
             // Instead we get the querystring from the original URL and append it to resolvedUrl
-            resolvedUrl: `${linkedChart.resolvedUrl}${queryString}`,
+            resolvedUrl,
         },
```
mlbrgl: comment needs updating

site/gdocs/utils.ts
```diff
@@ -152,12 +153,20 @@ export const useLinkedChart = (
...
     }
 
+    const parsedResolvedUrl = Url.fromURL(linkedChart.resolvedUrl)
+    const resolvedUrl = queryString
+        ? parsedResolvedUrl.setQueryParams({
+              ...parsedResolvedUrl.queryParams,
+              ...parsedOriginalUrl.queryParams,
+          }).fullUrl
+        : parsedResolvedUrl.fullUrl
+
```
mlbrgl: the decision to merge query params seems important enough to be documented, as it contrast with the old binary "use original params only if none present on the target" found in the WP code: https://github.com/owid/owid-grapher/blob/74009cda2266a7bde27c3531839de3f815c11dbe/baker/redirects.ts#L144-L149

I find the merge behavior less predictable and more difficult to reason with but ok if we have a good reason to introduce it.

mlbrgl:
> @mlbrgl I think we can change the function Codex found to merge the query params, right? We currently have no redirects in the DB which have query params on the source.

@rakyi yep we can harmonize. Technically, there could be a chain between a site redirect (/hello -> /grapher/xyz/?a=1) and a chart redirect (xyz -> abc) but this is almost irrelevant at this point given that this part of the codebase only handles prominent links in wordpress content (topic country profiles, and explorer blocks).

_I made a related comment above about the merging strategy vs overriding under condition of non existence on the target._

site/gdocs/utils.ts
marcelgerber: I'm on Martin's side here; merging query params is more reasonable to me than simply discarding them.

It at least gives the user a chance that, with an "incoming" query param of `?time=2010..2020`, they'll indeed end up seeing that time range in the target chart (it's not a guarantee as there might've been other changes when moving the chart to a mdim, like changing the default tab or sth like that).

rakyi:
FYI @mlbrgl not sure how you are getting the GitHub links to files you shared in your comments, but all of them are 404. Probably safest is to share a link to a commit on master, when possible.

adminSiteServer/apiRoutes/mdims.ts
rakyi: Well, that's wrong then, isn't it? For example, it can silently break iframe embeds, article embeds etc. I think manually deleting the redirects shows you which URLs will break, and you can make a conscious decision about it.

adminSiteServer/apiRoutes/mdims.ts
rakyi: You're right. I added this validation before I changed other constraints, and then didn't realize we no longer need it. Removing.

functions/explorers/[slug].ts
Comment (rakyi) [reply functions/explorers/[slug].ts]: This is a copy of an existing comment in `functions/grapher/[slug].ts`. I'm happy to rephrase it in both places. Can you tell me more about what you find unsatisfactory about it?

site/gdocs/utils.ts
rakyi: I think we can remove it and simplify the code as well.

site/gdocs/utils.ts
rakyi: We also merge the query params in the `functions/` already currently (this is the thing Codex pointed out in the review above), which happens quite often in practice, whereas the old WP redirects code doesn't currently handle any redirects with query params, AFAICT.

rakyi:
> @rakyi yep we can harmonize.

Cool, I just added that.

functions/explorers/[slug].ts
Comment (mlbrgl) [reply functions/explorers/[slug].ts]: Ah I get it now, I couldn't work out the grammar of the beginning of the sentence: "unified after the fact handling". Anyway, this rephrasing sounds more straightforward:

> Only check _explorerRedirects.json for redirects if a 404 occurs; otherwise, skip the extra fetch in the happy path.

what do you think?

site/gdocs/utils.ts
mlbrgl: right I see what you mean, although I'd say it's not so much about applying vs discarding (as both strategies may apply source params), but rather about under what condition query params should be applied. The netlify way was rather binary: you may apply source params but only if none exist on the target. The merge way is more granular: source params are always applied, but may be overriden by target params.

As a default strategy for cases where sources and targets might not be related, I can see why you'd want to keep source and target params separated. In this specific case, sources and targets share a similar API, and I can see why you'd favor merging, with sanitization taking care of the odd egde-case. You have a better read of this part of the codebase than I do, so I'll trust your judgement!

mlbrgl:
> FYI @mlbrgl not sure how you are getting the GitHub links to files you shared in your comments, but all of them are 404. Probably safest is to share a link to a commit on master, when possible.

ah sorry, yep I'll switch to something more rebasing-resistant. Thanks!

functions/explorers/[slug].ts
Comment (rakyi) [reply functions/explorers/[slug].ts]: Nice and simple, changing! Thanks.

rakyi:
### Merge activity

* **Jan 8, 9:12 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5812).
* **Jan 8, 9:12 AM UTC**: @rakyi merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5812).

PR #5851, opened 2025-12-22
feat(narrative-charts): return merged config in GET endpoint

Marigold:
## Motivation

I need to get full grapher config of narrative charts to be able to render them in chart-diff. Rather than rewriting the logic in ETL, I thought it'd be easier to return it from GET of narrative chart. If there's an easier way, let me know, please!

## Summary

Add `mergedConfig` to the GET `/narrative-charts/:id` endpoint response. This provides the fully resolved configuration by merging the parent chart config with the narrative chart's patch overrides, making it easier for clients to get the complete chart configuration without needing to perform the merge logic themselves.

## What changed

- Modified `getNarrativeChartById()` to calculate and return the merged config alongside the patch and parent configs
- Uses the existing `mergeGrapherConfigs()` utility to properly combine parent + patch


rakyi:
The `configFull` already contains the full config of the narrative chart.

Marigold:
Ha, that was dumb... how could I have missed that?

PR #5852, opened 2025-12-23
cleanup(wp): remove global entity selector, help block, and prominent link hydration

marcelgerber:
Removing these features, as they are no longer needed now that we don't have any WP pages published any more.

Note that there are still WP remnants out there: Explorers still show WP blocks underneath the chart. These ones use prominent links, but in this case they don't need to be client-hydrated, as the global entity selector is never present in this case.

### DIFF
--- a/site/owid.scss
+++ b/site/owid.scss
@@ -317,7 +317,7 @@ h6 {
 .site-header {
     position: relative;
     // Must be specified for the header (and its sub-elements, e.g. topics menu)
-    // to be on top of the sticky nav, charts and global entity selector, while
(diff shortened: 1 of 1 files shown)


packages/@ourworldindata/grapher/package.json
```diff
@@ -43,7 +43,6 @@
     "react-dom": "^19.2.0",
     "react-flip-toolkit": "^7.2.4",
     "react-move": "^6.5.0",
-    "react-select": "^5.10.2",
```
ikesau, line 46: 🥳

PR #5853, opened 2025-12-23
Chrome-plugin

danyx23:
## Context

This Chrome extension provides a live preview of OWID articles while editing in Google Docs. Opens as a sidepanel and renders the article using the same React components as the main site. It renders client side only (with just a raw GDoc API proxied via the admin to avoid having to store credentials in the client), which makes it faster and retain scroll position.

In the future, this extension might be an interesting CMS like surface for us where we could render our components differently, allow drag-and-drop etc. For now, it's only about the preview and bringing up the infrastructure.

### Key Features
- Sidepanel UI activated when viewing Google Docs
- Fetches raw Google Doc content via new server API endpoints
- Client-side parsing: gdocToArchie() → archieToEnriched()
- Renders preview using OwidGdoc component with full styling
- Manual refresh with auto-refresh toggle
- Authentication via admin session cookies

## Trying out the prototype

This prototype has a settings page (Navigate to the exteions -> Options) that allows you to specify http://localhost:3030 as the base url instead of the default admin.owid.io. If you create a new gdoc in google drive, you have to add it in the grapher admin before you can preview the page via the chrome plugin.

To install the chrome plugin:
- check out the `chrome-plugin` or `chrome-extension-component-gallery` branch locally
- run the dev server locally (`make up`)
- in a terminal, go into the `chrome-extension` subfolder
- run `yarn build`
- Open Chrome (this plugin does not work on Firefox or Safari)
- Click the three vertical dots at the top right corner of the chrome window, go to extensions, manage extensions
- In the top right of the Extensions screen, enable "Developer mode"
- Click the button at the top left "Load unpacked"
- Navigate to the the owid-grapher folder, then into `chrome-extension` and select the `dist` folder
- Open a Gdoc in chrome
- Open the "OWID Article Preview" extension - you should see a preview of the gdoc
- Drag the preview as wide as you like
- Make a change to a gdoc
- Click "refresh" after making a change - notice that the scroll position stays stable
- If you like, enable auto-refresh

### Architecture
- **Content fetching** (fast, frequent): `/admin/api/gdocs/:id/raw` - pure proxy to Google Docs API
- **Attachments** (slow, cached): `/admin/api/gdocs/:id/attachments` - loads charts, images, linked docs from DB
- Background service worker handles cookie-authenticated API requests
- Parsing pipeline runs entirely in the browser

### Server Changes
- Added `getGdocRaw` endpoint - proxies Google Docs API, no DB access
- Added `getGdocAttachments` endpoint - returns linked content metadata
- Added `getRouteWithoutTransaction` helper for routes without DB access

### Extension Structure
```
/chrome-extension/
├── manifest.json (Manifest V3)
├── src/background/service-worker.ts
├── src/sidepanel/ (React UI)
└── src/shared/api.ts
```

## Testing guidance

1. Start local admin server: `yarn startAdminDevServer`
2. Build extension: `cd chrome-extension && yarn build`
3. Load extension in Chrome: `chrome://extensions` → Load unpacked → select `chrome-extension/dist`
4. Log into local admin at `http://localhost:3030/admin`
5. Open a Google Doc shared with the OWID service account
6. Click extension icon to open sidepanel
7. Preview should render with proper styling

## Releases

This PR already sets up a CI pipeline to build and distribute the chrome extension in our internal OWID store. You'd trigger a workflow with a version number which is then used to update the manifest, tag the commit, build the extension and upload it to our store.

## Planning docs

The PR still contains the planning docs I used to create this (code was written by claude/codex). I will delete them before the merge, but they might provide useful context.


danyx23:
* **#5896** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5896?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5853** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

danyx23:
@codex review

docs/plans/chrome-plugin-review.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Scope
+
+- Compared branch against master base commit 502770eda78539e145d483334a9db9ef1b25bac1.
+- Reviewed docs/plans/chrome-plugin.md implementation plan and notes.
+
+## Findings
+
+1. High: `FunctionalRouter` double-send risk. `getGdocRaw` and `getGdocAttachments` call `res.json`, then `FunctionalRouter.wrap` will still call `res.send` with the callback return value. This can trigger "headers already sent" errors and unstable responses. Prefer `res.set(...)` + `return data` instead of `res.json(...)`. Files: `adminSiteServer/apiRoutes/gdocs.ts:406`, `adminSiteServer/apiRoutes/gdocs.ts:444`.
+2. High: Attachment data can drift from live content. `getGdocAttachments` uses `GdocsContentSource.Internal`, which loads attachments from stored DB content and ignores current Google Doc edits. New charts/images will not show until the doc is refreshed into the DB. Consider using `GdocsContentSource.Gdocs` on manual attachment refresh or accepting a `contentSource` param. File: `adminSiteServer/apiRoutes/gdocs.ts:471`.
```
rakyi, line 11: All the points here look valid, but this seems like the biggest issue.

eslint.config.mjs
```diff
@@ -166,6 +166,15 @@ export default tseslint.config(
...
     },
+    {
+        // The chrome extension uses @owid/* tsconfig path aliases to import
+        // from monorepo directories (db, site, etc.) which are not npm packages.
+        // The extraneous-dependencies rule can't resolve these aliases.
+        files: ["chrome-extension/**/*"],
+        rules: {
+            "import-x/no-extraneous-dependencies": "off",
+        },
+    },
```
rakyi: We should fix this in the long term instead of this exception, but it doesn't block merge.

chrome-extension/manifest.json
```diff
@@ -0,0 +1,36 @@
+{
+    "manifest_version": 3,
+    "name": "OWID Article Preview",
+    "version": "1.0.0",
```
rakyi: Start with some kind of preview/alpha/0.0.1?

chrome-extension/package.json
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "@owid/chrome-extension",
+    "version": "1.0.0",
```
rakyi: Also here, 0.0.1? Should stay in sync with the manifest, probably.

chrome-extension/vite.config.ts
```diff
@@ -0,0 +1,140 @@
...
+            const optionsDir = resolve(distDir, "options")
+            if (!existsSync(iconsDir)) {
+                mkdirSync(iconsDir, { recursive: true })
+            }
+            if (!existsSync(sidepanelDir)) {
+                mkdirSync(sidepanelDir, { recursive: true })
+            }
+            if (!existsSync(optionsDir)) {
+                mkdirSync(optionsDir, { recursive: true })
+            }
```
rakyi: `mkdirSync` with `{ recursive: true }` is idempotent, we don't need the `existsSync` checks.

chrome-extension/vite.config.ts
```diff
@@ -0,0 +1,140 @@
...
+            copyFileSync(
+                resolve(__dirname, "manifest.json"),
+                resolve(distDir, "manifest.json")
+            )
+
+            // Move HTML files to correct locations (Vite outputs to src/<dir>/)
+            // and fix the relative paths
+            const srcHtml = resolve(distDir, "src/sidepanel/index.html")
+            const destHtml = resolve(sidepanelDir, "sidepanel.html")
+            if (existsSync(srcHtml)) {
```
rakyi: It's probably better to remove the `existsSync` checks here and below and fail fast rather than have missing files in the build output.

chrome-extension/src/shims/crypto.ts
rakyi, line 1: Can you explain why we do this? It's quite hacky. It would be better to avoid shims and switch to an API available both in the browser and the server, or use a cross-platform package directly.

chrome-extension/src/sidepanel/App.tsx
```diff
@@ -0,0 +1,350 @@
...
+    const [state, setState] = useState<AppState>(initialState)
+    const [autoRefresh, setAutoRefresh] = useState(false)
+    const [adminBaseUrl, setAdminBaseUrl] = useState<string>("")
+    const contentIntervalRef = useRef<number | null>(null)
+    const attachmentsIntervalRef = useRef<number | null>(null)
+
+    // Load admin base URL for auth error link
+    useEffect(() => {
+        void getAdminBaseUrl().then(setAdminBaseUrl)
+    }, [])
```
rakyi: We should do this outside the component on the top level scope, since it's not reactive.

chrome-extension/src/sidepanel/App.tsx
```diff
@@ -0,0 +1,350 @@
...
+    const match = url.match(/\/document\/d\/([a-zA-Z0-9_-]+)/)
+    return match ? match[1] : null
+}
+
+export function App() {
+    const [state, setState] = useState<AppState>(initialState)
+    const [autoRefresh, setAutoRefresh] = useState(false)
+    const [adminBaseUrl, setAdminBaseUrl] = useState<string>("")
+    const contentIntervalRef = useRef<number | null>(null)
+    const attachmentsIntervalRef = useRef<number | null>(null)
```
rakyi: It looks like we don't need these refs, and they can be local variables in the `useEffect`.

chrome-extension/src/sidepanel/App.tsx
rakyi, line 1: Seems like react-query would work nicely here for managing the state coming from the server.

chrome-extension/src/sidepanel/App.tsx
```diff
@@ -0,0 +1,350 @@
...
+    if (state.authError) {
+        return (
+            <div className="owid-preview-extension">
+                <div className="auth-error">
+                    <h2>Authentication Required</h2>
+                    <p>Please log in to the OWID admin to preview documents.</p>
+                    <a
+                        href={`${adminBaseUrl}/admin`}
+                        target="_blank"
+                        rel="noopener noreferrer"
```
rakyi: In general, we don't want/need to use `noreferrer` for analytics reasons.

chrome-extension/src/sidepanel/App.tsx
```diff
@@ -0,0 +1,350 @@
...
+        void fetchContent()
+    }, [fetchContent])
+
+    const handleRefreshAttachments = useCallback(() => {
+        void fetchAttachments()
+    }, [fetchAttachments])
+
+    const handleToggleAutoRefresh = useCallback(() => {
+        setAutoRefresh((prev) => !prev)
+    }, [])
```
rakyi: We don't need these extra functions. The inner functions are either already wrapped in `useCallback` or don't need it (`setAutoRefresh`).

chrome-extension/src/sidepanel/Preview.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+    useEffect(() => {
+        const container = document.querySelector("#owid-document-root")
+        if (!container) return
+        runFootnotes(container)
+    }, [content])
+
+    useEffect(() => {
+        if (dodsLoadedRef.current) return
+        dodsLoadedRef.current = true
```
rakyi: Another good use case for react-query. This line looks like a bug, we set the ref to true before we even start loading.

rakyi (commented):
You might be already aware, but some things don't render correctly: refs in the text, OWID logo in the citation section.

> If you create a new gdoc in google drive, you have to add it in the grapher admin before you can preview the page via the chrome plugin.

The crucial consideration here is: how do we make the plugin render the same way as the admin (or how much does it matter that it renders the same). If we keep the rendering in the extension, we might need to publish a new version every time we change the related gdoc JS/CSS, which is likely not feasible.

It might be much easier to render the gdoc fully on the server. Did you try it? Is it too slow? If so, improving the performance might be possible.

rakyi:
@mlbrgl will ask authors about the extension when talking to them during his discovery work.

PR #5861, opened 2025-12-30
fix: bullet lists in API docs not rendering properly

lucasrodes:


lucasrodes:
not needed, fix in etl side https://github.com/owid/etl/pull/5394

Issue #5377, opened 2025-09-08
Rework admin authentication


rakyi:
See [the write-up in Notion](https://www.notion.so/owid/2025-09-08-Auth-refactoring-26874b71f92e80698b38e9cd51114998) (shouldn't be public).

rakyi:
Only switching ETL to token auth remains.

PR #5863, opened 2026-01-02
Rework admin authentication

rakyi:
## Context

Rework admin auth as planned in #5377, except the remaining work is to switch ETL to an API token based auth. For now, we keep checking the session cookie to keep ETL working, but we don't create new sessions anymore in admin.

## Testing guidance

I tested:

- dev auth locally
- Tailscale auth on the staging server
- CF auth locally

@Marigold could you verify that the ETL still works fine?


package.json
```diff
@@ -89,7 +89,6 @@
         "algoliasearch": "^5.42.0",
         "antd": "^5.24.0",
         "archieml": "^0.5.0",
-        "bcryptjs": "^3.0.2",
```
rakyi, line 93: We should switch to the built-in `crypto` module, if we ever need to do something similar again.

adminSiteServer/appClass.tsx
```diff
@@ -63,16 +65,18 @@
...
 
+        app.use("/admin", sessionCookieAuthMiddleware)
+
         if (ENV === "staging") {
-            // Try to log in with tailscale if we're in staging
-            app.use(tailscaleAuthMiddleware)
-        } else {
-            // In production use Cloudflare
-            app.use("/admin/login", authCloudflareSSOMiddleware)
+            app.use("/admin", tailscaleAuthMiddleware)
```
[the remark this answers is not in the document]
rakyi: Cloudflare/Google should do enough rate limiting for us in prod.

adminSiteServer/appClass.tsx
```diff
@@ -63,16 +65,18 @@
...
+        } else if (ENV === "production") {
+            app.use("/admin", cloudflareAuthMiddleware)
```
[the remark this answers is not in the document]
rakyi: Cloudflare/Google should do enough rate limiting for us in prod.

rakyi:
Yeah, I'm planning to do that next, but it might need to wait until cooldown.

rakyi:
@marcelgerber could you also review this, pls?

marcelgerber:
@codex review

adminSiteServer/authentication.ts
```diff
@@ -1,46 +1,51 @@
 import * as Sentry from "@sentry/node"
 import express from "express"
-import crypto from "crypto"
 import * as db from "../db/db.js"
-import {
-    CLOUDFLARE_AUD,
-    SECRET_KEY,
-    SESSION_COOKIE_AGE,
```
marcelgerber, line 8: We can now get rid of these two in serverSettings.

adminSiteServer/authentication.ts
```diff
@@ -1,46 +1,51 @@
...
-    ADMIN_BASE_URL,
-    ENV,
-} from "../settings/serverSettings.js"
-import { BCryptHasher } from "../db/hashers.js"
+import { CLOUDFLARE_AUD } from "../settings/serverSettings.js"
 import * as jose from "jose"
-import { DbPlainSession, DbPlainUser, JsonError } from "@ourworldindata/utils"
```
marcelgerber, line 14: types `DbPlainSession`, `DbInsertSession` are now also unused.

adminSiteServer/authentication.ts
```diff
@@ -78,278 +83,184 @@ export async function authCloudflareSSOMiddleware(
...
 
-    const sessionid = req.cookies["sessionid"]
-    if (sessionid) {
-        const userAndSession = await db.knexReadWriteTransaction(
-            async (trx) => {
-                // Expire old sessions
-                await db.knexRaw(
-                    trx,
-                    "DELETE FROM sessions WHERE expire_date < NOW()"
-                )
```
marcelgerber, line 132: ~Deleting old sessions might be important so that we (in ETL code) can actually create a new session, depending on how exactly the etl code works there. Can you check that the ETL code takes care of that already?~

marcelgerber: nevermind, it took me until now to remember that there can, of course, be multiple open sessions for a single user. this won't be a problem, then.

adminSiteServer/authentication.ts
rakyi: I think we might need the SECRET_KEY for the API token implementation in a future PR. Adding a note and will remove it if not.

PR #5874, opened 2026-01-07
Remove deferredState from Search

rakyi:
We no longer seem to need it for anything.

### DIFF
--- a/site/search/Search.tsx
+++ b/site/search/Search.tsx
@@ -5,7 +5,7 @@ import {
     SearchResultType,
 } from "@ourworldindata/types"
 import { LiteClient } from "algoliasearch/lite"
-import { useMemo, useDeferredValue } from "react"
+import { useMemo } from "react"
 import { match } from "ts-pattern"
 import { useIsFetching } from "@tanstack/react-query"
 
@@ -61,16 +61,12 @@ export const Search = ({
 
     const analytics = useMemo(() => new SiteAnalytics(), [])
 
-    const deferredState = useDeferredValue(state)
-
     // Handle analytics tracking (skips initial page load)
-    useSearchAnalytics(deferredState, analytics)
+    useSearchAnalytics(state, analytics)
 
     const isFetching = useIsFetching()
 
-    // Derived state for template configuration Use immediate state to avoid
-    // firing duplicate queries (one for the current (deferred) template, one for the
-    // target template after deferred value catches up)
+    // Derived state for template configuration
     const topicType = getSelectedTopicType(state.filters, eligibleAreas)
     const templateConfig: TemplateConfig = {
         resultType: getEffectiveResultType(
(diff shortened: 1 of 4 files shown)


rakyi:
* **#5874** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5874?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#5876](https://github.com/owid/owid-grapher/pull/5876) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5876?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#5877](https://github.com/owid/owid-grapher/pull/5877) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5877?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5874?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

mlbrgl (commented):
true, it is not as used as originally planned, but I was keeping it around for deprioritizing  country selector updates, which I noticed took quite a large chunk of processing while being often hidden. 

Using deferred state for analytics only would probably be overkill, but given we already had it, it was tempting to use it there too.

rakyi:
When I tested the change, I couldn't tell the difference, but maybe I don't know what to check. Do you still see a difference yourself?

mlbrgl:
Even in the busiest render when applying a country filter (the first render of the results), the country selector is dominating:
 
Screenshot 2026-01-07 at 16 37 23

Screenshot 2026-01-07 at 16 42 51

However, it seems like since the introduction of the `key` prop on `SearchBar`, that deferred state has no longer any positive effect on the country selector, so yes let's simplify 👍

from copilot:

> Two separate mechanics are at play:
> 
> key={stateToSearchParams(state)} on Searchbar: Whenever the serialized search state changes, React unmounts the current Searchbar tree and mounts a fresh instance. That means every child (including SearchCountrySelector) is recreated from scratch, with props based on the latest state. There’s no gradual transition where the old filters linger; React throws away the old subtree entirely.
> 
> useSelectedRegionNames(true): This hook asks for the deferred copy of state. But because the whole component tree is remounted after each meaningful change, by the time SearchCountrySelector renders, the provider is already passing the latest state as state and deferredState (the deferred value would only lag during a continuous render, which doesn’t happen here—the remount is discrete).
> 
> So the selector never actually benefits from the lagging data: it unmounts, remounts, and immediately sees the newest filters. Passing true just forces React to compute a deferred value and trigger an extra render later, without changing the rendered output or preventing flicker. Removing the true flag keeps behavior identical but avoids the overhead of the deferred copy.

Issue #5827, opened 2025-12-16
Index non-topic tags for charts in Algolia


ikesau:
We currently filter out any non-topic tags in a chart record's tags field via `getTopicHierarchiesByChildName`

Fiona would like non-topic tags to be usable.

We'd still need to filter out a few "admin" tags like `Unlisted` and `core-econ.org`

Ideally these tags also show up in the autocomplete (but not in the **Filter by Topic** list)

danyx23:
Fiona and Hannah would like to publish the Crime topic page in Jan/Feb. I've therefore preliminarily triaged this as prio high and we can revisit in the first triage meeting in the new year.

spoonerf:
Thanks @ikesau and @danyx23, I thought I'd just mention that I'm not 100% sure if we'd want to open this up to _all_ non-topic tags, we have some pretty niche ones. Perhaps an ideal solution would be that non-topic tags could be opted into Search, but that they wouldn't necessarily be included by default. 

If that makes things difficult, though, we could always approach it from the other angle of cleaning up non-topic tags we aren't happy with.

PR #5879, opened 2026-01-07
✨ add `searchableInAlgolia` to our tags table

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/5827

Adds a "Searchable in Algolia" checkbox to the tag admin page. When true, this field means a non-topic tag is for all intents and purposes treated as a topic tag in Search.

Admin page:
image

Searchbar:
image

## Testing guidance

Switch it on for a given tag, then write an index from `indexChartsToAlgolia.ts` to a local JSON file, and see the tag is now present in the records' tags arrays. Also confirm that the tag now works in autocomplete.

## Checklist

### DIFF
--- a/db/db.ts
+++ b/db/db.ts
@@ -563,7 +563,9 @@ export const getNonGrapherExplorerViewCount = (
 }
 
 /**
- * 1. Fetch all records in tag_graph, isTopic = true when there is a published TP/LTP/Article with the same slug as the tag
+ * 1. Fetch all records in tag_graph:
+ *    - isTopic = true when there is a published TP/LTP/Article with the same slug as the tag
+ *    - isSearchable = true when isTopic OR the tag has searchableInAlgolia set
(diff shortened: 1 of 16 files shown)


db/db.ts
```diff
@@ -581,7 +581,7 @@ export async function getFlatTagGraph(knex: KnexReadonlyTransaction): Promise<
             tg.weight,
             t.slug,
             t.name,
-            p.slug IS NOT NULL AS isTopic
+            (p.slug IS NOT NULL OR t.searchableInAlgolia = TRUE) AS isTopic
```
ikesau: I considered keeping these fields separate (because now the variable name is lying) but that would make this change so much bigger, and the only reason I can think of that we might want to know when something's a topic or a fake topic is to link to its topic page, but we can use the `tag.slug` to figure that out.

PR #5881, opened 2026-01-08
✨ db migration for `users`: add `slackId`, set `isActive` to false where applicable

lucasrodes:
Adding two migrations:
- Adding Slack IDs to `users` table in MySQL. This can help us improve processes (i.e., bots) that need to tag users on Slack.
- Set `isActive` to False for users not at OWID.

### DIFF
--- a/db/migration/1767863328279-AddSlackIdToUsers.ts
+++ b/db/migration/1767863328279-AddSlackIdToUsers.ts
@@ -1,8 +1,6 @@
 import { MigrationInterface, QueryRunner } from "typeorm"
 
-export class AddSlackIdToUsers1767863328279
-    implements MigrationInterface
-{
+export class AddSlackIdToUsers1767863328279 implements MigrationInterface {
     public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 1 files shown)


lucasrodes:
@marcelgerber! Good catch. The answer seems to have an update, mentioning that now one can use the slack handle as `<slack_handle>`. I tested it and seems to work!

marcelgerber:
@lucasrodes Hm, I just tested it inside #bot-testing, and it doesn't seem to work.
Or, rather, it works fine for `<@lucas>`, but doesn't work for `<@marcel>`.

I also found https://docs.slack.dev/changelog/2017-09-the-one-about-usernames, which says that this whole thing has been deprecated ever since 2017, and I think we really would be better off if we didn't rely on them.

I've assembled a list of Slack member IDs for all Slack members, will send it to you on Slack.

lucasrodes:
@marcelgerber thanks for your feedback, Marcel! I've changed the migration to now use IDs instead of usernames.

marcelgerber (approved):

Issue #5856, opened 2025-12-23
🐛 Grapher analytics:`owid.grapher_view` not being sent following click in all charts block


bnjmacdonald:
## Description

Since ~18 June 2025, the all charts block (on topic pages and data pages) and the key indicators block are NOT sending an `owid.grapher_view` for any chart in the block _except_ the first chart shown in the block by default. This bug might also be affecting similar blocks that render charts following a user click, such as the key insights block on modular topic pages.

## Expected behaviour

When a user switches the chart in the all charts block or opens an accordion in the key indicators block, an `owid.grapher_view` event should be sent for the new grapher shown.

This bug is happening on all devices and all browsers, as evidenced by the almost [0 recorded `owid.grapher_views` events of graphers on the home page except for /grapher/child-mortality between 18-30 June 2025](http://metabase.owid.io/question/112-views-of-each-grapher-on-page?page=https%3A%2F%2Fourworldindata.org&date_range=2025-06-18~2025-06-30&device_categories=smart%20tv&device_categories=mobile&device_categories=tablet&device_categories=desktop), which was not the case [1-17 June 2025](http://metabase.owid.io/question/112-views-of-each-grapher-on-page?page=https%3A%2F%2Fourworldindata.org&date_range=2025-06-18~2025-06-30&device_categories=smart%20tv&device_categories=mobile&device_categories=tablet&device_categories=desktop). 

## Steps to reproduce

Either in Google Tag Manager or with the network tab open in devtools, scroll down to the key indicators block on the home page and open any other accordion. An `owid.grapher_view` event should be sent on each open, but it is not. 

You can also confirm this behavior on the all charts block of a topic page ([example](https://ourworldindata.org/age-structure#all-charts)) or the all charts block of a data page ([example](https://ourworldindata.org/grapher/population#all-charts))

marcelgerber:
This will have to do with the charts refactor, where (I think) we don't instantiate a new grapher when switching between charts in the all charts block.

It's worth checking whether the chart "slideshow", that we commonly use in SDG pages, is also affected.

marcelgerber:
Daniel doesn't remember why we are now reusing a single grapher instance for the all charts block; it might be that we just want to change that back again.

PR #5882, opened 2026-01-08
🐛 fix GA grapher_view events not being sent in the All charts block

sophiamersmann:
Resolves https://github.com/owid/owid-grapher/issues/5856

I ended up just adding a react key which seems to work. I wasn't sure how else to fix this since I don't really get how `useMaybeGlobalGrapherStateRef` works.

### DIFF
--- a/site/blocks/RelatedCharts.tsx
+++ b/site/blocks/RelatedCharts.tsx
@@ -1,5 +1,5 @@
 import * as _ from "lodash-es"
-import { useState, useRef } from "react"
+import { useState } from "react"
 import * as React from "react"
 import { RelatedChart } from "@ourworldindata/utils"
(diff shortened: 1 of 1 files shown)


site/blocks/RelatedCharts.tsx
```diff
@@ -18,7 +18,6 @@ export const RelatedCharts = ({
     charts: RelatedChart[]
     showKeyChartsOnly?: boolean
 }) => {
-    const refChartContainer = useRef<HTMLDivElement>(null)
```
sophiamersmann, line 21: Unrelated; just noticed it's not used

ikesau (approved):
Tested, confirmed working.

Lmk if you want an explanation of the maybeGlobal hook sometime 🙂

PR #5883, opened 2026-01-08
Make multi-dim catalogPath not null

rakyi:
It was null temporarily until we fixed the existing data.


marcelgerber (approved):

rakyi:
The build failed, because the staging DB still has the data that contain null, but I verified it works locally and the prod DB is also fixed.

Issue #5582, opened 2025-10-27
Automatically create a redirect when published gdoc slug gets changed


rakyi:
## Core problem

Almost always when we change a gdoc slug of a published gdoc we want the old URL to be redirected to the new one.

## Proposed solution

Create a redirect automatically from the old slug to the new one. Validate it in the same way as when we create redirects manually. If the redirect is undesired, it can be deleted manually.

## Context

Requested by Charlie in [Slack](https://owid.slack.com/archives/C46U9LXRR/p1761567124088089?thread_ts=1761566480.046409&cid=C46U9LXRR).

PR #5889, opened 2026-01-08
✨ automatic redirect creation on published gdoc slug change

ikesau:
Fixes https://github.com/owid/owid-grapher/issues/5582

You can test it locally by publishing a local test document with a custom slug (e.g. `test-a`) and republishing it repeatedly with `test-b`, `test-c`, and back to `test-a`) checking the redirects table each time.

Doesn't support the case where an article is republished as a data insight, team page, or announcement (where its slug would stay the same, but its canonical URL would change) because I don't think an author would ever do that.


rakyi:
@codex review

rakyi:
@codex review

rakyi:
@codex review

PR #5893, opened 2026-01-09
Update CookieNotice to new copy and design

rakyi:
## Context

* [Slack](https://owid.slack.com/archives/C09QHUP0411)
* [Figma](https://www.figma.com/design/oqc0vVNdZJAB2n4QdTahDU/Cookies-Banner?node-id=0-1&t=4KHGFBnnOxamnvX6-1)

image

image

## Testing guidance

The banner is only shown when you don't have the `cookie_preferences` cookie set.

## Checklist

### Before merging

### DIFF
--- a/site/SiteFooter.tsx
+++ b/site/SiteFooter.tsx
@@ -37,6 +37,7 @@ const RESOURCE_LINKS: LinkData[] = [
 
 const LEGAL_LINKS: LinkData[] = [
     { title: "Privacy policy", url: "/privacy-policy" },
+    { title: "Cookie notice", url: "/cookie-notice" },
     { title: "Legal disclaimer", url: "/organization#legal-disclaimer" },
     {
         title: "Grapher license",


site/SiteFooter.tsx
```diff
@@ -37,6 +37,7 @@ const RESOURCE_LINKS: LinkData[] = [
 
 const LEGAL_LINKS: LinkData[] = [
     { title: "Privacy policy", url: "/privacy-policy" },
+    { title: "Cookie notice", url: "/cookie-notice" },
```
ikesau, line 40: Unless you've got a PR stacked on this, this isn't a real page? It should be `"privacy-policy#cookie-preferences"`

site/css/cookie-notice.scss
```diff
@@ -1,5 +1,8 @@
+$full-button-width-breakpoint: 400px;
+
 .cookie-notice {
```
ikesau, line 3: A nice extra might be to add extra padding to the footer if the banner is displayed, so that you can still see all the buttons when scrolled to the bottom.

| Before | After |
| --- | --- |
| Image | Image |

site/CookieNotice.tsx
```diff
@@ -14,65 +16,99 @@ export const CookieNotice = ({
...
+                            We use a small number of <strong>cookies</strong>{" "}
+                            and similar technologies to{" "}
+                            <strong>make our website work</strong>. Some of
+                            these are essential and can't be switched off. Other{" "}
+                            <strong>optional cookies</strong> help us better
+                            understand <strong>how our website is used</strong>{" "}
+                            — by accepting these, you can help us improve our
+                            publication for everyone. Read more in our{" "}
+                            <a href="/privacy-policy">Privacy Policy</a> or{" "}
+                            <a href="/cookie-notice#cookie-preferences">
```
ikesau, line 105: This should be updated, too

site/SiteFooter.tsx
rakyi: It is indeed a real page that hasn't been published yet in production, but it has on [staging](http://staging-site-update-cookie-notice/cookie-notice). My bad, I should have mentioned it.

site/css/cookie-notice.scss
rakyi: It would be nice, but since the banner can have variable height based on the screen size, I don't think it's worth the additional complexity.

site/SiteFooter.tsx
ikesau: Aaah, I see. Nice!

Issue #5773, opened 2025-12-05
Downloaded CSV has null values for `color` column for all but the last year


Marigold:
In this example

https://ourworldindata.org/grapher/life-expectancy-un-vs-gdp-per-capita-wb.csv

`owid_region` has null values for all years but 2023.

marcelgerber:
Related to #4595.

Issue #5645, opened 2025-11-10
Downloaded data from discrete bar chart contains multiple redundant `time` columns


Marigold:
I downloaded CSV for this chart https://ourworldindata.org/grapher/dietary-choices-uk?time=2025-01-01 using https://ourworldindata.org/grapher/dietary-choices-uk.csv?v=1&csvType=filtered&useColumnShortNames=true&time=2025-01-01&overlay=download-data

The data looks like this
```
Entity,Code,Day,flexitarian,none,vegan,meat_eater,pescetarian,vegetarian,time,time,time,time,time,time
All adults,,2025-01-01,13,3,3,72,2,6,2025-01-01,2025-01-01,2025-01-01,2025-01-01,2025-01-01,2025-01-01
```

Those `time` columns at the end seem useless.

sophiamersmann:
Related to #4595

danyx23:
We're closing this one in favour of #4595

PR #5894, opened 2026-01-09
🔨 clean up downloaded csv files

sophiamersmann:
Cleans up downloaded CSV files.

- Resolves https://github.com/owid/owid-grapher/issues/5645
- Resolves https://github.com/owid/owid-grapher/issues/5773
- Resolves https://github.com/owid/owid-grapher/issues/4595

Columns in downloaded CSV files:
- Entity: always present
- Code: always present (but might be empty if no code exists)
- Year or Day: always present
- Data columns: x, y, color, size (e.g. 'Life expectancy')
- Original time columns for each data column (e.g. 'Life expectancy (Original time)' (only original times that differ from the current time are listed)
    - If there's only a single y-column, then the original time column is just used as 'Year'/'Day' column
- Entity annotations column (if exists)

Notable changes:
- The full data table for download used to be `inputTable`, now it's `table`. Given that the author might have excluded certain times and entities for good reason, this seems more sensible to me
- The column name (if not short) used to be `column.name` but now is `col.nonEmptyDisplayName` since that's typically a bit nicer and corresponds to the names used in the chart
    - For original time colunns, the suffix '(Original Year/Day)' is added
    - For annotations columns, the suffix '(Annotations)' is added
    - For projections, the suffix '(Projected)' is added
    - For columns with a specified target time, the suffix 'in XXXX' is added
- I added a few sentences to the README that explains the Original time columns and the concept of tolerance

### DIFF
--- a/functions/_common/readmeTools.ts
+++ b/functions/_common/readmeTools.ts
@@ -246,8 +246,8 @@ function* activeFilterSettings(
         yield `A filtered subset of the full data was downloaded. The following filters were applied:`
         for (const [key, val] of Object.entries(filterSettings)) {
             if (key === "country")
-                yield `${key}: ${val.replaceAll("~", ", ")}` // country filter is separated with tilde
-            else yield `${key}: ${val}`
+                yield `- ${key}: ${val.replaceAll("~", ", ")}` // country filter is separated with tilde
+            else yield `- ${key}: ${val}`
         }
         yield ""
     }
@@ -270,7 +270,17 @@ export function constructReadme(
     const urlWithFilters = `${grapherState.canonicalUrl}`
 
     const downloadDate = formatDate(new Date()) // formats the date as "October 10, 2024"
-    if (isSingleColumn)
+
+    const shouldUseFilteredTable = searchParams.get("csvType") === "filtered"
+    const table = shouldUseFilteredTable
+        ? grapherState.filteredTableForDownload
+        : grapherState.tableForDownload
+    const hasOriginalTimeColumn = table.columnsAsArray.some(
+        (column) => column.def.derivedFrom?.relationship === "originalTime"
+    )
+
+    if (isSingleColumn) {
+        const toleranceExplanation = `\nThe data file includes an additional column suffixed with (Original Year) or (Original Day). This column appears when we use "tolerance" to display data for time points where exact data is unavailable. For example, if a country does not have data for one or more years, we use tolerance to fill the gaps using the closest available data points. These suffix columns show you which year (or day) the value really came from, allowing you to see when the data was actually measured.`
         readme = `# ${grapherState.displayTitle} - Data package
 
 This data package contains the data that powers the chart ["${grapherState.displayTitle}"](${urlWithFilters}) on the Our World in Data website. It was downloaded on ${downloadDate}.
@@ -279,11 +289,12 @@ ${[...activeFilterSettings(searchParams, multiDimAvailableDimensions)].join("\n"
 
 The high level structure of the CSV file is that each row is an observation for an entity (usually a country or region) and a timepoint (usually a year).
 
-The first two columns in the CSV file are "Entity" and "Code". "Entity" is the name of the entity (e.g. "United States"). "Code" is the OWID internal entity code that we use if the entity is a country or region. For normal countries, this is the same as the [iso alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the entity (e.g. "USA") - for non-standard countries like historical countries these are custom codes.
+The first two columns in the CSV file are "Entity" and "Code". "Entity" is the name of the entity (e.g. "United States"). "Code" is the OWID internal entity code that we use if the entity is a country or region. For most countries, this is the same as the [iso alpha-3](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) code of the entity (e.g. "USA") - for non-standard countries like historical countries these are custom codes.
 
 The third column is either "Year" or "Day". If the data is annual, this is "Year" and contains only the year as an integer. If the column is "Day", the column contains a date string in the form "YYYY-MM-DD".
 
 The final column is the data column, which is the time series that powers the chart. If the CSV data is downloaded using the "full data" option, then the column corresponds to the time series below. If the CSV data is downloaded using the "only selected data visible in the chart" option then the data column is transformed depending on the chart type and thus the association with the time series might not be as straightforward.
+${hasOriginalTimeColumn ? toleranceExplanation : ""}
 
 ## Metadata.json structure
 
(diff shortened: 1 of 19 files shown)


sophiamersmann:
* **#5913** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5913?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5894** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/core-table/src/CoreTable.ts
```diff
@@ -509,6 +508,40 @@ export class CoreTable<
...
+            this.get(OwidTableSlugs.entityId)
+        )
+    }
+
+    // todo: should be on owidtable
+    @imemo get entityIdSlug(): string {
+        return this.entityIdColumn.slug
+    }
+
+    // todo: should be on owidtable
```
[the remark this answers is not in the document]
sophiamersmann: Addressed in the next PR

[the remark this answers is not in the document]
sophiamersmann: Addressed in the next PR

functions/_common/readmeTools.ts
```diff
@@ -246,8 +246,8 @@ function* activeFilterSettings(
         yield `A filtered subset of the full data was downloaded. The following filters were applied:`
         for (const [key, val] of Object.entries(filterSettings)) {
             if (key === "country")
-                yield `${key}: ${val.replaceAll("~", ", ")}` // country filter is separated with tilde
-            else yield `${key}: ${val}`
+                yield `- ${key}: ${val.replaceAll("~", "")}` // country filter is separated with tilde
```
marcelgerber: we still want to comma-separate them
```suggestion
                yield `- ${key}: ${val.replaceAll("~", ", ")}` // country filter is separated with tilde
```

packages/@ourworldindata/grapher/src/core/GrapherState.test.ts
```diff
@@ -1854,3 +1854,430 @@ describe("tolerance is not applied twice (issue #4891)", () => {
...
+            selectedEntityNames: ["France", "United Kingdom"],
+        })
+
+        const downloadTable = grapher.filteredTableForDownload
+        const ukRows = downloadTable.rows.filter(
+            (r) => r.entityName === "United Kingdom"
+        )
+        // At least one UK row should have the code
+        expect(ukRows.length).toBeGreaterThan(0)
+        expect(ukRows.some((r) => r.entityCode === "GBR")).toBe(true)
```
marcelgerber: This test doesn't actually test the backfilling, we need to be using `.every()` instead.

packages/@ourworldindata/grapher/src/core/GrapherState.test.ts
```diff
@@ -1854,3 +1854,430 @@ describe("tolerance is not applied twice (issue #4891)", () => {
...
+        const originalTimes =
+            downloadTable.get("gdp-originalTime").valuesIncludingErrorValues
+        const times = downloadTable.get("year").values
+
+        // Find an index where originalTime === time
+        const matchingIndex = times.findIndex(
+            (time, i) => time === originalTimes[i]
+        )
+
+        if (matchingIndex >= 0) {
```
marcelgerber: Remove this if, otherwise this test will always silently pass if it **can't** find a matching index.

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -1057,6 +1062,120 @@ export class GrapherState
...
+        // the y-column's original time as main time column if it exists
+        if (
+            this.yColumnSlugs.length === 1 &&
+            !this.xColumnSlug &&
+            table.has(originalTimeSlugs[0])
+        ) {
+            table = table
+                .dropColumns([timeSlug])
+                .renameColumn(originalTimeSlugs[0], timeSlug)
+        }
```
marcelgerber: If we do this, we can end up with multiple values for the same year?

You can see this, for example, when you download filtered data here: https://improve-data-downloads.owid.pages.dev/grapher/deaths-registered-with-an-ill-defined-cause-of-death?time=2022..latest&overlay=download-data

packages/@ourworldindata/grapher/src/modal/DownloadModal.tsx
```diff
@@ -1044,9 +1044,14 @@ export const DownloadModalDataTab = (props: DownloadModalProps) => {
         downloadCtx.filteredTable.numRows
     )
 
+    // Only offer the filtered table for download if it's different from the full table
+    const shouldShowFilteredTableDownloadButton =
+        downloadCtx.filteredTable.numRows > 0 &&
+        downloadCtx.fullTable.numRows !== downloadCtx.filteredTable.numRows
```
marcelgerber: The filtered table can still be meaningfully different, especially if the chart is in relative mode.

It's very hard to properly communicate this (that's a transformation, not filtering it down), though.

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -1057,6 +1062,120 @@ export class GrapherState
...
+
+        // Annotation columns to include
+        const annotationSlugs = xySlugs
+            .map((slug) => makeAnnotationsSlug(slug))
+            .filter((slug) => table.has(slug))
+
+        // Select only relevant columns (in the right order)
+        table = table.select([
+            table.entityNameSlug, // Entity name column
+            table.entityCodeSlug, // Entity code column
```
marcelgerber, line 1102: Maybe if the entity code is empty for all rows, then we should also drop the entity code col, because it probably is confusing then.

Example chart: https://improve-data-downloads.owid.pages.dev/grapher/biomass-fish-stocks-taxa

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
sophiamersmann: Right, I think in that case I'll just drop this code without replacement

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
sophiamersmann: Yeah, I thought about it, then just kept it because the code column is mentioned in the readme, which is probably not a good enough reason. I'll just drop it

packages/@ourworldindata/grapher/src/modal/DownloadModal.tsx
sophiamersmann: Dropping it then

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
marcelgerber: I agree, my thinking is that an empty `code` column may just be adding more confusion than value.

marcelgerber (approved):

sophiamersmann:
### Merge activity

* **Feb 2, 10:08 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5894).
* **Feb 2, 10:09 AM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5894) rebased this pull request as part of a merge.
* **Feb 2, 10:09 AM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5894).

PR #5906, opened 2026-01-12
Add admin API key auth

rakyi:
- Add admin_api_keys table
- Add createAdminApiKey script to generate an API key
- Add apiKeyAuthMiddleware used in all envs

Part of #5377
Follow-up to #5863

@Marigold would you take care of the ETL changes after we merge this? In theory, we need to generate only a single key for the ETL user (using the provided script) and use it in all envs. Would that make sense? And when we switch to API key auth everywhere, we'll remove the remaining session auth middleware and table from the admin.

### DIFF
--- a/adminSiteServer/authentication.ts
+++ b/adminSiteServer/authentication.ts
@@ -17,6 +17,7 @@ export type Request = express.Request
 export type Response = express.Response<any, { user: DbPlainUser }>
 
 const API_KEY_HEADER = "authorization"
+const ACT_AS_USER_HEADER = "x-act-as-user"
 const CLOUDFLARE_COOKIE_NAME = "CF_Authorization"
 const CLOUDFLARE_TEAM_DOMAIN = "https://owid.cloudflareaccess.com"
 const DEV_ADMIN_EMAIL = "admin@example.com"
@@ -130,7 +131,31 @@ export async function apiKeyAuthMiddleware(
             return
         }
 
-        await setAuthenticatedUser(res, user, trx)
+        let authenticatedUser = user
+        const actAsUserId = getActAsUserIdFromRequest(req)
+        if (!user.isSuperuser && actAsUserId !== undefined) {
+            console.error("Non-superuser attempted to use x-act-as-user.", {
+                userId: user.id,
+                actAsUserId,
+            })
+            return
+        }
+
+        if (user.isSuperuser && actAsUserId !== undefined) {
+            const actAsUser = await trx<DbPlainUser>(UsersTableName)
+                .where({ id: actAsUserId })
+                .first()
+            if (!actAsUser) {
+                console.error(
+                    `User with id ${actAsUserId} not found. Please contact an administrator.`
+                )
+                return
+            }
+            logActAsUser(req, user.id, actAsUserId)
+            authenticatedUser = actAsUser
+        }
+
+        await setAuthenticatedUser(res, authenticatedUser, trx)
         await updateApiKeyLastUsed(apiKeyRow.id, trx)
     })
     return next()
@@ -326,6 +351,29 @@ function getApiKeyFromRequest(req: express.Request): string | undefined {
     if (!trimmed.startsWith(bearerPrefix)) return undefined
     const token = trimmed.slice(bearerPrefix.length).trim()
     return token.length ? token : undefined
+}
+
+function getActAsUserIdFromRequest(req: express.Request): number | undefined {
+    const header = req.get(ACT_AS_USER_HEADER)
+    if (!header) return undefined
+    const trimmed = header.trim()
+    if (!trimmed) return undefined
+    const parsed = Number.parseInt(trimmed, 10)
+    if (Number.isNaN(parsed) || parsed <= 0) return undefined
+    return parsed
+}
+
+function logActAsUser(
+    req: express.Request,
+    superuserId: number,
+    actAsUserId: number
+): void {
+    console.info("Admin API key act-as", {
+        superuserId,
+        actAsUserId,
+        method: req.method,
+        path: req.originalUrl,
+    })
 }
 
 async function findAdminApiKey(
--- a/adminSiteServer/tests/api-key-auth.test.ts
+++ b/adminSiteServer/tests/api-key-auth.test.ts
@@ -57,4 +57,103 @@ describe("Admin API key auth", { timeout: 10000 }, () => {
 
         expect(apiKeyRow?.lastUsedAt).toBeNull()
     })
+
+    it("allows superusers to act as another user via x-act-as-user", async () => {
+        const superuser = await env
+            .testKnex(UsersTableName)
+            .where({ isSuperuser: 1 })
+            .first<DbPlainUser>()
+        expect(superuser).toBeTruthy()
+
+        const [actAsUserId] = await env.testKnex(UsersTableName).insert({
+            email: "act-as-user@example.com",
+            fullName: "Act As User",
+            isActive: 1,
+            isSuperuser: 0,
+            createdAt: new Date(),
+            updatedAt: new Date(),
+            lastSeen: null,
+        })
+
+        await env
+            .testKnex(UsersTableName)
+            .where({ id: superuser!.id })
+            .update({ lastSeen: null })
+
+        const { apiKey } = await seedApiKeyForUser(superuser!.id)
+
+        const response = await fetch(`${env.baseUrl}/users.json`, {
+            headers: {
+                Authorization: `Bearer ${apiKey}`,
+                "x-act-as-user": String(actAsUserId),
+            },
+        })
+
+        expect(response.status).toBe(200)
+
+        const updatedActAsUser = await env
+            .testKnex(UsersTableName)
+            .where({ id: actAsUserId as number })
+            .first<DbPlainUser>()
+
+        const updatedSuperuser = await env
+            .testKnex(UsersTableName)
+            .where({ id: superuser!.id })
+            .first<DbPlainUser>()
+
+        expect(updatedActAsUser?.lastSeen).toBeTruthy()
+        expect(updatedSuperuser?.lastSeen).toBeNull()
+    })
+
+    it("rejects x-act-as-user for non-superusers", async () => {
+        const [primaryUserId] = await env.testKnex(UsersTableName).insert({
+            email: "non-superuser@example.com",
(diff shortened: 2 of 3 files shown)


adminSiteServer/authentication.ts
```diff
@@ -251,10 +294,9 @@ async function getOrCreateDevAdminUser(): Promise<DbPlainUser> {
             updatedAt: now,
         })
 
-        const created = await trx
-            .table("users")
+        const created = await trx<DbPlainUser>(UsersTableName)
```
rakyi, line 322: This infers better types. Before, `created` would not be possibly undefined.

serverUtils/apiKey.ts
```diff
@@ -0,0 +1,11 @@
+import crypto from "crypto"
+
+export const API_KEY_BYTES = 32
+
+export function createApiKey(): string {
+    return crypto.randomBytes(API_KEY_BYTES).toString("base64url")
+}
+
+export function hashApiKey(apiKey: string): string {
+    return crypto.createHash("sha256").update(apiKey).digest("hex")
```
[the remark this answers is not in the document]
rakyi: SHA-256 is enough in this case, because we are not hashing a password. Passwords can have low entropy and are not guaranteed to be unique, but here we control the secret generation, which has 32 bytes of entropy and is thus sufficiently random.

db/migration/1768210667836-CreateAdminApiKeys.ts
```diff
@@ -0,0 +1,24 @@
...
+export class CreateAdminApiKeys1768210667836 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`-- sql
+            CREATE TABLE admin_api_keys (
+                id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+                userId INT NOT NULL,
+                keyHash CHAR(64) NOT NULL,
+                createdAt DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+                lastUsedAt DATETIME NULL,
+                revokedAt DATETIME NULL,
```
marcelgerber: do we have a good case for using `revokedAt`, or would it be enough in our case to just remove the row in case a key needs to be revoked?

adminSiteServer/authentication.ts
```diff
@@ -3,13 +3,20 @@ import express from "express"
...
+} from "@ourworldindata/types"
 import { DbPlainUser, JsonError } from "@ourworldindata/utils"
 import { execWrapper } from "../db/execWrapper.js"
+import { hashApiKey } from "../serverUtils/apiKey.js"
 
 export type Request = express.Request
 
 export type Response = express.Response<any, { user: DbPlainUser }>
 
+const API_KEY_HEADER = "x-api-key"
```
marcelgerber: wonder if we should instead use [Bearer authentication](https://swagger.io/docs/specification/v3_0/authentication/bearer-authentication/); possibly using a very common scheme could make it slightly easier to LLM-generate scripts using auth?

adminSiteServer/authentication.ts
```diff
@@ -275,6 +317,42 @@ function getClientIp(req: express.Request): string | undefined {
...
+    trx: db.KnexReadWriteTransaction
+) {
+    const keyHash = hashApiKey(apiKey)
+    return await trx<DbAdminApiKey>(AdminApiKeysTableName)
+        .where({ keyHash })
+        .whereNull("revokedAt")
+        .first("id", "userId")
+}
+
+async function markApiKeyUsed(
```
marcelgerber: I'd go with 
```suggestion
async function updateApiKeyLastUsed(
```

adminSiteServer/tests/api-key-auth.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    })
+
+    it("does not update lastUsedAt when API key is missing", async () => {
+        const user = await env.testKnex(UsersTableName).first<DbPlainUser>()
+        expect(user).toBeTruthy()
+
+        const { id } = await seedApiKeyForUser(user!.id)
+
+        const response = await fetch(`${env.baseUrl}/users.json`)
+        expect(response.status).toBe(200)
```
marcelgerber: This one shouldn't be 200 but rather 401 (forbidden) or so, no?

adminSiteServer/authentication.ts
rakyi: I thought it doesn't really matter, but maybe you are right. Switching to the `authorization` header. Fun fact: OpenAI uses `authorization` and Anthropic uses `x-api-token` in their APIs.

adminSiteServer/tests/api-key-auth.test.ts
rakyi: It falls back to the dev auth in test, which adds a default admin user on every request. Doesn't matter anymore now, since I removed `revokedAt`.

db/migration/1768210667836-CreateAdminApiKeys.ts
rakyi: I thought it would be nice for bookkeeping purposes, but on a second thought it might be indeed better to go for simplicity. YAGNI. Changed.

adminSiteServer/tests/api-key-auth.test.ts
rakyi: Ah no, this is the other test. Well, I think this is good enough. We'd need to disable the dev auth middleware only for these tests, which is quite hacky.

marcelgerber: Ah, I see. Can we run the tests with `ENV=production` instead? that's probably more sensible overall, and I don't think there's a (DB) test that relies on it being the dev env.

marcelgerber: And besides, testing that a wrong token doesn't grant you a logged-in state seems important.

Marigold:
> In theory, we need to generate only a single key for the ETL user (using the provided script) and use it in all envs. Would that make sense?

Having a single key would be quite limiting. A nice thing about the current approach is that we can "impersonate" users (create a session id for a given `user_id`). This is especially useful when we sync chart updates from a staging server. The updated chart has `lastEditedByUserId` which we use to session and then use it to call Admin API (and get correct `lastEditedByUserId` in prod).

Can we implement "impersonation"? For instance, if we pass it in a header `X-Act-As-User-Id`, we could then have a single API key for local & staging servers associated with a single ETL user.

rakyi:
The natural way of doing it would be to have one API key per user. Since we don't have many users, that could be manageable, right?

Marigold:
> that could be manageable, right?

Each person making changes to charts on staging servers would need a special API key and we'd need to store all of them in `.env` (plus all data scientists would need to update their local envs with their own keys). That feels like an overkill. We need a way how to impersonate someone. It doesn't have to be via headers, we could also send it as part of payload to chart endpoints (not sure how hard would that be). Maybe there's a better solution that I'm completely missing.

adminSiteServer/tests/api-key-auth.test.ts
rakyi: 34 tests failed when I ran `make dbtest` with `ENV=production`. We can use the new API key auth for authentication, which fixes that. See the added commit.

rakyi:
> Each person making changes to charts on staging servers would need a special API key and we'd need to store all of them in .env

I can't tell what's bad about it. Could you explain?

We do it once, and then we are done until we have a new user. And adding a new user to our systems won't be a meaningfully bigger task if we need to generate one additional API key.

> (plus all data scientists would need to update their local envs with their own keys)

They will have to add one API key to their local env if they are going to be making requests against the Grapher admin API anyway, no?

rakyi:
> They will have to add one API key to their local env if they are going to be making requests against the Grapher admin API anyway, no?

Well, only if they make requests against prod or staging. When Grapher runs with `ENV=development` (which is the default locally), it doesn't require auth.

Marigold:
> I can't tell what's bad about it. Could you explain?

I don't know if there's a good way to store >10 {user_id: api_key} pairs in `.env` file. I think that having an option to "make an update to a chart as a certain user" is a useful functionality. We could grant some API keys (the ETL one) this "superuser" power.

Users who run `chart-sync` locally would also need to store all those {user_id: api_key} pairs in their `.env`. (It's run by ETL, but if it fails we have to re-run it locally).

> They will have to add one API key to their local env if they are going to be making requests against the Grapher admin API anyway, no?

Yeah, you're right, that's inevitable. (No one runs grapher locally, everyone runs etl against staging)

rakyi:
> Users who run `chart-sync` locally would also need to store all those {user_id: api_key} pairs in their `.env`.

This part I didn't expect. Ok, I'll add the feature based on `isSuperuser` (which the ETL user is currently not, I'll change that) and probably a special header.

Marigold:
Thanks! 🙏 I'll prepare ETL and test it against your staging server then.

rakyi:
Added the requested feature, header is called `x-act-as-user` and it should specify a user ID to act as.

rakyi:
@codex review

rakyi:
* **#5928** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5906** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5906?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5906?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rakyi:
Yeah, good idea. Changed.

rakyi:
### Merge activity

* **Jan 15, 12:50 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5906).
* **Jan 15, 12:51 PM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5906) rebased this pull request as part of a merge.
* **Jan 15, 12:51 PM UTC**: @rakyi merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/5906).

rakyi:
@Marigold let me know if you need any help switching to the API key auth in ETL. Once everything moves, we'll remove the old session auth.

PR #5929, opened 2026-01-15
🐛 fix missing country param in archive url

### LINKED ISSUE #5898: Explorer: Archive URL doesn't include country param
## Steps to reproduce

1. Go to https://ourworldindata.org/explorers/energy?tab=line&country=~GBR
2. Open the embed menu
3. Note that both embed URLs correctly include `&country=~GBR`
4. Switch to any other explorer view
5. Open the embed menu again
6. Note that the archived embed URL now _doesn't_ include `&country=~GBR`

## Screenshots

https://github.com/user-attachments/assets/ce0ce236-e87e-4f51-9ecb-963f0e92da66

## Additional context

This is possibly due to the way we explicitly pass the explorer's selection array into the grapher, and then possibly it is basing its `authoredVersion` on the selection's entities.

sophiamersmann:
Fixes #5898

When switching between explorer views, the archived embed URL would lose the country parameter, even though the regular embed URL preserved it correctly.

The updateGrapherFromExplorer methods were adding selectedEntityNames to the config passed to setAuthoredVersion(). This made the selection part of the authored config, so GrapherState's areSelectedEntitiesDifferentThanAuthors returned false, excluding the country parameter from changedParams and thus from the archived embed URL.

Don't know if there's a better way?

### DIFF
--- a/packages/@ourworldindata/explorer/src/Explorer.test.ts
+++ b/packages/@ourworldindata/explorer/src/Explorer.test.ts
@@ -60,6 +60,32 @@ describe(Explorer, () => {
             "Relative to world total": "false",
         })
     })
+
+    it("includes country param in archived embed URL after switching views", async () => {
+        const explorer = SampleExplorerOfGraphers({
+            queryStr: "?country=~GBR",
+            archiveContext: {
+                type: "archived-page-version",
+                archiveUrl: "https://archive.org/example",
+                archivalDate: "20240101-000000",
+            },
+        })
+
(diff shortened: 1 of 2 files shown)


marcelgerber:
I put up an alternative fix at #5938, which doesn't solve the underlying issue of `changedParams` being wrong, but also solves the issue in what I find to be a nicer way. Your choice!

sophiamersmann:
I like yours better. Less risky and less ugly

PR #5932, opened 2026-01-15
🐛 fix missing chart -> variable relation

### LINKED ISSUE #5923: Inheritance is not respected for `note` in Grapher
## Description

Indicator inheritance is not respected for the note field, and it can't be linked to the ETL.

## Expected behaviour

We expect that, after setting the inheritance and linking the `note` field back, we are able to show the metadata set in the ETL, but the configuration stays unlinked.

## Steps to reproduce

This is happening in [this PR](https://github.com/owid/etl/pull/5460), for [this chart](http://staging-site-style-ilo-metadata-update-8a/admin/charts/8940/edit), as shown in the video. I want to merge to check if it happens in production as well (at least for this case)

## Screenshots

In this video I show how I can't set the note to have inheritance

https://github.com/user-attachments/assets/38b9b71b-1c3a-482e-9295-5090cb9d21aa

## Environment

For desktop, provide:

- OS: macOS Tahoe 26.2
- Browser Chrome
- Version 143.0.7499.193 (Official Build) (arm64)

## Additional context

Mojmir mentions it might be an admin issue (and not something with staging).

sophiamersmann:
Fixes  #5923

Inheritance is currently broken for [this chart](https://admin.owid.io/admin/charts/8940/edit) (and all map-only charts) because the `charts_x_parents` db view doesn't match them correctly to the parent variable id.

The charts_x_parents view was incorrectly excluding charts that have chartType = NULL (i.e., charts with chartTypes=[], which are map-only charts without a chart tab). This affected 35 charts with isInheritanceEnabled = TRUE that should have been included in the view.

### DIFF
--- a/db/migration/1768494494681-FixChartsXParentsViewNullChartType.ts
+++ b/db/migration/1768494494681-FixChartsXParentsViewNullChartType.ts
@@ -1,0 +1,93 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class FixChartsXParentsViewNullChartType1768494494681
+    implements MigrationInterface
+{
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        // Fix the charts_x_parents view to include charts with NULL chartType
+        // These are charts with chartTypes=[] (no chart tab, only map),
+        // which were being incorrectly excluded because NULL != 'ScatterPlot'
+        // evaluates to NULL
+        await queryRunner.query(`-- sql
+            ALTER VIEW charts_x_parents AS (
+              WITH y_dimensions AS (
+                SELECT
+                  *
+                FROM
+                  chart_dimensions
+                WHERE
+                  property = 'y'
+              ),
+              single_y_indicator_charts AS (
+                SELECT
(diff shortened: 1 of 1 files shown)


db/migration/1768494494681-FixChartsXParentsViewNullChartType.ts
```diff
@@ -0,0 +1,93 @@
...
+                  -- should only be one
+                  max(yd.variableId) as variableId
+                FROM
+                  charts c
+                  JOIN chart_configs cc ON cc.id = c.configId
+                  JOIN y_dimensions yd ON c.id = yd.chartId
+                WHERE
+                  -- scatter plots can't inherit settings
+                  -- NULL chartType means no chart tab (chartTypes=[]), which should be included
+                  (cc.chartType != 'ScatterPlot' OR cc.chartType IS NULL)
```
sophiamersmann, line 34: This is the line that changed. It used to be just `cc.chartType != 'ScatterPlot'`, which excluded map-only charts as well

Issue #5901, opened 2026-01-09
PUBLIC_TAG_PARENT_IDS still drive charts indexability


mlbrgl:
### The problem

The `Religion` tag has no parentId, this makes any chart tagged with it non indexable (isIndexable 0, https://github.com/owid/owid-grapher/blob/religion-algolia/db/model/Chart.ts#L394)

### The immediate solution

Setting a parentId for `Religion` (from one of the hardcoded PUBLIC_TAG_PARENT_IDS) wouldn't solve the problem for old charts, as `isIndexable` is computed only when chart tags are being set. So, in addition, we need every chart tagged with `Religion` to be marked indexable (to reflect the new state of `Religion`, now child of a public tag).

👉 see https://github.com/owid/owid-grapher/pull/5895

### The long term solution

`parentId` are deprecated in favor of the tag graph, so we need:
- to update `setChartTags` to read parents from the tag graph
- a migration that derives `isIndexable` columns from the tag graph hierarchy.

However, this doesn't solve the following issue:
- [ ] #5902

Issue #5902, opened 2026-01-09
If a tag is removed from the tag graph, all charts with that tag remain indexable


mlbrgl:
see #5901 for context

rakyi:
Fixed in #5935

PR #5935, opened 2026-01-15
🔨 Remove deprecated parentId and isIndexable properties from the tags and charts table

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/5901 and https://github.com/owid/owid-grapher/issues/5902

- Removes `parentId` from the `tags` table and either removes this dependency, or switches it to use the tag graph where needed
- Removes `isIndexable` from the charts table (because it was getting out of sync) and adds a bit of SQL in a few places to calculate it dynamically
 
`isIndexable = true` when the chart is tagged with a topic tag OR `searchableInAlgolia = true` AND not tagged with `Unlisted`

https://github.com/user-attachments/assets/7b99d40e-9e5b-4295-94d0-d9ff8228071b

## Testing guidance

Go to the tag edit page:
1. Set a slug that doesn't match a published gdoc - you'll get a warning to check for typos.
2. Set a slug that _does_ match a published gdoc - `Searchable in Algolia` will switch on
3. Confirm that the tag edit page and dataset edit page still work as expected

## Checklist

If DB migrations exists:

### After merging

### DIFF
--- a/adminSiteClient/TagEditPage.tsx
+++ b/adminSiteClient/TagEditPage.tsx
@@ -182,8 +182,7 @@ class TagEditor extends Component<{
                             />
                             <small className="form-text text-muted">
                                 The slug for this tag's topic page, e.g.
-                                trade-and-globalization. If specified, we assume
-                                this tag is a topic. Must be unique
+                                trade-and-globalization.
                             </small>
                         </div>
                         <Toggle
@@ -200,8 +199,8 @@ class TagEditor extends Component<{
                             disabled={this.slugMatchesPublishedTopicPage}
                             secondaryLabel={
                                 this.slugMatchesPublishedTopicPage
-                                    ? "This slug matches a published gdoc, so charts with this tag will be indexed in Algolia"
-                                    : "When enabled, charts with this tag will be indexed in Algolia even without a matching published gdoc"
+                                    ? "This slug matches a published topic page, so charts with this tag will be indexed in Algolia"
+                                    : "When enabled, charts with this tag will be indexed in Algolia even without matching a published topic page"
                             }
                         />
                         <div style={{ marginTop: 16 }}>
--- a/adminSiteServer/apiRoutes/tags.ts
+++ b/adminSiteServer/apiRoutes/tags.ts
@@ -188,13 +188,13 @@ export async function updateTag(
                         SELECT 1
                         FROM posts_gdocs_x_tags gt
                         WHERE pg.id = gt.gdocId AND gt.tagId = ?
-                ) AND pg.published = TRUE AND pg.slug = ?`,
+                ) AND pg.published = TRUE AND pg.slug = ? AND pg.type IN ('topic-page', 'linear-topic-page')`,
             [tagId, tag.slug]
         )
         if (!gdoc.length) {
             return {
                 success: true,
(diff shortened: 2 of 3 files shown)


ikesau:
@codex review

adminSiteClient/TagEditPage.tsx
```diff
@@ -40,14 +40,17 @@ class TagEditable {
 }
 
 @observer
-class TagEditor extends Component<{ tag: TagPageData }> {
+class TagEditor extends Component<{
+    tag: TagPageData
+    publishedGdocSlugs: string[]
```
ikesau: This allows us to show to the user if they've added a slug that actually matches a published gdoc, which means that it will become a topic tag if they hit save.

adminSiteServer/mockSiteRouter.ts
```diff
@@ -341,7 +341,9 @@ getPlainRouteWithROTransaction(
 
         if (topicTag && topicTag.slug !== undefined) {
             // if topic slug is not a valid topic, render all data insights
-            const validTopicSlugs = await db.getAllTopicSlugs(trx)
+            const validTopicSlugs = await db
+                .getAllTopicTags(trx)
```
ikesau, line 345: I changed this function's signature, so here we need to map it down again

baker/algolia/utils/charts.ts
```diff
@@ -117,7 +117,26 @@ export const getChartsRecords = async (
                      LEFT JOIN charts_x_entities ce ON c.id = ce.chartId
                      LEFT JOIN entities e ON ce.entityId = e.id
             WHERE cc.full ->> "$.isPublished" = 'true'
-                AND c.isIndexable IS TRUE
+            -- NOT tagged "Unlisted"
```
ikesau, line 120: This is the SQL snippet that replaces `isIndexable` - there are 3 instances of it (including the reverse migration)

Marigold:
@ikesau merge [this one in ETL](https://github.com/owid/etl/pull/5497) after this one please (and let me know if anything else changes).

db/model/Chart.ts
```diff
@@ -27,12 +27,6 @@ import { zodResponseFormat } from "openai/helpers/zod"
 import { OPENAI_API_KEY } from "../../settings/serverSettings.js"
 import * as z from "zod"
 
-// XXX hardcoded filtering to public parent tags
-export const PUBLIC_TAG_PARENT_IDS = [
```
mlbrgl, line 31: 🥳!

adminSiteClient/TagEditPage.tsx
mlbrgl: We could use an autocomplete box here:
- autocomplete lists eligible gdocs
- authors can select one gdoc in the list
- authors can deselect/remove the gdoc association
- "searchable in Algolia" syncs (unchanged behaviour)

Adding a referential integrity constraint adds friction to the "create tag then topic page" flow but we already discourage authors from typing an unpublished slug (see tags.ts), so we might as well bake that constraint in and remove that path?

baker/algolia/utils/charts.ts
mlbrgl: I wonder to what extent "unlisted" overlaps with the definition of indexable/searchable tags. It seems like we could get rid of one of these abstractions (not necessarily in this PR though.)

ikesau: Yeah, definitely.

PR #5936, opened 2026-01-16
🐛 (analytics) circumvent truncated grapher slugs

### LINKED ISSUE #5860: 🐛 Grapher analytics: `grapher` dimension is being truncated at 100 characters
## One-liner

[Slack context](https://owid.slack.com/archives/C08U7KHAN3U/p1767103454637389). Some grapher URLs are > 100 characters in length, which means that when we send the URLs in the `grapher` custom dimension of GA4 events (like for `owid.grapher_click` and `owid.grapher_view`), the `grapher` field is truncated.

[Affected graphers (ranked by their data page views)](http://metabase.owid.io/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImxpYi90eXBlIjoibWJxbC9xdWVyeSIsImRhdGFiYXNlIjoyLCJzdGFnZXMiOlt7Im5hdGl2ZSI6Ii0tIGxpc3Qgb2YgY2hhcnRzIFxud2l0aFxuXG50IGFzIChcblx0c2VsZWN0IHVybCwgdmlld3NfMzY1ZCwgcmFuaygpIG92ZXIgKG9yZGVyIGJ5IHZpZXdzXzM2NWQgZGVzYykgYXMgcmFua18zNjVkXG5cdGZyb20gY2hhcnRzXG4pXG5cbnNlbGVjdCAqXG5mcm9tIHRcbndoZXJlIGxlbih1cmwpID4gMTAwXG5vcmRlciBieSB2aWV3c18zNjVkIGRlc2NcbiIsImxpYi90eXBlIjoibWJxbC5zdGFnZS9uYXRpdmUiLCJ0ZW1wbGF0ZS10YWdzIjp7fX1dfSwiZGlzcGxheSI6InRhYmxlIiwicGFyYW1ldGVycyI6W10sInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9LCJ0eXBlIjoicXVlc3Rpb24ifQ==)

## Potential solutions

1. [As we do with `user_agent`](https://github.com/owid/owid-grapher/blob/3369f90048d5f531add03175d4e7356c81445596/functions/_common/analytics.ts#L64), we could send a `grapher_next` custom dimension alongside `grapher` to capture characters 101-200 (then do the concatenation in downstream analytics). There are currently no graphers with url > 200 characters, but there's no guarantee this will always be true.

2. ...?

sophiamersmann:
Resolves #5860

Adds `grapherPathNext` to send along the rest of the slug if `grapherPath` is truncated after 100 characters

### DIFF
--- a/packages/@ourworldindata/types/src/analyticsTypes/analyticsTypes.ts
+++ b/packages/@ourworldindata/types/src/analyticsTypes/analyticsTypes.ts
@@ -165,6 +165,8 @@ export interface SiteGuidedChartLinkClickParams {
     eventAction: "click"
     /** Target URL or chart path */
     eventTarget: string
+    /** Continuation of eventTarget for URLs > 100 chars (characters 101-200) */
+    eventTargetNext?: string
     /** Grapher path of the linked chart (optional - may use eventTarget instead) */
     grapherPath?: string
 }
@@ -174,6 +176,8 @@ export interface SiteChartPreviewMouseoverParams {
     eventAction: "mouseover"
     /** Target chart URL */
     eventTarget: string
+    /** Continuation of eventTarget for URLs > 100 chars (characters 101-200) */
+    eventTargetNext?: string
     /** Explorer path if in explorer */
     explorerPath?: string
     /** Grapher path being previewed */
@@ -184,7 +188,9 @@ export interface SiteChartPreviewMouseoverParams {
 
 export interface GrapherViewParams {
     /** Grapher chart path (e.g., '/grapher/life-expectancy') */
-    grapherPath: string
+    grapherPath?: string
+    /** Continuation of grapherPath for URLs > 100 chars (characters 101-200) */
+    grapherPathNext?: string
     /** View configuration ID for multi-dimensional data pages */
     viewConfigId?: string
     /** Name of the narrative chart if embedded */
@@ -196,6 +202,8 @@ export interface GrapherClickParams {
     eventAction: GrapherImageDownloadEvent | string
     /** Grapher chart path */
     grapherPath?: string
+    /** Continuation of grapherPath for URLs > 100 chars (characters 101-200) */
+    grapherPathNext?: string
     /** View configuration ID */
     viewConfigId?: string
     /** Name of narrative chart */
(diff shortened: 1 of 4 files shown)


packages/@ourworldindata/types/src/analyticsTypes/analyticsTypes.ts
```diff
@@ -184,7 +184,9 @@ export interface SiteChartPreviewMouseoverParams {
 
 export interface GrapherViewParams {
     /** Grapher chart path (e.g., '/grapher/life-expectancy') */
-    grapherPath: string
+    grapherPath?: string
```
ikesau, line 191: There's also `SiteGuidedChartLinkClick` and `SiteChartPreviewMouseover` which we should think about.

@bnjmacdonald - these two events ☝️ could be for explorers, too. 

Should we use the same event and have `grapherPath` and `grapherPathNext` if it's for a grapher, and `explorerPath` if it's an explorer? 

Right now we're just sending `eventTarget: url` which will be getting truncated for long grapher URLs.

bnjmacdonald: My preferences are really weak here, if existent at all (but thanks for checking!):

1. Re `eventTarget` vs `grapherPath`/`explorerPath` for `SiteGuidedChartLinkClick` and `SiteChartPreviewMouseover`, it makes no difference to me. So, maybe fine to keep it as `eventTarget`? Otherwise you have to introduce a bit of extra logic. Happy with either.
- I haven't been doing any analytics on `SiteGuidedChartLinkClick` or `SiteChartPreviewMouseover` events to date, so there's no concern about breaking anything downstream (assuming no one else has been doing analytics on these events either).

2. Re adding `eventTargetNext` to `SiteGuidedChartLinkClick` and `SiteChartPreviewMouseover`, sure ya that doesn't hurt and there's no substantive downside. (GA4 does have [limits](https://support.google.com/analytics/answer/14240153?hl=en) on the number of event-scoped custom dimensions, but I'm not sure if that affects us at all b/c we do all our analytics on the raw bigquery export rather than in the GA4 web dashboard or via the GA4 API).

sophiamersmann: Thank you, @bnjmacdonald! In that case I'll add `eventTargetNext` to `SiteGuidedChartLinkClick` and `SiteChartPreviewMouseover`

PR #5960, opened 2026-01-21
🎉 Data callouts

ikesau:
A new archie component that allows us to make data statements in our gdocs.

The syntax looks like this:

```
{.data-callout}
url: https://ourworldindata.org/grapher/life-expectancy?time=earliest..1970&country=NZL
[.+content]
New Zealand’s life expectancy in $latestYear(life_expectancy_0) was $latestValue(life_expectancy_0)

{.chart}
url: https://ourworldindata.org/grapher/life-expectancy?time=earliest..1970&country=NZL
{}
[]
{}
```

Or you can write one in a country profile like this:
```
{.data-callout}
url: https://ourworldindata.org/grapher/life-expectancy?time=earliest..1970&country=$entityCode
[.+content]
$EntityName’s life expectancy in $latestYear(life_expectancy_0) was $latestValue(life_expectancy_0)

{.chart}
url: https://ourworldindata.org/grapher/life-expectancy?time=earliest..1970&country=$entityCode
{}

[]
{}
```

This will render as:
image

Graphers, multi-dims, and (non-CSV-based) explorers are all supported.

Values are generated via `constructGrapherValuesJsonFromTable` which was written entirely by Claude Code based on my QA feedback. It's done this way so that values can be efficiently generated en masse during baking for profiles.

Other features:
- Coverage Matrix (for country profiles, to show how many callouts will show per country. **Accessed in the Gdocs Preview settings drawer, beneath the entity selector**)
- Country Profile `exclude` property so that authors can prevent specific country profiles from baking if their callout coverage is low
- Callout function generation (for authors to quickly generate the strings they need for a particular chart (because shortNames can be quite unwieldy))
- Skipping baking profiles for entities if it has data callouts and they're all cleared

[Test Profile](https://docs.google.com/document/d/1EXoMpjf9okN0ELfoUitxdVtIOKQmzc3xoHP_8Y1RdbE/edit?tab=t.0)
[Test Article](https://docs.google.com/document/d/1HxX5aPRRfex7G90wRELwLkcyzpFXneKN-gzO11MvDQ0/edit?tab=t.25uj1m6o9ezn)

Features to QA:
- Profiles
  - Do Graphers/Explorers/Mdims all work in the preview and mockSiteRouter
  - Do they bake
  - When callouts are specified but a country has no data for them, does its profile _not_ bake
  - Does the `exclude` property work
- Articles
  - Do Graphers/Explorers/Mdims all work in the preview and mockSiteRouter
  - Do articles with callouts in them serialize to JSON correctly for Algolia
- General
  - Parsing and error validation
  - No JS rendering
- Functions page
  - Does it work for Graphers/Explorers/Mdims
- Coverage Matrix
  - Is it accurate and usable

### DIFF
--- a/adminSiteServer/apiRoutes/gdocs.ts
+++ b/adminSiteServer/apiRoutes/gdocs.ts
@@ -232,11 +232,10 @@ export async function getGdocCalloutCoverage(
             }
             // Use a shallow copy with the instantiated URL so the key lookup matches
             const instantiatedBlock = { ...block, url: instantiatedUrl }
-            rowCoverage[rows[index].id] =
-                checkShouldDataCalloutRender(
-                    instantiatedBlock,
-                    linkedCallouts
-                )
+            rowCoverage[rows[index].id] = checkShouldDataCalloutRender(
+                instantiatedBlock,
+                linkedCallouts
+            )
         })
 
         coverageByEntity[entity.code] = rowCoverage
@@ -285,6 +284,7 @@ export async function getCalloutFunctionStrings(
         if (columnInfo?.shortName) {
             functionStringsByName[columnInfo.name] = [
                 `$latestValue(${columnInfo.shortName})`,
+                `$latestValueWithUnit(${columnInfo.shortName})`,
                 `$latestTime(${columnInfo.shortName})`,
             ]
         }
--- a/packages/@ourworldindata/utils/src/profiles.ts
+++ b/packages/@ourworldindata/utils/src/profiles.ts
@@ -536,8 +536,10 @@ export function getCalloutValue(
 ): string | undefined {
     return match(functionName)
         .with("latestValue", () => {
-            return findYDataPoint(values, parameters[0])
-                ?.formattedValueShort
+            return findYDataPoint(values, parameters[0])?.formattedValue
+        })
+        .with("latestValueWithUnit", () => {
+            return findYDataPoint(values, parameters[0])?.formattedValueShort
         })
         .with("latestTime", () => {
             return findYDataPoint(values, parameters[0])?.formattedTime
--- a/db/model/Gdoc/enrichedToMarkdown.test.ts
+++ b/db/model/Gdoc/enrichedToMarkdown.test.ts
@@ -125,7 +125,7 @@ describe("enrichedBlocksToMarkdown", () => {
                     { spanType: "span-simple-text", text: "CO2 emissions: " },
                     {
                         spanType: "span-callout",
-                        functionName: "latestValue",
+                        functionName: "latestValueWithUnit",
                         parameters: ["co2_emissions"],
                         children: [{ spanType: "span-simple-text", text: "" }],
                     },
--- a/packages/@ourworldindata/types/src/gdocTypes/Spans.ts
+++ b/packages/@ourworldindata/types/src/gdocTypes/Spans.ts
@@ -32,7 +32,11 @@ export type SpanGuidedChartLink = {
     url: string
 }
 
-export const CALLOUT_FUNCTIONS = ["latestTime", "latestValue"] as const
+export const CALLOUT_FUNCTIONS = [
+    "latestTime",
+    "latestValue",
+    "latestValueWithUnit",
+] as const
 export type CalloutFunction = (typeof CALLOUT_FUNCTIONS)[number]
 
 export type SpanCallout = {
--- a/packages/@ourworldindata/utils/src/Util.ts
+++ b/packages/@ourworldindata/utils/src/Util.ts
@@ -1979,7 +1979,8 @@ export const guidedChartRegex = /#guide:(https?:\/\/[^\s]+)/
  * Group 1: function name (e.g., "latestTime", "latestValue")
  * Group 2: parameters (e.g., "shortName")
  */
-export const plaintextCalloutRegex = /\$(latestValue|latestTime)\(([^)]*)\)/g
+export const plaintextCalloutRegex =
+    /\$(latestValueWithUnit|latestValue|latestTime)\(([^)]*)\)/g
 
 export function extractDetailsFromSyntax(str: string): string[] {
     return [...str.matchAll(new RegExp(detailOnDemandRegex, "g"))].map(


ikesau:
@codex review

ikesau:
@codex review

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
+import * as _ from "lodash-es"
```
ikesau, line 1: Lines 0-225 is cut-paste from the CF functions directory

ikesau:
@codex review please 🙂

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+import { CoreColumn, OwidTable } from "@ourworldindata/core-table"
+import { GrapherState } from "./GrapherState"
+import { makeChartState } from "../chart/ChartTypeMap"
+import { MapChartState } from "../mapCharts/MapChartState"
+import { buildSourcesLineFromColumns } from "./sourcesLine"
+
+export function constructGrapherValuesJson(
+    grapherState: GrapherState,
+    entityName: EntityName,
+    timeQueryParam?: string
```
sophiamersmann, line 36: Doesn't seem to be used. Did you add this in an earlier iteration and forgot to remove?

db/model/Gdoc/dataCallouts.ts
```diff
@@ -0,0 +1,530 @@
...
+
+    // Group by chart key to prepare each chart once
+    // Use USA as a representative entity code to create a fetchable URL
+    const urlsByChartKey = new Map<string, string>()
+    for (const templateUrl of templateUrls) {
+        // Replace $entityCode with a real code to create a fetchable URL
+        // Also handle URL-encoded $ (%24) in case the URL was encoded
+        const fetchableUrl = templateUrl
+            .replace(/\$entityCode/g, "USA")
+            .replace(/%24entityCode/gi, "USA")
```
sophiamersmann: Is dropping the country param an option? Or maybe even using stripGrapherQueryParams, i.e. dropping all params, and then fetchableUrl = chartKey?

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+
+/**
+ * Prepare a table and metadata for batch extraction of callout values.
+ * Call this once per unique chart configuration, then use
+ * constructGrapherValuesJsonFromTable for each entity.
+ */
+export function prepareCalloutTable(
+    inputTable: OwidTable,
+    config: GrapherInterface,
+    formatValueForTooltip?: (value: PrimitiveType) => string | undefined
```
sophiamersmann: Drop since unused?

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+        return omitUndefinedValues({
+            entityName,
+            columns,
+            source: sourcesLine,
+        })
+    }
+
+    // Get entity-specific times from the y columns
+    // This matches how GrapherState.times works - filtered to selected entity
+    const entityTimes = getEntityTimesFromTable(
```
sophiamersmann: Not sure this is an issue in practice but not all columns have values for all years, right? I think it would be easier if you resolved the given time against a single column. For example:
- Resolve `configMaxTime` and `timeQueryParam` here (against all available times, i.e. not entity-specific)
- Pass the resolved time to `makeDimensionValuesForTimeDirect`
- In `makeDimensionValueForColumnAndTime`, grab all available years for the given entity and column (you can use `column.owidRowsByEntityName`), and then find the year closest to the target year you passed in

I don't think it's worth doing this for the kind of bug it catches, but it might make the code easier to follow?

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+        inputTable,
+        yColumnSlugs,
+        entityName
+    )
+    const effectiveTimes = entityTimes.length > 0 ? entityTimes : times
+
+    // Resolve time bounds against entity-specific times
+    const resolveTimeBound = (bound: Time): Time | undefined => {
```
sophiamersmann: [This is the function Grapher uses](https://github.com/owid/owid-grapher/blob/d3bcfff7e15dc984ffd163ddb35e8a38902e90cb/packages/%40ourworldindata/utils/src/Util.ts#L647), which might be helpful to you here?

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+ * chart's transformed table includes transformations that are not
+ * applied to grapherState.transformedTable (e.g. relative mode in
+ * line charts).
+ */
+const getTransformedColumn = (
+    grapherState: GrapherState,
+    slug?: string
+): CoreColumn => grapherState.chartState.transformedTable.get(slug)
+
+export function isValuesJsonValid(valuesJson: GrapherValuesJson): boolean {
```
sophiamersmann: Maybe a different name since an (almost) empty GrapherValuesJson object is technically valid?

sophiamersmann: Ah now I get it, an almost empty GrapherValuesJson object would pass, too. Idk, this function looks a bit like Claude churn to me, maybe give it a less ambitious name and only check what you need?

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+export function prepareCalloutTable(
+    inputTable: OwidTable,
+    config: GrapherInterface,
+    formatValueForTooltip?: (value: PrimitiveType) => string | undefined
+): PreparedCalloutTable {
+    const dimensions = config.dimensions ?? []
+
+    // Compute column slug the same way ChartDimension does:
+    // slug ?? variableId.toString()
+    const getColumnSlug = (d: (typeof dimensions)[0]): string | undefined =>
```
sophiamersmann: Alternatively, you could create [ChartDimension objects](https://github.com/owid/owid-grapher/blob/master/packages/%40ourworldindata/grapher/src/chart/ChartDimension.ts) and then use its slug, so for example something like

```
const dimensions = config.dimensions.map((dimensionConfig) => new ChartDimension(dimensionConfig, { table: inputTable }))
```

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -0,0 +1,545 @@
...
+    // Build sources line from columns
+    const sourcesLine =
+        config.sourceDesc ??
+        buildSourcesLineFromColumns(inputTable.getColumns(_.uniq(yColumnSlugs)))
+
+    // Get sorted unique times from the table
+    const times = inputTable.getTimesUniqSortedAscForColumns(yColumnSlugs)
+
+    // Parse configured time bounds (handle both numeric and string formats)
+    const parseTimeBound = (
```
sophiamersmann: I think `minTimeBoundFromJSONOrNegativeInfinity` and `maxTimeBoundFromJSONOrPositiveInfinity` is probably what you want to use here

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
sophiamersmann: Or maybe you don't need this at all because you have `checkShouldDataCalloutRender`? But I might be missing something

db/model/Gdoc/dataCallouts.ts
```diff
@@ -0,0 +1,530 @@
...
+    if (!inputTable) return undefined
+
+    return { config, inputTable }
+}
+
+/**
+ * Prepare a chart config and input table for an explorer URL.
+ * This resolves the explorer's config based on query params and fetches the data.
+ */
+async function prepareCalloutTableForExplorer(
```
sophiamersmann, line 132: I wonder if we should just grab the config from the `explorer_views` table instead of doing this manually. I'm a bit hesitant to suggest this because it's going to put [this rewrite](https://github.com/owid/owid-grapher/pull/5336) of mine to the test. It's a rewrite of indicator-based explorers that I did when I wanted to use the stored explorer view configs for search and realised they're a bit too far from the truth to be usable. I didn't end up actually relying on the stored explorer view configs, but _I think_ the rewrite solved the biggest offenders.

sophiamersmann: Or maybe you've considered using the `explorer_views` table and it doesn't work for some reason?

packages/@ourworldindata/utils/src/profiles.ts
```diff
@@ -403,3 +423,116 @@ export function checkShouldConditionalSectionRender({
...
+                      ([_, col]) => col.shortName === shortName
+                  )?.[0]
+                : undefined
+            if (!columnSlug) return undefined
+
+            const dataPoint = values.endValues?.y?.find((dp) => {
+                return dp.columnSlug === String(columnSlug)
+            })
+
+            return dataPoint?.formattedValue
```
sophiamersmann: I realise now this is a bit confusing, but maybe you want to use `formattedValueShort` here so that it comes with a short unit, e.g. for life expectancy, `formattedValueShort` would be `80 years` and `formattedValue` would be just `80`, I think

adminSiteServer/apiRoutes/gdocs.ts
```diff
@@ -99,6 +113,174 @@ export async function getIndividualGdoc(
...
+
+            // Extract values using the optimized method
+            const values = constructGrapherValuesJsonFromTable(
+                prepared,
+                entity.name,
+                url.queryParams.time
+            )
+
+            // Check if the values are valid (has data)
+            const isValid = isValuesJsonValid(values)
```
sophiamersmann: Wouldn't it be safer to attempt extracting the value and checking if it exists?

packages/@ourworldindata/utils/src/profiles.ts
```diff
@@ -403,3 +423,116 @@ export function checkShouldConditionalSectionRender({
...
+        })
+        .with("latestYear", () => {
```
sophiamersmann: Maybe a bit of a misnomer since this might also refer to dates?

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
ikesau: Oops, yes!

PR #5977, opened 2026-01-22
Add make refresh.swap for atomic database refresh

Marigold:
## Motivation

`staging-site-master` runs `make refresh` after every deploy to get the freshest data possible. The problem with `make refresh` is that it takes a long time during which there might be active connections to the db (grapher build, etl, etc.). I tried solving that with locks, but that didn't really work.

## Summary

- Add new `make refresh.atomic` command that refreshes the grapher database without killing active connections
- Uses atomic table swap via MySQL's `RENAME TABLE` instead of dropping/recreating the database
- Other processes (ETL, analytics, etc.) can continue using the database during refresh

## How it works

Instead of `DROP DATABASE` + `CREATE DATABASE` + import (which kills all connections), this approach:

1. Imports fresh data into a temporary database (`grapher_new`)
2. Atomically swaps all tables using a single `RENAME TABLE` statement:
   ```sql
   RENAME TABLE 
     grapher.charts TO grapher_old.charts,
     grapher_new.charts TO grapher.charts,
     ...
   ```
3. Drops both temporary databases (`grapher_new` and `grapher_old`)

The `RENAME TABLE` statement is atomic - all renames happen together, so there's never a moment where tables are missing. Existing connections may get an error on their next query, but can immediately retry and succeed.

## Test plan

- [x] Run `make refresh.atomic` with no existing database (first-time setup path)
- [x] Run `make refresh.atomic` with an existing database (swap path)
- [x] Verify other processes connected to the database survive the refresh

## TODO before merging

- [x] Grant the following to DBs on all `staging-site-*`

```
GRANT ALL PRIVILEGES ON *.* TO 'owid'@'%';
FLUSH PRIVILEGES;
```

### DIFF
--- a/devTools/docker/atomic-grapher-data.sh
+++ b/devTools/docker/atomic-grapher-data.sh
@@ -82,6 +82,16 @@ swapGrapherDb() {
     # Drop old backup database if exists, create fresh one
     _mysql --database="" -e "DROP DATABASE IF EXISTS $OLD_DB; CREATE DATABASE $OLD_DB CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;"
 
+    # Get view definitions from new DB BEFORE the swap (while tables still exist there)
+    echo "==> Extracting view definitions"
+    NEW_VIEWS=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$NEW_DB' AND TABLE_TYPE='VIEW'")
+    declare -A VIEW_DEFS
+    for VIEW in $NEW_VIEWS; do
+        # Get the CREATE VIEW statement, strip DEFINER clause, and replace database references
+        VIEW_DEF=$(_mysql -N -e "SHOW CREATE VIEW $NEW_DB.$VIEW" | cut -f2 | sed 's/ DEFINER=`[^`]*`@`[^`]*`//' | sed "s/\`$NEW_DB\`\./\`$GRAPHER_DB_NAME\`./g")
+        VIEW_DEFS[$VIEW]="$VIEW_DEF"
+    done
+
     # Get list of tables in new DB (incoming data)
     NEW_TABLES=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$NEW_DB' AND TABLE_TYPE='BASE TABLE'")
 
@@ -125,8 +135,8 @@ swapGrapherDb() {
     echo "==> Executing atomic table swap"
     _mysql --database="" -e "$RENAME_CMD"
 
-    # Handle views: drop existing views and recreate from new DB
-    # Views can't be renamed across databases, so we extract and recreate them
+    # Handle views: drop existing views and recreate from extracted definitions
+    # Views can't be renamed across databases, so we recreate them
     echo "==> Updating views"
 
     # Drop all existing views in main database
@@ -135,14 +145,9 @@ swapGrapherDb() {
         _mysql --database="$GRAPHER_DB_NAME" -e "DROP VIEW IF EXISTS $VIEW"
     done
 
-    # Get view definitions from new DB and recreate them in main DB
-    NEW_VIEWS=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$NEW_DB' AND TABLE_TYPE='VIEW'")
-    for VIEW in $NEW_VIEWS; do
-        # Get the CREATE VIEW statement and adapt it for the main database
-        VIEW_DEF=$(_mysql -N -e "SHOW CREATE VIEW $NEW_DB.$VIEW" | cut -f2)
-        # Replace references to new DB with main DB in the view definition
-        VIEW_DEF="${VIEW_DEF//$NEW_DB/$GRAPHER_DB_NAME}"
-        _mysql --database="$GRAPHER_DB_NAME" -e "$VIEW_DEF"
+    # Recreate views in main DB using the extracted definitions
+    for VIEW in "${!VIEW_DEFS[@]}"; do
+        _mysql --database="$GRAPHER_DB_NAME" -e "${VIEW_DEFS[$VIEW]}"
     done
 
     echo "==> Cleaning up"


devTools/docker/atomic-grapher-data.sh
```diff
@@ -0,0 +1,136 @@
...
+    _mysql --database="" -e "DROP DATABASE IF EXISTS $OLD_DB; CREATE DATABASE $OLD_DB CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;"
+
+    # Get list of tables in new DB (incoming data)
+    NEW_TABLES=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$NEW_DB' AND TABLE_TYPE='BASE TABLE'")
+
+    # Get list of tables in main (current data)
+    MAIN_TABLES=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$GRAPHER_DB_NAME' AND TABLE_TYPE='BASE TABLE'")
+
+    # Build the atomic rename command
+    RENAME_CMD="SET FOREIGN_KEY_CHECKS=0; RENAME TABLE "
```
marcelgerber: out of interest, I tried out running the atomic swap without `FOREIGN_KEY_CHECKS=0`, and it still worked.

Given that, I'd wager to remove the line, to ensure we get nice integrity checks.

Marigold: Nice catch!

Marigold:
> have you checked that Codex's review comment really does not apply to us? it sounds valid, and we indeed have 3 views that we would also like to be updated when running this command.

> Just now, I tried deleting a view, and indeed running make refresh.atomic didn't properly recreate it.

I checked it, but messed up when testing it and thought it's not real (dumb, I know...). It's fixed now.

devTools/docker/atomic-grapher-data.sh
```diff
@@ -122,11 +122,29 @@ swapGrapherDb() {
...
+        _mysql --database="$GRAPHER_DB_NAME" -e "DROP VIEW IF EXISTS $VIEW"
+    done
+
+    # Get view definitions from new DB and recreate them in main DB
+    NEW_VIEWS=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$NEW_DB' AND TABLE_TYPE='VIEW'")
+    for VIEW in $NEW_VIEWS; do
+        # Get the CREATE VIEW statement and adapt it for the main database
+        VIEW_DEF=$(_mysql -N -e "SHOW CREATE VIEW $NEW_DB.$VIEW" | cut -f2)
+        # Replace references to new DB with main DB in the view definition
+        VIEW_DEF="${VIEW_DEF//$NEW_DB/$GRAPHER_DB_NAME}"
```
marcelgerber: I've looked at the output of `SHOW CREATE VIEW`, and it doesn't even have a reference to `NEW_DB`.

But on the flipside we should rather strip out the definer, which is a regex that's less likely to wreak havoc than this one is.

see https://github.com/owid/owid-grapher/blob/b6addb486b6c7f1f4071908675510116db5bc29a/devTools/docker/refresh-grapher-data.sh#L26-L27.


`SHOW CREATE VIEW` output for reference:
```sql
CREATE ALGORITHM=UNDEFINED DEFINER=`grapher`@`%` SQL SECURITY DEFINER VIEW `active_datasets` AS
SELECT `datasets`.`id` AS `id`,
       `datasets`.`name` AS `name`,
       `datasets`.`description` AS `description`,
       `datasets`.`createdAt` AS `createdAt`,
       `datasets`.`updatedAt` AS `updatedAt`,
       `datasets`.`namespace` AS `namespace`,
       `datasets`.`isPrivate` AS `isPrivate`,
       `datasets`.`createdByUserId` AS `createdByUserId`,
       `datasets`.`metadataEditedAt` AS `metadataEditedAt`,
       `datasets`.`metadataEditedByUserId` AS `metadataEditedByUserId`,
       `datasets`.`dataEditedAt` AS `dataEditedAt`,
       `datasets`.`dataEditedByUserId` AS `dataEditedByUserId`,
       `datasets`.`nonRedistributable` AS `nonRedistributable`,
       `datasets`.`isArchived` AS `isArchived`,
       `datasets`.`sourceChecksum` AS `sourceChecksum`,
       `datasets`.`shortName` AS `shortName`,
       `datasets`.`version` AS `version`,
       `datasets`.`updatePeriodDays` AS `updatePeriodDays`
FROM `datasets`
WHERE (0 = `datasets`.`isArchived`)
```

marcelgerber (approved):

devTools/docker/atomic-grapher-data.sh
Marigold: Oh, missed that, thanks! I've only tested whether it works and didn't realise there are redundant commands.

Marigold:
I also ran some ad-hoc comparisons of tables & views and it seems to be working. I'll use it only on staging servers and see whether it helps with the race conditions and locking hell we're experiencing in ops.

devTools/docker/atomic-grapher-data.sh
```diff
@@ -82,6 +82,16 @@ swapGrapherDb() {
     # Drop old backup database if exists, create fresh one
     _mysql --database="" -e "DROP DATABASE IF EXISTS $OLD_DB; CREATE DATABASE $OLD_DB CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_as_cs;"
 
+    # Get view definitions from new DB BEFORE the swap (while tables still exist there)
+    echo "==> Extracting view definitions"
+    NEW_VIEWS=$(_mysql -N -e "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA='$NEW_DB' AND TABLE_TYPE='VIEW'")
+    declare -A VIEW_DEFS
+    for VIEW in $NEW_VIEWS; do
+        # Get the CREATE VIEW statement, strip DEFINER clause, and replace database references
+        VIEW_DEF=$(_mysql -N -e "SHOW CREATE VIEW $NEW_DB.$VIEW" | cut -f2 | sed 's/ DEFINER=`[^`]*`@`[^`]*`//' | sed "s/\`$NEW_DB\`\./\`$GRAPHER_DB_NAME\`./g")
```
marcelgerber, line 91: still in there 🍃

```bash
| sed "s/\`$NEW_DB\`\./\`$GRAPHER_DB_NAME\`./g"
```

Marigold: So I tried it without `sed` and got an error during migration
```
==> Updating views
ERROR 1146 (42S02) at line 1: Table 'grapher_new.charts' doesn't exist
```

It seems to add database name in `SHOW CREATE VIEW`. You can test it on this example
```
mysql> CREATE DATABASE test_new;
mysql> CREATE TABLE test_new.my_table (id INT);
mysql> CREATE VIEW test_new.my_view AS SELECT id FROM test_new.my_table;
mysql> SHOW CREATE VIEW test_new.my_view\G
         Create View: CREATE ... VIEW `test_new`.`my_view` AS 
                      select `test_new`.`my_table`.`id` AS `id` 
                      from `test_new`.`my_table`
```

Maybe the reason the final view shows unqualified names (just datasets not grapher.datasets) is because MySQL optimizes them away when the view and tables are in the same database - but that only happens after our sed replaces grapher_new with grapher.

marcelgerber: okay, my last attempt at this is to instead use `mysql --database=$NEW_DB`.
In my local env, whether I run `USE grapher` or `USE other_db` beforehand makes a difference in the output.

But if that also doesn't help, then please just go ahead and merge it as-is.

PR #5978, opened 2026-01-22
🐛 info tooltip rendered behind settings popup

sophiamersmann:
The info tooltip is currently shown behind the settings popup. Fixing this with Tippy was surprisingly annoying, so I just switched to React aria.

[Prod](https://ourworldindata.org/grapher/types-violent-crime-rate-us) / [Staging](http://staging-site-fix-tooltip-in-settings-popu/grapher/types-violent-crime-rate-us)

Screenshot 2026-01-21 at 23 10 05

### DIFF
--- a/packages/@ourworldindata/components/package.json
+++ b/packages/@ourworldindata/components/package.json
@@ -23,6 +23,7 @@
         "mobx": "^6.13.7",
         "mobx-react": "^7.6.0",
         "react": "^19.2.0",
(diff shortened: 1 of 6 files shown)


sophiamersmann:
@rakyi, is [this](https://github.com/owid/owid-grapher/pull/5978#issuecomment-3789761735) something to be concerned about? 👆🏻

rakyi:
I don't think so. It's very likely a false positive.

PR #5982, opened 2026-01-22
🔨 use internal set for SelectionArray

sophiamersmann:
Refactors SelectionArray to use a set internally (instead of a list). Seems like the more appropriate data model to me. I can't think of any reason we'd ever want duplicate entries in the SelectionArray. And sets do respect insertion order, so I think we should be ok there too. I can't think of any other way this can hurt us


sophiamersmann:
* **#5982** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/5982?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #6023, opened 2026-01-28
✨ static viz analytics

ikesau:
Adds analytic events for static viz downloads and fixes some pluralization issues when there's only one image specified for a static viz.


ikesau:
@codex review

PR #6039, opened 2026-01-30
✨ use more standard keyboard navigation for content switchers

sophiamersmann:
The keyboard navigation of the content switchers (to switch between views in Grapher) is a bit weird because the +X 'tab' is actually a trigger button to open a popup. I think keeping the tabs and the +X button separate is less surprising for keyboard users.

- On [prod](https://ourworldindata.org/grapher/children-per-woman-un): You can navigate to +X with the arrow key, then the popup automatically opens
- On [staging](http://staging-site-content-switchers-a11y/grapher/children-per-woman-un): Arrow keys only navigate between chart tabs, tabbing brings you to the +X button

---
- <kbd>&nbsp;2&nbsp;</kbd> #6039 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #6038


sophiamersmann:
@codex review

PR #6042, opened 2026-02-02
Add Scheduled publish status and label

edomt:
- Introduce a 'Scheduled' publish status to the gdocs index:
  - add GdocPublishStatus.Scheduled
  - include it in the publish-status select
  - detect scheduled docs (published with a future publishedAt).
- Update filtering to handle Scheduled and exclude scheduled items from Published results.
- When viewing Scheduled, sort items by publishedAt ascending.
- In the UI, show a yellow 'Scheduled for [date]' label for scheduled docs and add corresponding CSS styling in GdocsIndexPage.scss.

### DIFF
--- a/adminSiteClient/GdocsIndexPage.tsx
+++ b/adminSiteClient/GdocsIndexPage.tsx
@@ -217,10 +217,7 @@ export class GdocsIndexPage extends React.Component<RouteComponentProps> {
                       // show this document if its type is active
                       this.filters[gdoc.type]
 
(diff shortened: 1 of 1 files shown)


edomt:
@ikesau Would you be able to review this, please? 🙂

No rush, it's just a little quality-of-life update that would be nice — mostly for me (I think) when I want to double-check that I've properly scheduled DIs and articles.

Capture d’écran 2026-02-02 à 09 18 56

PR #6044, opened 2026-02-02
Remove getPrimaryTopic error logging

rakyi:
We now get a ton of false positives from this check when authors work on a new topic page, which introduces a new tag that gets a slug added only after publication, thus the check is no longer useful.

[Sentry](https://our-world-in-data.sentry.io/issues/37405224/)

### DIFF
--- a/baker/DatapageHelpers.ts
+++ b/baker/DatapageHelpers.ts
@@ -20,7 +20,6 @@ import {
 } from "@ourworldindata/types"
 import { KnexReadonlyTransaction } from "../db/db.js"
 import { parseFaqs } from "../db/model/Gdoc/rawToEnriched.js"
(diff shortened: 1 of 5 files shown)


baker/DatapageHelpers.ts
```diff
@@ -135,8 +134,7 @@ export const resolveFaqsForVariable = (
 
 export const getPrimaryTopic = async (
     knex: KnexReadonlyTransaction,
-    tags: string[] = [],
```
ikesau, line 138: Hm okay, fair point.

May as well remove the error from `getSlugForTopicTag` then and remove the related try/catch blocks, no? 🙂

ikesau (approved):

baker/DatapageHelpers.ts
rakyi: Yeah, good point!

Issue #5209, opened 2025-07-30
Enable scatter plots as secondary charts (scatter vs gdp per capita by default)


danyx23:
We are now able to show multiple chart types in one chart, but one important one that is still missing is the scatter plot (vs gdp per capita by default).

We should add this as an option.

When we do this, the x axis will be a bit of an issue - both the scatter an marimekko charts use this but they would prefer different x axis if specified (e.g. gdp per capita for scatter, population for marimekko). Marimekko charts without an x axis are still very useful (and often preferred), so secondary Marimekko charts should just ignore the x axis if scatter is also enabled.

PR #6046, opened 2026-02-02
🎉 support scatter plots as secondary tab

sophiamersmann:
Resolves https://github.com/owid/owid-grapher/issues/5209

Adds a scatter tab to line charts & co.

This is the trickiest addition so far because of the x-axis:
- If a Mariemkko tab is present as well, the Marimekko chart ignores the x-axis (because Marimekkos are typically clearer without an x-axis)
- Line charts ignore x-axis settings when a scatter is present (i.e. the x-axis settings are thus only applied to the scatter's x-axis, not the line chart's time axis); in practice this means that authors can't set a custom label for line charts when a scatter is enabled (which is probably fine)

If a scatter tab is added, then the x, size and color dimensions are automatically prefilled (with GDP per capita, population, and the continents indicators, resp).

The Tabs Section in the admin now looks quite bad, which is why I redesigned it in a [later pr](https://github.com/owid/owid-grapher/pull/6136)

### Testing

It's best to use the the last pr in the stack for testing: http://staging-site-admin-tabs

---
This is **part 1 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #6152 
- <kbd>&nbsp;3&nbsp;</kbd> #6136 
- <kbd>&nbsp;2&nbsp;</kbd> #6135 
- <kbd>&nbsp;1&nbsp;</kbd> #6046 👈


adminSiteClient/EditorBasicTab.tsx
```diff
@@ -602,36 +592,14 @@ export class EditorBasicTab<
...
             value === this.chartTypeOptionNone
-                ? []
-                : [value as GrapherChartType]
+                ? undefined
+                : (value as GrapherChartType)
 
-        if (grapherState.isMarimekko) {
-            grapherState.hideRelativeToggle = false
-            grapherState.stackMode = StackMode.relative
-        }
```
sophiamersmann, line 612: I accidentally dropped this but brought it back in a stacked pr (I just didn't want to go through the rebase)

packages/@ourworldindata/grapher/src/stackedCharts/MarimekkoChartState.ts
```diff
@@ -445,6 +448,8 @@ export class MarimekkoChartState implements ChartState, ColorScaleManager {
             axis.updateDomainPreservingUserSettings(this.xDomainDefault)
         }
 
+        // Marimekko charts should always use linear scale
+        axis.scaleType = ScaleType.linear
```
ikesau, line 452: This isn't strictly related to the PR's main changes, right?

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -2883,7 +2894,15 @@ export class GrapherState
...
             return false
 
-        if (isOnMarimekkoTab && xColumnSlug === undefined) return false
+        // Disable relative mode for Marimekko charts without an x dimension
+        if (
+            isOnMarimekkoTab &&
+            (xColumnSlug === undefined ||
+                // If a scatter is present, the x-axis is ignored
+                (xColumnSlug && hasScatter))
+        )
```
ikesau: Are you specifically trying to guard against xColumnSlug being `""` or could this simplify to `(!xColumnSlug || hasScatter)` ?

packages/@ourworldindata/grapher/src/stackedCharts/MarimekkoChartState.ts
sophiamersmann: nope, drive-by change

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
sophiamersmann: yup, can be simplified. I just tagged on the extra condition. thanks!

sophiamersmann:
> I tried testing with [charts with projections](http://staging-site-admin-tabs/admin/charts/8561/edit) and they didn't work / it was a little glitchy. Maybe something to fix, but not a blocker I don't think.

good point! I first thought I would just hide the chart types that don't support projected data in the admin, but that wouldn't really work because then authors can't switch to them. So, I just left it as is for now since it's a bit of an edge case and I don't want to introduce that kind of complexity to the code. But I realised that there is a somewhat related bug where the chart tab validation was buggy when projected indicators were present, which is fixed [here](https://github.com/owid/owid-grapher/pull/6152)

vpicouet:
Hi! Thanks for this addition. I'm using scatter plots in an explorer with indicators that have projections (e.g. population with UN projections), and I'm running into an issue I think mentioned by @ikesau : the scatter plot stops at the last historical data point (~2025) and doesn't allow browsing into projection years.
I was wondering if there was indeed a plan to support projected data in scatter plots? Even a basic behavior could already be very useful. Thanks a lot!

sophiamersmann:
Hey @vpicouet, supporting projected data in scatter plots is definitely more interesting now that scatter plots can be added as an additional tab to line and bar charts, but we don’t have plans to add support for this at the moment.

danyx23:
@vpicouet out of curiousity, can you say more about how you are using our explorer? We always had the impression that our tools are not re-used much and are weighing whether our current approach to re-use makes sense. It would therefore be valuable for us know more about projects that benefit from our tools.

PR #6050, opened 2026-02-03
feat: add gdocs component for bespoke viz

marcelgerber:
Adds a `{.bespoke-component}` component.

It has an in-code registry of component "bundles" (which are just pointers to JS & CSS URLs), and the JS entrypoint needs to export a
```ts
mount: (
    container: HTMLElement,
    opts: { variant?: string; config: Record<string, unknown> }
)
```
method.

The component is a client-only component; it currently doesn't do anything before hydration and just renders an (infinite) loading indicator.

### What is `variant`?
The `variant` is an optional pointer for multiple visualizations to be part of the same bundle, but to be embedded in different spots inside the gdoc - but still being able to share state.
In theory, this could also be achieved by just using `config`, but I feel like it's a common-enough use case that we want to have an "approved" way to do this.

### How this works

Technically, we load the JS file using dynamic await, and then call `mount` on a div we create inside of a Shadow DOM.
The CSS is also loaded into the Shadow DOM, and thus the bespoke component is fully encapsulated when it comes to CSS.

### Drawbacks of Shadow DOM
This creates a bunch of small problems also, though: We need to use `:host` rather than `:root` inside the CSS to define global variables, and need to ensure that any code that mounts elements inside portals (e.g. tooltip code like `floating-ui` in my case) mounts its element inside the Shadow DOM, because if it gets mounted outside it doesn't have CSS styles.

### Try it out
You can [preview this here](http://staging-site-gdocs-bespoke-component/admin/gdocs/15Lq5MoWF8ZaqHnN2Ld4Dv6AtOyT-om1F-TjeHqz271M/preview) ([this is the doc](https://docs.google.com/document/d/15Lq5MoWF8ZaqHnN2Ld4Dv6AtOyT-om1F-TjeHqz271M/edit))

In the demo, you can easily see how the two charts are connected in state. It's easy to disconnect their state (but the bespoke implementation needs to handle that), but the ability to have them connected is really useful in some of our use cases.

## TODOs
- [ ] Remove test component before merging
- [ ] Make the error message display more end-user-friendly

## Open questions
- [ ] What's a good way to support sizing, and avoid layout reflows?
- [ ] What changes would we need to make to allow for articles containing bespoke components to be archived?
- [ ] Should we remove the `.script` component now?
- [ ] Should we add a way to mount a component without using Shadow DOM (as an option in the component registry; we can also easily do this later on)?

## Screenshot
CleanShot 2026-02-03 at 17 16 51

### DIFF
--- a/site/bespokeComponentRegistry.ts
+++ b/site/bespokeComponentRegistry.ts
@@ -23,16 +23,6 @@ export const BESPOKE_COMPONENT_REGISTRY: Record<
     //     scriptUrl: "/assets/bespoke/my-widget.mjs",
     //     cssUrl: "/assets/bespoke/my-widget.css",
     // },
-    test: {
-        scriptUrl:
-            "https://owid-public.owid.io/marcel-bespoke-data-viz-02-2026/poverty-plots/income-plots.mjs",
-        cssUrl: "https://owid-public.owid.io/marcel-bespoke-data-viz-02-2026/poverty-plots/income-plots.css",
-    },
-    "how-rich-am-i": {
-        scriptUrl:
-            "https://owid-public.owid.io/marcel-bespoke-data-viz-02-2026/how-rich-am-i/how-rich-am-i.mjs?nocache",
-        cssUrl: "https://owid-public.owid.io/marcel-bespoke-data-viz-02-2026/how-rich-am-i/how-rich-am-i.css?nocache",
-    },
 }
 
 /**


db/model/Gdoc/enrichedToMarkdown.ts
```diff
@@ -505,5 +505,9 @@ ${links}`
                 .map((link) => `* [${link.text}](${link.url})`)
                 .join("\n")
         })
+        .with(
+            { type: "bespoke-component" },
+            (_): string | undefined => undefined
```
marcelgerber, line 614: Open question: Do we want _anything_ to be inclueded in the markdown where there is a bespoke component?

db/model/Gdoc/rawToArchie.ts
```diff
@@ -1000,6 +1001,24 @@ function* rawBlockSocialsToArchieMLString(
...
+    yield* propertyToArchieMLString("variant", block.value)
+    yield* propertyToArchieMLString("size", block.value)
+    if (block.value.config) {
+        // TODO this doesn't support deep nesting, which in theory the type Record<string, unknown> allows for
+        yield `{.config}`
+        for (const [key, value] of Object.entries(block.value.config)) {
+            yield* propertyToArchieMLString(key, { [key]: value })
+        }
+        yield "{}"
+    }
```
marcelgerber, line 1048: Not sure whether there is a better way to convert an arbitrary key-value object into Archie; or whether we have had that case before for any other component?

site/gdocs/components/BespokeComponent.tsx
marcelgerber, line 1: This is where the action is

site/bespokeComponentRegistry.ts
```diff
@@ -0,0 +1,45 @@
...
+    //     scriptUrl: "/assets/bespoke/my-widget.mjs",
+    //     cssUrl: "/assets/bespoke/my-widget.css",
+    // },
+    test: {
+        scriptUrl:
+            "https://owid-public.owid.io/marcel-bespoke-data-viz-02-2026/poverty-plots/income-plots.mjs",
+        cssUrls: [
+            "https://owid-public.owid.io/marcel-bespoke-data-viz-02-2026/poverty-plots/income-plots.css",
+        ],
+    },
```
marcelgerber: This is an example registry entry; and what is powering the demo that you're seeing. I'll remove it before we merge the PR.

site/gdocs/components/BespokeComponent.tsx
```diff
@@ -0,0 +1,180 @@
...
+                const jsPromise = import(
+                    definition.scriptUrl
+                ) as Promise<BespokeComponentModule>
+
+                if (abortController.signal.aborted) return
+
+                // Create a container div inside the shadow root for the component to render into
+                const mountContainer = document.createElement("div")
+                mountContainer.className = "bespoke-container"
+                shadowRoot!.appendChild(mountContainer)
```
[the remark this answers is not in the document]
marcelgerber: How would I go about doing this?

db/model/Gdoc/enrichedToMarkdown.ts
ikesau: We currently only use the markdown representation for search snippets. Seems unlikely we could think of some textual representation of the component here that wouldn't be redundant in the context of the article. I say skip it.

db/model/Gdoc/rawToArchie.ts
ikesau: Can't think of one! 

edit: the closest we come to it is arbitrary nesting of freeform content fields e.g. a sticky-right section inside a gray section inside a guided-chart section. But all those keys are known in advance.

db/model/Gdoc/rawToArchie.ts
```diff
@@ -1000,6 +1001,24 @@ function* rawBlockSocialsToArchieMLString(
...
 
+function* rawBlockBespokeComponentToArchieMLString(
+    block: RawBlockBespokeComponent
+): Generator<string, void, undefined> {
+    yield "{.bespoke-component}"
+    yield* propertyToArchieMLString("bundle", block.value)
+    yield* propertyToArchieMLString("variant", block.value)
+    yield* propertyToArchieMLString("size", block.value)
+    if (block.value.config) {
+        // TODO this doesn't support deep nesting, which in theory the type Record<string, unknown> allows for
```
ikesau: I wonder if it'd be smart to just decide now that nesting isn't allowed, and check for it in parsing/validation.

i.e. we'd force:
```
{.config}
aspect-ratio-small: 1
aspect-ratio-large: 2
{}
```

instead of 
```
{.config}
{.aspect-ratio}
small: 1
large: 2
{}
{}
```

I don't have a strong opinion on this, but maybe it makes it a little easier to work with and a little harder to write? (Though, the most common archieML breakage we encounter is authors mistakenly missing/closing brackets, so this might make that failure mode a little rarer)

db/model/Gdoc/rawToEnriched.ts
```diff
@@ -463,6 +466,44 @@ const parseBlockquote = (raw: RawBlockBlockquote): EnrichedBlockBlockquote => {
...
+        parseErrors: [error],
+    })
+
+    if (!raw.value.bundle) {
+        return createError({
+            message: "Bundle property is required for bespoke-component",
+        })
+    }
+
+    const size = checkIsBlockSize(raw.value.size)
```
ikesau: I find this pattern better so that typos are caught:

```
// Default to wide
    const size = image.value.size ?? BlockSize.Wide
    if (!checkIsBlockSize(size)) {
        return createError({
            message: `Invalid size property: ${size}`,
        })
    }
```

site/gdocs/components/BespokeComponent.tsx
```diff
@@ -0,0 +1,180 @@
...
+                const cssPromise = loadCssIntoShadow(
+                    shadowRoot!,
+                    definition.cssUrl
+                )
+                // Dynamically import the JS module
+                const jsPromise = import(
+                    definition.scriptUrl
+                ) as Promise<BespokeComponentModule>
+
+                if (abortController.signal.aborted) return
```
ikesau: Can this (and the one on line 79) ever do anything? 

`abort()` can only get called by the cleanup function, and the cleanup function can only run when the thread is yielded by `await`.

ikesau (commented):
All the Archie stuff looks good! Will QA this tomorrow before full approval 👍

site/gdocs/components/ArticleBlock.tsx
```diff
@@ -939,6 +940,15 @@ function ArticleBlockInternal({
                 links={block.links}
             />
         ))
+        .with({ type: "bespoke-component" }, (block) => (
+            <BespokeComponent
+                className={getLayout(
+                    `bespoke-component--${block.size}`,
```
ikesau, line 1037: What if we just rendered the container at 12 columns no matter what, and the viz itself took care of its sizing?

Doing it this way (with `block.size`) means that the logic for sizing a bespoke data viz will exist in at least 2 places:
1. The archie wrapper component (adding a dependency on our site's CSS (unlikely to change, but technically possible)
2. The bespoke viz rendering & CSS

Given that there's a config object that we can pass down (if we want a big and small version of the same viz) I feel like we might as well centralize?

site/gdocs/components/BespokeComponent.tsx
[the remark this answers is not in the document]
ikesau: Maybe you could still fix this, but it seems like... not a big deal to me, given that `block.config` and `variant` won't change, due to how we only ever hydrate the page once.

site/gdocs/components/BespokeComponent.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (error) {
+        return <BespokeError className={className} message={error} />
+    }
+
+    return (
+        <div
+            className={className}
+            style={{
+                position: "relative",
+                minHeight: 300,
```
ikesau, line 142: So this is where we could implement the aspect ratio system we were talking about.

If we go with the "centralize sizing information in one place" idea that I mentioned in the other comment, does that mean we'll also need a `max-width` attribute for each variant? 🤔

```
{.config}
desktop-aspect-ratio: 1.618
max-width: 900px
mobile-aspect-ratio: 1
// we'd probably want mobile max width to always be 100%
{}
```

Or if we want it to be more consistent, use a `size` enum.

site/bespokeComponentRegistry.ts
ikesau: Right so we still need to figure out the way we write and generate these bundles. I guess for starters it doesn't need to be too systematized.

site/gdocs/components/ArticleBlock.tsx
marcelgerber: Hm, I'm not sure how this will play along with putting this component into other, small components - whether it's an aside or side-by-side or anything else - but I'll keep it in mind when thinking about grid-aware sizing.

marcelgerber:
@codex review

Issue #6047, opened 2026-02-03
Making tags 'Searchable in Algolia' doesn't seem to work as expected


spoonerf:
## Description

There is a new tick box in the tag admin to make tags searchable in Algolia, without them needing to be a topic tag, but it doesn't seem to be working as expected. 

It has been ticked for [Sexual & Reproductive Health](https://admin.owid.io/admin/tags/1864) for a couple of weeks now, but this tag doesn't appear in Search. 

However, ['Crime'](https://admin.owid.io/admin/tags/284) which has the same characteristics does appear.

## Expected behaviour

Sexual & Reproductive Health should appear as a tag in Search. 

## Steps to reproduce

1. Go to this URL : https://ourworldindata.org/search
2. Type in: Sexual & Reproductive Health
3. Notice how the tag for this doesn't appear. Compare by typing in 'Crime' to see expected behaviour. 

## Screenshots

<img width="664" height="324" alt="Image" src="https://github.com/user-attachments/assets/9cf01ebd-7b02-491c-8680-bf2ff6269095" />
<img width="738" height="377" alt="Image" src="https://github.com/user-attachments/assets/a1a3c026-5d53-4cb8-8d49-4fddfedf4c04" />

mlbrgl:
@spoonerf 
For a tag to show up in search, it needs to be:
- "searchable": either a topic tag (i.e. a tag with a slug matching a topic page) or a regular tag with the new "searchable in algolia" box ticked
- AND in the tag graph: only tags in the tag graph are considered, and then filtered out if they don't match the the condition above 👆 That's the missing bit, and where "searchable in algolia" promises a bit too much 🙂

I'm setting up a staging server with the tag attached to the tag graph so you can check that everything looks fine in search, and then we can consider running another indexing job.

I'll keep this open until we triage it, to see if we can make this a bit less confusing.

spoonerf:
Ah thanks @mlbrgl! I didn't realise that was the missing step, good to know. I think just a note next to the 'Searchable in Algolia' box would be very helpful in improving clarity here.

mlbrgl:
@spoonerf sounds good! Here is the [staging server](http://staging-site-feat-add-noop-file/search?topics=Sexual+%26+Reproductive+Health) for review.

Please note that:
- I didn't update prod
- the top level category matters, as it's being used by this component:
[...]

PR #6053, opened 2026-02-03
✨ add a note to the searchable in algolia checkbox

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/6047

@spoonerf would this be sufficient for you?

image


spoonerf:
Yeh that looks good, thanks @ikesau!

PR #6060, opened 2026-02-04
✨ improve discrete bar charts with projections

sophiamersmann:
In discrete bar charts, if there's exactly one projected column with a historical counterpart, then these columns are stitched together and plotted as a single time series.

Examples:
- [prod](https://ourworldindata.org/grapher/life-expectancy-at-birth-including-the-un-projections?tab=discrete-bar&time=latest&facet=entity) / [staging](http://staging-site-discrete-bar-projections/grapher/life-expectancy-at-birth-including-the-un-projections?tab=discrete-bar&time=latest&facet=entity)
- [prod](https://ourworldindata.org/grapher/population-with-un-projections?tab=discrete-bar&time=latest&country=OWID_WRL~DEU~UN_EUR) / [staging](http://staging-site-discrete-bar-projections/grapher/population-with-un-projections?tab=discrete-bar&time=latest&country=OWID_WRL~DEU~UN_EUR)

In Search: [prod](https://ourworldindata.org/search) / [staging](http://staging-site-discrete-bar-projections/search) (see the small discrete bar chart in the first result)

### DIFF
--- a/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
+++ b/packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
@@ -1,9 +1,15 @@
 import { ChartManager } from "../chart/ChartManager"
 import { CoreColumn } from "@ourworldindata/core-table"
 import { ChartSeries } from "../chart/ChartInterface"
-import { Color, CoreValueType, Time } from "@ourworldindata/types"
+import {
+    Color,
+    CoreValueType,
+    ProjectionColumnInfo,
+    Time,
+} from "@ourworldindata/types"
 import { TextWrap } from "@ourworldindata/components"
 import { InteractionState } from "../interaction/InteractionState.js"
+import { ColumnSlug } from "@ourworldindata/utils"
 
 export interface DiscreteBarSeries extends ChartSeries {
     entityName: string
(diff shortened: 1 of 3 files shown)


packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartState.ts
```diff
@@ -100,8 +145,48 @@ export class DiscreteBarChartState implements ChartState, ColorScaleManager {
...
+        if (!projectionColumnInfoBySlug)
+            return { type: "single", slugs: ySlugs }
+
+        const projectionSlugs = ySlugs.filter(
+            (slug) => this.inputTable.get(slug).isProjection
+        )
+
+        // We only support combining projected and historical data
+        // if there is exactly one column pair to combine
+        if (ySlugs.length !== 2 || projectionSlugs.length !== 1)
```
sophiamersmann, line 172: We could also allow stitching multiple pairs of historical+projected data, but I can't really think of when this would be useful and it would make the code more complex, so I opted for stitching a single column pair only since that's the main use case

packages/@ourworldindata/grapher/src/barCharts/DiscreteBarChartConstants.ts
```diff
@@ -48,8 +56,20 @@ export interface DiscreteBarItem {
     time: number
     colorValue?: CoreValueType
     color?: Color
+    isProjection?: boolean
 }
 
+/**
+ * Describes how Y columns are handled.
+ *
+ * - "single": Each Y column is treated independently and plotted as its own series.
```
marcelgerber: "single" makes it sound like it is only ever one column, which is not the case.
Can we rename it to "independent" or "separate", maybe?

sophiamersmann: Good point!

PR #6063, opened 2026-02-04
fix: fix full-width grid sizing in Chrome 145+

### LINKED ISSUE #6036: (Chrome Beta only) Opening dropdowns on a wide screen causes content reflow
## Description

On a wide screen (the relevant breakpoint seems to be about 1330px, which is a bit odd), opening a dropdown (e.g. "Sort by" in the entity picker) causes the whole data page grid to widen a bit, causing all sorts of content reflows.

**This happens in Egde Beta (145.0.3800.22), Chrome Beta (145.0.7632.26), Chrome Dev (146.0.7655.0), but not on any non-beta browsers (I've tried Chrome 144.0.7559.110, Safari, Firefox).**

## Screenshots

https://github.com/user-attachments/assets/212be0e3-ef8e-4525-843d-05e71554380a

## Additional context

This is not a recent change on our end; I can repro it on [an archived page from October](https://archive.ourworldindata.org/20251022-222946/grapher/life-expectancy.html), for example.

marcelgerber:
Fixes #6036.

[In Chrome 145+](https://www.bram.us/2026/01/15/100vw-horizontal-overflow-no-more/), the value of `50vw` used in the column width definition would change depending on whether we have `html { scrollbar-gutter: stable }` present or not (which React Aria puts when the dropdown is open), which would then cause a content reflow.

Turns out, there is a more straightforward way to write the grid col definition in the first place!

There is a slight visual change now - but that's just because the inner grid cols are now _actually_ 1280px, not 1265 which they were before (because the scrollbar width got substracted).

### DIFF
--- a/site/css/grid.scss
+++ b/site/css/grid.scss
@@ -64,10 +64,18 @@ $grid-responsive: (lg, md, sm);
 // Use this when you want a full-page-width grid container
 // 1 (auto) | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 (auto)
 .grid-cols-12-full-width {
-    --outer-column-width: calc((50vw - 640px - var(--grid-gap)));
-    grid-template-columns: var(--outer-column-width) repeat(12, 1fr) var(
-            --outer-column-width
-        );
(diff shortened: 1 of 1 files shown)


ikesau:
Thanks Marcel! This works great and is far more intuitive!

PR #6064, opened 2026-02-04
🐛 fix entity selector dropdowns

### LINKED ISSUE #6020: Entity selector dropdowns are cut off (unusable) when they open to the top
## Description

In windows with small height - when the "Filter by" and "Sort by" dropdowns are relatively close to the bottom of the viewport - they open to the top, and are then cut off and pretty much unusable.

## Screenshots

https://github.com/user-attachments/assets/9968ce2d-41e9-4cd5-8c4d-75df709cc467

## Additional context

[Slack](https://owid.slack.com/archives/C46U9LXRR/p1769596352411899)

possibly we could fix it by using a different element that it's using for its portal

ikesau:
## Context

Resolves #6020 

Something broke in a recent version of `react-aria-components` causing any but the first Popover to incorrectly measure its available space inside the `.scrollable` container.

I tried to fix that in a couple of ways to no avail, and so instead removed the portal so that the popover renders on the document body directly. This required a followup change so that the drawer's `onDocumentClick` didn't trigger when the popover was clicked.

## Video
https://github.com/user-attachments/assets/175c46ef-2189-4935-9946-3ccd2ee4102d

## Testing guidance
Visit various grapher pages and articles and confirm that the entity selector works.

### DIFF
--- a/packages/@ourworldindata/grapher/src/slideInDrawer/SlideInDrawer.tsx
+++ b/packages/@ourworldindata/grapher/src/slideInDrawer/SlideInDrawer.tsx
@@ -30,14 +30,24 @@ export class SlideInDrawer extends React.Component<SlideInDrawerProps> {
 
     override componentDidMount(): void {
         document.addEventListener("keydown", this.onDocumentKeyDown)
-        document.addEventListener("mousedown", this.onDocumentClick, {
+        document.addEventListener("pointerdown", this.onDocumentPointerDown, {
+            capture: true,
+        })
+        document.addEventListener("click", this.onDocumentClick, {
             capture: true,
         })
     }
 
     override componentWillUnmount(): void {
         document.removeEventListener("keydown", this.onDocumentKeyDown)
-        document.removeEventListener("mousedown", this.onDocumentClick, {
+        document.removeEventListener(
+            "pointerdown",
+            this.onDocumentPointerDown,
+            {
+                capture: true,
+            }
+        )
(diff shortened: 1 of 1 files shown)


packages/@ourworldindata/grapher/src/slideInDrawer/SlideInDrawer.tsx
```diff
@@ -30,14 +30,14 @@ export class SlideInDrawer extends React.Component<SlideInDrawerProps> {
 
     override componentDidMount(): void {
         document.addEventListener("keydown", this.onDocumentKeyDown)
-        document.addEventListener("click", this.onDocumentClick, {
+        document.addEventListener("mousedown", this.onDocumentClick, {
```
ikesau: rac mounts the popover on `mousedown` which lead to a race condition with this `click` handler (slow clicks would fire after the `.underlay` div had mounted, confusing this handler into thinking the user had clicked outside of the drawer.

Checking on `mousedown` fixes this

marcelgerber: phew, yeah that's annoying. I also tried somehow solving this to use `click` instead (mostly to solve the other problem Codex brought up), but to no avail.

marcelgerber (approved):
good, this works well.

I'd like it to be a little bit less fragile, but also couldn't find a good way for that.

Codex's comment is also correct, but for that I'm also not sure what'd be a good way to solve it.

PR #6073, opened 2026-02-09
Remove unused AlertBanner

rakyi:
Looks like a dead code we didn't need for a while.


mlbrgl (approved):
makes sense, thanks! 👍

PR #6079, opened 2026-02-12
Devcontainer modernization

danyx23:
This PR updates our devcontainer setup to work nicely again. Running owid-grapher in a devcontainer is a nice way to be able to run agents without any permission checks but with guardrails in place so it can't do arbitrary things to your main system.

The devcontainer setup works like this:
- Make sure docker and the devcontainer VS Code extension are installed
- Open the project in a devcontainer
- A docker compose setup is used to bring up mysql and load the dev database on first start
- Run `make up.devcontainer` to bring up tmux with the admin etc
- Use your normal browser on your host system to connect to the admin

The docs have been updated and contain important information on how to change the `.env` file when you have already set that up for normal host machine access.

An optional user script has been set up so you can install/upgrade your coding agent of choice or do other customizations of the docker container on startup that we don't want to burn into the docker image.

The devcontainer setup ensures that agents only have access to the project directory, but several caveats apply:
- The current setup mounts agent dot directories (`~/.claude`, `~/.codex`, ...) into the container. This is conveniet as the authentication doesn't have to be redone, your user skills are available etc - but it is also a security loophole if you use agents also outside a container environment, because a prompt injection attack could install a malicious skill into your agent dot directory
- Network access is not restricted - this means that prompt injection attacks could lead to private resources being accessed via tailscale. In theory this shouldn't be a big concern because they would miss necessary authentication like ssh keys, but there might be a few non-critical systems that could still leak private information.
- Related to the above, MCPs or skills that you authenticate inside the docker container can do whatevery the have permission to. So if you connect, say, a notion MCP with wide-ranging permissions, the agent might be prompt injected to lead information from that system or make changes to that system.

Still, this seems like a useful upgrade to just running agents on your host machine without any additional layer of protection.

This is just a first version. In the future, we might want to:
- add support for using different git worktrees inside one container, and make forwarded ports configurable
- build the container once and upload it to the github container registry so you don't need to rebuild the container at first start or when we update it

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,19 +6,11 @@
     "workspaceFolder": "/workspace",
     "customizations": {
         "vscode": {
-            "extensions": [
-                "dbaeumer.vscode-eslint",
-                "esbenp.prettier-vscode"
-            ]
+            "extensions": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
         }
     },
     // Forward the ports used by the dev environment so they are accessible on the host.
-    "forwardPorts": [
-        3030,
-        8090,
-        8788,
-        3307
-    ],
+    "forwardPorts": [3030, 8090, 8788, 3307],
     "portsAttributes": {
         "3030": {
             "label": "Admin Server"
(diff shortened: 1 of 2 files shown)


mlbrgl:
Worth checking if we can solve #6095 here. Please report back in #6095 either way.

PR #6081, opened 2026-02-12
🐛 fix admin API returns

ikesau:
We occasionally get `Error: Cannot remove headers after they are sent to the client` in Sentry.

This PR fixes this genre of mistake once and for all by removing the `res.send()` method from the `Response` object that the `FunctionalRouter` passes, forcing devs to _return_ the response object from their API callbacks.


rakyi (approved):
I was hoping we could get rid of the `FunctionalRouter` because AFAICT it is an unnecessary layer of abstraction. When working with our Express app, one has to know both how Express and `FunctionalRouter` work.

This change makes it more useful by helping to prevent mistakes, but I'm not sure it's worth it.

Issue #6016, opened 2026-01-27
Search: Featured Metrics dominate results for simple text queries


mlbrgl:
### Problem

For simple, unstructured search queries (e.g. **“population”**), search results are often dominated by **Featured Metrics** that are only loosely related to the user’s intent. This makes it difficult to find the most obvious or canonical chart.

Example:

* Searching for **“population”** does not reliably surface the chart titled **“Population”** near the top.
* Instead, results are dominated by charts like *“Share of the population using the internet”* or *“…cases per million people”*.
* This happens because:

  * Featured Metrics are strongly promoted in search.
  * Broad synonyms (e.g. `people ↔ population`) cause many unrelated charts to match.
  * The intended chart may itself be a Featured Metric, but ranked lower than others.

### Preferred solution

**Only expose or promote Featured Metrics when the search state consists solely of topic and/or country filters, with no free-text query.**

As soon as the user enters any text query:

* Featured Metrics should no longer be promoted 
* Results should be ranked purely by textual and semantic relevance.

Rationale:

* Featured Metrics are explicitly designed and curated for **topic–country contexts**
* Free-text search represents a different mode of interaction (lookup rather than browsing).

   
* This creates a clear and consistent mental model across the product:

  * **Linear Topic Pages / topic–country browsing → Featured Metrics**
  * **Free-text search → Regular search results**

This preserves the importance of Featured Metrics where they are meaningful, while preventing them from dominating simple search queries.

see [slack](https://owid.slack.com/archives/C06UA61V85U/p1768906124503929)

---

#### 1. Demote Featured Metrics in all non-browse search scenarios

-> see #5956 

Featured Metrics would be generally downranked or removed from search results.

* **Pros**

  * Improves relevance for simple queries.
  * Reduces Featured Metric dominance.
* **Cons**

  * [Weakens continuity between Linear Topic Pages and search](https://github.com/owid/owid-grapher/pull/5956#discussion_r2728958785).
  * Makes the role of Featured Metrics feel arbitrary rather than context-driven.

#### 2. Promote Featured Metrics in search only when a topic tag is present

Featured Metrics remain promoted in search, but only if a topic filter is active.

* **Pros**

  * Preserves Featured Metrics in structured exploration.
* **Cons**

  * Still allows Featured Metrics to appear alongside free-text queries.
  * Blurs the boundary between topic–country browsing and lookup.

---

### Going further: boost exact chart title matches

[...]

PR #6082, opened 2026-02-12
✨ better featured metrics search behaviour

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/6016

Adds a new facetable property to our records in the `explorer-views-and-charts` index - `isFM`

When a query exists, sets `isFM:false`, otherwise default behaviour.

| Before | After |
|--------|--------|
| image | image | 

## Testing guidance

On staging, try setting a topic filter, and confirm the results prioritize featured metrics for that topic.

Then try searching for a word that one of the featured metrics contains, and see if the chart still shows up at the top still. (it might, depending on the chart and its views, but there should also be queries where other charts show up first which confirms that you're not seeing the FM version of the record)

### DIFF
--- a/adminSiteServer/tests/slug-validation.test.ts
+++ b/adminSiteServer/tests/slug-validation.test.ts
@@ -3,6 +3,7 @@ import { getAdminTestEnv } from "./testEnv.js"
 import {
     ChartConfigsTableName,
     ExplorersTableName,
+    MultiDimDataPagesTableName,
 } from "@ourworldindata/types"
 import {
     knexReadonlyTransaction,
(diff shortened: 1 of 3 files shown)


ikesau:
@mlbrgl presumably integration tests will fail until the prod index is updated with this new attribute and faceting config?

db/migration/1770917617246-SetSearchableInAlgoliaForFMTags.ts
```diff
@@ -0,0 +1,29 @@
...
+        // 160: Human Development Index (HDI)
+        // 206: Government Spending
+        // 293: Education Spending
+        // 294: Global Education
+        // 302: Literacy
+        // 1579: Medicine & Biotechnology
+        await queryRunner.query(`
+            UPDATE tags
+            SET searchableInAlgolia = TRUE
+            WHERE id IN (160, 206, 293, 294, 302, 1579)
```
mlbrgl: are you sure about this list? e.g. "literacy" is alredy searchable (same slug as literacy LTP)

site/search/queries.ts
```diff
@@ -98,9 +99,14 @@ export async function queryDataTopics(
         getFilterNamesOfType(state.filters, FilterType.COUNTRY),
         state.requireAllCountries
     )
+    const fmFacetFilter = formatFeaturedMetricFacetFilter(state.query)
```
mlbrgl: this is a bit misleading at the moment, and will never run because as soon as there is a query, the template changes and SearchDataTopicResults unmount. Either way, I would at least add a comment.

db/migration/1770917617246-SetSearchableInAlgoliaForFMTags.ts
ikesau: Ah damn, I forgot I had edited my local DB for other tag work. Will check on prod for the right ones to use.

Issue #6054, opened 2026-02-03
Add support for multi-dims in featured metrics


rakyi:
Featured metrics were developed before multi-dims had been launched, and so aren't supported. We now would like to support them.

It's probably just a matter of making [the API less strict](https://github.com/owid/owid-grapher/blob/54176aac326da3997b56fd82ea59fffb5a999834/adminSiteServer/apiRoutes/featuredMetrics.ts#L21) and then making sure that during indexing, the multi-dim URL is correctly matched to what exists in the `featured_metrics` table.

[Slack](https://owid.slack.com/archives/C46U9LXRR/p1770028794547299)

rakyi:
Done in https://github.com/owid/owid-grapher/pull/6083

PR #6083, opened 2026-02-12
🎉 multi-dims in featured metrics

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/6054

`validateChartSlug` is called on every FM before we create the row, but it wasn't checking the multi-dim table to see if the chart existed there.

Also adds a `?` before the query string (like we do for explorer records) which allows `findMatchingRecordByPathnameAndQueryParams` to work.

## Testing guidance

1. Check this code out locally
2. Add a multi-dim featured metric
3. Confirm that it can be added
4. Run `indexExplorerViewsMdimViewsAndChartsToAlgolia` on your staging index
5. Confirm that the multi-dim shows up as an FM

### DIFF
--- a/baker/algolia/utils/mdimViews.ts
+++ b/baker/algolia/utils/mdimViews.ts
@@ -193,11 +193,11 @@ async function getRecords(
         return {
             type: ChartRecordType.MultiDimView,
             objectID: `mdim-view-${id}`,
-            id: `mdim/${slug}${queryStr}`,
+            id: `mdim/${slug}${queryStr ? `?${queryStr}` : ""}`,
             chartId: -1,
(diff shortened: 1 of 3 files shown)


baker/algolia/utils/mdimViews.ts
```diff
@@ -140,7 +140,7 @@ async function getRecords(
             chartId: -1,
             chartConfigId: view.fullConfigId,
             slug,
-            queryParams: queryStr,
+            queryParams: queryStr ? `?${queryStr}` : "",
```
mlbrgl, line 200: cosmetic: maybe worth adding it to the id too for legibility

PR #6084, opened 2026-02-12
🐛 fix filtered profile headings showing in toc

ikesau:
## Context

When a profile is instantiated for an entity, `data-callout` blocks that lack data for that entity are cleared (their content is set to []). However, the sidebar TOC is generated _before_ this clearing happens, so headings inside cleared data-callout blocks still appear in the sidebar navigation even though they don't render.

This PR fixes that by moving `generateToc()` from `instantiateProfile()` to `instantiateProfileForEntity()`, after `loadAndClearLinkedCallouts()` returns.

| Before | After |
|--------|--------|
| image | image | 

## Testing guidance

You can use this [test gdoc](https://docs.google.com/document/d/1oSxIPftAu6-nsJqfrP0LkwH7fdGmeEMabMQswTETMfU/edit?tab=t.0)

Confirm that when on an entity that doesn't have data for the rubella vaccine callout, the Vaccination heading doesn't show up, but it does if on an entity with data.

### DIFF
--- a/db/model/Gdoc/GdocProfile.ts
+++ b/db/model/Gdoc/GdocProfile.ts
@@ -8,6 +8,7 @@ import {
     OwidGdocProfileScope,
 } from "@ourworldindata/types"
 import {
+    generateToc,
     getRegionByNameOrVariantName,
     removeTrailingParenthetical,
     instantiateProfile,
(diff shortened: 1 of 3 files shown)


sophiamersmann (approved):
Tested 👍🏻

Issue #5918, opened 2026-01-13
Cycle 2026.1 Country Profiles Enhancements


ikesau:
Enhancements to [Country Profiles](https://ourworldindata.org/profile/energy/canada)

## Motivation

Last cycle, we migrated these profiles from the old WP-based system we'd had for years. These pages are now powered by Google Docs, using a novel kind of templating syntax. We suspect that now that it's easier to write and publish these, we'll want to create a few more, and if we're going to do that, we should add a few more features to them first.

## Proposed scope

**Must do**

- [ ] Data callouts (plus hiding sections if the chart doesn't have data) (1w)
- [ ] Search integration (1w)
- [ ] Topic page integration (3d)

**Could do**

- [ ] Entity selection synchronization

**Won't do**

- [ ] Navigating between countries on the same profile

PR #6094, opened 2026-02-13
🎉 Country profiles in search

ikesau:
## Context

Part of https://github.com/owid/owid-grapher/issues/5918

[Figma](https://www.figma.com/design/50j5XbYEx9YuFcD7GZcd6F/Country-Profiles?node-id=1536-3845&p=f&t=TKbv0jzR9Um8nDs9-0)

image

image

## Testing guidance

- **Indexing**
  - Do profiles get indexed on (re)publish
    - Do pages that don't render (due to all invalid callouts) not get indexed
  - Do profiles get indexed by `indexPagesToAlgolia`
    - Do pages that don't render (due to all invalid callouts) not get indexed
  - Do profiles get the correct featured image indexed
- **Searching**
  - Do country profiles show up in the results
    - When a country filter is applied and R&W is shown
    - When a country filter is applied and query text exists
    - Not when _only_ a topic filter is applied
  - Do country profiles show up in the autocomplete
    - When a country is typed
    - When a topic is typed
  - Are clicks tracked in GA?

### Before merging

### DIFF
--- a/site/search/SearchWritingResults.tsx
+++ b/site/search/SearchWritingResults.tsx
@@ -218,6 +218,7 @@ export const SearchWritingResults = ({
     })
 
     const profileSlots = Math.min(profilesQuery.totalResults, 2)
+    const profilesRequestDone = !showProfiles || !profilesQuery.isLoading
 
     const articlesQuery = useInfiniteSearchOffset<
         SearchFlatArticleResponse,
@@ -229,14 +230,14 @@ export const SearchWritingResults = ({
         },
         firstPageSize: 4 - profileSlots,
         laterPageSize: 6,
+        enabled: profilesRequestDone,
     })
 
     const noArticles = articlesQuery.totalResults === 0
     const articleSlots = Math.min(articlesQuery.totalResults, 4 - profileSlots)
     const topicFirstPageSize = 6 - profileSlots - articleSlots
 
-    const dependenciesLoaded =
-        !articlesQuery.isLoading && (!showProfiles || !profilesQuery.isLoading)
+    const dependenciesLoaded = !articlesQuery.isLoading && profilesRequestDone
 
     const topicsQuery = useInfiniteSearchOffset<
         SearchTopicPageResponse,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -25,5 +25,8 @@
     "Prettier-SQL.keywordCase": "upper",
     "Prettier-SQL.SQLFlavourOverride": "mysql",
     "Prettier-SQL.expressionWidth": 80,
-    "oxc.formatting.enable": true
+    "oxc.formatting.enable": true,
+    "[typescriptreact]": {
+        "editor.defaultFormatter": "oxc.oxc-vscode"
+    }
 }
--- a/db/model/Gdoc/GdocBase.ts
+++ b/db/model/Gdoc/GdocBase.ts
@@ -817,6 +817,14 @@ export class GdocBase implements OwidGdocBaseInterface {
                 }
                 return links
             })
+            .with({ type: "country-profile-selector" }, (block) => [
+                createLinkFromUrl({
+                    url: block.url,
+                    sourceId: this.id,
+                    componentType: block.type,
+                    text: block.title ?? "Country profile selector",
+                }),
+            ])
             .with(
                 {
                     // no urls directly on any of these blocks
@@ -1133,10 +1141,12 @@ export class GdocBase implements OwidGdocBaseInterface {
                     }
 
                     // Validate profile links: must have ?entity=X with a valid, available entity
+                    // (skip for country-profile-selector, which links to the profile itself)
                     if (
                         linkedDoc?.type === OwidGdocType.Profile &&
                         doesGdocExist &&
-                        isGdocPublished
+                        isGdocPublished &&
+                        link.componentType !== "country-profile-selector"
                     ) {
                         const queryParams = Url.fromURL(
                             link.queryString
--- a/db/model/Gdoc/enrichedToIndexableText.ts
+++ b/db/model/Gdoc/enrichedToIndexableText.ts
@@ -425,7 +425,8 @@ export function enrichedBlockToIndexableText(
                         "sdg-grid",
                         "sdg-toc",
                         "socials",
-                        "subscribe-banner"
+                        "subscribe-banner",
+                        "country-profile-selector"
                     ),
                 },
                 (): undefined => undefined
--- a/db/model/Gdoc/enrichedToMarkdown.ts
+++ b/db/model/Gdoc/enrichedToMarkdown.ts
@@ -608,5 +608,6 @@ ${links}`
                 nestedOptions
             )
         })
+        .with({ type: "country-profile-selector" }, () => undefined)
         .exhaustive()
 }
--- a/db/model/Gdoc/enrichedToRaw.ts
+++ b/db/model/Gdoc/enrichedToRaw.ts
@@ -64,6 +64,7 @@ import {
     RawBlockStaticViz,
     RawBlockLTPToc,
     RawBlockConditionalSection,
+    RawBlockCountryProfileSelector,
 } from "@ourworldindata/types"
 import { spanToHtmlString } from "./gdocUtils.js"
 import { match, P } from "ts-pattern"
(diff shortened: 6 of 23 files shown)


site/search/Autocomplete.tsx
```diff
@@ -170,21 +172,71 @@ const FeaturedSearchesSource: AutocompleteSource<BaseItem> = {
...
-            query: state.query,
-            url: itemUrl,
-            position: String(state.activeItemId),
-        })
-        navigator.navigate({ itemUrl, item, state })
-    },
-    getItemUrl({ item }) {
-        const itemUrl = prependSubdirectoryToAlgoliaItemUrl(item)
-        return itemUrl
+const algoliaItemTemplate: AutocompleteSource<BaseItem>["templates"] = {
```
ikesau, line 177: Lifted this out of `AlgoliaSource` so that it can be reused by `createProfileSource`

baker/algolia/utils/pages.ts
```diff
@@ -33,30 +40,51 @@ import {
...
+} from "../../../db/model/Gdoc/dataCallouts.js"
 
 const computePageScore = (record: Omit<PageRecord, "score">): number => {
     const { importance, views_7d } = record
     return importance * 1000 + views_7d
 }
 
 const getThumbnailUrl = (
-    gdoc: OwidGdocPostInterface | OwidGdocDataInsightInterface,
+    gdoc: OwidGdocBaseInterface,
```
mlbrgl: we could keep a tighter type here to benefit maintain compile type checking

adminSiteServer/apiRoutes/gdocs.ts
mlbrgl, line 1: we should delete profile records in `deleteGdoc` too

mlbrgl: we might want to either add support for profiles in `getPreviewGdocIndexRecords` or update the message that says "not indexed" (maybe something the lines of "profiles are not supported in preview mode currently")

site/search/Autocomplete.tsx
```diff
@@ -426,10 +491,39 @@ export function Autocomplete({
...
-                        AlgoliaSource
+                        filtersSource,
+                        AlgoliaPagesSource,
+                        AlgoliaChartsSource
                     )
+
+                    const countryName =
+                        detectedCountry?.name ?? userCountryNameRef.current
+                    if (countryName) {
+                        const profileSource = createProfileSource(countryName)
```
mlbrgl: nit: for symmetry `AlgoliaProfileSource`

site/search/Autocomplete.tsx
```diff
@@ -170,21 +172,71 @@ const FeaturedSearchesSource: AutocompleteSource<BaseItem> = {
...
+        query: state.query,
+        url: itemUrl,
+        position: String(state.activeItemId),
+    })
+    navigator.navigate({ itemUrl, item, state })
+}
+
+const algoliaGetItemUrl: AutocompleteSource<BaseItem>["getItemUrl"] = ({
+    item,
+}) => prependSubdirectoryToAlgoliaItemUrl(item)
```
mlbrgl, line 246: this lead to doubling `/profile` in the item's path

site/search/Autocomplete.tsx
```diff
@@ -426,10 +491,39 @@ export function Autocomplete({
...
+                        query,
+                        listedRegionsNames(),
+                        allTopics,
+                        [],
+                        { threshold: 0.75, limit: 3 },
+                        synonymMap
+                    )
+                    const detectedCountry = detectedFilters.find(
+                        (f) => f.type === FilterType.COUNTRY
+                    )
```
mlbrgl: one of the reasons for not running `extractFiltersFromQuery` on keystroke was keeping the UI snappy on low powered devices. If we decide the trade-off is acceptable, then we probably should compute the filters once and use it for `filtersSource` too (with some adpatations around exact matches). This will also make the results easier to reason with.

Alternatively, just letting algolia match on the title attribute also works and we can reorder results by type using `reshape`.

site/search/SearchWritingResults.tsx
```diff
@@ -148,10 +181,26 @@ function MultiColumnResults({
...
+
+    const profilesQuery = useInfiniteSearchOffset<
+        SearchProfileResponse,
+        ProfileHit
+    >({
+        queryKey: (state) => searchQueryKeys.profiles(state),
+        queryFn: (liteSearchClient, state, offset, length) => {
+            return queryProfiles(liteSearchClient, state, offset, length)
+        },
+        firstPageSize: 4,
```
mlbrgl: this throws off the display of the topic/profile section, which should render - for instance, in the absence of articles - no more than 6 boxes (e.g. "china" + co2 now renders 8).

site/search/SearchWritingResults.tsx
```diff
@@ -4,62 +4,105 @@ import * as _ from "lodash-es"
...
 import { SearchResultHeader } from "./SearchResultHeader.js"
 import { useInfiniteSearchOffset } from "./searchHooks.js"
 import { SearchFlatArticleHit } from "./SearchFlatArticleHit.js"
+import { SearchProfileHit } from "./SearchProfileHit.js"
 import { SearchTopicPageHit } from "./SearchTopicPageHit.js"
 import { SearchWritingResultsSkeleton } from "./SearchWritingResultsSkeleton.js"
 import { SearchHorizontalDivider } from "./SearchHorizontalDivider.js"
 import { useSearchContext } from "./SearchContext.js"
 
+type PageHit = TopicPageHit | ProfileHit
```
mlbrgl: A "Page" concept that excludes "Articles" feels a bit foreign to me - any chance we could rename it to something more transparent?

mlbrgl:
nice, thanks for the integration! Nothing major but there are a couple of things we might to look at before merging.

site/search/Autocomplete.tsx
ikesau: Hmm, I can't find a way to do this cleanly without parsing the query.

Here's the behaviour we want:
- only show one country profile in the autosuggest
- If no country is typed, apply geolocated country as a facet
- If a country is typed, use that country as the facet

I can't find a way to get all 3, except by using [optionalFilters](https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters), which requires changing the pages ranking criteria to include `filters` and would presumably break all of the precisely arranged results we get on the pages index currently.

site/search/Autocomplete.tsx
```diff
@@ -383,13 +382,18 @@ const createFiltersSource = (
...
- * query is forwarded to Algolia so that matching words are highlighted.
+ * Creates a profile source that boosts the user's geolocated country
+ * using Algolia's `optionalFilters`. This avoids running expensive
+ * client-side country detection on every keystroke while still ensuring:
+ * - "energy" → "Energy in Canada" (boosted by geolocation)
+ * - "canada" → Canada profiles (matched by Algolia on title)
+ * - "energy france" → "Energy in France" (matched naturally)
+ *
+ * Requires the `filters` ranking criterion in the index settings
+ * (see configureAlgolia.ts).
```
ikesau, line 393: @mlbrgl here's the best version I can get without parsing the query. I've updated the staging index with the new ranking criteria so you can see it work, but presumably also notice the ways in which other aspects of search will break.

mlbrgl: that looks pretty good! As far as I understand it, `filters` in the ranking algo only impacts queries using [filter scoring](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/filter-scoring) or optional filters. Given that we don't use these features (except for this query), there won't be any impact on any other queries. Did you see anything suspicious?

ikesau: ah, I had missed that bit of text. Nice, then! I guess this'll work :)

site/search/SearchWritingResults.tsx
ikesau: True! Looking into this, I'm not sure it's working as intended?

I thought I could fix this issue by writing (pseudocode):

```

const profiles = profilesQuery(size: 2)

slotsLeft = 6 - Math.min(profiles.totalResults, 2)

const articles = articlesQuery(size: 2)

slotsLeft -= Math.min(articles.TotalResults, 2)

const topicPages = topicQuery(size: slotsLeft)

```

but topicPages is always fetching 6, so I think in the current (production) code, `noArticles` is probably not working correctly, either?

ikesau: image

mlbrgl: If no articles are returned (e.g. ["co2" + tag:china](https://ourworldindata.org/search?q=co2&countries=China&resultType=writing))
- requests 2 articles (gets none) and consequently fetches 6 topic pages.

If articles are returned (e.g. ["co2"](https://ourworldindata.org/search?q=co2&resultType=writing)), then only 2 topic pages are fetched initially, then 4.

So that works as expected as far as I can tell.

And the interleaving logic makes sure than the amount of topics vs articles is always balanced in the multi column layout (e.g. if there were 2 articles and 10 topics total for a query, then most topics would end up in the overflow).

ikesau:
@codex review

PR #6125, opened 2026-02-20
Don't focus search input on result type change

rakyi:
Fixes #5930

PR #6131, opened 2026-02-20
Preserve searchbar remount while preventing autofocus on mobile

### LINKED ISSUE #5930: Keyboard autofocuses unexpectedly when selecting "data" or "writing" filters on mobile
A bit of friction on search on mobile: when a user clicks on “data” or “writing”, the keyboard pops up unsolicited and blocks the content.

[Slack Message](https://owid.slack.com/archives/C06UA61V85U/p1768468782595599)

mlbrgl:
Fixes #5930

## Summary

The previous fix excluded `resultType` from the searchbar's React `key`, which prevented the mobile keyboard from popping up on tab switches. However, this also meant the searchbar wouldn't remount when the result type changed, so local input state (e.g. an uncommitted query) wouldn't reset.

This takes a different approach: keep the full state in the key so remounts work as expected, but control `autoFocus` explicitly — only autofocus the search input on the initial page load, not on re-mounts triggered by state changes.

- Adds an `autoFocus` prop threaded through `Search` → `Searchbar` → `SearchInput`
- Uses a ref in `Search` to track the first mount and pass `autoFocus={true}` only once
- Reverts the key back to the full `stateToSearchParams(state)`

## Test plan

- [x] On mobile (or with mobile emulation), go to `/search?q=climate`
- [x] Switch between All / Data / Writing tabs
- [x] Verify the keyboard does **not** pop up on tab switch
- [x] Type a partial query without submitting, then switch tabs — verify the uncommitted text resets

### DIFF
--- a/site/search/Search.tsx
+++ b/site/search/Search.tsx
@@ -3,11 +3,9 @@ import {
     FilterType,
     TemplateConfig,
     SearchResultType,
-    SearchState,
-    SearchUrlParam,
 } from "@ourworldindata/types"
 import { LiteClient } from "algoliasearch/lite"
-import { useMemo } from "react"
+import { useMemo, useRef } from "react"
 import { match } from "ts-pattern"
 import { useIsFetching } from "@tanstack/react-query"
 
@@ -35,15 +33,6 @@ import { buildSynonymMap } from "./synonymUtils.js"
 import { SiteAnalytics } from "../SiteAnalytics.js"
 import { PoweredBy } from "react-instantsearch"
 import { listedRegionsNames } from "@ourworldindata/utils"
-
-function makeSearchbarKey(state: SearchState): string {
-    const searchParams = stateToSearchParams(state)
-    // Don't re-render the searchbar when the result type changes to prevent
-    // autofocusing the input. It opened the keyboard on mobile.
-    // https://github.com/owid/owid-grapher/issues/5930
-    searchParams.delete(SearchUrlParam.RESULT_TYPE)
-    return searchParams.toString()
-}
 
 export const Search = ({
     topicTagGraph,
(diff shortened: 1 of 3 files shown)


mlbrgl:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6131?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6131** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6131?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6131?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6125** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

rakyi:
> However, this also meant the searchbar wouldn't remount when the result type changed, so local input state (e.g. an uncommitted query) wouldn't reset.

I think this is good. I as a user find it annoying, that filtering the results I already got erases my in-progress query. Switching between the result types doesn't trigger a new search, only what we show. I think we should remount the input only when we get new search results, e.g. clicking the back button or adding a country/tag, which do trigger a new search.

Do you have a different intuition?

rakyi:
Now when I play with it more I also don't like that adding a country from the autocomplete losses the input focus, but deleting it focuses the input. Those two should be IMO consistent.

rakyi:
We resolved this in a huddle.

rakyi:
Closing in favor of https://github.com/owid/owid-grapher/pull/6155

PR #6137, opened 2026-02-23
:test_tube: Experiment: all charts vs featured metrics block on modular topic pages

bnjmacdonald:
## Context

This PR makes 2 main changes:

1. Implements an experiment on a sample of modular topic pages that swaps the all charts block with the featured metrics block. 
- Also implements the experiment on a sample of data pages.
- Experimental conditions:

    - (a) status quo (all charts block). 70% of visitors
    - (b) treatment (featured metrics block). 30% of visitors

2. Removes deprecated code from experiment `data-page-insight-btns-2`.
- Note: does NOT remove topic-filtered view of /data-insights, which is currently undiscoverable and so can be dealt with in possible upcoming changes to topic filtering capabilities elsewhere on the site.

### Before merging

- [x] Update experiment `expires` date

### DIFF
--- a/site/gdocs/components/ArticleBlock.tsx
+++ b/site/gdocs/components/ArticleBlock.tsx
@@ -114,7 +114,6 @@ function ArticleBlockInternal({
             setExperimentState(s)
         }
     }, [])
-    const { isPageInExperiment, assignedExperiments } = experimentState
 
     if (block.parseErrors.filter(({ isWarning }) => !isWarning).length > 0) {
         return (
@@ -173,14 +172,13 @@ function ArticleBlockInternal({
                     <AllCharts
                         {...block}
                         id={
-                            isPageInExperiment &&
-                            assignedExperiments &&
-                            assignedExperiments[
+                            experimentState &&
+                            experimentState[
                                 `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
-                            ] &&
-                            assignedExperiments[
+                            ]?.isPageInExperiment &&
+                            experimentState[
                                 `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
-                            ] !== "all-charts"
+                            ]?.arm !== "all-charts"
                                 ? ""
                                 : "all-charts"
                         }
@@ -193,14 +191,13 @@ function ArticleBlockInternal({
                         <BlockQueryClientProvider>
                             <FeaturedMetrics
                                 id={
-                                    isPageInExperiment &&
-                                    assignedExperiments &&
-                                    assignedExperiments[
+                                    experimentState &&
+                                    experimentState[
                                         `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
-                                    ] &&
-                                    assignedExperiments[
+                                    ]?.isPageInExperiment &&
+                                    experimentState[
                                         `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
-                                    ] === "featured-metrics"
+                                    ]?.arm === "featured-metrics"
                                         ? "all-charts"
                                         : ""
                                 }
--- a/packages/@ourworldindata/utils/src/experiments/config.ts
+++ b/packages/@ourworldindata/utils/src/experiments/config.ts
@@ -20,7 +20,7 @@ export const experiments: Experiment[] = [
         id: "all-charts-vs-featured-v1",
         expires: "2026-03-24T00:00:00.000Z",
         arms: [
-            { 
+            {
                 id: "all-charts",
                 fraction: 0.7,
                 replaysSessionSampleRate: 0.2,
--- a/packages/@ourworldindata/utils/src/experiments/state.ts
+++ b/packages/@ourworldindata/utils/src/experiments/state.ts
@@ -1,17 +1,13 @@
 import Cookies from "js-cookie"
 import { EXPERIMENT_PREFIX } from "./constants.js"
-import { Experiment } from "./Experiment.js"
 import { experiments } from "./config.js"
 
-export type ExperimentState = {
-    assignedExperiments: Record<string, string>
-    isPageInExperiment: boolean
-}
+export type ExperimentState = Record<
+    string,
+    { arm: string; isPageInExperiment: boolean }
+>
 
-export const defaultExperimentState: ExperimentState = {
-    assignedExperiments: {},
-    isPageInExperiment: false,
-}
+export const defaultExperimentState: ExperimentState = {}
 
 /**
  * Gets the experiment state for the current page.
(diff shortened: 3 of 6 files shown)


baker/GrapherBaker.tsx
bnjmacdonald, line 1: all edits to this file just relate to removal of `exp-data-page-insight-btns-2` experiment

baker/GrapherBakingUtils.ts
bnjmacdonald, line 1: all edits to this file just relate to removal of `exp-data-page-insight-btns-2` experiment (`getTagsWithDataInsights` function is not being used anywhere)

packages/@ourworldindata/types/src/gdocTypes/Datapage.ts
bnjmacdonald, line 1: all edits to this file just relate to removal of `exp-data-page-insight-btns-2` experiment

site/DataPageContent.scss
bnjmacdonald, line 1: all edits to this file just relate to removal of `exp-data-page-insight-btns-2` experiment

functions/_common/experiments.ts
bnjmacdonald, line 1: edits to this file relate mostly to moving the `addClassNamesToBody` logic from `functions/grapher/[slug].ts` to here so that experiment class names are added to `<body>` for topic pages too.
- @rakyi : this shouldn't break anything for the `user-survey-role-v1` experiment, right?

also includes some general refactoring.

functions/_common/experiments.ts
```diff
@@ -64,12 +60,46 @@ export const experimentsMiddleware = async (
...
 
     const response = await context.next()
-    const headers = new Headers(response.headers)
+    let responseWithBodyClasses = response
+    const contentType = response.headers.get("content-type")
+    const isHtmlResponse = contentType?.includes("text/html") ?? false
+    if (
+        isHtmlResponse &&
+        experimentClassNames.length &&
+        response.status === 200
```
bnjmacdonald: @rakyi I'm actually a little unsure if we actually want to include `response.status === 200` here, as then cached 304 responses aren't getting the experiment class names added to the body. Which in theory means that a returning visitor to a page in the experiment (e.g. `/poverty`) can get "assigned" to the experiment but the class names don't get added to the body (so they just see the status quo), right? 

Not exactly sure how the caching works here for a user who visited `/poverty` before the experiment began and then visited again during the experiment.

rakyi:
@codex review

functions/_common/experiments.ts
rakyi: 304 responses have no body at all, since the point of caching is to avoid sending content the client already has. However, Cloudflare doesn't currently (by default) cache HTML responses so that shouldn't be an issue.

functions/_common/experiments.ts
```diff
@@ -64,12 +60,46 @@ export const experimentsMiddleware = async (
...
+
+    if (!cookiesToSet.length && !experimentClassNames.length) {
         return context.next()
     }
 
     const response = await context.next()
-    const headers = new Headers(response.headers)
+    let responseWithBodyClasses = response
+    const contentType = response.headers.get("content-type")
+    const isHtmlResponse = contentType?.includes("text/html") ?? false
```
rakyi: I think we should skip processing non-HTML responses completely, including setting the cookies and just return the response without additional processing.

functions/_common/experiments.ts
rakyi: Yeah, that should be fine.

site/gdocs/components/ArticleBlock.tsx
```diff
@@ -97,6 +102,19 @@ function ArticleBlockInternal({
...
+        defaultExperimentState
+    )
+    useEffect(() => {
+        if (typeof window !== "undefined") {
+            const s = getExperimentState()
+            setExperimentState(s)
+        }
+    }, [])
+    const { isPageInExperiment, assignedExperiments } = experimentState
+
```
rakyi, line 117: Since experiment state is not reactive and won't change, we should assign it to a variable at a top level. If we need it in multiple places, we should do it only once in an appropriate file and then import it where needed.

site/gdocs/components/ArticleBlock.tsx
```diff
@@ -125,12 +143,70 @@ function ArticleBlockInternal({
...
+                        }}
+                    />
+                )
+            }
+
+            return (
+                <>
+                    <AllCharts
+                        {...block}
+                        id={
```
rakyi, line 174: Why do we change the `id` here and below based on the experiment state? It should not be necessary, since the `id` should only be used for programmatic purposes, i.e. permalinks, and not for styling.

site/gdocs/components/ArticleBlock.tsx
```diff
@@ -125,12 +143,70 @@ function ArticleBlockInternal({
...
+                                id={
+                                    isPageInExperiment &&
+                                    assignedExperiments &&
+                                    assignedExperiments[
+                                        `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
+                                    ] &&
+                                    assignedExperiments[
+                                        `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
+                                    ] === "featured-metrics"
+                                        ? "all-charts"
```
rakyi, line 201: We should not use `all-charts` ID for featured metrics, that's confusing and not true.

site/FeaturedMetrics.tsx
```diff
@@ -46,29 +50,41 @@ export const FeaturedMetrics = ({
...
         0,
         MAX_MEDIUM_RESULTS + MAX_SMALL_RESULTS
     )
+    // for all-charts-vs-featured-v1 experiment, exclude a hit that
+    // links to the page we are already on.
+    if (isDataPage) {
+        hits = hits.filter(
+            (hit) =>
+                hit.type !== "chart" ||
+                `/grapher/${hit.slug}` !== window.location.pathname
```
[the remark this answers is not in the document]
rakyi: Seems like this might not be a problem in practice, but still good to do just in case.

site/gdocs/components/ArticleBlock.tsx
bnjmacdonald: it's so that we don't have 2 elements both with `id="all-charts"`, which breaks the sticky nav click behavior (browser won't scroll to the displayed element even though the other has `display: none`). 

An alternative solution would be to give each element their own id (e.g. "#all-charts" vs "#featured-metrics"), but that requires other new logic so that the sticky nav links change appropriately (so that "#featured-metrics" is linked in the sticky nav instead of "#all-charts" depending on the experiment arm). 

The current solution feels the least invasive (easiest to remove the code later) and we don't have to worry about breaking any existing links on other pages/websites to `{url}#all-charts`, so I went with that.

site/gdocs/components/ArticleBlock.tsx
bnjmacdonald: b/c that would require changes to the sticky nav links (so that `#featured-metrics` is linked appropriately in the sticky nav) and would require an assumption that no other page (on our site or elsewhere) is linking to `{url}#all-charts` (which would be broken by `#featured-metrics`). Might be an ok assumption, but it seems just as reasonable to stick with `#all-charts` across both experimental conditions.

In a user's eyes, `#all-charts` is presumably just as good a descriptor for either of these two blocks (i.e. both blocks list a bunch of charts and then link to "see all").

PR #6145, opened 2026-02-25
🔨 rename hideLegend to hideSeriesLabels

sophiamersmann:
The chart config's `hideLegend` is a misnomer and its overuse lead to a bug where hiding entity names in a line chart would also hide the map legend.

For line charts, slope charts and stacked area charts, `hideLegend` is renamed to `hideSeriesLabels`. It hides the line legend labels on the right. For all other chart types, `hideLegend` is dropped.

We had a few stacked bar charts and stacked discrete bar charts that used `hideLegend` to hide the color legend:

- [x] https://ourworldindata.org/grapher/countries-with-minimum-urban-threshold - changed from a StackedDiscreteBar chart to a DiscreteBar chart that automatically hides color legends with a single item
- [x] https://ourworldindata.org/grapher/maize-exports-ukraine-russia-perspective - nothing to do, redirects to an explorer
- [x] https://ourworldindata.org/grapher/wheat-exports-ukraine-russia-perspective - Pablo updated the chart
- [x] https://ourworldindata.org/grapher/annual-reported-ai-incidents-controversies - hideLegend setting makes no difference because single item legends are hidden by default
- [x] https://ourworldindata.org/grapher/passenger-miles-traveled-self-driving-taxis - hideLegend setting makes no different because single item legends are hidden by default

(In almost all cases `hideLegend` was used to hide a legend with a single item, which Grapher already does automatically)

ETL PR: https://github.com/owid/etl/pull/5719


sophiamersmann:
* **#6145** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6145?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6148](https://github.com/owid/owid-grapher/pull/6148) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6148?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6149](https://github.com/owid/owid-grapher/pull/6149) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6149?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -517,6 +520,9 @@ export class GrapherState
 
     variant = GrapherVariant.Default
 
+    /** Whether to hide a chart's legends */
+    hideLegend: boolean | undefined = false
```
marcelgerber, line 524: ~Is it intentional that we add this here? probably not, right?~

marcelgerber: Ah no, I (think?) I just realised it: `hideLegend` can still be used to hide the *map* legend, is that correct?

sophiamersmann: I just moved it down so that it's not in the 'persisted state' section anymore. `hideLegend` is still internal state that hides the map legend, for example, but also hides legends when faceted, etc

PR #6146, opened 2026-02-25
Migrate from prettier to oxfmt

rakyi:
oxfmt is a rust-based substitute for prettier, which is now very close to feature parity for our use case and runs ~5x as fast on the whole project on my machine.


package.json
```diff
@@ -255,12 +255,8 @@
         "preact@npm:^10.13.2": "patch:preact@npm%3A10.17.0#~/.yarn/patches/preact-npm-10.17.0-e831e619a5.patch",
         "preact@npm:^10.10.0": "patch:preact@npm%3A10.17.0#~/.yarn/patches/preact-npm-10.17.0-e831e619a5.patch"
     },
-    "prettier": {
-        "trailingComma": "es5",
-        "semi": false
-    },
     "lint-staged": {
-        "*.{tsx,ts,jsx,js,json,md,html,css,scss,yml}": "yarn prettier --write"
+        "*": "yarn oxfmt --no-error-on-unmatched-pattern"
```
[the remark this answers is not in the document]
rakyi: That's wrong, `no-error-on-unmatched-pattern` is there specifically to prevent such errors.

.oxfmtrc.json
```diff
@@ -0,0 +1,22 @@
+{
+    "$schema": "./node_modules/oxfmt/configuration_schema.json",
+    "trailingComma": "es5",
+    "semi": false,
+    "printWidth": 80,
+    "sortPackageJson": false,
```
marcelgerber: do we not want this? is it producing a different order than yarn's inbuilt sorting? otherwise, I would be in favor of this.

rakyi: I just wanted to minimize the changes and prettier didn't do that. I now enabled it, but it didn't change anything, so not sure what's going on.

PR #6159, opened 2026-02-27
enhance: make sources line specific to the active tab

sophiamersmann:
For multi-chart Graphers, the sources line should be specific to the current view. For example, if the scatter uses a population indicator as the size dimension, then the sources line on the line chart tab doesn’t need to include the sources for that population indicator.

This also turned into a bit of a refactor of how we're dealing with dimensions specified in the config that are not currently in use. This is now centralised in GrapherState (as opposed to doing this in the individual chart components). Now there are two layers of column slugs:
- `inputYColumnSlugs`, `input{X,Size,Color}ColumnSlug` - passed in by the author via the config
- `yColumnSlugs`, `{x,size,color}ColumnSlug` - columns currently used in the active chart tab
    - Here is where we apply exceptions, e.g. Marimekkos don't use the x-dimension if there's also a scatter tab

We should use the actual slugs in almost all cases, except for specific reasons, e.g. data downloads.

The SVG tester reports differences because some chart configs specified unused dimensions, and those sources ended up in the sources line (now they don't).

### DIFF
--- a/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
+++ b/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
@@ -113,6 +113,7 @@ export interface EntitySelectorManager {
     selection: SelectionArray
     entityType?: string
     entityTypePlural?: string
+    // TODO: Use the column slugs currently in use, if possible
     inputColumnSlugs?: ColumnSlug[]
     isEntitySelectorModalOrDrawerOpen?: boolean
     canChangeEntity?: boolean


sophiamersmann:
* **#6159** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/grapher/src/chart/ChartTabs.ts
```diff
@@ -177,3 +179,44 @@ export const isMapTab = (
...
+ */
+export function resolveDefaultTab(config: GrapherInterface): GrapherTabName {
+    return match(config.tab)
+        .returnType<GrapherTabName>()
+        .with(GRAPHER_TAB_CONFIG_OPTIONS.table, () => GRAPHER_TAB_NAMES.Table)
+        .with(GRAPHER_TAB_CONFIG_OPTIONS.map, () => GRAPHER_TAB_NAMES.WorldMap)
+        .with(GRAPHER_TAB_CONFIG_OPTIONS.chart, undefined, () => {
+            // Skips a validation step that GrapherState usually does
+            return config.chartTypes?.[0] ?? GRAPHER_TAB_NAMES.LineChart
+        })
```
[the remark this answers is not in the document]
sophiamersmann: Just dropped the helper

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -2536,11 +2535,28 @@ export class GrapherState
...
     }
 
-    @computed get categoricalColorColumnSlug(): string | undefined {
+    @computed private get categoricalColorColumnSlug(): string | undefined {
         if (!this.colorColumnSlug) return undefined
         return this.numericColorColumnSlug ? undefined : this.colorColumnSlug
     }
 
+    /** Color column slug used by the active tab */
+    @computed get activeColorColumnSlug(): string | undefined {
```
sophiamersmann: This is now centralised here (it used to be done in the individual chart components before)

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -113,7 +113,7 @@ export interface EntitySelectorManager {
     selection: SelectionArray
     entityType?: string
     entityTypePlural?: string
-    activeColumnSlugs?: string[]
+    inputColumnSlugs?: ColumnSlug[]
```
sophiamersmann, line 117: I think this should be the actual column slugs (not the input ones), but I'm not sure if switching to the actual ones could have any adverse effects, so I'm keeping it as is for now

sophiamersmann:
@codex review

packages/@ourworldindata/grapher/src/core/GrapherValuesJson.ts
```diff
@@ -260,28 +257,22 @@ export function prepareCalloutTable(
     inputTable: OwidTable,
     config: GrapherInterface
 ): PreparedCalloutTable {
-    const chartDimensions = (config.dimensions ?? []).map(
-        (d) => new ChartDimension(d, { table: inputTable })
-    )
+    const grapherState = new GrapherState(config)
```
sophiamersmann, line 260: If we don't attach the table, then this should be lightweight

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
ikesau: Maybe worth a comment? Seems like the sort of thing that could get lost

sophiamersmann: Good thinking! Will do :)

PR #6166, opened 2026-02-27
🐝 separate archieMlComponents.ts into many files

ikesau:
## Context

- Splits `ArchieMlComponents.ts` into individual files in a new `archieMLComponents/` directory
- `ArchieMlComponents.ts` becomes a barrel file that re-exports everything from the individual files and defines the union types (`OwidRawGdocBlock`, `OwidEnrichedGdocBlock`, `OwidEnrichedGdocBlockTypeMap`)
- Shared types like `ParseError`, `BlockSize`, `Ref`, etc. live in `archieMLComponents/generic.ts`
 
This is one of a few similar refactors I intend to do, to make it easier for LLMs to investigate a component without burning up tokens.

## Tradeoffs

- Adding a new component now requires adding an `export *` line to `ArchieMlComponents.ts` (in addition to adding it to the union types, which was already required).
- `ArchieMlComponents.ts` re-exports everything via `export *`, so there's a theoretical risk of name collisions with other exports in `index.ts`. TypeScript catches these at compile time, and there are currently no conflicts.
- There's a circular type dependency between `generic.ts` → `ArchieMlComponents.ts` → individual files → `generic.ts`, but it's type-only (import type) so it has no runtime impact

### DIFF
--- a/packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/CountryProfileSelector.ts
+++ b/packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/CountryProfileSelector.ts
@@ -1,4 +1,4 @@
-import { EnrichedBlockWithParseErrors } from "./generic.js";
+import { EnrichedBlockWithParseErrors } from "./generic.js"
 
 export type RawBlockCountryProfileSelector = {
     type: "country-profile-selector"


site/gdocs/components/topic-page.scss
marcelgerber, line 1: Not sure if you want this change in here?

ikesau: Oops, jujutsu mistake 👼

marcelgerber: 🥋

PR #6183, opened 2026-03-04
🐝 more oxfmt vscode settings

ikesau:
Adds a `defaultFormatter` rule for TS files, which VSCode seems to need for format on save to work as expected.

I also noticed our SQL formatter (only used for `db/migration/pre-migrations-schema.sql`) was no longer supported, so I updated it to a [newer one](https://marketplace.visualstudio.com/items?itemName=ReneSaarsoo.sql-formatter-vsc) and configured it to minimize the diff as best I could.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,8 @@
     "editor.tabSize": 4,
     "editor.detectIndentation": false,
     "editor.defaultFormatter": "oxc.oxc-vscode",
+
+    "oxc.fmt.configPath": ".oxfmtrc.json",
     "files.insertFinalNewline": false,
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -4,6 +4,8 @@
     "editor.tabSize": 4,
     "editor.detectIndentation": false,
     "editor.defaultFormatter": "oxc.oxc-vscode",
```
rakyi, line 6: So this line is not enough?

ikesau: Seems not. When I ran `> Format Document` it would prompt me to set this first, and didn't work without it.

PR #6185, opened 2026-03-04
Improve HTML semantics

rakyi:
Fixes issues I noticed in the markdown generated by Cloudflare across our major content types.

- Changes several components to be HTML lists where appropriate.
- Adds actual space between inline elements where CSS was used to add only visual space. This should fix missing whitespace between these elements in the markdown.
- Changes links that contain block elements to be a block themselves. This might fix missing spaces between the elements inside the link in the markdown.
- Render admin links only on the client - to exclude them from the markdown.

https://developers.cloudflare.com/fundamentals/reference/markdown-for-agents/

## Testing guidance

You can double-check that the changed components remain visually same/very similar as before.

### Before merging

### DIFF
--- a/site/NewsletterSubscription.tsx
+++ b/site/NewsletterSubscription.tsx
@@ -142,7 +142,7 @@ export const NewsletterSubscriptionForm = ({
                 <label htmlFor={idBiweekly}>
                     <span className="newsletter-subscription-form__label-title">
                         The OWID Brief
-                    </span>
+                    </span>{" "}
                     <span className="newsletter-subscription-form__label-frequency note-12-medium">
                         Twice a month
                     </span>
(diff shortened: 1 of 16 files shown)


ikesau (approved):
Visually the elements look good on Firefox, Chrome and Safari 👍 

I'm not sure how to test the markdown itself. My local wrangler and staging don't seem to handle the header.

Two other things I noticed while testing this out myself: 
1. `AdminLinks` component is shown (`[Gdoc](#)/[Admin](#)`)
2. Anchor links on headings confuse it

Is there anything we can do about those? All I can see is `cssSelector` in the docs, which I suppose we could figure out, but that's higher maintenance.

rakyi:
Yeah, unfortunately I think we have to test the markdown output only in prod.

I (hopefully) fixed the admin links by rendering them only on the client.

Anchor links on headings seem fine to me. Cursor even correctly scrolls to the Markdown heading when I click on the link, so maybe it works as intended? At least I think it's not too bad.

PR #6186, opened 2026-03-04
Move country detection from owid.io to ourworldindata.org

### LINKED ISSUE #6101: move country detection from owid.io to ourworldindata.org
Related to #6099 - during a review it was flagged that the detect country request looks suspicious. It would be better if this came under the ourworldindata domain (currently it is on owid.io) and the code be moved to the owid-grapher repo so that the code can be inspected

mlbrgl:
## Context

Closes #6101 — the detect-country request previously hit `detect-country.owid.io`, a separate Cloudflare Worker. This looked suspicious to reviewers and the code wasn't inspectable alongside the rest of the site.

This PR moves country detection into a CF Pages Function at `/api/detect-country`, using the existing `regions.data.ts` as the direct source of truth (no more separate `country_codes.json` that needs periodic syncing from the ETL).

**Deployment sequence:** This PR should be deployed **first**, then [owid/cloudflare-workers#25](https://github.com/owid/cloudflare-workers/pull/25) (delete old worker), then [owid/ops#405](https://github.com/owid/ops/pull/405) (update docs).

### Changes

1. **New endpoint** `functions/api/detect-country/index.ts` — reads `request.cf?.country` (alpha-2), looks it up in the regions data via `getParentRegions()`, returns `{ country: { code, name, short_code, slug, regions } | null }`
2. **Client URL** `packages/@ourworldindata/utils/src/Util.ts` — changed from `https://detect-country.owid.io` to `/api/detect-country` (relative URL works across all environments)
3. **Removed redirect** `baker/redirects.ts` — `/detect-country → detect-country.owid.io` no longer needed
4. **Updated workflow** `.github/workflows/update-regions.yml` — removed TODO checkbox about updating the old worker separately

### Data verification

The new implementation produces **identical output** to the old `country_codes.json` (251 entries, 0 differences). The only non-semantic difference is the ordering of the parent regions array. Verified with this comparison script:

<details>
<summary>compareJson.ts</summary>

```typescript
/**
 * Compare old and new country codes JSON files using remeda's isDeepEqual.
 *
 * Usage: yarn tsx --tsconfig tsconfig.tsx.json functions/api/detect-country/compareJson.ts
 */
import { isDeepEqual } from "remeda"
import fs from "fs"

interface CountryResponse {
    code: string
    name: string
    short_code: string
    slug: string
    regions: string[] | null
}

type CountryMap = Record<string, CountryResponse>

const normalize = (data: CountryMap): CountryMap =>
    Object.fromEntries(
        Object.entries(data).map(([k, v]) => [
            k,
            { ...v, regions: v.regions?.slice().sort() ?? null },
        ])
    )

const oldData: CountryMap = normalize(
    JSON.parse(fs.readFileSync("/tmp/old_country_codes.json", "utf-8"))
)
const newData: CountryMap = normalize(
    JSON.parse(fs.readFileSync("/tmp/new_country_codes.json", "utf-8"))
)

if (isDeepEqual(oldData, newData)) {
    console.log(
        `Identical (${Object.keys(oldData).length} entries, ignoring region array order)`
    )
} else {
    console.log("Differences found:")
    for (const key of new Set([
        ...Object.keys(oldData),
        ...Object.keys(newData),
    ])) {
        if (!isDeepEqual(oldData[key], newData[key])) {
            console.log(`  ${key}:`, oldData[key], "→", newData[key])
        }
    }
}
```

Output: `Identical (251 entries, ignoring region array order)`

</details>

## Testing guidance

- Deploy to a staging/preview environment and hit `/api/detect-country` — should return JSON with your country info
    - https://move-detect-country-to-graph.owid.pages.dev/api/detect-country
- Locally, `request.cf?.country` is simulated by Wrangler
- Verify that grapher components using `getUserCountryInformation()` still work (entity selector, map zoom, donate form, autocomplete) - although they won't be promoting the user's country until the function is deployed to production

### DIFF
--- a/functions/README.md
+++ b/functions/README.md
@@ -93,6 +93,58 @@ The analytics system requires these environment variables:
 
 # Our dynamic routes
 
+## `/api/detect-country`
+
+This route detects the visitor's country using Cloudflare's `cf.country` property (derived from the request's IP address) and returns structured country information.
+
+This was previously handled by a separate service at `detect-country.owid.io` and has been moved in-house to reduce external dependencies.
+
+### Response
+
+The response is a JSON object with a single `country` key, which is either `null` (if the country could not be detected or is not in our regions dataset) or an object with the following fields:
+
+| Field        | Type             | Description                            |
+| ------------ | ---------------- | -------------------------------------- |
+| `code`       | string           | OWID region code (e.g. `"DEU"`)        |
+| `name`       | string           | Country name (e.g. `"Germany"`)        |
+| `short_code` | string           | ISO 3166-1 alpha-2 code (e.g. `"DE"`)  |
+| `slug`       | string           | URL slug (e.g. `"germany"`)            |
+| `regions`    | string[] \| null | Parent region codes, or `null` if none |
+
+### Headers
+
+- `Access-Control-Allow-Origin: *` — accessible from any origin
+- `Cache-Control: public, s-maxage=0, max-age=7200` — cached in the browser for 2 hours, not cached on the CDN edge
+
+### Example
+
+```
+GET /api/detect-country
+
+200 OK
+{
+  "country": {
+    "code": "DEU",
+    "name": "Germany",
+    "short_code": "DE",
+    "slug": "germany",
+    "regions": [
+      "OWID_EU27",
+      "OWID_EUR",
+      "OWID_HIC",
+      "OWID_NH",
+      "PEW_EUR",
+      "UNM49_EUR",
+      "UNM49_WEU",
+      "UNSDG_ENA",
+      "UN_EUR",
+      "WB_ECA",
+      "WHO_EUR"
+    ]
+  }
+}
+```
+
 ## `/api/search`
 
 This route provides a search API for both charts and pages (articles, about pages).


mlbrgl:
* **#6186** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6186?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6177](https://github.com/owid/owid-grapher/pull/6177) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6177?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6195](https://github.com/owid/owid-grapher/pull/6195) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6195?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -532,7 +532,7 @@ export async function fetchWithTimeout(
 const _getUserCountryInformation = async (): Promise<
     UserCountryInformation | undefined
 > =>
-    await fetchWithRetry("https://detect-country.owid.io")
+    await fetchWithRetry("/api/detect-country")
```
mlbrgl: it would have been nice to have a relative URL here - and we still could for staging and prod, but supporting dev got tricky for call sites in the grapher side of the codebase which don't depend on settings. This could probably be injected as a dependency but that feels like a bigger refactor, that is not warranted if done only for this dev use case.

marcelgerber: I'm a bit torn on this.

- In the prod admin we don't necessarily need this (it's not adding much, and maybe having entity selection etc. be "unbiased" in the admin would actually be nice).
- On staging servers, it would just work (incl. in the admin), and it has the added benefit that any regions changes would be properly reflected on these staging servers already.
- In local dev, it would mostly not work. But maybe we can have a local-dev-only redirect going to the wrangler instance, or to https://ourworldindata.org/api/detect-country directly.
- The archive is another so-so thing, because it for sure wouldn't work if we have relative URLs. The open question there is whether it is actually worth it to have country detection there, or whether for things like the Wikipedia archive, we want as few outgoing requests as possible.
  - But then again, for that case it would be way better if we just didn't send that request at all, rather than sending a request that we expect to fail.


Overall, I think I'm inclined to keep it as you have it now, and use the absolute URL here.

mlbrgl:
### Merge activity

* **Mar 6, 3:49 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6186).
* **Mar 6, 3:49 PM UTC**: @mlbrgl merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6186).

PR #6191, opened 2026-03-05
🎉 Dynamic RSS

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/5467

- Adds dynamic RSS feed generation via a Cloudflare Pages function
- Adds a new Algolia index `pages-chronological`

The RSS function is called on requests to `/atom.xml`

You can pass `topics` and `type` query params.

If present, the function will query the new chronological index, faceting on these properties.

If not present, the function falls back to the baked version of the feed.

**weirdness:** `atom.xml` only shows articles and [linear] topic pages. `atom.xml?topics=Energy` will show those, plus DIs and announcements. So the filtered view is a superset.

This is because we already have subscribers on our default RSS feed and I wouldn't want to change their experience.

We could start versioning the endpoint or something, but doing it this way seems fine. (this PR doesn't add a way to discover this functionality, anyway, so..)

The new index only has ~1000 records in it (one per published gdoc, and it excludes profiles) so we've still got plenty of space left in the plan's budget. I think we might also not want to have this run every monday - it only tracks metadata which will almost never update, so once we do the initial index run, we can rely on incremental updates (added to the `gdocs.ts` API file) to keep it in sync.

## Testing guidance

1. Check out http://staging-site-dynamic-rss/atom.xml and confirm it's the same RSS feed as before
2. Check out http://staging-site-dynamic-rss/atom.xml?topics=Energy and confirm it's only energy posts
3. Check out http://staging-site-dynamic-rss/atom.xml?topics=Energy,Medicine%20%26%20Biotechnology and confirm it's both kinds of posts
4. Check out http://staging-site-dynamic-rss/atom.xml?type=article and confirm it's only articles
5. Check out http://staging-site-dynamic-rss/atom.xml?type=article&topics=Energy and confirm it's the conjunction

### DIFF
--- a/functions/atom.xml.ts
+++ b/functions/atom.xml.ts
@@ -1,5 +1,9 @@
 import * as Sentry from "@sentry/cloudflare"
-import { OwidGdocType, SearchIndexName } from "@ourworldindata/types"
+import {
+    OwidGdocType,
+    SearchIndexName,
+    CHRONOLOGICAL_INDEX_TYPES,
+} from "@ourworldindata/types"
 import { getCanonicalUrl } from "@ourworldindata/components"
 import { Env } from "./_common/env.js"
 import {
@@ -8,15 +12,6 @@ import {
     getIndexName,
 } from "./api/search/algoliaClient.js"
 import { formatTopicFacetFilters } from "./api/search/searchApi.js"
-
-const ALLOWED_FEED_PAGE_TYPES = new Set<string>([
-    OwidGdocType.Article,
-    OwidGdocType.TopicPage,
-    OwidGdocType.LinearTopicPage,
-    OwidGdocType.DataInsight,
-    OwidGdocType.AboutPage,
-    OwidGdocType.Announcement,
-])
 
 const HITS_PER_FEED = 50
 
@@ -88,9 +83,9 @@ async function queryChronologicalPages(
         throw new Error(`Algolia search failed: ${response.statusText}`)
     }
 
-    const data = (await response.json()) as {
+    const data: {
         results: [{ hits: FeedHit[] }]
-    }
+    } = await response.json()
 
     return data.results[0].hits
 }
@@ -185,7 +180,7 @@ function parsePageTypes(param: string | null): string[] | undefined {
     return param
         .split(",")
         .map((t) => t.trim())
-        .filter((t) => ALLOWED_FEED_PAGE_TYPES.has(t))
+        .filter((t) => CHRONOLOGICAL_INDEX_TYPES.has(t))
 }
 
 export const onRequestGet: PagesFunction<Env> = async (context) => {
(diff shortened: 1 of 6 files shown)


packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -2047,6 +2049,19 @@ export function checkIsHomepage(
     return gdoc.content.type === OwidGdocType.Homepage
 }
 
+/**
+ * Posts that should show up in the chronological algolia index
+ * for dynamic RSS feeds & /latest page
```
ikesau, line 2054: Assuming you'll use this type at some point, Matthieu :)

baker/algolia/utils/pagesChronological.ts
```diff
@@ -0,0 +1,160 @@
...
+        records.push({
+            objectID: gdoc.id,
+            type: gdoc.content.type,
+            slug: gdoc.slug,
+            title: gdoc.content.title || "",
+            excerpt: getExcerptFromGdoc(gdoc),
+            date: gdoc.publishedAt.toISOString(),
+            modifiedDate: (gdoc.updatedAt ?? gdoc.publishedAt).toISOString(),
+            authors: gdoc.content.authors ?? [],
+            tags: [...topicTags],
```
mlbrgl, line 154: `topicTags` might have been a set at some point, but spreading is now unnecessary

adminSiteServer/apiRoutes/gdocs.ts
```diff
@@ -362,6 +371,9 @@ async function indexAndBakeGdocIfNeccesary(
             if (isProfile) {
                 await indexIndividualProfile(nextGdoc as GdocProfile, trx)
             }
+            if (isChronologicalPost) {
+                await indexIndividualGdocInChronological(nextJson, trx)
```
mlbrgl, line 375: Codex found an edge case here, whereby if a page was previously eligible but changes between a chronological type to a non-chronological type (and vice-versa), the record stays in the original index until a full reindex. Given you hinted at running full reindexing less frequently, that might be worth addressing (although it is a rather unlikely scenario)

functions/atom.xml.ts
mlbrgl, line 1: on the [cloudlfare preview](https://dynamic-rss.owid.pages.dev/atom.xml?topics=Energy), the endpoint returns:
```
<?xml version="1.0" encoding="utf-8"?>
<feed xmlns="http://www.w3.org/2005/Atom">
<title>Error</title>
<subtitle>An error occurred while generating the feed</subtitle>
</feed>
````

in the [function logs](https://dash.cloudflare.com/078fcdfed9955087315dd86792e71a7e/pages/view/owid/88574763-48a1-4cb4-a846-d179cfa808a7/functions):

```
"logs": [
    {
      "message": [
        "Atom feed error:",
        "Error: Algolia search failed: Not Found"
      ],
      "level": "error",
      "timestamp": 1773831554721
    },
```

File: functions/atom.xml.ts [hunk 4]
```diff
@@ -0,0 +1,250 @@
...
+<link type="application/atom+xml" rel="self" href="${escapeXml(feedUrl)}"/>
+<updated>${updated}</updated>
+${entries}
+</feed>`
+}
+
+function parseTopics(param: string | null): string[] {
+    if (!param) return []
+    return param
+        .split(",")
```
mlbrgl, line 173: could we use the "~" convention for topics and types? (similar to the handling of multiple countries in grapher/explorer and search query params)

adminSiteServer/apiRoutes/gdocs.ts
ikesau: I don't think we'll be moving a data insight into an about page any time soon 🤷

PR #6192, opened 2026-03-05
🐛 fix entity selector sort column after switching mdims

sophiamersmann:
Resolves https://github.com/owid/owid-grapher/issues/5242

Example: http://staging-site-fix-mdim-entity-selector-sor/grapher/religious-composition?tab=discrete-bar&time=latest&religion=jews&indicator=share (switch between the religions)

### DIFF
--- a/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
+++ b/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
@@ -4,6 +4,7 @@ import { observer } from "mobx-react"
 import {
     computed,
     action,
+    comparer,
     reaction,
     when,
     IReactionDisposer,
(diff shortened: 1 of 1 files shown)


packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -247,8 +248,12 @@ export class EntitySelector extends React.Component<EntitySelectorProps> {
         // we need to change the sort config accordingly
         this.disposers.push(
             reaction(
-                () => this.sortOptions,
-                () => this.updateSortConfigIfOptionHasBecomeUnavailable()
+                () => ({
+                    slugs: this.sortOptions.map((option) => option.slug),
+                    isReady: this.manager.isReady,
```
sophiamersmann, line 253: Previously, if `isReady` was false, `updateSortConfigIfOptionHasBecomeUnavailable` was called, but it was a no-op and then wasn’t called again.

An alternative would be to make `sortOptions` depend on `isReady`, for example by returning no options when it isn’t ready, but then the dropdown briefly disappears when switching views, which feels a bit jumpy.

PR #6199, opened 2026-03-06
🔨 Remove country indexes

ikesau:
## Context

Removes https://ourworldindata.org/countries (redirects to `/search`)

Removes indices like https://ourworldindata.org/country/spain (redirects to `/search?countries=Spain`)

The country redirects are done with a function because the query params are case sensitive, so a regular dynamic redirect wouldn't work.

### Before merging

If DB migrations exists:

- [x] If columns have been added/deleted, all necessary views were recreated and ETL and Analytics team members have been informed of the incoming changes

### After merging

### DIFF
--- a/adminSiteServer/apiRoutes/bulkUpdates.ts
+++ b/adminSiteServer/apiRoutes/bulkUpdates.ts
@@ -129,7 +129,6 @@ export async function updateBulkChartConfigs(
             user: res.locals.user,
             newConfig,
             existingConfig: oldValuesConfigMap.get(id),
-            referencedVariablesMightChange: false,
         })
     }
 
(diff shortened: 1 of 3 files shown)


Marigold:
Here's the ETL part. We should merge it after this one https://github.com/owid/etl/pull/5755.

functions/country/[slug].ts
Comment (marcelgerber) [inline functions/country/[slug].ts:1]: Just putting this out here, we could also just set up 200-ish redirects for each country?
Not sure how that would affect our redirect limits (of 2,000 static redirects). Potentially, we could also use [Bulk redirects](https://developers.cloudflare.com/rules/url-forwarding/bulk-redirects/) for this case.

Comment (ikesau) [reply functions/country/[slug].ts:1]: Ah true! I hadn't wanted to use up 200 of our budget on something so trivial, but I'd forgotten about bulk redirects. Created a new list (so we're now at 4/5)

PR #6204, opened 2026-03-06
🎉 Image text extraction

ikesau:
## Context

- Adds a new column to our images table `extractedText`
- Adds functionality to extract text from every uploaded image
- Adds a script to backfill prexisting images

This will be useful for images in Algolia, and could also potentially be used to create better sources for our images (by linking to the `origins` table)

### DIFF
--- a/adminSiteServer/apiRoutes/images.ts
+++ b/adminSiteServer/apiRoutes/images.ts
@@ -15,6 +15,7 @@ import * as lodash from "lodash-es"
 
 import { Request } from "../authentication.js"
 import { HandlerResponse } from "../FunctionalRouter.js"
+import { logErrorAndMaybeCaptureInSentry } from "../../serverUtils/errorLog.js"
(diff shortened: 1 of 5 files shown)


adminSiteServer/apiRoutes/images.ts
```diff
@@ -294,3 +302,23 @@ export async function getImageUsageHandler(
...
+            await fetchGptGeneratedTextFromImage(imageUrl)
+        if (extractedText !== null) {
+            await db.knexReadWriteTransaction(async (trx) => {
+                await trx("images")
+                    .where({ id: imageId })
+                    .update({ extractedText })
+            })
+        }
+    } catch (error) {
+        console.error(`Failed to extract text for image ${imageId}:`, error)
```
rakyi: We probably want `logErrorAndMaybeCaptureInSentry` here.

Issue #6111, opened 2026-02-18
User feedback on mobile


rakyi:
## Core problem

The user currently doesn't have a way to submit general feedback on mobile, we don't show the floating "Feedback" button there.

## Proposed solution

Add a new item to mobile site nav (below "About") and to the footer (below "About us") that should open a full-screen feedback modal (same component as on the desktop) if easy to make work or link to https://ourworldindata.org/feedback (there we should capture the previous page via `Document.referrer` (not entirely reliable) or a query param).

Instead/in addition to adding the link to mobile site nav, we could add it to the top bar:

<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/93252a43-239a-4517-8207-917517041be9" />

## Context

[Slack](https://owid.slack.com/archives/C5BDCB2R3/p1771342188689849)

PR #6231, opened 2026-03-13
🎉 mobile feedback form

ikesau:
## Context

Resolves https://github.com/owid/owid-grapher/issues/6111

Not a beautiful redesign, but functional. Marwa signs off.

https://github.com/user-attachments/assets/93e7060c-8e1f-4fe9-8e2d-c18d11b59d5b

## Testing guidance

- Open the menu with a narrow viewport with and confirm the feedback button is there, and opening it works as expected.
- Increasing the viewport width repositions the modal.
- You can't scroll the body content when the modal is open.

## Checklist

### DIFF
--- a/site/SiteMobileMenu.scss
+++ b/site/SiteMobileMenu.scss
@@ -11,7 +11,8 @@
     }
 
     > ul {
-        > li {
+        // Don't add a border to the last two items (Donate and Send Feedback)
+        > li:not(:nth-last-child(2)) {
             border-bottom: rgba(152, 169, 189, 0.3) 1px solid;
         }
     }
@@ -28,6 +29,7 @@
         display: flex;
         align-items: center;
         justify-content: space-between;
+        cursor: pointer;
 
         &.active {
             color: $blue-90;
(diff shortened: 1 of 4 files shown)


site/SiteMobileMenu.scss
```diff
@@ -70,4 +72,70 @@
         color: $white;
         background-color: $vermillion;
     }
+
+    .donate {
+        padding: 9.5px 24px;
```
mlbrgl: this overrides the previous padding. Should these be merged?

site/SiteMobileMenu.scss
```diff
@@ -70,4 +72,70 @@
...
+}
+.SiteMobileFeedbackModal__overlay {
+    position: absolute;
+    inset: 0;
+    background-color: rgba(0, 0, 0, 0.5);
+}
+
+.SiteMobileFeedbackModal__box {
+    position: relative;
+    width: 100vw;
```
mlbrgl: this is always overriden by the mixin

site/SiteMobileMenu.scss
```diff
@@ -70,4 +72,70 @@
...
+    border-radius: 0;
+
+    @include sm-up {
+        top: 0;
+        left: 0;
+        position: absolute;
+        width: 100vw;
+        height: 100vh;
+        border-radius: 3px;
+    }
```
mlbrgl: there is no (visual) difference across viewports: the modal covers 100% in both directions, which differs from the demo video.

Regardless there are some avoidable repetitions of width/height declarations.

ikesau:
Also addresses https://github.com/owid/owid-grapher/issues/5013

Issue #4804, opened 2025-04-23
DI index page shows outdated data


sophiamersmann:
## Description

The DI index page sometimes shows outdated data. That's especially confusing for fields like the Figma URL because then chart image can't be updated.

## Expected behaviour

The DI index page shouldn't show outdated info.

## Steps to reproduce

- Go to the DI index page
- Go to some Gdoc and make a change (update the Figma URL, for example)
- Go back to the DI index page
- Note that the Figma URL is stale
- Visit the admin preview of the DI you updated
- Go back to the DI index page
- Note that the Figma URL is now updated

## Additional info

Refreshing the DI preview page is a workaround.

danyx23:
What would be easy to do is to add a refresh when the "Update image" is clicked which is probably the point where this is the most annoying.

PR #6232, opened 2026-03-13
🎉 (admin) add refresh DIs button

sophiamersmann:
Resolves https://github.com/owid/owid-grapher/issues/4804

There’s currently a limitation where new content from gdocs is only fetched when the admin preview is visited, so DIs shown on the DI index page in the admin can go stale. 

Authors sometimes update the `figma-url` field in the gdoc and expect the ‘Figma URL’ and ‘Update image’ buttons to appear on the DI index page. One workaround is to visit the DI preview page in the admin, but this keeps coming up, so I thought it was time to fix it.

I went with a ‘Refresh’ button, instead of refreshing automatically in the background on page load, because it’s much simpler and means I don’t have to worry about things like debouncing while typing in the search input, etc

### DIFF
--- a/adminSiteClient/DataInsightIndexPage.tsx
+++ b/adminSiteClient/DataInsightIndexPage.tsx
@@ -82,6 +82,7 @@ type PublicationFilter = "published" | "scheduled" | "draft"
 type Layout = "list" | "gallery"
 
 const DEFAULT_LAYOUT: Layout = "list"
+const PAGE_SIZE = 10
 
 const editIcon = <FontAwesomeIcon icon={faPen} size="sm" />
 const linkIcon = <FontAwesomeIcon icon={faUpRightFromSquare} size="sm" />
@@ -307,11 +308,15 @@ export function DataInsightIndexPage() {
         PublicationFilter | undefined
     >()
     const [layout, setLayout] = useState<Layout>(DEFAULT_LAYOUT)
+    const [currentPage, setCurrentPage] = useState(1)
 
     const [dataInsightForImageUpload, setDataInsightForImageUpload] =
         useState<DataInsightIndexItemThatCanBeUploaded>()
 
     const [isCreateModalOpen, setIsCreateModalOpen] = useState(false)
+
+    // True while new content is being refreshed from Google Docs
+    const [isRefreshing, setIsRefreshing] = useState(false)
 
     const [notificationApi, notificationContextHolder] =
         notification.useNotification()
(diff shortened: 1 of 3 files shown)


adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -383,6 +388,56 @@ export function DataInsightIndexPage() {
...
+        () =>
+            filteredDataInsights.slice(
+                (currentPage - 1) * PAGE_SIZE,
+                currentPage * PAGE_SIZE
+            ),
+        [filteredDataInsights, currentPage]
+    )
+
+    const refreshDataFromGdocs = useCallback(async () => {
+        const gdocIds = visibleDataInsights.map((di) => di.id)
```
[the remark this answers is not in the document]
sophiamersmann: Fixed by hiding the Refresh button in Gallery mode

adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -383,6 +388,56 @@ export function DataInsightIndexPage() {
...
+        setIsRefreshing(true)
+        try {
+            const result = await admin.requestJSON(
+                "/api/dataInsights/refresh",
+                { gdocIds },
+                "POST"
+            )
+
+            // Fetch the latest data if any items were updated
+            if (result.updated > 0) await refreshDataInsights()
```
sophiamersmann, line 424: Fetches the complete set, not just the updated DIs, meaning new DI documents may be added. If you're not on the first page, this can be a bit confusing since the DIs are shifted, but I don't expect the Refresh button to be used often when not on the first page.

adminSiteServer/apiRoutes/dataInsights.ts
```diff
@@ -273,3 +279,62 @@ export function getChartTypeFromConfigAndQueryParams(
...
+    )
+
+    let updated = 0
+    const errors: Array<{ id: string; message: string }> = []
+
+    // Fetch fresh content from Google Docs for each data insight
+    await Promise.all(
+        records.map(async (record) => {
+            try {
+                const gdoc = await loadGdocFromGdocBase(
```
ikesau, line 317: Should we maybe only do this on unpublished gdocs? Otherwise "refresh" is actually "republish"

sophiamersmann: I didn't realise this would republish DIs. Should def only do this for unpublished ones then

PR #6245, opened 2026-03-17
Update to Vite 8

marcelgerber:
Found a way to make it work!

Combining React and the SWC plugin for decorator support wasn't exactly easy, but I found a way, after all.

Things to test:
- [x] Click around the dev site (site + admin)
- [x] Click around the deployed site (site + admin)
- [x] Browser testing on BrowserStack

Before merging:
- [ ] SWC has now [added support for 2023-11 decorators](https://github.com/swc-project/swc/pull/11686). If that's released by the day of the merge, update to that.

### DIFF
--- a/vite.config-common.mts
+++ b/vite.config-common.mts
@@ -86,8 +86,7 @@ export const defineViteConfigForEntrypoint = (entrypoint: ViteEntryPoint) => {
                                 decorators: true,
                             },
                             transform: {
-                                // NOTE: SWC doesn't support the latest '2023-11' version yet.
-                                decoratorVersion: "2022-03",
+                                decoratorVersion: "2023-11" as any, // @swc/types@0.1.25 doesn't have the now-supported decorator version 2023-11 yet, but swc supports it
                                 useDefineForClassFields: true,
                             },
 


vite.config-common.mts
```diff
@@ -72,19 +76,41 @@ export const defineViteConfigForEntrypoint = (entrypoint: ViteEntryPoint) => {
...
+            withFilter(
+                optimizeReactAriaLocales.vite({
                     locales: ["en-US"],
                 }),
-                enforce: "pre",
-            },
+                {
+                    transform: {
+                        id: /[/\\](@react-stately|@react-aria|@react-spectrum|react-aria-components)[/\\]/,
+                    },
```
rakyi, line 112: Could you add a comment explaining why we do this? The plugin docs don't mention this.

marcelgerber:
> It's also interesting that BundleMon reports -36 KB reduction in the bundle size here in the CI results, which is a little suspicious.

Yep, I was also surprised about that! Part of that (I think ~6kB) is due to now excluding `dotenv`, which also excludes the fillers for `fs`, `path`, etc.

But the rest of it really is just the switch to Vite 8 - I assume that it has better tree shaking, and is more aggressive in removing unused methods/properties.

The release post also states this:
> And by integrating deeply across layers, we can pursue optimizations that were previously out of reach, such as leveraging Oxc's semantic analysis for better tree-shaking in Rolldown.

It's also not that the case that we now emit some syntax that we previously didn't - I've tested on BrowserStack in all our supported browsers, and all of them work just fine still.



> If/when they make the decorators work in rolldown, we'd only need to remove the SWC plugin?

Yes, indeed - I've added a comment for that now!

Issue #6169, opened 2026-03-02
Cycle 2026.2: Dumbbell plots


sophiamersmann:
_Add dummbell plots as a new chart type_

## Proposed scope

Two modes:
- Compare two indicators (two indicators, single time point)
- Compare two time points (single indicator, time range)

**Must do**

- [ ] Option: Value labels (absolute, (percentage) increase/decrease, off)
- [ ] Option: End points (dots connected by line, dots connected by arrow, only arrow)
- [ ] Legend: within chart area (preferred) or horizontal colour legend
- [ ] Tooltips
- [ ] Dumbbell plots in search
- [ ] Dumbbell plots as secondary tab
- [ ] Support for other features:
    - [ ] Tolerance
    - [ ] Selection mode (change-country disabled)
    - [ ] Log/linear axis
    - [ ] Row sorting order
    - [ ] Comparison lines (only vertical)
    - [ ] Missing data strategy

**Could do**

- [ ] Option: Custom indicator colours

**Won't do**

- [ ] Dot plots (single indicator, single time point)
- [ ] Relative mode
- [ ] Faceting
- [ ] Colour customisation

PR #6254, opened 2026-03-18
🎉 dumbbell chart state and layout

sophiamersmann:
[Cycle issue](https://github.com/owid/owid-grapher/issues/6169) / [Designs](https://www.figma.com/design/dCB69o19cAVtuk283STAcY/Dumbbell-Plots?node-id=1-31&m=dev)

Simple version of dumbbell plots in Grapher. Not feature complete yet (see below for follow-up work).

The focus of this PR is on chart state (transform functions) and layout (x and y axes).

### Description

Dumbbell plots have two modes:

* Compare values from the same column across two time points (single indicator, time range)
* Compare values from two different columns at a single time point (two indicators, single time)

This PR constructs the sized and placed series for both strategies and plots the x- and y-axis, as well as a first version of dumbbells with value labels on each side.

Making space for the value labels is a bit intricate:
* If the user hasn’t specified a fixed min/max domain value, I think extending the domain a bit so the value label ends up inside the chart area looks nicest
* If the user has specified a fixed min/max domain value, we respect that and simply add a bit of left or right padding so the value label doesn’t overflow

### To do

- [x] Not enough space for entity labels: http://localhost:3030/admin/charts/2838/edit (discrete bar chart looks good)
- [x] Drop support for log axis

### Follow-up work

* Dumbbell elements (only a basic version in this PR)
* In-chart legend, aligned with the top-most dots
* Color (hard-coded for now)
* Interaction (focus, hover, tooltips)
* More entity sorting options
* New dumbbell-specific options: value labels, end points
* Dumbbell plots in search
* Dumbbell chart thumbnails

Open questions:
* Dumbbell plots support log scales, but we don’t currently indicate this with a ‘log axis’ notice, as we do for other chart types. Should we add this?
* Not sure if adding `transformTableForSelection` is worth it. I looked at the slope chart’s transforms, and some of that would make sense for dumbbell plots too, but it’s also quite expensive (calculating tolerance on an unfiltered table). I think I’d rather wait until this becomes an actual problem.

### DIFF
--- a/packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartHelpers.ts
+++ b/packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartHelpers.ts
@@ -22,14 +22,12 @@ export function calculateAxisLayout({
     series,
     domain,
     width,
-    isLog,
     minFixed,
     maxFixed,
 }: {
     series: SizedDumbbellSeries[]
     domain: [number, number]
     width: number
-    isLog: boolean
     minFixed: boolean
     maxFixed: boolean
 }): AxisLayout {
@@ -47,11 +45,8 @@ export function calculateAxisLayout({
         },
     }))
 
-    const toScale = (v: number): number => (isLog ? Math.log(v) : v)
-    const fromScale = (v: number): number => (isLog ? Math.exp(v) : v)
-
-    let domainStart = toScale(domain[0])
-    let domainEnd = toScale(domain[1])
+    let domainStart = domain[0]
+    let domainEnd = domain[1]
     let padLeft = 0
     let padRight = 0
 
@@ -67,7 +62,7 @@ export function calculateAxisLayout({
         const pixelsPerUnit = effectiveWidth / domainSpan
         const domainStartPixel = padLeft - domainStart * pixelsPerUnit
         const toPixel = (value: number): number =>
-            domainStartPixel + toScale(value) * pixelsPerUnit
+            domainStartPixel + value * pixelsPerUnit
 
         // Find the worst-case label overflow on each side
         let leftOverflow = 0
@@ -101,7 +96,7 @@ export function calculateAxisLayout({
     }
 
     return {
-        domain: [fromScale(domainStart), fromScale(domainEnd)],
+        domain: [domainStart, domainEnd],
         pad: { left: padLeft, right: padRight },
     }
 }
--- a/packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartHelpers.ts
+++ b/packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartHelpers.ts
@@ -1,0 +1,107 @@
+import { PadObject, Position, RequiredBy } from "@ourworldindata/utils"
+import { SizedDumbbellSeries } from "./DumbbellChartConstants.js"
+
+export interface AxisLayout {
+    domain: [number, number]
+    pad: RequiredBy<PadObject, Position.left | Position.right>
+}
+
+/**
+ * Ensures every row's value labels fit inside the chart area by adjusting
+ * the axis domain and/or adding padding.
+ *
+ * On each side, one of two strategies is used depending on whether the author
+ * has specified a fixed domain edge:
+ * - Unfixed: widen the domain so dots shift inward, making room for labels
+ * - Fixed: preserve the user's domain edge; add padding instead
+ *
+ * Because the two sides interact (widening one shifts all dots), we iterate
+ * until overflow is below a tolerance.
+ */
+export function calculateAxisLayout({
+    series,
+    domain,
+    width,
+    isLog,
+    minFixed,
+    maxFixed,
+}: {
+    series: SizedDumbbellSeries[]
+    domain: [number, number]
+    width: number
+    isLog: boolean
+    minFixed: boolean
+    maxFixed: boolean
+}): AxisLayout {
+    if (series.length === 0 || width <= 0)
+        return { domain, pad: { left: 0, right: 0 } }
+
+    const rows = series.map((series) => ({
+        left: {
+            value: series.left.value,
+            labelWidth: series.left.label.width + series.left.label.padding,
+        },
+        right: {
+            value: series.right.value,
+            labelWidth: series.right.label.width + series.right.label.padding,
+        },
+    }))
+
+    const toScale = (v: number): number => (isLog ? Math.log(v) : v)
+    const fromScale = (v: number): number => (isLog ? Math.exp(v) : v)
+
+    let domainStart = toScale(domain[0])
+    let domainEnd = toScale(domain[1])
+    let padLeft = 0
+    let padRight = 0
+
+    const MAX_ITERATIONS = 5
+    const TOLERANCE = 0.5 // px
+
+    for (let i = 0; i < MAX_ITERATIONS; i++) {
+        const domainSpan = domainEnd - domainStart
+        const effectiveWidth = width - padLeft - padRight
+        if (domainSpan <= 0 || effectiveWidth <= 0) break
+
+        // Helper to convert a value to its pixel position along the axis
+        const pixelsPerUnit = effectiveWidth / domainSpan
+        const domainStartPixel = padLeft - domainStart * pixelsPerUnit
+        const toPixel = (value: number): number =>
+            domainStartPixel + toScale(value) * pixelsPerUnit
+
+        // Find the worst-case label overflow on each side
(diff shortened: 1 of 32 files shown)


sophiamersmann:
* **#6475** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6475?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6476](https://github.com/owid/owid-grapher/pull/6476) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6476?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6484](https://github.com/owid/owid-grapher/pull/6484) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6484?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#6474** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6254** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6254?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6254?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6461** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6461?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

marcelgerber (commented):
> Dumbbell plots support log scales, but we don’t currently indicate this with a ‘log axis’ notice, as we do for other chart types. Should we add this?

Do we imagine this being useful and being used? I'm asking because we removed support for log axes from discrete bar at some point, because it was more confusing than it was useful.
I could see a similar argument applying here, and can't currently see a concrete use case for a log axis.
We could still enable them later on when we see a use.

marcelgerber:
One UI design thing that stood out is that there is quite little space for entity names, which can then lead to them flowing into 3 lines [as seen here](http://localhost:3030/admin/charts/2838/edit).
CleanShot 2026-05-11 at 20 17 34

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartHelpers.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    for (let i = 0; i < MAX_ITERATIONS; i++) {
+        const domainSpan = domainEnd - domainStart
+        const effectiveWidth = width - padLeft - padRight
+        if (domainSpan <= 0 || effectiveWidth <= 0) break
+
+        // Helper to convert a value to its pixel position along the axis
+        const toPixel = (value: number): number =>
+            padLeft +
+            ((toScale(value) - domainStart) / domainSpan) * effectiveWidth
```
marcelgerber: This we can optimize by pulling out constants:

```suggestion
        const pixelsPerUnit = effectiveWidth / domainSpan
        const domainStartPixel = padLeft - domainStart * pixelsPerUnit
        const toPixel = (value: number): number =>
            domainStartPixel + toScale(value) * pixelsPerUnit
```

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+        const chartState = new DumbbellChartState({ manager })
+
+        const riser = chartState.series.find((s) => s.entityName === "Riser")!
+        const faller = chartState.series.find((s) => s.entityName === "Faller")!
+
+        expect(riser.connector.direction).toEqual("right")
+        expect(riser.color).toEqual(POSITIVE_CHANGE_COLOR)
+        expect(faller.connector.direction).toEqual("left")
+        expect(faller.color).toEqual(NEGATIVE_CHANGE_COLOR)
```
marcelgerber: can add in a small assertion here that even for a falling series, we have `left <= right`
```suggestion
        expect(riser.connector.direction).toEqual("right")
        expect(riser.color).toEqual(POSITIVE_CHANGE_COLOR)
        expect(riser.left.value).toBeLessThanOrEqual(riser.right.value)
        expect(faller.connector.direction).toEqual("left")
        expect(faller.color).toEqual(NEGATIVE_CHANGE_COLOR)
        expect(faller.left.value).toBeLessThanOrEqual(faller.right.value)
```

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.ts
```diff
@@ -0,0 +1,325 @@
...
+        const sortColumnSlug = this.sortConfig.sortColumnSlug
+        if (!sortColumnSlug) return undefined
+
+        const sortColumn = this.transformedTable.get(sortColumnSlug)
+        if (sortColumn && !sortColumn.isMissing) return sortColumn
+
+        return undefined
+    }
+
+    /** Constructs series by comparing compares the same column's values across two time points */
```
marcelgerber:
```suggestion
    /** Constructs series by comparing the same column's values across two time points */
```

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.ts
```diff
@@ -0,0 +1,325 @@
...
+                    return undefined
+
+                // The color of the dumbbell is determined by whether
+                // the value has increased or decreased over time
+                const direction =
+                    startRow.value <= endRow.value ? "right" : "left"
+                const directionColor =
+                    direction === "right"
+                        ? POSITIVE_CHANGE_COLOR
+                        : NEGATIVE_CHANGE_COLOR
```
marcelgerber: It might be worth to add an "unchanged" variant, where the two values are exactly unchanged. Styling them with `POSITIVE_CHANGE_COLOR` is a bit confusing, and possibly it would be better if we could keep them more neutral (grey, blue, ...).

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.ts
```diff
@@ -0,0 +1,325 @@
...
+
```
marcelgerber: I find the nomenclature here a bit confusing.
We show `POSITIVE_CHANGE_COLOR` when the change (`endValue - startValue`) is **negative**.

I guess "positive" here refers more to "good" (with it being green, and with poverty going down being a good thing), but since that's also not the case for all charts I'd go with some more neutral name.

marcelgerber (approved):
real nice, and very much exciting that we're getting a new chart type! there are obviously a good bunch of UI issues at this point, but I'm guessing that you're aware of all of them, and will address them in future PRs. this is certainly a good basis for all of this! 🙌

sophiamersmann:
> One UI design thing that stood out is that there is quite little space for entity names, which can then lead to them flowing into 3 lines [as seen here](http://localhost:3030/admin/charts/2838/edit).

Not sure why that happens since it uses the same code as DiscreteBar and StackedDiscreteBar charts. Will look into it!

> Do we imagine this being useful and being used?

I remember going back and forth on it in the shaping session. I can't quite recall now what made us want to support it, but neither of us had a strong opinion. I think I might as well drop support now and, as you say, easy to add support if we do find use cases

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.ts
sophiamersmann: Yup, I'll chat with Marwa about this!

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.ts
sophiamersmann: I'll just use increase/decrease 👍🏻

marcelgerber:
> I remember going back and forth on it in the shaping session. I can't quite recall now what made us want to support it, but neither of us had a strong opinion. I think I might as well drop support now and, as you say, easy to add support if we do find use cases

makes sense to me! we could even keep support there in code (as you have that already), and then just force the chart to always go into linear mode always.

sophiamersmann:
### Merge activity

* **Jun 3, 12:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6254).
* **Jun 3, 12:43 PM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6254).

PR #6259, opened 2026-03-19
🎉🤖 Revamp Latest page with Algolia-powered SPA search

mlbrgl:
## Motivation

`/latest` and `/data-insights` were two server-rendered, paginated listings with separate code paths and no shared filtering. There was no way to browse all chronological content together; topic filtering only existed on `/data-insights` and was broken on hydration; pagination required a baker pass per page.

This PR collapses both into a single Algolia-powered SPA at `/latest` with topic + content-type filters, infinite scroll, and unified card rendering for articles, data insights, announcements, data updates, and website upgrades. Old entry points (`/data-insights`, `/blog/*`, `/latest/page/N`, header nav, footer, hero, breadcrumbs) redirect or rewire to filtered `/latest` URLs.

Architecture documented in [site/latest/README.md](site/latest/README.md).

## Redirects

Every entry point on the site (header, mobile nav, hero pill, footer, breadcrumb, "see all" buttons, redirect rules) now builds `/latest` URLs through one `latestUrl()` helper. Retired routes — `/data-insights`, `/data-insights/N`, `/blog/*`, `/latest/page/N` — redirect to filtered `/latest` URLs.

Pass criterion: URL lands on a working destination (200, or 301 → 200). Fail: 404 or 301 → 404.

| URL | Production (current) | Staging (this PR) |
|---|---|---|
| `/data-insights` | ✅ [200](https://ourworldindata.org/data-insights) (SSR index) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights) → `/latest?type=data-insight` |
| `/data-insights/1` | ❌ [404](https://ourworldindata.org/data-insights/1) (page 1 = bare on prod) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/1) → `/latest?type=data-insight` |
| `/data-insights/2` | ✅ [200](https://ourworldindata.org/data-insights/2) (SSR page 2) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/2) → `/latest?type=data-insight` |
| `/data-insights/25` | ❌ [404](https://ourworldindata.org/data-insights/25) (no such page) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/25) → `/latest?type=data-insight` |
| `/data-insights/26` (boundary) | ✅ [404](https://ourworldindata.org/data-insights/26) | ✅ [404](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/26) (intentional) |
| `/data-insights/100` | ✅ [404](https://ourworldindata.org/data-insights/100) | ✅ [404](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/100) |
| `/data-insights/foo` (non-numeric) | ✅ [404](https://ourworldindata.org/data-insights/foo) | ✅ [404](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/foo) (rule doesn't shadow real slugs) |
| `/data-insights/teenage-pregnancy-rates-have-fallen-across-the-world` (real slug) | ✅ [200](https://ourworldindata.org/data-insights/teenage-pregnancy-rates-have-fallen-across-the-world) | ✅ [200](https://auto-claude-001-revamp-lates.owid.pages.dev/data-insights/teenage-pregnancy-rates-have-fallen-across-the-world) (permalinks unaffected) |
| `/atom-data-insights.xml` | ✅ [200](https://ourworldindata.org/atom-data-insights.xml) | ✅ [200](https://auto-claude-001-revamp-lates.owid.pages.dev/atom-data-insights.xml) (atom feed unaffected) |
| `/latest/page/1` | ❌ [404](https://ourworldindata.org/latest/page/1) (page 1 = bare on prod) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/latest/page/1) → `/latest` |
| `/latest/page/2` | ✅ [200](https://ourworldindata.org/latest/page/2) (SSR pagination) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/latest/page/2) → `/latest` |
| `/latest/page/99` | ❌ [404](https://ourworldindata.org/latest/page/99) | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/latest/page/99) → `/latest` |
| `/blog` | ✅ [301](https://ourworldindata.org/blog) → `/latest` | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/blog) → `/latest` |
| `/blog/foo` | ❌ [301](https://ourworldindata.org/blog/foo) → `/latest/foo` → 404 | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/blog/foo) → `/latest` |
| `/blog/some-old-post` | ❌ [301](https://ourworldindata.org/blog/some-old-post) → `/latest/some-old-post` → 404 | ✅ [301](https://auto-claude-001-revamp-lates.owid.pages.dev/blog/some-old-post) → `/latest` |
| `/latest` | ✅ [200](https://ourworldindata.org/latest) (SSR feed) | ✅ [200](https://auto-claude-001-revamp-lates.owid.pages.dev/latest) (SPA shell) |

**Production: 11/16 ✅, 5 ❌**. **Staging: 16/16 ✅** — the 5 prod failures are exactly the cases this PR fixes.

## Why no redirect for `/data-insights?topic=...`

The `topic=` query param isn't exposed anywhere on `/data-insights` or individual data insight pages — there are no links on the site producing these URLs (related-topic chips link to topic-page slugs like `/internet`, not `?topic=...`). The values seen in analytics match topic-page names, suggesting they're scraped from page text rather than clicked.

The traffic also looks synthetic: 100% null `page_referrer` (vs ~38% baseline), uniform desktop/Chrome/Mac fingerprint, sub-second hops between unrelated pages, and malformed URLs like `/data-insights?+Globalization&topic=Trade+` and topic values with trailing whitespace (`"CO2 "`, `"War "`, `"Military Personnel "`) that no real link would produce. Data insights filtered by topic page also break on hydration today (e.g. `/data-insights?topic=Internet`), so there's no working destination to preserve.

## Test plan

_Gdocs on staging with "DEMO" in their title can be used for testing_

- [x] `/latest` renders all five content types; topic + type filters work; zero-match options disable; "Load More" paginates
- [x] Articles with `latest-featured-image` / `latest-excerpt` render the override
- [x] `/data-insights`, `/data-insights/N`, `/blog/*`, `/latest/page/N` redirect to `/latest` (with appropriate type filter where applicable)
- [x] Legacy `/latest?topic=…`-style URLs are sanitized on first paint
- [x] ~~Does the change work in the archive?~~

## Checklist

- [x] Google Analytics events adapted
- [ ] `PagesChronological` Algolia index reindexed at deploy time
- [x] CSS/HTML checked on Desktop and Mobile Safari at all three breakpoints
- [x] HTML checked for accessibility


mlbrgl:
* **#6259** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6259?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6262** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

baker/algolia/utils/pagesChronological.ts
```diff
@@ -2,45 +2,194 @@ import * as _ from "lodash-es"
...
-        | OwidGdocDataInsightInterface
-        | OwidGdocAnnouncementInterface,
+/**
+ * Class-aware variant of `checkIsChronologicalGdoc`: narrows the
+ * broad return of `gdocFromJSON` to a chronological-feed class instance.
+ * Used (instead of the interface-narrowing predicate from
+ * `@ourworldindata/utils`) by the bulk indexing path below, which needs
+ * the `loadLinkedX(knex)` methods that live on `GdocBase` — the bare
+ * interface union doesn't carry them.
+ */
```
ikesau, line 36: This comment feels excessive. Personally I'd put this function adjacent to `checkIsChronologicalGdoc` and just say `// Use this when you're working with classes instead of JSON`

baker/algolia/utils/pagesChronological.ts
```diff
@@ -2,45 +2,194 @@ import * as _ from "lodash-es"
...
+function isChronologicalGdocInstance(
+    gdoc: ReturnType<typeof gdocFromJSON>
+): gdoc is (GdocPost | GdocDataInsight | GdocAnnouncement) & ChronologicalGdoc {
+    return checkIsChronologicalGdoc(gdoc)
+}
+
+/**
+ * Load the linked-content fields each chronological gdoc type needs
```
ikesau: "linked-content fields" i.e. attachments? can we just call them that?

baker/algolia/utils/pagesChronological.ts
```diff
@@ -2,45 +2,194 @@ import * as _ from "lodash-es"
...
-    const topicTags = getUniqueNamesFromTagHierarchies(
-        originalTagNames,
-        topicHierarchiesByChildName
-    )
+/**
+ * Build type-specific attachment fields for a PageChronologicalRecord.
+ * Dispatches on `content.type` so it works both for class instances
+ * (bulk path) and plain objects from `toJSON()` (individual path).
+ */
+function buildAttachment(
```
ikesau: I've always thought of an attachment as _one_ `ImageMetadata` object; _one_ `LinkedAuthor`, etc (like email)

So I think this should be called `buildAttachments` or `buildChronologicalRecordAttachments`

baker/algolia/utils/pagesChronological.ts
```diff
@@ -2,45 +2,194 @@ import * as _ from "lodash-es"
...
+ * Load the linked-content fields each chronological gdoc type needs
+ * for indexing. Mirrors the per-type contract enforced by `buildAttachment`
+ * — keep these two functions in sync when adding a new type.
+ *
+ * Takes `ChronologicalGdoc & GdocBase` because we need both the discriminated
+ * `content` shape (from the interface union) and the `loadLinkedX(knex)`
+ * methods (declared on the GdocBase class). Class instances from
+ * `gdocFromJSON` satisfy both.
+ */
+async function loadAttachmentsForChronologicalIndexing(
```
ikesau, line 58: I guess, more generally, could we just call `loadState` instead of this function? Is performance a significant enough problem in the bulk indexing path that we need to pare it down?

That way we wouldn't need to worry about the code drift either 🙂

baker/algolia/utils/pagesChronological.ts
```diff
@@ -2,45 +2,194 @@ import * as _ from "lodash-es"
...
+            },
+            _.noop
+        )
+        .exhaustive()
+}
 
-    const topicHierarchiesByChildName =
-        await db.getTopicHierarchiesByChildName(knex)
+/** Copy linkedCharts / linkedDocuments into the attachment if non-empty. */
+function copyLinkedContentIfPresent(
```
ikesau: and here also, `copyAttachmentsIfPresent`

packages/@ourworldindata/types/src/domainTypes/Search.ts
```diff
@@ -41,6 +49,31 @@ export type PageChronologicalRecord = {
     authors: string[]
     tags: string[]
     thumbnailUrl: string
+    // Drives /latest's type filter pill (data-insight / article /
```
ikesau: This current definition allows us to create invalid states and TS wouldn't catch it. I wonder if it'd be worth it to split this into a discriminated union instead of one "bag of optionals" type.

```
type PageChronologicalRecord = {
  // shared base...
} & (
  | { type: "data-insight"; body: OwidEnrichedGdocBlock[]; etc... }
  | { type: "article"; featuredImage?: string; etc... }
  // etc...
)
```

baker/algolia/utils/pagesChronological.ts
```diff
@@ -2,45 +2,194 @@ import * as _ from "lodash-es"
...
-    )
+/**
+ * Build type-specific attachment fields for a PageChronologicalRecord.
+ * Dispatches on `content.type` so it works both for class instances
+ * (bulk path) and plain objects from `toJSON()` (individual path).
+ */
+function buildAttachment(
+    gdoc: ChronologicalGdoc
+): Partial<PageChronologicalRecord> {
+    const attachment: Partial<PageChronologicalRecord> = {}
```
ikesau: I posted a comment on the definition of this type which was prompted by seeing this function first: we could define this type as a discriminated union, which would allow these instantiations to be more rigorous and validated with zod.

It's probably completely fine the way it is now, but it could be _even more_ fine if you feel motivated to refactor/tighten it up, and presumably would lead to fewer optional checks later on.

baker/algolia/utils/pagesChronological.ts
```diff
@@ -70,8 +216,24 @@ export async function indexIndividualGdocInChronological(
         return
     }
 
+    const cloudflareImagesByFilename = await db
+        .getCloudflareImages(knex)
+        .then((images) => _.keyBy(images, "filename"))
```
ikesau: kinda tempted to write `db.getCloudflareImagesByFilename`...

Image

baker/algolia/utils/pagesChronological.ts
```diff
@@ -132,28 +294,39 @@ export async function getPagesChronologicalRecords(
...
+    // linkedImageFilenames can pull in author avatars and prominent-link
+    // thumbnails.
     const records: PageChronologicalRecord[] = []
     for (const gdoc of gdocs) {
-        if (!gdoc.content.type || !gdoc.publishedAt) continue
+        // Narrow gdocFromJSON's broad return to a chronological class
+        // instance with a content.type literal-narrowed to ChronologicalGdoc's
+        // 5-type set. The DB query above already constrains rows to those
+        // types, so this is primarily a static-typing hatch — and a
+        // defensive guard against future additions to the query.
```
ikesau: This comment feels unnecessary to me.

baker/algolia/utils/pagesChronological.ts
```diff
@@ -132,28 +294,39 @@ export async function getPagesChronologicalRecords(
...
+        if (!isChronologicalGdocInstance(gdoc)) continue
+
+        await loadAttachmentsForChronologicalIndexing(gdoc, knex)
+        gdoc.imageMetadata = _.pick(
+            cloudflareImagesByFilename,
+            gdoc.linkedImageFilenames
+        )
```
ikesau: (If we're going to have it,) shouldn't this be done by `loadAttachmentsForChronologicalIndexing` ?

site/latest/latestUtils.ts
```diff
@@ -0,0 +1,93 @@
...
+    LatestFeedGdoc,
+    LatestType,
+    LatestUrlParam,
+    PageChronologicalRecord,
+} from "@ourworldindata/types"
+import { OwidGdocType, slugify } from "@ourworldindata/utils"
+import { match } from "ts-pattern"
+
+/** Build a URL to the /latest page, optionally pre-filtered by type. */
+export function latestUrl(type?: LatestType): string {
```
ikesau: You know I'm a stickler `verbNoun` function names 😅 

`buildLatestPagePath`? `makeLatestPagePath`?

(path not URL because it doesn't have the base url in it)

baker/redirects.ts
```diff
@@ -45,9 +66,21 @@ export const getCloudflarePagesRedirects = async (
...
+        // number is dropped on redirect.
+        "/latest/page/* /latest 301",
+
+        // Was `/blog/* /latest/:splat 301`, presumably to forward
+        // /blog/page/N to the SSR /latest's matching /latest/page/N. With
+        // those paginated routes gone (see above) and /latest having no
+        // /:slug route, the :splat lands on 404 for every input. Drop it.
+        // Specific /blog/<slug> recoveries live in DB-backed siteRedirects
+        // (emitted before this rule) and still win.
+        "/blog/* /latest 301",
```
ikesau, line 83: At what point do we retire redirects? On one hand, they're generally not expensive to keep around, on the other.... eh, who's landing on `/blog/` these days?

baker/SiteBaker.tsx
```diff
@@ -1097,61 +1045,8 @@ export class SiteBaker {
...
-            )
-
-            const outPath = path.join(
-                this.bakedSiteDir,
-                `latest/page/${pageNum}.html`
-            )
-            await fs.mkdirp(path.dirname(outPath))
-            await this.stageWrite(outPath, pageHtml)
-        }
+        const html = await renderLatestPage(knex)
```
ikesau, line 1064: One of those things that's hard to put a value on, but we generally try to have graceful no-JS fallbacks for our site.

Would it be difficult to have a latest page with 10 items on it and a "Javascript is required to filter / see more" notice for those who don't use it?

db/model/Gdoc/GdocAnnouncement.ts
```diff
@@ -33,6 +34,34 @@ export class GdocAnnouncement
                     type: OwidGdocErrorMessageType.Error,
                 })
             }
+            // Body and a top-level {.cta} are mutually exclusive: the
+            // standalone page renders excerpt + CTA in CTA mode and hides
+            // the body, so any body content would be invisible to readers
+            // on the canonical announcement page.
```
ikesau: I feel like the error message explains this already ✂️

db/model/Gdoc/GdocAnnouncement.ts
```diff
@@ -33,6 +34,34 @@ export class GdocAnnouncement
...
+                    message: `An announcement with a top-level {.cta} block must have an empty body. Either remove the body content (and keep the CTA), or remove the {.cta} block (and keep the body).`,
+                    type: OwidGdocErrorMessageType.Error,
+                })
+            }
+        }
+
+        // The kicker drives the announcement's category on /latest. Reject
+        // unrecognized values at publish time so the indexer never has to
+        // guess a fallback.
+        if (this.content.kicker) {
```
ikesau, line 49: Blegh, it's so inconsequential but using `kicker` as an enum will bother me forever 😂

db/model/Gdoc/GdocFactory.ts
```diff
@@ -520,25 +520,15 @@ async function getAndLoadPublishedGdocs<T extends GdocBase>(
 
 export async function getAndLoadPublishedDataInsightsPage(
     knex: KnexReadonlyTransaction,
-    page?: number, // 1-indexed
-    topicSlug?: string
+    page?: number // 1-indexed
```
ikesau: based on callsites, looks like we can just get rid of this param? If specified, it's only ever called with `1`

`LATEST_INDEX_PAGE_SIZE` could be removed also seeing as we don't need it to calculate any UI anymore. Just hard-code it to 20 and rename this function `getAndLoadLastPublishedDataInsights`

db/model/Gdoc/GdocPost.ts
```diff
@@ -98,6 +91,12 @@ export class GdocPost extends GdocBase implements OwidGdocPostInterface {
             )
         }
 
+        if (content["latest-excerpt"]) {
+            content["latest-excerpt"] = content["latest-excerpt"].map(
+                (html: RawBlockText) => htmlToEnrichedTextBlock(html.value)
```
ikesau: Should we validate this and filter nulls?

db/db.ts
```diff
@@ -450,7 +417,14 @@ export const getHomepageAnnouncements = (
         ORDER BY pg.publishedAt DESC
         LIMIT 3
         `
-    ).then((rows) => rows.map(rawGdocToMinimalPost))
+    ).then(
+        // SQL filters `type = 'announcement'`, so every row is an
+        // announcement at runtime. Promote to the static type.
```
ikesau: I don't think this comment is necessary

packages/@ourworldindata/types/src/gdocTypes/Gdoc.ts
```diff
@@ -381,6 +389,58 @@ export type OwidGdoc =
...
 
+export const CHRONOLOGICAL_INDEX_TYPE_VALUES = [
+    OwidGdocType.Article,
+    OwidGdocType.LinearTopicPage,
+    OwidGdocType.TopicPage,
+    OwidGdocType.DataInsight,
+    OwidGdocType.Announcement,
+] as const
+
+export const LATEST_FEED_TYPE_VALUES = [
```
ikesau, line 396: We should update the dynamic RSS endpoint to support these also. Should be simple to update `parsePageTypes`

site/SiteNavigation.tsx
```diff
@@ -155,7 +156,7 @@ export const SiteNavigation = ({
                                     <a href={SEARCH_BASE_PATH}>Data</a>
                                 </li>
                                 <li>
-                                    <a href="/data-insights">Insights</a>
+                                    <a href={latestUrl()}>Latest</a>
```
ikesau: We don't want to keep a link directly to data insights around?

site/SiteConstants.ts
```diff
@@ -109,21 +109,3 @@ export enum GalleryArrowDirection {
     prev = "prev",
     next = "next",
 }
-
-export const MOST_RECENT_DATA_INSIGHT = "most-recent-data-insight"
```
ikesau, line 113: 🪦🥳

site/latest/AnnouncementContent.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+/** Shared inner content of an announcement, used by both /latest feed hits
+ * (Algolia-backed) and the standalone preview page (gdoc-backed). Always
+ * needs a wrapper that supplies the AttachmentsContext and surrounding
+ * chrome — it is not a standalone component on its own. The two call sites
+ * differ only in heading level (h1 vs h2) and whether <ExpandableText>
+ * truncates with a Read more toggle (feed) or renders fully (standalone via
+ * alwaysExpanded). The shared <ExpandableText> wrapper keeps the DOM and
+ * grid layout identical between the two so editors previewing an
+ * announcement see what readers will see in the feed. */
```
ikesau: I don't understand what `shared <ExpandableText> wrapper` means here. 

Is it like: the standalone version passes `alwaysExpanded` instead of bypassing `<ExpandableText>` so the DOM and grid layout stay identical to the feed?

site/latest/ExpandableText.scss
```diff
@@ -0,0 +1,24 @@
+.expandable-text__lead--collapsed {
+    > p:last-of-type {
+        display: inline;
```
ikesau, line 3: The "jump" in content you were observing when expanding this comes from this line: switching from `display: inline` to unset, fwiw.

site/latest/LatestContext.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext } from "react"
+import { SiteAnalytics } from "../SiteAnalytics.js"
+
+interface LatestContextType {
+    analytics: SiteAnalytics
```
ikesau, line 5: Any particular reason to provide this with context compared to the way we normally do it?

site/latest/latestHooks.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+export function useLatestAnalytics(
+    state: LatestState,
+    analytics: SiteAnalytics
+): void {
+    const lastLoggedStateRef = useRef<LatestState | null>(null)
+
+    useEffect(() => {
+        // Skip analytics for default/empty state (initial page load).
+        if (R.isDeepEqual(state, DEFAULT_LATEST_STATE)) return
```
ikesau: If a user goes from `A → default state → A` it won't log the second A because the `lastLoggedStateRef` doesn't get updated when navigating to the default.

Maybe we should instead just not log the initial default load, but subsequent ones might be interesting. Idk.

site/latest/latestUtils.test.ts
```diff
@@ -0,0 +1,27 @@
+import { describe, expect, it } from "vitest"
+import { deriveAnnouncementLatestType } from "./latestUtils.js"
+
+describe(deriveAnnouncementLatestType, () => {
+    it.each(["data-update", "website-upgrade", "announcement"])(
+        "passes through canonical slug %s",
```
ikesau, line 6: Neat templating feature!

site/latest/LatestSearch.tsx
```diff
@@ -0,0 +1,220 @@
...
+    }, [searchParams, state, setSearchParams])
+
+    const updateParams = useCallback(
+        (updater: (current: LatestState) => LatestState) => {
+            setSearchParams(stateToSearchParams(updater(state)))
+        },
+        [setSearchParams, state]
+    )
+
+    const onTopicsChange = useCallback(
```
ikesau: I don't think `useCallback` is doing anything with these, because `updateParams` depends on `state`. `state` updates each time `searchParams` changes, so all of these get recreated.

site/latest/latestState.ts
```diff
@@ -0,0 +1,36 @@
...
+// `stateToSearchParams` lets the SPA canonicalize the URL the same way
+// /search does (see site/search/searchState.ts).
+export function searchParamsToState(
+    searchParams: URLSearchParams,
+    allAreas: string[]
+): LatestState {
+    const allAreasSet = new Set(allAreas)
+    const topics = [
+        ...deserializeSet(searchParams.get(LatestUrlParam.TOPICS)),
+    ].filter((t) => allAreasSet.has(t))
```
ikesau: Could use `Set.intersection` here 🤓 

```ts
const topics = [
    ...deserializeSet(searchParams.get(LatestUrlParam.TOPICS)).intersection(
        new Set(allAreas)
    ),
]  
```

site/latest/LatestTopicFacets.tsx
```diff
@@ -0,0 +1,317 @@
...
+                    </button>
+                    {isContentTypeDropdownOpen && (
+                        <>
+                            <div
+                                className="latest-topic-facets__content-type-dropdown-backdrop"
+                                onClick={() =>
+                                    setIsContentTypeDropdownOpen(false)
+                                }
+                            />
+                            <ul
```
ikesau: I can open this but can't enter it with my keyboard

site/latest/LatestTopicFacets.tsx
```diff
@@ -0,0 +1,317 @@
...
+                                                {
+                                                    "latest-topic-facets__content-type-dropdown-item--active":
+                                                        isActive,
+                                                    "latest-topic-facets__content-type-dropdown-item--disabled":
+                                                        isDisabled,
+                                                }
+                                            )}
+                                            role="option"
+                                            aria-selected={isActive}
+                                            aria-disabled={isDisabled}
```
ikesau: I think the best way to fix it would just be to make the `<li>` wrap `<button>`s and use `disabled={isDisabled}` on that instead of `aria-disabled`

ikesau (changes requested):
🎉

Nicely done!

I would say most of my comments are nitpicks and minor comments. I really would like to ship this with a no-JS page though, and for the keyboard navigation on the page to work.

Also, I don't think this was due to your changes, but the email signup input is 14px on mobile, which causes iOS to zoom in. Might wanna fix that lest we suffer marwa rage.

baker/redirects.ts
mlbrgl: moz inbound links tool reports 24K backlinks to /blog over 113 domains. So I'd keep it for now. But it's true that we don't have a systematic process to evaluate and deprecate them.

baker/algolia/utils/pagesChronological.ts
mlbrgl: ok, simplified. Fn left in their respective packages due to TS project dependencies.

db/model/Gdoc/GdocAnnouncement.ts
mlbrgl: ...or until we get gdoc migration capabilities 🤞 Well let me know if there is anything you'd do in the meantime.

packages/@ourworldindata/types/src/gdocTypes/Gdoc.ts
```diff
@@ -441,6 +501,8 @@ export interface OwidGdocPostContent {
     refs?: { definitions: RefDictionary; errors: OwidGdocErrorMessage[] }
     summary?: EnrichedBlockText[]
```
mlbrgl, line 498: rename to `latest-feed-featured-image`

packages/@ourworldindata/types/src/gdocTypes/Gdoc.ts
```diff
@@ -441,6 +501,8 @@ export interface OwidGdocPostContent {
...
     "deprecation-notice"?: EnrichedBlockText[]
+    "latest-featured-image"?: string
+    "latest-excerpt"?: EnrichedBlockText[]
```
mlbrgl: rename to `latest-feed-excerpt`

site/latest/AnnouncementContent.tsx
mlbrgl: yep, that's correct. Rephrased accordingly.

site/latest/latestHooks.ts
mlbrgl: good call! That would also help us decide down the line whether to add a reset button if we see a significant amount of manual resets. Search analytics adapted in the same way.

site/latest/LatestContext.tsx
mlbrgl: Mostly out of parity with search, where the context is more flesched out and makes it a more natural home for analytics. A more opinionated reason is that - even though the instanciation of SearchAnalytics is cheap - it feels slightly wrong to have that code run no matter the page, on module load. So I'm limiting the blast radius by having it run once per SPA (now moved out of the useMemo and into module scope for both search and latest)

baker/algolia/utils/pagesChronological.ts
mlbrgl: yes 😅

baker/algolia/utils/pagesChronological.ts
mlbrgl: ok I reversed it and moved it close to the DB to warn against drift

baker/algolia/utils/pagesChronological.ts
mlbrgl: Profiling it returns the following results (ran on staging):

loadAttachmentsForChronologicalIndexing (selective) (5 runs, seconds): 10.7 / 11.5 / 11.7 / 11.8 / 12.3

-> median 11.7s

loadState (3 runs, seconds): 617.0 / 641.7 / 749.4

-> median 641.47s

`loadState` simplifies maintenance in one way (we don't have to worry about drift), but makes further work on this process more painful. I'll keep the selective approach for now - let's see if future edits tip the scale the other way.

db/db.ts
```diff
@@ -1238,3 +1212,25 @@ export const getUniqueTopicCount = async (
...
+          AND published IS TRUE
+          AND publishedAt <= NOW()`,
+        { type: OwidGdocType.Article }
+    )
+        .then((res) => (res ? res.count : 0))
+        .catch((e) => {
+            console.error("Failed to get published article count", e)
+            throw e
+        })
+    if (!count) throw new Error("Failed to get published article count")
```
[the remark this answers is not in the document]
mlbrgl: @ikesau I thought you'd done it this way in `getUniqueTopicCount` as no published topics would mean in our case that something went really wrong and was worth throwing for (favoring production bakes over hypothetical empty local envs, which are seeded from prod anyway). Do you feel differently now or did I misread your intentions?

mlbrgl:
thanks for the feedback @ikesau, some good stuff in there! I now applied react-aria to both filtering widgets, which gives us better accessibility and keyboard friendliness.

Overall you should find all your points addressed in separate commits, on top of the the previous commits, which are unchanged in content.

As discussed, the no-JS alternative won't fit in this PR, but could be picked up at a later time.

PR #6266, opened 2026-03-23
update supported browsers

marcelgerber:
This updates the supported browsers to ones that have been released in July 2023 at the latest.

This also aligns with [Vite's new default range](https://vite.dev/guide/build#browser-compatibility), which I believe is a reasonable choice to make at this point.

Incidentally, I've run tests on old Safari versions recently, and we have an SVG rendering issue on Safari <16 anyway:
CleanShot 2026-03-23 at 09 36 21

This also means that we can get rid of the polyfill, since there is nothing it'll do for these browsers.
In the future, we can think about using core-js or other polyfill approaches for modern features we'd like to use, like `Set.intersection()` and others.

I've also checked our Google Analytics data, and there's a negligible amount of users on browsers versions that we now drop support for.

Please note that our code will now fail in Safari < 16.4 with a SyntaxError.

<details>
<summary><i>Unchecked</i> AI summary of newly-supported features</summary>

So we've bumped from **Chrome 91 / Firefox 91 / Safari 14.1** (mid-2021) to **Chrome 111 / Firefox 114 / Safari 16.4** (early-mid 2023). Here are notable JS and CSS features that became available across all major browsers in that range:

### JavaScript

-   **`Array.prototype.at()`** --- negative indexing like `arr.at(-1)` (Chrome 92+, Firefox 90+, Safari 15.4+)
-   **`Object.hasOwn()`** --- cleaner alternative to `Object.prototype.hasOwnProperty.call()` (Chrome 93+, Firefox 92+, Safari 15.4+)
-   **`structuredClone()`** --- deep cloning without hacks (Chrome 98+, Firefox 94+, Safari 15.4+)
-   **`Array.prototype.findLast()` / `findLastIndex()`** --- find from the end (Chrome 97+, Firefox 104+, Safari 15.4+)
-   **`Array.prototype.toReversed()` / `toSorted()` / `toSpliced()` / `with()`** --- non-mutating array methods (Chrome 110+, Firefox 115+, Safari 16+)
-   **`Array.prototype.group()` / `Map.groupBy()`** --- native grouping (Chrome 117+, Firefox 119+ --- note: not yet in our Firefox 114 target)
-   **Top-level `await`** in modules (Chrome 89+, Firefox 89+, Safari 15+)
-   **`crypto.randomUUID()`** (Chrome 92+, Firefox 95+, Safari 15.4+)
-   **`AbortSignal.abort()` / `AbortSignal.timeout()`** --- static constructors (Chrome 93+/103+, Firefox 100+/100+, Safari 15+/16+)
-   **Error `.cause`** --- `new Error("msg", { cause: err })` (Chrome 93+, Firefox 91+, Safari 15+)
-   **`String.prototype.replaceAll()`** --- already in old Safari/Firefox but Chrome 85+ (fully available)
-   **Private class fields and methods** (`#field`) --- syntax that Vite can now pass through without downleveling (Chrome 84+, Firefox 90+, Safari 14.1+)
-   **RegExp match indices** (`/pattern/d`) (Chrome 90+, Firefox 88+, Safari 15+)
-   **RegExp lookbehind** (Safari 16.4+)

### CSS

-   **CSS `:has()` selector** --- parent selector (Chrome 105+, Safari 15.4+; Firefox 121+ --- beyond our FF target, so degrades gracefully)
-   **Container queries (`@container`)** (Chrome 105+, Firefox 110+, Safari 16+)
-   **CSS `@layer` (cascade layers)** (Chrome 99+, Firefox 97+, Safari 15.4+)
-   **`:where()` / `:is()` selectors** --- already partially available, but fully stable across all targets now
-   **`accent-color`** --- style form controls (Chrome 93+, Firefox 92+, Safari 15.4+)
-   **`color-mix()`** (Chrome 111+, Firefox 113+, Safari 16.2+)
-   **Small/large/dynamic viewport units** (`svh`, `lvh`, `dvh`) (Chrome 108+, Firefox 101+, Safari 15.4+)
-   **CSS `translate` / `rotate` / `scale` as individual properties** (Chrome 104+, Firefox 72+, Safari 14.1+)
-   **`inert` attribute** (Chrome 102+, Firefox 112+, Safari 15.5+)
-   **`overscroll-behavior`** (Chrome 63+, Firefox 59+, Safari 16+)
-   **`:focus-visible`** (Chrome 86+, Firefox 85+, Safari 15.4+)
-   **CSS `lch()`, `oklch()`, `lab()`, `oklab()` color functions** (Chrome 111+, Firefox 113+, Safari 15.4+)

The most impactful ones for this codebase are likely the non-mutating array methods (`toReversed()` etc.), `structuredClone()`, `Object.hasOwn()`, container queries, `:has()`, and `color-mix()` --- several of which are already in use.

</details>

### DIFF
--- a/docs/browser-support.md
+++ b/docs/browser-support.md
@@ -1,11 +1,13 @@
 ## Supported browsers
 
-As of **2026-03-23**, we support the following browsers:
+As of **2026-03-24**, we officially support the following browsers:
 
 - Safari 16.0+ [September 2022]
 - iOS Safari 16.0+ [September 2022]
 - Chrome/Edge 106+ [September 2022] (note that Chrome 109 is the last supported version on Windows 7/8, so we want to support that for a tad bit longer)
 - Firefox 110+ [February 2023] (note that 115 is an ESR version)
+
+Older versions of these browsers may still work, but we don't actively test or support them, and the site may be seriously broken.
 
 ### "Most breaking" features
 
@@ -18,10 +20,10 @@ As of **2026-03-23**, we support the following browsers:
 
 Vite/Rolldown may emit syntax in the bundled output that we don't actively write ourselves. Vite will use syntax that are not supported in older browsers. The most notable examples are:
 
-- [Public class fields](https://caniuse.com/mdn-javascript_classes_public_class_fields) that contain parentheses (`class Foo { bar = _.noop() }`) — Safari 16.0+.
-- [Static class initialization blocks](https://caniuse.com/mdn-javascript_classes_static_initialization_blocks) (`class Foo { static { ... } }`) — Chrome 91+, Firefox 93+, Safari 16.4+.
+- [Private class methods](https://caniuse.com/mdn-javascript_classes_private_class_methods) - Chrome 84+, Safari 15+, Firefox 90+.
+- [Nullish coalescing assignment operator (`??=`)](https://caniuse.com/mdn-javascript_operators_nullish_coalescing_assignment) — Chrome 85+, Safari 14+, Firefox 79+.
 
-Both are safe for our current targets, but they will causes issues (incl. SyntaxErrors) when our code is run in older browsers, especially in Safari < 16.4.
+Both are safe for our current targets, but they will causes issues (incl. SyntaxErrors) when our code is run in older browsers, especially in Safari < 16.0.
 
 ### Features we can't yet use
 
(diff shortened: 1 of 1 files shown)


rakyi:
Hmm, I'm not sure about this, less than 3 years seems too little. Do we have any reasons for bumping the versions and removing the polyfill (other than it's nice to have simpler code)?

> I've also checked our Google Analytics data, and there's a negligible amount of users on browsers versions that we now drop support for.

Can you say how many? Even if it's 0.1% it can mean breaking the site for hundreds of users, right? I'd rather stay conservative and make sure the site works for most users. I think even now some features are broken for older browsers, like the Safari example you pointed out, but it's better if fewer things are broken than many.

marcelgerber:
@rakyi 

> Do we have any reasons for bumping the versions and removing the polyfill (other than it's nice to have simpler code)?

Not any particular one, but Daniel recently mentioned that Wikipedia would like there to be as few as possible external domains that we send data to, and so if we could get rid of the external polyfill link that'd be great.

And on the other hand, it would be great if we can "officially" use newer features, like container queries (which we already use, but in a way where it's not bad if they fail).

> Can you say how many?

I had another look now, and apparently, I did look at the wrong numbers this morning; there's more people on old Chrome versions than I thought.

(table stripped)

So, overall, this means we would end support for a bit over 1% of our users.

All other Chrome versions really are neglibible, we only have a decent number of users of these two particular versions (and of v58 for some reason, which there's no way in hell we will support!).

For Safari, it'd be easy to set the target to 16.0+ (September 2022) instead.
For Chrome, maybe it's better to include an _explicit_ polyfill for `.toReversed()` etc., which would mean that thereby we can lower the target to include Chrome 109, at least. I'm not sure we should continue supporting 91.

Do you agree, @rakyi?

PR #6290, opened 2026-03-26
🔨 remove chart intro animations

ikesau:
Animations work best when you need to make sure your user notices a subtle change to the UI as a result of one of their actions. Having the entire chart wipe/fade in doesn't serve that purpose.

It also makes various slide transitions in my slideshow builder look worse. 

I think we should remove them 🙂 

**Before**

https://github.com/user-attachments/assets/ec53f76a-41fa-4411-8e7d-ee644ac2f937

**After**

https://github.com/user-attachments/assets/a6477209-750f-4b64-bb64-a79f415bb86e


ikesau:
@codex review

PR #6294, opened 2026-03-27
🎉 Add pull-chart and chart-rows block types for inline chart thumbnails

mlbrgl:
## Summary

Two new ArchieML block types that embed chart thumbnail images inline in articles. Thumbnails link to full interactive charts, with a shared `ChartThumbnail` component providing the image + "Click to explore" hover overlay.

> **Note:** Earlier commits in this PR implemented both as a single `small-chart` block with a `variant` discriminant. The final commit splits them into two independent types — the conceptual overlap between the two was too thin to justify keeping them together.

🎨 [Figma](https://www.figma.com/design/p5dMrFkbKoNa3n0bzi7w11/More-views-of-this-data?node-id=0-1)

### `chart-rows` — vertical list of chart thumbnails

- Supports **kicker**, **title**, **source**, and per-row rich text content with horizontal dividers
- **Guided chart integration**: when inside a `[.+guided-chart]` block, clicking a thumbnail updates the parent chart instead of navigating. Title and source are hidden in this mode.

```
{.chart-rows}
kicker: More views of this data
title: Daily incomes by decile
source: Global Carbon Budget (2025)

[.rows]
  image: life-expectancy-thumbnail.png
  url: /grapher/life-expectancy
  [.+content]
    Life expectancy has more than doubled.
  []

  image: gdp-per-capita-thumbnail.png
  url: /grapher/gdp-per-capita
  [.+content]
    Some content
  []
[]
{}
```

### `pull-chart` — single chart thumbnail floating within text

- Supports **left-center** and **right-center** alignment, floating the thumbnail outside the content column
- Flat structure: `image`, `url`, and `content` directly on the block (no rows)
- No guided chart integration (always renders as a plain link)

```
{.pull-chart}
align: left-center
image: chart.png
url: /grapher/test

[.+content]
Content that wraps around the floating thumbnail.
[]
{}
```

### Shared

- **`ChartThumbnail`** component: static chart image with hover overlay, reused by both block types
- **Improved guided chart scroll behavior**: scrolls to chart when not fully visible, uses `block: "nearest"` for minimal scroll distance

### Bug fix

- **`guided-chart` children now extracted in `posts_gdocs_components`**: `guided-chart` was incorrectly in the "no children" group, so nested blocks (chart, chart-rows, text, etc.) were never indexed. Fixed to traverse its `content` property.

## Test plan

_Test document:_ http://staging-site-small-chart/admin/gdocs/1JBIpGt3WD5bl2YQ8_OcBGre7YLc4CeYdIJrtsj7pc90/preview

- [x] Verify chart-rows renders with kicker, title, source, thumbnail+text pairs, and dividers
- [x] Verify pull-chart floats thumbnail within text at left-center and right-center alignments
- [x] Verify "Click to explore" overlay appears on hover over any thumbnail
- [x] Test guided chart integration: clicking a chart-rows thumbnail inside `[.+guided-chart]` updates the parent chart
- [x] Verify title/source are hidden in guided chart mode, kicker is still shown
- [x] Test on mobile breakpoint — rows stack vertically, pull-chart floats collapse

## After deploy
- [ ] Update the reference Gdoc component documentation

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,6 +19,7 @@ When you want to create a git commit, refer to docs/agent-guidelines/commit-mess
 - Avoid the use of the `any` type. Only use it if you have to and ask for permission.
 - In Grapher and the admin, where we use MobX 6, we use a somewhat nonstandard setup. We use class based components with TC-39 stage 3 decorators, but only for @computed and @action properties. The observable props are not marked with @observable, but are instead listed in the constructor in a `makeObservable` call. The `makeObservable` call must mention all obserable props, but none of the @computed or @action ones.
 - For CSS, we mostly use named style classes following the BEM conventions in separate .scss files. We usually avoid inline styles - only use those if the component you are working on already makes use of them for a similar use case. Components usually have a companion scss file with the same name. The entry point for our site styles is /site/owid.scss, the entry point for grapher styles is /packages/@ourworldindata/grapher/src/core/grapher.scss
+- In SCSS files, do NOT use parent selector nesting (`&__element`, `&--modifier`). Write out full class names (`.block__element`) so it's easy to search/navigate between JSX and SCSS.
 - Check [docs/browser-support.md](./docs/browser-support.md) before using modern JS or CSS features. It lists our supported browsers, the "most breaking" features we rely on, and features we can't yet use.
 
 # Codebase overview


mlbrgl:
* **#6294** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6294?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

db/model/Gdoc/rawToEnriched.ts
```diff
@@ -1187,6 +1196,139 @@ const parsePullQuote = (raw: RawBlockPullQuote): EnrichedBlockPullQuote => {
...
+            parseErrors.push({
+                message: "chart-rows row item missing url property",
+            })
+            continue
+        }
+
+        const enrichedContent: EnrichedBlockText[] = row.content
+            ? row.content
+                  .map(parseRawBlocksToEnrichedBlocks)
+                  .filter((b): b is EnrichedBlockText => b?.type === "text")
```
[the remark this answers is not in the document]
mlbrgl: Fixed in 1568e48 — both parsers now emit a parse warning listing the count of dropped non-text blocks.

db/model/Gdoc/rawToEnriched.ts
```diff
@@ -1187,6 +1196,139 @@ const parsePullQuote = (raw: RawBlockPullQuote): EnrichedBlockPullQuote => {
...
+        parseErrors.push({ message: "pull-chart missing image property" })
+    }
+    if (!url) {
+        parseErrors.push({ message: "pull-chart missing url property" })
+    }
+
+    const enrichedContent: EnrichedBlockText[] = content
+        ? content
+              .map(parseRawBlocksToEnrichedBlocks)
+              .filter((b): b is EnrichedBlockText => b?.type === "text")
```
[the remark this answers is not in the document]
mlbrgl: Fixed in the same commit (1568e48).

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
```diff
@@ -0,0 +1,33 @@
...
+import { EnrichedBlockWithParseErrors } from "./generic.js"
+import { EnrichedBlockText } from "./Text.js"
+
+export type RawChartRowItem = {
+    image?: string
+    url?: string
+    content?: OwidRawGdocBlock[]
+}
+
+export type ChartRowItem = {
```
rakyi: Shouldn't this be `EnrichedChartRowItem`? It seems inconsistent, but maybe I'm just missing something.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/PullChart.ts
```diff
@@ -0,0 +1,24 @@
...
+        align?: string
+        image?: string
+        url?: string
+        content?: OwidRawGdocBlock[]
+    }
+}
+
+export type EnrichedBlockPullChart = {
+    type: "pull-chart"
+    align?: PullChartAlignment
```
rakyi, line 19: Since we set one even when the user doesn't specify it, this should never be undefined, right?

db/model/Gdoc/rawToEnriched.ts
```diff
@@ -1187,6 +1196,155 @@ const parsePullQuote = (raw: RawBlockPullQuote): EnrichedBlockPullQuote => {
...
+        }
+        if (!row.url) {
+            parseErrors.push({
+                message: "chart-rows row item missing url property",
+            })
+            continue
+        }
+
+        const parsedContent = row.content
+            ? row.content.map(parseRawBlocksToEnrichedBlocks)
```
rakyi: Should we use the more specific `parseText` here and for `pull-chart` content? Then we should not need to reimplement the error handling for non-text.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
```diff
@@ -0,0 +1,33 @@
...
+        kicker?: string
+        title?: string
+        source?: string
+        rows?: RawChartRowItem[]
+    }
+}
+
+export type EnrichedBlockChartRows = {
+    type: "chart-rows"
+    kicker?: string
```
rakyi: Since we set one even when the user doesn't specify it, this should never be undefined, right?

Although, I think in this case it might be better to leave it empty (`""`) in the DB and let the component decide on the fallback, rather than us setting it during parsing. Capturing that something wasn't specified gives us more flexibility in the future.

Might be better also for `align` of the `pull-chart`, but I'm less sure there.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
```diff
@@ -0,0 +1,33 @@
...
+    title?: string
+    source?: string
```
rakyi: I like that we fall back to `""` for these other optional attributes in the `pull-chart`. It's better to have just a single representation of an empty value (`""` in this case) rather than the possibility of also `undefined` (like here), unless the `""` and `undefined` have different meaning, which I think is not the case here.

db/model/Gdoc/enrichedToMarkdown.ts
rakyi, line 1: It should be easy to add the other attributes (at minimum as paragraphs), so we might as well do it, right?

site/gdocs/components/ChartRows.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    {thumbnail}
+                </button>
+                {content}
+            </div>
+        )
+    }
+
+    return (
+        <div className="chart-rows__row">
+            <a href={row.url} className="chart-thumbnail">
```
rakyi: We should probably add `aria-labelledby`, similar to `DataInsightCard`. Same for `PullChart`.

site/gdocs/components/ChartRows.tsx
```diff
@@ -0,0 +1,81 @@
...
+    d: EnrichedBlockChartRows
+    className?: string
+}) {
+    const isInGuidedChart = !!useGuidedChartLinkHandler()
+
+    return (
+        <div
+            className={cx(
+                "chart-rows",
+                isInGuidedChart && "chart-rows--guided",
```
rakyi: Nit: I find the object with boolean values syntax we use elsewhere more readable.

site/gdocs/components/ChartRows.tsx
```diff
@@ -0,0 +1,81 @@
...
+                <h4 className="chart-rows__title">{d.title}</h4>
+            )}
+            <div className="chart-rows__rows-container">
+                {d.rows.map((row, i) => (
+                    <ChartRow key={i} row={row} />
+                ))}
+            </div>
+            {!isInGuidedChart && d.source && (
+                <p className="chart-rows__source">
+                    <strong>Data source: </strong>
```
rakyi: I think this doesn't match the semantics of the [`<strong>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/strong) element. We can use CSS to make it bold.

site/gdocs/components/ChartRows.scss
```diff
@@ -0,0 +1,109 @@
+.chart-rows {
+    margin-top: 16px;
+    margin-bottom: 32px;
+    border-top: 1px solid $blue-10;
+    border-bottom: 1px solid $blue-10;
+    padding-top: 24px;
+    padding-bottom: 24px;
+
+    &__kicker {
```
rakyi: AFAIK we decided against using this syntax because it makes it harder to navigate between JSX and SCSS. See e.g. `site/search/SearchWritingResults.scss`.

Same in the other new SCSS files.

site/gdocs/components/ChartRows.scss
```diff
@@ -0,0 +1,109 @@
...
+        @include h5-black-caps;
+        & {
+            color: $blue-50;
+            margin: 0 0 24px;
+        }
+
+        // Tighten gap when followed by a title
+        &:has(+ .chart-rows__title) {
+            margin-bottom: 2px;
```
rakyi: We don't really need the `:has` selector. It makes the styles complex and harder to understand IMO. There should be multiple ways of achieving the same without it.

site/gdocs/components/ChartRows.scss
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    .chart-thumbnail {
+        flex-shrink: 0;
+        max-width: 300px;
+
+        @include sm-only {
+            max-width: unset;
+            width: calc(10 / 12 * 100%);
```
rakyi: Could you add a code comment explaining why we do this?

site/gdocs/components/ChartRows.scss
```diff
@@ -0,0 +1,109 @@
...
+        > *:last-child {
+            margin-bottom: 0;
+        }
+    }
+
+    &__row-content .article-block__text {
+        @include body-3-regular;
+    }
+
+    // Inside a guided chart: no outer borders, no margins, kicker only
```
rakyi: It would be clearer if we had separate classes for this case instead of resetting many props.

rakyi (commented):
Nice work overall!

I'm particularly keen on tightening the types/data model though.

db/model/Gdoc/enrichedToMarkdown.ts
mlbrgl: sure, makes sense 👍

db/model/Gdoc/rawToEnriched.ts
mlbrgl: yep let's do that. I guess I'm a bit phased out by all these parse* functions accepting narrow types, only to check for the validity of the type within the function (effectively meaning that the input type is wider than advertised). But that's a fight for another day 🙂

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
mlbrgl: Yep, I specifically didn't use the `enriched` prefix here because this pair is not an archie component, but rather a subtype within a parsed archie component - and in my mental model the raw/enriched convention is for parsed Archie component (`OwidRawGdocBlock`/`OwidEnrichedGdocBlock`). I could find at least one example of this raw/enriched convention being applied to non-archi component so happy to go with the less surprising option here.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
mlbrgl: I agree, there shouldn't be a reason to override the default with an empty value in this case and show nothing instead of the default value. Let's combine the semantics of undefined and "".

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/PullChart.ts
mlbrgl: I followed the same logic here and let the component set the default. `undefined` feels more appropriate for an enum than "", but I could be persudaded otherwise.

db/model/Gdoc/enrichedToMarkdown.ts
mlbrgl: Addressed in [68f6373b](https://github.com/owid/owid-grapher/commit/68f6373bb) — kicker, title, and source are now included in the markdown output.

db/model/Gdoc/rawToEnriched.ts
mlbrgl: Addressed in [70a93b1d](https://github.com/owid/owid-grapher/commit/70a93b1d1) — using parseText directly, with parse errors surfaced with block-type context.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
mlbrgl: Addressed in [d2c21160](https://github.com/owid/owid-grapher/commit/d2c21160e) — renamed to EnrichedChartRowItem for consistency, though see my earlier reply about the rationale for not using the prefix originally.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
mlbrgl: Addressed in [a810d730](https://github.com/owid/owid-grapher/commit/a810d730f) — kicker, title, source are now required strings defaulting to "". The "More views of this data" fallback moved to the component.

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/ChartRows.ts
mlbrgl: Addressed in [a810d730](https://github.com/owid/owid-grapher/commit/a810d730f) — all three fields default to "" consistently. Also made pull-chart return early with createError when image/url is missing (instead of rendering half-broken).

packages/@ourworldindata/types/src/gdocTypes/archieMLComponents/PullChart.ts
mlbrgl: Addressed in [a810d730](https://github.com/owid/owid-grapher/commit/a810d730f) — removed the parser default for align; the component already handles it with `d.align ?? "left-center"`.

site/gdocs/components/ChartRows.tsx
mlbrgl: Addressed in [74f46c97](https://github.com/owid/owid-grapher/commit/74f46c976) — replaced `<strong>` with a `.chart-rows__source-label` span styled bold via CSS.

site/gdocs/components/ChartRows.scss
mlbrgl: yes you're right, that's the convention we reestablished back then. With AI writing and maintaining more and more of this code, I'm needing less and less of the direct search/navigation capabilities and so gave in to its preference for this style of writing. I'll realign for now.

mlbrgl: Addressed in [f9ae3807](https://github.com/owid/owid-grapher/commit/f9ae3807a) — expanded all `&__element` nesting to fully qualified class names in ChartRows.scss and ChartThumbnail.scss.

site/gdocs/components/ChartRows.scss
mlbrgl: Addressed in [9f38329c](https://github.com/owid/owid-grapher/commit/9f38329cd) — replaced `:has` with a negative `margin-top` on the title.

site/gdocs/components/ChartRows.scss
mlbrgl: Addressed in [13f4ae6d](https://github.com/owid/owid-grapher/commit/13f4ae6de) — added `// emulating 10 columns without grid` comment.

site/gdocs/components/ChartRows.scss
mlbrgl: Addressed in [13f4ae6d](https://github.com/owid/owid-grapher/commit/13f4ae6de) — split into `.chart-rows--standalone` and `.chart-rows--guided` so neither resets the other.

site/gdocs/components/ChartRows.scss
mlbrgl: Addressed in [13f4ae6d](https://github.com/owid/owid-grapher/commit/13f4ae6de) — used `margin-top` on the title instead.

site/gdocs/components/ChartRows.tsx
mlbrgl: Addressed in [13f4ae6d](https://github.com/owid/owid-grapher/commit/13f4ae6de) — using object syntax now.

site/gdocs/components/ChartRows.tsx
mlbrgl: good point. I went with `aria-describedby` to complement a shorter (generic) label since we don't have short titles per row, which as I understand it (and by testing with VoiceOver) is a more appropriate setup than having the screen reader read the whole paragraph as a label.

mlbrgl: Addressed in [046d9499](https://github.com/owid/owid-grapher/commit/046d9499e) — added `aria-label` for the action name and `aria-describedby` pointing at the adjacent text content, for both ChartRows and PullChart.

mlbrgl:
@rakyi thanks for the feedback! I've addressed all comments, with separate commits referenced in each thread. Let me know if you're doing another review.

rakyi (approved):
Looks great now!

PR #6298, opened 2026-03-30
🔨 remove animation-only clipPaths

marcelgerber:
## Summary
- Follow-up to #6290
- Removes clipPaths from LineChart, StackedAreaChart, and StackedBarChart that only existed to support the wipe-in intro animation (rect width from 0 → full)
- Their static renderings already worked without clipping, confirming these clipPaths served no other purpose
- Functional clipPaths in ScatterPlotChart, MapChart, and CustomComparisonLine are left intact

## Test plan
- [x] Verify line charts render correctly without visual clipping issues
- [x] Verify stacked area charts render correctly
- [x] Verify stacked bar charts render correctly
- [x] Check edge cases: charts with data near axis boundaries

### DIFF
--- a/packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
+++ b/packages/@ourworldindata/grapher/src/lineCharts/LineChart.tsx
@@ -48,11 +48,9 @@ import {
 } from "./LineChartConstants"
 import { CoreColumn } from "@ourworldindata/core-table"
 import {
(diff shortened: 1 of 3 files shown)


marcelgerber:
@sophiamersmann can you think of any problems this could cause? Given that for all these chart types the `renderStatic` call doesn't include a clip path I think this should be safe, no?

PR #6302, opened 2026-03-31
Add JSON-LD to data insight

rakyi:
[Slack](https://owid.slack.com/archives/C0149NKLZAM/p1774891532589769)

Can you think of any problems with this?


site/gdocs/OwidGdocPage.tsx
```diff
@@ -237,8 +245,15 @@ function isProfilePredicate(
 
 function isArticleLikePredicate(
```
ikesau: Maybe rename this to tie it more to the JSON-LD sense of article. Otherwise I could imagine people getting confused.

Issue #6189, opened 2026-03-05
Cycle 2026.2: Surfacing our work in LLMs


rakyi:
_We want to improve how our work surfaces in answers given by LLMs._

## Motivation

- Minimize LLMs using our data incorrectly and/or attributing us incorrectly
- Increase the quality, availability, and level of detail of LLM responses for topics we have data for
- Surface data provenance in chats

## Proposed scope

Design and run experiments on data pages to improve how OWID appears in LLM answers. We are focusing on inference, not pretraining.

**Must do**

- [x] Choose data pages to use for interventions     
- [x] Get a sense of baseline by running experiments without interventions first
  - In web interfaces (ChatGPT, Claude)
  - In coding agents (Codex, Claude)
    - With and without custom agent skills (optional)
- [x] Implement the interventions and compare the results
- [x] Check if chat surfaces are willing to use query params when instructed (could be a data exfiltration risk)
  - Both ChatGTP and Claude know and will use our query params, but they can make mistakes/hallucinate and clicking the link asks the user for confirmation, saying the URL might not be safe

Interventions:

- [x] Add instructions for how to download the data/metadata
- [ ] Add instructions for how to download the graph as an image and/or make sure the image is present
- [ ] Add instructions for how to use search to find more data
- [ ] Create a Markdown/txt version of the data page with inline data
  - This could use query params to configure the data table
  - Link to the page in HTML

**Could do**

- [ ] Try JSON-LD with a dataset metadata, which would include data download URLs

**Won't do**

- Embed the data directly in the HTML
  - This would likely make the pages too big

## Technical notes

We recently enabled Markdown for Agents in Cloudflare which might help, but likely only with token efficiency/performance.

PR #6312, opened 2026-04-02
Improve no-JS multi-dim data page

rakyi:
Include a fallback image and the metadata section for the default multi-dim view.

I added dynamic rewriting of the baked download URLs similar to what we do for images, to make sure the image and the URL match. I'm not happy about how hacky it is, and it's increasing our worker CPU usage, but it felt important enough to no confuse humans and especially LLMs. Curious what you think.

Part of #6189

## Testing guidance

You can try disabling JS on multi-dim data pages with various query params.

- [ ] Does the change work in the archive?
  - It does not, archive is out of scope

### DIFF
--- a/functions/_common/grapherTools.ts
+++ b/functions/_common/grapherTools.ts
@@ -1,19 +1,25 @@
 import * as _ from "lodash-es"
 import {
+    CsvDownloadType,
+    type DataDownloadContextBase,
     generateGrapherImageSrcSet,
+    getDownloadUrl,
     Grapher,
     GrapherState,
     loadCatalogData,
 } from "@ourworldindata/grapher"
 import {
+    type DownloadRewriteTarget,
     GrapherInterface,
     MultiDimDataPageConfigEnriched,
     R2GrapherConfigDirectory,
     AdditionalGrapherDataFetchFn,
+    GRAPHER_QUERY_PARAM_KEYS,
 } from "@ourworldindata/types"
 import {
     excludeUndefined,
     Bounds,
+    makeDownloadCodeExamples,
     searchParamsToMultiDimView,
 } from "@ourworldindata/utils"
 import { StatusError } from "itty-router"
@@ -359,6 +365,7 @@ export function rewriteMetaTags(
     let origin = ""
 
     const thumbnailUrl = `${url.pathname}.png${url.search}`
+    const downloadCtxBase = getDownloadContextBase(url)
 
     const rewriter = new HTMLRewriter()
         .on("picture[data-owid-populate-url-params] source", {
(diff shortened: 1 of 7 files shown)


marcelgerber:
@codex review

functions/_common/grapherTools.ts
```diff
@@ -400,10 +406,96 @@ export function rewriteMetaTags(
...
+        ...downloadCtxBase,
+        csvDownloadType: CsvDownloadType.Full,
+        shortColNames: false,
+    })
+    const metadataUrl = getDownloadUrl("metadata.json", {
+        ...downloadCtxBase,
+        csvDownloadType: CsvDownloadType.Full,
+        shortColNames: false,
+    })
+    const codeExamples = makeDownloadCodeExamples(csvUrl, metadataUrl)
```
marcelgerber, line 468: In theory, we could also include a `<span data-owid-download-url-target>` **inside** the code example block, and then only replace the span content rather than the whole code example. Doesn't make a massive difference, but maybe it's a bit nicer and makes the CF code a bit cleaner (or at least doing a bit less)?

Example:
```html
<code class="wp-code-snippet__code">
=IMPORTDATA("<span data-owid-download-url-target="api-csv">https://multi-dim-no-js.owid.pages.dev/grapher/vaccination-coverage-who-unicef.csv?v=1&amp;csvType=full&amp;useColumnShortNames=false&amp;metric=coverage&amp;antigen=comparison&amp;x=1</span>")
</code>
```

functions/_common/grapherTools.ts
```diff
@@ -400,10 +406,96 @@ export function rewriteMetaTags(
...
+    return {
+        slug: url.pathname.split("/").at(-1) ?? "",
+        searchParams,
+        externalSearchParams,
+        baseUrl: `${url.origin}${url.pathname}`,
+    }
+}
+
+function getRewrittenDownloadUrl(
+    target: string,
```
marcelgerber: Could we make target a shared union type between the functions code and the data page code?

functions/_common/grapherTools.ts
```diff
@@ -400,10 +406,96 @@ export function rewriteMetaTags(
...
 
     return rewriter.transform(page)
 }
 
+export function getDownloadContextBase(url: URL): DataDownloadContextBase {
+    const searchParams = new URLSearchParams(url.searchParams)
+    const externalSearchParams = new URLSearchParams()
+    const grapherQueryParamKeys = new Set<string>(GRAPHER_QUERY_PARAM_KEYS)
+
+    for (const [key, value] of url.searchParams.entries()) {
```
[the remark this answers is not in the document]
rakyi: It's not worth it to fetch a multi-dim config just to avoid this edge case.

functions/_common/grapherTools.ts
rakyi: True, but I don't feel it's worth the additional complexity. Concatenating strings should be fast enough.

PR #6333, opened 2026-04-09
Fix author roles not showing for authors without a published author page

edomt:
## Summary
- When an author has a role (e.g. "visualization") but no published author page in the database, the role was silently dropped from the article byline
- The existing code applies roles to `LinkedAuthor` objects fetched from the DB — but authors without a published page aren't in that result, so their roles are lost
- This passes `authorRoles` from the article content directly through the Byline → LinkedAuthor component chain, bypassing the database lookup

## Test plan
- [x] `yarn testFormatChanged` passes
- [x] `yarn typecheck` passes
- [x] All unit tests pass (1800/1800, 1 pre-existing timezone failure)
- [x] Verify on staging that an article with a mix of authors (some with pages, some without) shows all roles correctly

### DIFF
--- a/site/gdocs/components/Byline.tsx
+++ b/site/gdocs/components/Byline.tsx
@@ -1,7 +1,13 @@
 import { Fragment } from "react"
 import LinkedAuthor from "./LinkedAuthor.js"
 
-export const Byline = ({ names }: { names: string[] }) => {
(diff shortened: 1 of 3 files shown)


edomt:
@codex review

PR #6344, opened 2026-04-10
🐛 fix explorer prominent links and thumbnails

ikesau:
The SQL query that resolves explorer configs in the site baker remaps the title and subtitle, so `makeExplorerLinkedChart` was failing, leading to prominent links with no title/subtitle on the baked site. This PR fixes that.

Also updates the code so that prominent links / recircs will use the dynamic explorer thumbnail endpoint instead of the deprecated `thumbnail` field on the explorer program.

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -394,7 +394,15 @@ export class SiteBaker {
                 .getAllPublishedExplorersBySlugCached(knex)
                 .then((results) =>
                     _.mapValues(results, (explorer) => {
-                        return makeExplorerLinkedChart(explorer, explorer.slug)
+                        return makeExplorerLinkedChart(
+                            {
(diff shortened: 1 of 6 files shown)


rakyi (approved):
Good change in general, but I think it's possible that dynamic thumbnails of the big covid explorers currently (almost) always crashes with OOM. At least [it used to](https://github.com/owid/owid-grapher/issues/4577) — could you double-check that?

ikesau:
@rakyi yeah that might still be the case. I just sampled 10 and they all loaded, but I don't think we've changed anything about how they load so idk, I guess I'm just lucky?

There are 2 prominent-links that point to explorers and zero recircs. I think it's okay and the status quo is definitely worse

PR #6345, opened 2026-04-10
🔨 remove old explorer thumbnail code

ikesau:
Removes outdated and mostly unused explorer thumbnail code seeing as we use the dynamic thumbnail endpoint for thumbnail previews now.

### DIFF
--- a/site/ExplorerPage.tsx
+++ b/site/ExplorerPage.tsx
@@ -74,7 +74,7 @@ export const ExplorerPage = (props: ExplorerPageSettings) => {
         urlMigrationSpec,
         archiveContext,
     } = props
-    const { explorerTitle, explorerSubtitle, slug, thumbnail } = program
+    const { explorerTitle, explorerSubtitle, slug } = program
 
     const isOnArchivalPage = archiveContext?.type === "archive-page"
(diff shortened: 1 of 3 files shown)


site/ExplorerPage.tsx
```diff
@@ -125,7 +125,6 @@ window.Explorer.renderSingleExplorerOnExplorerPage(
                 canonicalUrl={`${baseUrl}/${EXPLORERS_ROUTE_FOLDER}/${slug}`}
                 pageTitle={`${explorerTitle} Data Explorer`}
                 pageDesc={explorerSubtitle}
-                imageUrl={thumbnail}
```
rakyi, line 128: Are you saying that `thumbnail` is not set for explorers currently? Does it mean they get the default thumbnail for social media preview? If so, consider using the dynamic thumbnail for those or creating an issue for that.

rakyi (approved):

site/ExplorerPage.tsx
ikesau: We already use the dynamic thumbnail for explorers.

Image

The baked HTML will have the default thumbnail but it will be overwritten when served.

PR #6346, opened 2026-04-10
🐛 fix CaptionedChart chartHeight calculation

ikesau:
I've added a few more directives recently and one of them wasn't factored into the `chartHeight` calculation in CaptionedChart.

This ties it all together so that if title, subtitle, and logo are all hidden, the `VerticalSpace` won't render and `header.height` will be 0.

### DIFF
--- a/packages/@ourworldindata/grapher/src/header/Header.tsx
+++ b/packages/@ourworldindata/grapher/src/header/Header.tsx
@@ -288,7 +288,9 @@ abstract class AbstractHeader<
         )
     }
 
-    override render(): React.ReactElement {
+    override render(): React.ReactElement | null {
+        const hasContent = !!this.logo || this.showTitle || this.showSubtitle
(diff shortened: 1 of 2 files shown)


ikesau:
@codex review

packages/@ourworldindata/grapher/src/header/Header.tsx
```diff
@@ -288,7 +288,9 @@ abstract class AbstractHeader<
         )
     }
 
-    override render(): React.ReactElement {
+    override render(): React.ReactElement | null {
+        const hasContent = !!this.logo || this.showTitle || this.showSubtitle
+        if (!hasContent) return null
```
sophiamersmann, line 293: I think it makes sense to do this for the `StaticHeader` below as well

sophiamersmann (approved):

PR #6358, opened 2026-04-14
Exclude thumbnail pageviews from static image view counts

veronikasamborska1994:
## Context

Some older articles use static images as thumbnails in R&W blocks. This is an outdated practice but it significantly inflates view counts for a small number of affected images. For example, `Famine-victims-and-world-population-since-1860.png` currently accumulated views from the Population Growth topic page simply because it appears there as a small, unreadable thumbnail. These are not meaningful views.

Although we no longer do this, these instances still exist and we should exclude them so we can correctly identify our most viewed static images. 

This PR makes them easier to identify and excludes their views from the pageview count. 

## What this PR does

1. Adds a `context` column (`'content'` | `'article-thumbnail'`) to `posts_gdocs_x_images`, recording how each image is used per article. The same image can be `'content'` in one article and `'article-thumbnail'` in another.
2. `views_365d` now only sums pageviews where `context = 'content'`. Featured/cover images are also classified as `'article-thumbnail'`.
3. Adds a **"Used in both content and thumbnail"** filter to the admin images page so we can easily find these cases.

## Testing guidance

1. Go to `/admin/images`
2. Images with "thumbnail" in their name (e.g. `risk-ratios-thumbnail.png`) should appear only under **Thumbnails or featured images** with **Views per day** of 0
3. `Population-cartogram_World.png` — used as inline content on its own article but also as an R&W thumbnail on Population Growth — should appear under **Content** but now the views should reflect only its own article (much lower in many cases as in admin). A few more examples to check:
   - `Population-Pyramid-1950-to-2100.jpg`
   - `Emissions-by-sector-–-pie-charts.png`
   - `Demographic-Impact-of-China's-Great-Leap-Forward-Famine.png`

4. **"Used in both content and thumbnail"** filter should show the same images as above genuine content in one article but a thumbnail reference in another. This filter is also useful for finding instances where we may want to replace the thumbnails on those articles. 

Note: the `'article-thumbnail'` classification also covers featured images on author pages, which are not really the problem we're solving here. It's unclear whether this is worth solving given there are relatively few such cases. 

This code was largely written by Claude. I reviewed it by looking for edge cases and verifying the exclusions, but a thorough review would be appreciated and any improvements are welcome.

### DIFF
--- a/adminSiteClient/ImagesIndexPage.tsx
+++ b/adminSiteClient/ImagesIndexPage.tsx
@@ -51,17 +51,13 @@ type ImageMap = Record<string, DbEnrichedImageWithPageviews>
 
 type ImageTypeFilter =
     | "all"
-    | "featured-thumbnail-rw"
+    | "article-thumbnail"
     | "content"
     | "content-and-thumbnail"
     | "other"
 
 function isUsedAsThumbnail(image: DbEnrichedImageWithPageviews): boolean {
-    return (
-        image.isFeaturedImage === 1 ||
-        image.filename.toLowerCase().includes("thumbnail") ||
-        image.isInResearchAndWriting === 1
-    )
+    return image.isArticleThumbnail === 1
 }
 
 function getImageType(image: DbEnrichedImageWithPageviews): ImageTypeFilter {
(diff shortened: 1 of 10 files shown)


rakyi (commented):
The overall idea makes sense to me. I pushed a commit with a few fixes.

The code is pretty complicated, and I'm not so deep into images — @ikesau could you review it as well?

PR #6368, opened 2026-04-15
Backfill and constrain createdAt/updatedAt timestamps

rakyi:
## Context

As discussed on a dev meeting recently, we can tighten these types in the DB. It also fixes a few small issues where we assume `updatedAt` can't be null.

## Testing guidance

Mostly a sanity check on the changes

## Checklist

If DB migrations exists:

- [x] If columns have been added/deleted, all necessary views were recreated and ETL and Analytics team members have been informed of the incoming changes
- [x] The DB types in the ETL have been updated


mlbrgl (approved):
makes sense!  
  
Should we constrain `updatedAt`  to be NOT NULL for all cases? 👇 (an exception would have to be made for `files` as it doesn't have any `updatedAt` column)

| Invariant | Group 1 createdAt | Group 1 updatedAt | Group 2 createdAt | Group 2 updatedAt | Group 3 createdAt | Group 3 updatedAt |
| --- | --- | --- | --- | --- | --- | --- |
| Cannot be NULL | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ ⚠️ |
| Has a sensible DEFAULT | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| Auto-bumped on row update | ❌ (correct) | ✅ | ❌ (correct) | ❌ (intentional — app-managed) | ❌ (correct) | ✅ |
| Schema type uniform with siblings | ✅ DATETIME | ✅ DATETIME | ✅ DATETIME | ✅ DATETIME | ✅ TIMESTAMP | ✅ TIMESTAMP |

Group 1: nullableAutoUpdatedUpdatedAtTables
Group 2 manuallyManagedUpdatedAtTables
Group 3: nullableTimestampCreatedAtTables

rakyi:
Good catch! I somehow missed those. Added them to the migration and fixed a few more related issues.

rakyi:
@codex review

rakyi:
@codex review

db/migration/1776260942258-BackfillAndConstrainTimestamps.ts
```diff
@@ -0,0 +1,163 @@
...
+                WHERE createdAt IS NULL`
+            )
+        }
+
+        await queryRunner.query(
+            `-- sql
+            UPDATE explorers
+            SET createdAt = CURRENT_TIMESTAMP
+            WHERE createdAt IS NULL`
+        )
```
mlbrgl, line 64: any particular reason to not touch explorers.updatedAt?

mlbrgl (approved):
thanks for tightening this up!

db/migration/1776260942258-BackfillAndConstrainTimestamps.ts
rakyi: Just another miss, good catch! Added.

Issue #5951, opened 2026-01-20
Replace redirected chart URLs in gdocs at creation time


rakyi:
## Core problem

Currently, chart/multi-dim redirects affect the following facets of our site:

- HTTP redirects in the browser
- **URLs in ArchieML blocks**
  - Rendering the gdoc
  - Archive dependency/change tracking for gdoc posts
- Page views in Algolia score calculation

Redirects for URLs in ArchieML blocks are currently resolved at the time of use, i.e. during baking and archiving. This adds a lot of complex code also means we have to repeat the resolution calculation every time.

For HTTP and Algolia score calculation, resolving redirects must be done at runtime (i.e. time of use).

## Proposed solution

Replace URLs in gdocs with their fully resolved target after a redirect is created. For each chart type, we should be able to query for all gdocs that use its URL and then replace it using the gdocs API, e.g. as a string search-and-replace.

Deleting a redirect would mean to do the same operation in reverse — change the URLs in gdocs to the source of the redirect before it's deleted.

This is not fully reversible in the sense that some of the gdocs might have already used the target URL and thus reverting the change after a redirect is deleted would change them as well, but that's unlikely to be a problem in practice.

Or we would disallow deleting a redirect that's in use.

## Alternative

We don't strictly need to go all the way back to gdocs. We could instead resolve the redirects before we solve the gdoc ArchieML content into DB and keep the original URLs in the actual gdoc.

## Context

In either case, updating the URLs in the ArchieML would trigger re-archiving those articles (and other pages?) even though they didn't meaninfully change.

This came up in https://github.com/owid/owid-grapher/issues/5211 when I realized how much complexity are redirects responsible for.

rakyi:
Note, we should at the latest do this the next time we are substantially touching the redirects code.

rakyi:
Next steps: validate that the proposed solution (with updating actual gdocs) is workable

- possible witht the gdocs API
- not too slow
- we don't loose valuable metadata
- it doesn't break if someone else has the gdoc open at the time of replacement

Based on that decide which path to take.

marcelgerber:
Martin and I have talked about this in the context of the explorer repotting project, where we'd migrate every existing explorer to an mdim, and want good redirects for that.

PR #6371, opened 2026-04-15
🐛 Resolve multi-dim redirects in SiteBaker prefetch step

ikesau:
## Context

The prefetcher wasn't interacting with `multi_dim_redirects`, so gdoc charts using URLs that had since been redirected to multi-dims were broken.

This PR fixes that by also resolving redirects in `getAllLinkedPublishedMultiDimDataPages` :

1. We find all multi-dim links in the `posts_gdocs_links` table
2. Resolve the corresponding multi-dim configs, keyed by their slugs
3. Select all multi-dim redirects whose `source` column matches a `posts_gdocs_link` target
4. Resolve the corresponding multi-dim configs, keyed by the original slug

This means that a gdoc that links to `/grapher/some-old-slug` will still resolve to `/grapher/the-new-multi-dim-slug`

But now I really want to work on https://github.com/owid/owid-grapher/issues/5951 next 😫 

## Testing guidance

Possibly fastest to test locally:
1. (On master) publish a test document with a `chart` component that links to a grapher
2. `yarn buildLocalBake --steps gdocPosts` and `yarn startLocalBakeServer`
4. Visit http://localhost:3000/your-test-article-slug and confirm the grapher renders
5. Create a redirect for that slug to some multi-dim in http://localhost:3030/admin/multi-dim-redirects
6. `yarn buildLocalBake --steps gdocPosts` and `yarn startLocalBakeServer`
7. Visit http://localhost:3000/your-test-article-slug and confirm that the chart doesn't resolve (it won't appear; you can also check `_OWID_GDOC_PROPS.linkedCharts` in the browser console)
8. Checkout this branch
9. `yarn buildLocalBake --steps gdocPosts` and `yarn startLocalBakeServer`
10. Visit http://localhost:3000/your-test-article-slug and confirm the chart resolves (It won't load properly because the localBakeServer doesn't serve multi-dim configs, but an empty Grapher component will appear and it will exist on `_OWID_GDOC_PROPS.linkedCharts`)

- [x] Does the change work in the archive? **I think the archive system works correctly**

### DIFF
--- a/baker/SiteBaker.tsx
+++ b/baker/SiteBaker.tsx
@@ -471,11 +471,19 @@ export class SiteBaker {
             }
 
             const multiDims = await getAllLinkedPublishedMultiDimDataPages(knex)
-            for (const { id, slug, config } of multiDims) {
+            for (const {
+                id,
+                slug,
+                config,
+                originalSlug,
+                queryStr,
+            } of multiDims) {
                 publishedCharts.push(
-                    makeMultiDimLinkedChart(config, slug, {
+                    makeMultiDimLinkedChart(config, originalSlug, {
                         archivedPageVersion:
                             archivedVersions.multiDims[id] || undefined,
+                        queryStr,
+                        resolvedSlug: originalSlug !== slug ? slug : undefined,
                     })
                 )
             }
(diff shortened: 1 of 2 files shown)


rakyi (approved):
It's so ugly, right? Thanks for the fix though!

Have you considered using `getMultiDimRedirectTargets` in `getAllLinkedPublishedMultiDimDataPages`? Not sure if it would be worth it, your version is good.

ikesau:
@rakyi yeah - considered it, but it would have required more plumbing to resolve the configs I think.

PR #6401, opened 2026-04-23
Improve GrapherWithFallback logic

rakyi:
- Show fallback image only when JS is disabled
- Otherwise, render Grapher, which has its own loading indicator

This should improve the UX especially on data pages, where the image previously flashed for a brief moment before the Grapher started loading.

## Context

[Slack](https://owid.slack.com/archives/C46U9LXRR/p1775055275597579)

## Testing guidance

Check that the behavior is correct in data pages and articles.

  - Yes, but we don't load fallback images there.

### DIFF
--- a/site/GrapherWithFallback.tsx
+++ b/site/GrapherWithFallback.tsx
@@ -1,4 +1,4 @@
-import { GRAPHER_PREVIEW_CLASS } from "@ourworldindata/types"
+import { HIDE_IF_JS_ENABLED_CLASSNAME } from "@ourworldindata/types"
 import { GrapherFigureView } from "./GrapherFigureView.js"
 import cx from "classnames"
 import GrapherImage from "./GrapherImage.js"
@@ -19,9 +19,7 @@ export interface GrapherWithFallbackProps {
     isPreviewing?: boolean
 }
 
-export function GrapherWithFallback(
-    props: GrapherWithFallbackProps
-): React.ReactElement {
+export function GrapherWithFallback(props: GrapherWithFallbackProps) {
     const { slug, className, id, config, queryStr, isPreviewing } = props
 
     if (!slug && !props.configUrl) {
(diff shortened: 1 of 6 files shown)


site/blocks/RelatedCharts.test.tsx
rakyi, line 1: We no longer render the figure with fallback image by default, so we have to adjust the test.

rakyi:
@codex review

rakyi:
@codex review

site/GrapherWithFallback.tsx
```diff
@@ -85,8 +56,34 @@ export function GrapherWithFallback(
...
+                        "chart",
+                        "GrapherWithFallback__fallback",
+                        className
+                    )}
+                    aria-hidden={isClient}
+                >
+                    {!isClient &&
+                        (props.configUrl ? (
+                            <GrapherImage
+                                className={HIDE_IF_JS_ENABLED_CLASSNAME}
```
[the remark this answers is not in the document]
rakyi: This is an acceptable edge case.

rakyi:
@codex review

site/GrapherWithFallback.tsx
marcelgerber, line 1: There is an issue going on where when switching between related charts, it shows a loading indicator, then shows the chart from before, and only then loads the new chart.
This is actually also somewhat of an issue on prod right now, but it's not quite as visible there.

This is on http://staging-site-fix-grapher-fallback/life-expectancy#all-charts.

https://github.com/user-attachments/assets/4ae27b50-ddf0-407c-9bc4-efebb5550872

marcelgerber (approved):
Thank you! Yeah, it's nice that we now have less of a layout jump upon load.

Sadly, I found an issue - it's actually also there on prod and we should fix it either way, but it's now become more pronounced.

site/GrapherWithFallback.tsx
rakyi: Good catch! Should be fixed now.

PR #6403, opened 2026-04-24
Switch from MIT to custom transparent license

danyx23:
This PR switches our license to a bespoke license that leaves the code open for transparency reasons but does not allow arbitrary re-use. We intend to grant rights to projects that are interested in re-using our code, but we want to have better visibility and reserve the right to reject re-use scenarios that are not in line with our mission.


danyx23:
* **#6403** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

.github/workflows/cla.yml
```diff
@@ -0,0 +1,74 @@
...
+#      branch (Settings → Actions → General → Workflow permissions).
+#   3. If the repo's default workflow permissions are read-only, they will
+#      be overridden by the explicit `permissions:` block below.
+#   4. Verify the allowlist below — add any missing OWID staff or bot
+#      accounts that legitimately push changes.
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
```
rakyi: Coincidentally, just yesterday I removed the one usage of `pull_request_target` we had due to it being considered dangerous.

Could we achieve what we need with just `pull_request` as suggested in [this article](https://astral.sh/blog/open-source-security-at-astral)? Otherwise, we could implement the workflow in a custom GitHub App (also as suggested in the article), i.e. the existing owidbot or a new one.

.github/workflows/cla.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+  CLAAssistant:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "CLA Assistant"
+        if:
+          (github.event.comment.body == 'recheck' || github.event.comment.body == 'I
+          have read the Contributor Agreement and I hereby sign it') ||
+          github.event_name == 'pull_request_target'
+        uses: contributor-assistant/github-action@v2.6.1
```
rakyi: For security, we should depend on a specific commit SHA, since tags are mutable. (We should switch the existing action dependencies to that pattern as well, but that's out of scope here.)

See [the same article as above](https://astral.sh/blog/open-source-security-at-astral) for rationale.

Also, please read the source code of the action at the exact commit to make sure it is safe. Especially important since we give the job very broad `permissions`.

.github/workflows/cla.yml
danyx23: Thank you for the review, Martin. I've decided to remove the automation for now. The CAA is not yet finalized, and we only very rarely get third party PRs, so it's probably best to handle this manually. I will merge the license change now and we can reconsider how best to add automation around this later.

PR #6409, opened 2026-04-26
Prevent Grapher initial bounds flash

rakyi:
Measure embedded Grapher figures before mounting and sync external bounds before paint, so data page and GDoc embeds don't briefly render at the default 850px width.

### DIFF
--- a/packages/@ourworldindata/grapher/src/hooks.ts
+++ b/packages/@ourworldindata/grapher/src/hooks.ts
@@ -10,6 +10,7 @@ import {
     type DataDownloadContextBase,
     type DataDownloadContextServerSide,
 } from "./download.js"
+import { useIsomorphicLayoutEffect } from "usehooks-ts"
 
 /**
  * Like useState, but clearing (setting to undefined) is debounced.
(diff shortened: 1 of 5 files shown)


packages/@ourworldindata/grapher/src/hooks.ts
```diff
@@ -91,20 +92,53 @@ export function useDataApiDownloadConfig({
...
+    ref: RefObject<HTMLElement | null>,
+    ...options:
+        | []
+        | [initialValue: Bounds, throttleTime?: number]
+        | [initialValue: undefined, throttleTime?: number]
+): Bounds | undefined {
+    const initialValue =
+        options.length === 0 ? DEFAULT_GRAPHER_BOUNDS : options[0]
+    const throttleTime = options[1] ?? 100
+    const [bounds, setBounds] = useState<Bounds | undefined>(initialValue)
```
marcelgerber: This here doesn't have to be this complicated; this solution that uses generics also works (we need to pass `null` rather than `undefined`, but that's easy enough to do):
```suggestion
//
// The `T` type parameter narrows the return type based on `initialValue`:
// passing `null` widens the return to `Bounds | null` (so callers can detect
// the pre-measurement state); omitting it keeps the return as `Bounds`.
export function useElementBounds<T extends Bounds | null = Bounds>(
    ref: RefObject<HTMLElement | null>,
    initialValue: T = DEFAULT_GRAPHER_BOUNDS as T,
    throttleTime: number = 100
): Bounds | T {
    const [bounds, setBounds] = useState<Bounds | T>(initialValue)
```

marcelgerber (approved):
Great work, this has (slightly) bothered me for a good while now!

packages/@ourworldindata/grapher/src/hooks.ts
rakyi: That's much nicer, thanks!

rakyi:
@codex review

PR #6411, opened 2026-04-27
Always hide controls of charts in ChartStory

rakyi:
Even if the user doesn't specify it manually, we hide the controls (dropdowns of explorers and multi-dims) since they compete with ChartStory's own controls (arrows).  
  
[Slack](https://owid.slack.com/archives/C03NV9Z3YSV/p1776718193993799?thread_ts=1776624040.140859&cid=C03NV9Z3YSV)

PR #6394, opened 2026-04-20
Fix multi dim chart story

rakyi:
## Context

This fixes the chart story arrows not working for multi-dims. Expect follow-up fixes in future PRs.

[Slack](https://owid.slack.com/archives/C03NV9Z3YSV/p1776624040140859)

## Testing guidance

Check that the chart story arrows work with multi-dims (e.g. second chart story) on http://localhost:3030/sdgs/quality-education

- [x] Does the change work in the archive?
  - Archive doesn't have the bug

PR #6412, opened 2026-04-27
Simplify article chart sizing

rakyi:
- Use a shared article chart frame for Graphers, Explorers, and multi-dims.
- Let multi-dims fill their parent while the frame owns embed sizing. This fixes a bug where multi-dims could have extreme height in a
ChartStory.
- Adjust public iframe snippets so visible multi-dim and Explorer controls still leave about 600px for the Grapher.

## Context

[Slack ](https://owid.slack.com/archives/C03NV9Z3YSV/p1776624040140859)

## Testing guidance

In theory, check that all internal and external embeds of all chart types work correctly. I checked most of them already. Biggest difference is the ChartStory usage in SDGs, where we fix the sizing of normal graphers and multi-dims.

## Checklist

### Before merging

### DIFF
--- a/site/gdocs/components/Chart.tsx
+++ b/site/gdocs/components/Chart.tsx
@@ -11,6 +11,9 @@ import { MultiDimEmbed } from "../../MultiDimEmbed.js"
 import { useEmbedChart } from "../../hooks.js"
 import { useDocumentContext } from "../DocumentContext.js"
 
+const DEFAULT_CHART_HEIGHT = "575px"
+const CHART_WITH_EXTERNAL_CONTROLS_HEIGHT = "700px"
+
 export default function Chart({
     d,
     className,
@@ -75,6 +78,7 @@ export default function Chart({
         resolvedUrlParsed.queryParams.hideControls !== "true"
     const isExplorerWithControls = isExplorer && controlsAreVisible
     const isMultiDimWithControls = isMultiDim && controlsAreVisible
+    const hasExternalControls = isExplorerWithControls || isMultiDimWithControls
 
     const archivedChartVersion = linkedChart?.archivedPageVersion
     const shouldRenderArchiveEmbed =
@@ -120,7 +124,9 @@ export default function Chart({
 
     if (shouldRenderArchiveEmbed && archiveUrl) {
         const defaultHeight =
-            isMultiDimWithControls || isExplorerWithControls ? "680px" : "600px"
+            isExplorerWithControls || isMultiDimWithControls
+                ? CHART_WITH_EXTERNAL_CONTROLS_HEIGHT
+                : DEFAULT_CHART_HEIGHT
         return (
             <div
                 className={cx(className, {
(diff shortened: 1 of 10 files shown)


rakyi:
* **#6412** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6412?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6412?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6411** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6411?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6394** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6394?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

site/multiDim/hooks.ts
```diff
@@ -28,6 +28,7 @@ export const useBaseGrapherConfig = (
             adminBaseUrl: ADMIN_BASE_URL,
             dataApiUrl: DATA_API_URL,
             canHideExternalControlsInEmbed: true,
+            recommendedIframeEmbedHeight: 674,
```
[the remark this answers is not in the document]
rakyi: It works as is.

site/gdocs/components/Chart.tsx
```diff
@@ -158,45 +159,48 @@ export default function Chart({
...
-                    config={chartConfig}
-                    queryStr={queryStr}
-                    isEmbeddedInAnOwidPage={true}
-                    isEmbeddedInADataPage={false}
-                    isPreviewing={isPreviewing}
-                />
-            )}
+            <div
+                className={cx("owid-chart-frame", {
+                    "owid-chart-frame--with-controls": hasExternalControls,
```
mlbrgl: this is only having an effect for explorers (given that the single CSS property - height - is overriden in the mdim case), so it might be clearer to have mutually exclusive classes here, one for mdim, one for explorers.

site/gdocs/components/centered-article.scss
```diff
@@ -880,8 +899,36 @@ div.raw-html-table__container {
...
+    }
+
+    .ExplorerFigure {
+        min-height: 0;
+        max-height: none;
+    }
+}
+
+.owid-chart-frame--with-controls {
+    height: 680px;
```
mlbrgl: explorers break out of this through https://github.com/owid/owid-grapher/blob/2b954c31d8643efe2465c29612cc74a38d304044/site/gdocs/components/Chart.scss#L10

mlbrgl (approved):
Sweet!

It would be nice to add explanations about how these height numbers relate to other hard coded dimensions (and/or how they were calculated), so we surface those implicit dependencies.

I had claude create a three-way comparison between staging, prod and staging-archive to see the impact on sites using the current (height: 600px) and future (600/674/696px) embed codes. Everything lines up 👍 

Image

Source: [test-embeds-pr-6412.html](https://github.com/user-attachments/files/27201568/test-embeds-pr-6412.html) (open locally behind tailscale for testing)

To note: I couldn't repro the long chart failure mode on live (the test case in slack, on SDG pages).

site/gdocs/components/centered-article.scss
rakyi: True, moved to a single place.

site/gdocs/components/Chart.tsx
rakyi: Good idea, changed.

rakyi:
I think Claude messed something up in the comparison. Some charts were incorrectly set to 600px and some to 575px, but that didn't surface in the comparison. I now hopefully fixed everything to be 575px in articles (excluding controls), except explorers with controls where I kept the 700px.

rakyi:
Actually, I just misunderstood the comparison. It was comparing only iframes, which were all correct.

PR #6413, opened 2026-04-27
Convert a few admin components to vanilla React

rakyi:
I dipped my toes a bit into this to get a basic sense of how we could convert MobX class components to vanilla functional React components. These should all be low risk.


rakyi:
@codex review

sophiamersmann (approved):

PR #6416, opened 2026-04-28
🔨🤖 Restrict fetchAllWork to article/topic-page types and include titles

mlbrgl:
## Summary
- Restrict the `/api/all-work` endpoint to `article`, `topic-page`, and `linear-topic-page` (previously excluded `data-insight` and `fragment` only, which let through unrelated types like `about-page`, `author`, `homepage`, `announcement`, etc.)
- Pull the gdoc title from `content ->> '$.title'` and emit it as a `:skip` / `:endskip` ArchieML comment above each url, so the output is easier to scan when populating the `[.secondary]` section of an author page

## Context
The endpoint is used by editors to bulk-populate the `[.secondary]` list of an author's *Research & writing* block (alternate template, which highlights topics rather than articles). Previously the output was a flat list of urls with no titles, and included gdoc types that don't belong in that section (e.g. data insights were excluded but about-pages and announcements still slipped through). With this change the editor gets a titled, scannable list limited to articles and topic pages, ready to paste in.

## Test plan
- [x] Hit the endpoint for an author ([example](http://staging-site-restrict-fetch-all-work-type/admin/api/all-work?author=Pablo%20Rosado)) and confirm the output now contains `:skip` / `:endskip` blocks with titles, and excludes about/author/announcement/etc. types

### DIFF
--- a/adminSiteServer/apiRoutes/misc.ts
+++ b/adminSiteServer/apiRoutes/misc.ts
@@ -1,31 +1,38 @@
-// Get an ArchieML output of all the work produced by an author. This includes
-// gdoc articles, gdoc modular/linear topic pages and wordpress modular topic
-// pages. Data insights are excluded. This is used to manually populate the
-// [.secondary] section of the {.research-and-writing} block of author pages
-
-import { DbRawPostGdoc, JsonError } from "@ourworldindata/types"
+import {
+    DbRawPostGdoc,
+    JsonError,
+    OwidGdocPostContent,
+} from "@ourworldindata/types"
 
 import * as db from "../../db/db.js"
 import * as lodash from "lodash-es"
 import { expectInt } from "../../serverUtils/serverUtil.js"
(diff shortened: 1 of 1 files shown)


rakyi (approved):
The change makes sense!

However, I wonder why do we copy-paste ArchieML around instead of generating the "Research & writing" block automatically. Is it because there is some editorial step where someone has to decide which of the generated URLs exclude?

mlbrgl:
@rakyi thanks!

> Is it because there is some editorial step where someone has to decide which of the generated URLs exclude?

yes that was to give full editorial control on the curation and ordering for a wider definition of authors (beyond those who regularly publish articles), without being too prescriptive early on. This could now be reevaluated in light of how author pages have effectively been used since launch, but the gains are probably marginal. Let's see what reception we get from announcing this slight update - maybe it'll trigger some conversations.

PR #6417, opened 2026-04-28
Trigger Buildkite deploys directly

Marigold:
## Motivation

Buildkite is now the deploy orchestrator for these environments, so we should not keep a second custom deploy orchestrator in front of it. The legacy .queue/.pending files made sense for local baking, but in Buildkite-backed deploys they can fight with Buildkite's own queueing/skipping behavior. In particular, when Buildkite restarts services, restarting deploy-queue can make the legacy pending-file recovery trigger another Buildkite build.

## Summary

- Trigger Buildkite deploy-content builds directly from admin deploy requests when Buildkite is configured.
- Read unfinished Buildkite builds for the admin deploy status page instead of relying on local queue files in Buildkite-backed environments.
- Make the deploy queue worker a no-op in Buildkite-backed environments.
- Keep the legacy .queue/.pending deploy worker path for non-Buildkite/local baking only.

## Testing

- yarn test run --reporter dot baker/BuildkiteTrigger.test.ts baker/DeployQueueServer.test.ts
- yarn testLintChanged
- yarn testFormatChanged

I attempted ./db/tests/run-db-tests.sh locally, but Docker is not running in this environment, so I used the CI failure logs to fix the import/coupling issue instead.

### DIFF
--- a/adminSiteServer/apiRouter.ts
+++ b/adminSiteServer/apiRouter.ts
@@ -2,6 +2,7 @@
 
 import { TaggableType } from "@ourworldindata/types"
 import { DeployQueueServer } from "../baker/DeployQueueServer.js"
+import { BuildkiteTrigger } from "../baker/BuildkiteTrigger.js"
 import {
     updateVariableAnnotations,
     getChartBulkUpdate,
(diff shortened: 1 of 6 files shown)


Marigold:
It'd be nice cleanup, but likely not worth it at the moment.

PR #6418, opened 2026-04-28
Fix DoDs of the same level overlapping

### LINKED ISSUE #5917: DoDs can overlap eachother when moused over quickly
From this [explorer](https://ourworldindata.org/explorers/co2?tab=map&Gas+or+Warming=CO%E2%82%82&Accounting=Territorial&Fuel+or+Land+Use+Change=All+fossil+emissions&Count=Per+capita)

If you're mousing over a "base level" DoD (i.e. not a nested one) other DoDs should disappear.

Recursive DoDs should still work, only multiple DoDs on the same level should not be possible.

https://github.com/user-attachments/assets/976d5457-4a2b-44ff-b8a9-45aac4a5adf3

rakyi:
Nested DoDs should still work.

Fixes #5917

### DIFF
--- a/site/detailsOnDemand.tsx
+++ b/site/detailsOnDemand.tsx
@@ -56,6 +56,8 @@ export async function runDetailsOnDemand(
         },
     }).then((res) => res.json())
 
+    const dodDepthsByInstance = new Map<Instance, number>()
+
     document.addEventListener("mouseover", handleEvent, { passive: true })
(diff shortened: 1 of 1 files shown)


ikesau (approved):
Nice! Fixes the issue on the CO2 Explorer

Also doesn't break dods that are infinitely recursive e.g. https://ourworldindata.org/grapher/global-temperature-anomalies-by-el-nino-la-nina-and-month

PR #6438, opened 2026-04-30
Switch useTopicTagGraph to useQuery

rakyi:
Avoids duplicate requests from two call sites and duplicating the data on the application level.


mlbrgl (approved):
beautiful! 🤩

PR #6445, opened 2026-05-05
Mirror analytics fixes from cloudflare-workers PR #27

Marigold:
## Summary

Ports the GA4 analytics fixes from owid/cloudflare-workers#27 so the two mirrored `analytics.ts` files stay in sync.

The headline change is namespacing Grapher's known query params under `q_*` in `cf_function_invocation` events. Previously, the merge loop in `getCommonEventParams` would let any URL query-string key overwrite the structured params we'd just built — most importantly, `?country=DE` (a real Grapher param used by chart embeds, thumbnails, and downloads) clobbered the `country` field sourced from `cf-ipcountry`. The `country` column in dbt's `raw_traffic` / `worker_traffic` has been silently picking up these client-supplied values for some time.

Other query params pass through bare, except for collisions with structured names (`host`, `pathname`, `referrer`, etc.) and `status_code` (added in `analyticsMiddleware`). Reserved set is built from `Object.keys(params)` so it stays in sync if structured params are added/removed later. Grapher key list comes from `GRAPHER_QUERY_PARAM_KEYS` in `@ourworldindata/types` — same source-of-truth used by `functions/_common/urlTools.ts`.

Other changes bundled in:

- Add a `host` param so the GA4 event records which hostname served the request. Drop the `pathname` parameter from `getCommonEventParams` — it's derivable from `request.url` and was always passed by the single call site as such.
- Switch from the hand-rolled `parseCookies` in `cookieTools.ts` to `parseCookie` from the [cookie](https://github.com/jshttp/cookie) package. Bumps the functions workspace dep to `^1.1.1` (the `parse` export is deprecated upstream in favor of `parseCookie`).
- Add the reverse-link header comment pointing at the cloudflare-workers mirror, so future readers know to update both.

## Test plan

- [ ] Functions still build (`yarn build:functions` / Cloudflare Pages preview).
- [ ] Smoke-test a Grapher chart URL with `?country=DE&tab=chart` on a preview deploy and inspect the GA4 MP request body — expect `q_country: "DE"`, `q_tab: "chart"`, structured `country` from `cf-ipcountry`, structured `host` matching the deploy URL.
- [ ] Confirm `_ga` cookie parsing still extracts the client_id correctly via the new `parseCookie` import.

## Notes for data-eng

The `country` column in `raw_traffic` / `worker_traffic` will start showing the real `cf-ipcountry` value consistently once this lands (today it's overwritten by `?country=` for Grapher embed traffic). Worth flagging to anyone who has been treating that column as authoritative.


functions/_common/analytics.ts
```diff
@@ -1,5 +1,12 @@
+// This file is mirrored in cloudflare-workers/utils/analytics.ts so the
+// proxy workers emit the same GA4 events. Keep the two in sync; see
+// https://github.com/owid/cloudflare-workers/issues/26 for the rationale
+// against extracting a shared library.
+
 import * as Sentry from "@sentry/cloudflare"
-import { extractClientIdFromGACookie, parseCookies } from "./cookieTools.js"
```
rakyi, line 2: I was hoping we could delete the old helper altogether. Could you change the other call sites as well?

functions/_common/analytics.ts
rakyi, line 68: Should we also add `sampling` to be on par with the workers format?

functions/_common/analytics.ts
```diff
@@ -64,9 +72,19 @@ export function getCommonEventParams(request: Request, pathname: string) {
...
+    // Grapher params get namespaced under q_* so a request like `?country=DE`
+    // doesn't clobber the structured `country` (sourced from cf-ipcountry).
+    // Other query params pass through bare, except for collisions with
+    // structured names (host/pathname/etc.) and `status_code` (added by the
+    // caller in analyticsMiddleware below).
+    const grapherKeys = new Set<string>(GRAPHER_QUERY_PARAM_KEYS)
+    const reserved = new Set([...Object.keys(params), "status_code"])
+    for (const [key, value] of url.searchParams) {
+        if (grapherKeys.has(key)) {
+            params[`q_${key}`] = value.slice(0, 100)
```
rakyi: As discussed, we can prefix all the query params instead.

functions/_common/analytics.ts
Marigold: Killed.

functions/_common/analytics.ts
Marigold: Done (same in cloudflare-workers repo)

rakyi (approved):
LGTM!

PR #6460, opened 2026-05-06
🐛 center narrative chart in articles

mlbrgl:
## Summary

- Wrap the narrative-chart figure in `.owid-chart-frame` so it inherits the same centering rules as regular `Chart` embeds.
- Without the wrapper, the live `.GrapherComponent` kept its default `inline-block` ([grapher.scss:98-99](https://github.com/owid/owid-grapher/blob/master/packages/%40ourworldindata/grapher/src/core/grapher.scss#L98-L99)) and anchored at the left edge of the column. The `.owid-chart-frame .GrapherComponent { display: block; margin: 0 auto }` rule in `Chart.scss` now centers it the same way it does for regular charts.

## Pages with narrative charts to compare

Each of these articles uses both narrative charts and regular grapher embeds, so you can verify centering and that the two block types visually agree:

- https://narrative-chart-centering.owid.pages.dev/why-do-women-live-longer-than-men
- https://narrative-chart-centering.owid.pages.dev/democratic-rights
- https://narrative-chart-centering.owid.pages.dev/what-is-the-gini-coefficient

## Test plan

- [ ] On staging, confirm narrative charts on the three articles above are horizontally centered in their grid column (both before JS hydration, briefly visible, and after, when the live Grapher mounts).
- [ ] On the same articles, confirm regular Chart embeds, explorers, and any multi-dim charts are unchanged — they already use `.owid-chart-frame`, so this should be a no-op for them.
- [ ] Check at desktop and mobile widths (the outer wrapper still has `full-width-on-mobile`).

### DIFF
--- a/site/gdocs/components/NarrativeChart.tsx
+++ b/site/gdocs/components/NarrativeChart.tsx
@@ -79,31 +79,31 @@ export default function NarrativeChart({
             })}
             ref={refChartContainer}
         >
-            <figure
-                key={metadataStringified}
-                className={cx(GRAPHER_PREVIEW_CLASS, "chart")}
-                style={{
-                    width: "100%",
-                    border: "0px none",
-                    height: d.height,
-                }}
-                // MultiEmbedder should not kick in on archival pages.
-                {...(!isOnArchivalPage && {
-                    [GRAPHER_NARRATIVE_CHART_CONFIG_FIGURE_ATTR]:
-                        metadataStringified,
-                })}
+            <div
+                className="owid-chart-frame"
+                style={d.height ? { height: d.height } : undefined}
             >
-                <a href={linkTarget}>
-                    <img
-                        className="GrapherImage"
-                        src={imageSrc}
-                        alt={viewMetadata.title}
-                        width={DEFAULT_GRAPHER_WIDTH}
-                        height={DEFAULT_GRAPHER_HEIGHT}
-                        loading="lazy"
-                    />
-                </a>
-            </figure>
+                <figure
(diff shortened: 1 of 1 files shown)


rakyi (approved):
I think the comment from Codex is likely not a real issue, but worth double-checking.

mlbrgl:
* **#6460** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

mlbrgl:
@rakyi turns out codex was onto something! Given narrative charts are not hydrated on the archive, we should let the fallback image drive the height, and not arbitrarily force it to the (hydrated) grapher height, which creates gaps in the layout (in particular when narrative charts are in two-col layouts).

I have a handy gdoc for testing all scenarios (you still need to manually disable the multiembedder to simulate an archive scenario): https://docs.google.com/document/d/1cQ2aEhdPpK7s0evEEILq6MDfeMQaQyO3GYfBhB_GKpQ/edit?tab=t.0

Mind giving this a quick sanity check?

rakyi (approved):

mlbrgl:
@rakyi as discussed, merging now. Feel free to pick it up and refactor to avoid loading transient chart thumbnail images for narrative charts on the live site (as opposed to the archive)

PR #6473, opened 2026-05-11
Fix useTopicTagGraph in admin preview, no. 2

rakyi:
Use a same-origin admin API endpoint for preview mode instead of fetching the public baked JSON, avoiding CORS issues on admin. We take the same approach for details on demand.

This is a second attempt at the fix.

## Testing guidance

It's very hard to test it outside of prod because only there we run the admin on a different domain, but it should be a low risk change.

### DIFF
--- a/site/SiteNavigation.tsx
+++ b/site/SiteNavigation.tsx
@@ -35,13 +35,17 @@ const HAS_DONATION_FLAG = false
 export const SiteNavigation = ({
     hideDonationFlag,
     isOnHomepage,
+    isPreviewing,
 }: {
     hideDonationFlag?: boolean
     isOnHomepage?: boolean
+    isPreviewing?: boolean
 }) => {
     const [menu, setActiveMenu] = useState<Menu | null>(null)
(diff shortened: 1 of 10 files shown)


site/SiteNavigation.tsx
```diff
@@ -35,13 +35,15 @@ const HAS_DONATION_FLAG = false
 export const SiteNavigation = ({
     hideDonationFlag,
     isOnHomepage,
+    isPreviewing,
 }: {
     hideDonationFlag?: boolean
     isOnHomepage?: boolean
+    isPreviewing?: boolean
```
ikesau, line 42: Maybe we should make this non-optional so that we don't ever forget it and reintroduce this bug?

ikesau (approved):
Ah damn, yeah okay - makes sense.

site/SiteNavigation.tsx
rakyi: Yeah, changed.

rakyi:
@codex review

PR #6461, opened 2026-05-07
🔨 standardise AxisView components

sophiamersmann:
A few minor changes to make AxisView components more predictable:
- Prefer `axis` as prop name over `verticalAxis` and `horizontalAxis`
- Make the `bounds` prop non-optional for consistency
- Remove dead code

PR #6475, opened 2026-05-11
✨ dumbbell shapes

sophiamersmann:
[Cycle issue](https://github.com/owid/owid-grapher/issues/6169) / [Designs](https://www.figma.com/design/dCB69o19cAVtuk283STAcY/Dumbbell-Plots?node-id=1-31&m=dev)

Implements the dumbbell shapes and adds two new options for dumbbell plots. Not feature complete yet (see below for follow-up work).

This PR is only about plotting the actual dumbbell shapes.

### Description

* When the series strategy is entity (i.e. single indicator, time range), dumbbells plot a dot at the start value and point to the end value with an arrow.
* When the series strategy is column (i.e. two indicators, single time), dumbbells plot two dots, one at each end, connected by an arrow by default. The arrow can be changed to a straight line via a new setting.

New settings:

* `dumbbell.connectorStyle`: controls whether the two dots in a two-indicator dumbbell are connected by an arrow or a line (only applies when the series strategy is column).
* `dumbbell.valueLabelMode`: controls what is displayed as value labels. Supported options are absolute values, absolute change, percentage change, or no labels.
* I thought it made sense to group both options under a `dumbbell` namespace since they're specific to this chart type
* The settings are exposed in the admin under the 'Customize' tab on the very bottom

Edge cases:
- If the series strategy is entity, and the two values are close together, then the arrow is plotted above the dot with a bit of a different colour
- If the series strategy is column, and the two values are close together, then the dots are vertically offseted a bit, so that they're both visible

Other changes:
- I refactored the series to include start/end instead of left/right, which is a bit more convenient overall

### Follow-up work

* ~Dumbbell elements~ (addressed in this PR)
* ~New dumbbell-specific options: value labels, end points~ (addressed in this PR)
* In-chart legend, aligned with the top-most dots
* Color (hard-coded for now)
* Interaction (focus, hover, tooltips)
* More entity sorting options
* Dumbbell plots in search
* Dumbbell chart thumbnails


sophiamersmann:
* **#6475** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6475?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6476](https://github.com/owid/owid-grapher/pull/6476) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6476?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6484](https://github.com/owid/owid-grapher/pull/6484) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6484?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6475?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6474** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6254** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6461** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6461?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

marcelgerber:
One thing I just noticed: If the arrow is very short, then it can be plotted right atop the dot and look a bit odd. It also has a slightly different color (or opacity).
Not sure what best to do about it - maybe plot it behind the dot, make it the same color & opacity, and make sure the "long" (here: left) arrow edge doesn't show behind the dot?

[This is the chart in question.](http://staging-site-dumbbell-elements/admin/charts/486/edit?tab=customize)

CleanShot 2026-05-13 at 10 53 17

sophiamersmann:
> One thing I just noticed: If the arrow is very short, then it can be plotted right atop the dot and look a bit odd. It also has a slightly different color (or opacity).

That’s actually intentional. It’s the solution Marwa and I settled on. If you look at the code, you’ll see there’s an extra code path for it. We wanted the second dot (i.e. the arrow) to be visible, even when both elements are rendered at the same position. I don't think we're married to this, but seemed better than the alternatives

marcelgerber:
Hm, I'm not so sure about it. Two issues:
1. When you're not zoomed into the chart, it is really hard to see that this is an arrow (at least in this case), because there's little contrast and the shape is not clear. (see screenshot below at "chart size").
2. In the case above, it looks like the arrow is to the left of the dot, when it really is to the right. That's probably because the arrow head's center of mass is to the left.

CleanShot 2026-05-13 at 13 24 08

sophiamersmann:
What do you suggest to do instead?

marcelgerber:
It might be fine to just have a single dot, especially since usually, we'll have the label(s) next to it that make it clear that there's a second, obscured dot nearby.

packages/@ourworldindata/grapher/src/dumbbellCharts/Dumbbell.tsx
```diff
@@ -1,56 +1,223 @@
...
+    series: RenderDumbbellSeries
+    connectorStyle: DumbbellConnectorStyle
+}): React.ReactElement {
+    return match(connectorStyle)
+        .with(DumbbellConnectorStyle.Arrow, () => (
+            <TwoColumnArrowDumbbell series={series} />
+        ))
+        .with(DumbbellConnectorStyle.Line, () => (
+            <TwoColumnLineDumbbell series={series} />
+        ))
```
marcelgerber, line 48: Nothing to do here: But I could see a chart variation without a connecting line be useful, also.

For example, in the below chart, the line doesn't really add much value (and the arrow doesn't make sense).

Obviously, it's also very easy to add this later on when we need it.

Image

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChart.tsx
```diff
@@ -84,15 +91,71 @@ export class DumbbellChart
...
+
+                return { start: formattedStartValue, end: formattedEndValue }
+            })
+            .with(DumbbellValueLabelMode.Change, () => {
+                const diff = endValue - startValue
+                return { end: this.formatValue(diff, { showPlus: true }) }
+            })
+            .with(DumbbellValueLabelMode.PercentChange, () => {
+                if (startValue === 0) return {}
+                const change = ((endValue - startValue) / startValue) * 100
```
marcelgerber: This calculation goes wrong when one of the two numbers is negative:

[(chart in question)](http://staging-site-dumbbell-elements/admin/charts/5370/edit)

Image

Or when both numbers are negative:

Image

Not sure what best to do in these cases, we might just show no value?

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChart.tsx
```diff
@@ -84,15 +91,71 @@ export class DumbbellChart
...
+        return match(this.chartState.valueLabelMode)
+            .with(DumbbellValueLabelMode.Absolute, () => {
+                const formattedStartValue = this.formatValue(startValue)
+                const formattedEndValue = this.formatValue(endValue)
+
+                // Only show one label if the labels are the same
+                // (i.e. the values are the same after rounding)
+                if (formattedStartValue === formattedEndValue) {
+                    return { start: formattedStartValue }
+                }
```
marcelgerber: In the case where they exactly overlap, in 2-column mode, I would prefer if we actually did label both points.

[(chart in question)](http://staging-site-dumbbell-elements/admin/charts/5393/edit?tab=customize)

Image

marcelgerber (approved):
This is real nice!

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChart.tsx
marcelgerber: okay, this formula will properly take care of negative values instead:
```suggestion
                const change = ((endValue - startValue) / Math.abs(startValue)) * 100
```

sophiamersmann: yup, that's what I just did!

packages/@ourworldindata/grapher/src/dumbbellCharts/Dumbbell.tsx
sophiamersmann: Makes sense! But the line also doesn't hurt, so let's wait for the feature request

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChart.tsx
sophiamersmann: Agreed!

sophiamersmann:
@marcelgerber, as discussed, this is how it looks with arrows only. I think it works!

Screenshot 2026-05-13 at 16 32 44

marcelgerber:
yep, that works!

sophiamersmann:
### Merge activity

* **Jun 3, 12:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6475).
* **Jun 3, 12:47 PM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6475) rebased this pull request as part of a merge.
* **Jun 3, 12:47 PM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6475).

PR #6479, opened 2026-05-12
🐛 add charts index shrinking to fit algolia's max record size

ikesau:
The [fish stocks explorer](https://ourworldindata.org/explorers/fish-stocks) was recently migrated to be indicator-based, which fixed its metadata such that its `availableEntities` are now included in its Algolia records. This drove its records' size over Algolia's maximum of 20,000 bytes, which crashed the indexing script.

This PR fixes that by adding a "shrink records" function which serializes records and, if the record is too large, clears `availableEntities` and adds them one-by-one till it's at the max size.

For now, this will only affect the fish stocks explorer, which is already a weird case to support because no one will search `country:Albacore Tuna`, but at least we're future-proofed against such cases in the future.


marcelgerber (approved):
I guess Codex has a good point here, but otherwise this looks good! 👍🏻

PR #6480, opened 2026-05-12
🐝🤖 cancel stale PR CI runs

rakyi:


marcelgerber (approved):
Cool, this'll work.

I think Codex's comment is valid, but only for the case where there's previous runs that are pending (i.e., haven't started yet).
We could solve those [using `queue: max`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency), which would mean that all these pending jobs would be run one-after-the other.
I guess another way would be to just not enable concurrency groups for master.

The one thing I'm missing in this PR is the motivation for this change? Did we run into any issues where we were using too much GH CI capacity? I know Buildkite is a bottleneck of sorts, but that's not part of this CI job.

rakyi:
Good catch. I thought it would be an easy way to make the PR feedback a bit faster, but seeing the issues, I don't think it's worth it.

PR #6493, opened 2026-05-15
✨ Unify algolia charts index score formulae

ikesau:
We currently have three different scoring formulae for charts, explorer views, and multi-dim views. These get applied to each record array, and are then each normalized per-type between a range of 1000 to 11000.

There are 2 big problems with this approach:

1. The multi-dim formula is very dependent on pageviews, and our analytics table aggregates all pageviews for multi-dims into a single slug. This means multi-dims all get boosted identically, creating a large wall of mdim slices in our results
2. We normalize each of these 3 tracks separately. Grapher records are very long-tail and so "low-value" multi-dim views get normalized to a score that's higher than "medium-value" graphers. This happens even if we solve the problem of multi-dims all getting an exaggerated bonus from aggregated pageviews.
    1. example: `life-expectancy` (grapher) gets a pre-adjusted score of 100,000 and `life-expectancy-by-sex` (grapher) gets a pre-adjusted score of 20,000. In both cases the high score is due to a high number of pageviews in absolute terms. These are both high quality charts.
    2. `years-of-schooling?plus-some-params` (mdim) gets a pre-adjusted score of 5,000. A close variant of this chart gets a pre-adjusted score of 4,000.
    3. After normalization, `life-expectancy` is scored at 10,000. `life-expectancy-by-sex` is scored at 2,000. `years-of-schooling?plus-some-params` is scored at 10,000 and its variant gets a score of 8,000
    4. So despite both multi-dims getting a lower absolute score than the grapher records, its normalized scores are higher.

A smaller problem is that `numRelatedArticles` also wasn't disaggregated for our explorers and multi-dims, so linking to any view would boost the score for all of them.

## The fixes
1. Disaggregates pageviews for explorer and multi-dims using the new `analytics_chart_views` table (keyed by `chart_slug` for grapher and `view_config_id` for explorers and multi-dims)
2. Disaggregates related articles for explorer and multi-dims
3. Normalizes grapher, explorer, and multi-dim records all at once
4. Adds a small post-normalization boost for records that are FMs
5. Adds a more severe "boost" button to specific featured metrics if we want them to have a high score even in regular search (this is mostly to fix the GDP chart problem)

## Other changes
1. A `--dry-run` flag for the chart indexing script to write the records locally instead of indexing to Algolia.


ikesau:
@codex review

baker/algolia/indexExplorerViewsMdimViewsAndChartsToAlgolia.ts
```diff
@@ -55,24 +52,28 @@ const indexExplorerViewsMdimViewsAndChartsToAlgolia = async () => {
...
+            // Non-default explorer views are scaled to [0, 1000] to bury
+            // duplicative explorer variants under higher-quality results.
+            const [explorerFirstViews, explorerOtherViews] = _.partition(
+                explorerViews,
+                (view) => view.isFirstExplorerView
+            )
+            const scaledExplorerOtherViews = scaleRecordScores(
+                explorerOtherViews,
+                [0, 1000]
+            )
```
ikesau: Is this actually still necessary? Given we can do per-chart-view scores

baker/algolia/utils/mdimViews.ts
```diff
@@ -277,58 +273,9 @@ export async function getMdimViewRecords(
         )
     ).filter((m) => id === undefined || m.multiDim.id === id)
 
-    const [grapherRedirects, explorerRedirects] = await Promise.all([
```
ikesau, line 280: This shouldn't have been done. We still need to use grapher pageviews as a signal when creating new multi-dims.

ikesau:
@Martin I did those 3 extra things:

1. Removed explorer first-view splitting
2. Added a grapher→multi-dim redirects resolution to use the grapher's pageviews if higher
3. Removed the local index comparer tool

baker/algolia/utils/pageviews.ts
```diff
@@ -1,13 +1,72 @@
...
+): Promise<ChartViewsMap> {
+    const rows = await db.knexRaw<{
+        chart_slug: string
+        view_config_id: string
+        views_7d: number
+    }>(
+        trx,
+        `SELECT chart_slug, view_config_id, views_7d
+         FROM analytics_chart_views`
+    )
```
rakyi: Nit: For simple queries like this (and below), the Knex query builder works well because it allows us to reuse a type we only need to define once in a canonical place and properly narrows it (using `Pick`) for us when using e.g. `select`. That way when the type of column changes we (in theory) only need to change it in the canonical type definition once instead of doing a full text search for it through the code base. See e.g. `getChartConfigsByIds`.

baker/algolia/utils/context.ts
```diff
@@ -45,13 +51,45 @@ export async function createExplorersIndexingContext(
     return baseContext ?? createBaseIndexingContext(knex)
 }
 
+/**
+ * Inverts the grapher→mdim redirect map into a lookup keyed by the mdim target
+ * slug, so each mdim can find the grapher slugs that now redirect into it.
+ * Skipping explorers, because multi_dims_redirects doesn't have query params:
+ * We'd only be able to redirect /explorers/base to /graphers/base
```
rakyi: We currently support redirecting from an explorer slug to either a specific multi-dim view or the default view. In the future we'll also support a specific explorer view as the source, so we'll modify the existing handling of redirects here at some point (CC @marcelgerber).

Could you, please, restore the explorer -> multi-dim redirects handling, as well?

adminSiteClient/FeaturedMetricsPage.tsx
```diff
@@ -106,6 +110,22 @@ function FeaturedMetricList({
...
+                                className="featured-metrics-page__checkbox-tooltip"
+                            >
+                                <Checkbox
+                                    checked={!!featuredMetric.boostInSearch}
+                                    onChange={(e) =>
+                                        updateBoost.mutate({
+                                            id: featuredMetric.id,
+                                            boostInSearch: e.target.checked,
+                                        })
+                                    }
```
rakyi, line 118: At minimum, we should disable the checkbox when the mutation is ongoing. Adding a loading indicator would be even nicer.

adminSiteClient/FeaturedMetricsPage.tsx
```diff
@@ -521,6 +586,17 @@ export function FeaturedMetricsPage() {
...
+    const updateBoost = useMutation({
+        mutationFn: ({ id, boostInSearch }) =>
+            admin.requestJSON(
+                `/api/featured-metrics/${id}/boost`,
+                { boostInSearch },
+                "PUT"
+            ),
+        onSuccess: () =>
+            queryClient.invalidateQueries({ queryKey: ["featuredMetrics"] }),
+    }) as UpdateBoostMutation
```
rakyi: We don't need the type cast (which loses type safety). We can use a type parameter instead:

```ts
useMutation<
    Json,
    unknown,
    { id: number; boostInSearch: boolean },
    unknown
>
```

baker/algolia/utils/context.ts
```diff
@@ -1,20 +1,26 @@
...
 ): Promise<IndexingContext> {
-    const [pageviews, topicHierarchies] = await Promise.all([
-        getAnalyticsPageviewsByUrlObj(knex),
+    const [views, topicHierarchies] = await Promise.all([
+        getAnalyticsChartViews(knex),
         db.getTopicHierarchiesByChildName(knex),
     ])
 
-    return { pageviews, topicHierarchies }
+    return { views, topicHierarchies }
```
rakyi: `pageviews` was nice because you could tell it's not explorer/multi-dim views (configuration of dimensions). Could we improve the name? Maybe `analyticsViews`?

baker/algolia/utils/mdimViews.ts
```diff
@@ -126,8 +123,36 @@ async function getRecords(
...
         await getDatasetDimensionsByVariableIds(trx, [...relevantVariableIds])
+    const linksFromGdocs = await getPublishedLinksTo(
+        trx,
+        [slug],
+        ContentGraphLinkType.Grapher
+    )
+    const numRelatedArticles = linksFromGdocs.length
+    // Related-article links point at the bare mdim slug, which resolves to the
+    // default view. Attribute the count to that view only, so a popular mdim
+    // doesn't boost every one of its views equally.
```
rakyi: If this is true, then we have a deficiency in our data modeling. We don't literally link only to default views from articles. We don't have to change the code, but could you expand this comment with explaining that we do it just for a lack of better option? Or is there a way to get just link to a specific view?

packages/@ourworldindata/types/src/domainTypes/Search.ts
```diff
@@ -359,6 +359,26 @@ export type ChartsIndexingContext = IndexingContext & {
...
 }
 
+/**
+ * A grapher slug that previously served traffic now redirected to a mdim view.
+ * `queryStr` identifies the specific mdim view; when undefined, the redirect
+ * points at the mdim's default view.
+ */
+export interface MdimRedirectSource {
+    sourceSlug: string
+    queryStr?: string
```
rakyi: This is misleading. Elsewhere, we use it as `source.queryStr` which strongly implies that it's the query string of the redirect source. Could we change the interface name and its use e.g. to `MultiDimRedirect` and  `redirect.targetQueryStr` etc.?

baker/algolia/indexExplorerViewsMdimViewsAndChartsToAlgolia.ts
```diff
@@ -46,27 +56,35 @@ const indexExplorerViewsMdimViewsAndChartsToAlgolia = async () => {
...
+            // Scale charts, mdim views, and explorer views together so scores are directly comparable.
+            // Makes it easier to intuit what bonuses and boosts will do
+            const scaledRecords = scaleRecordScores(
+                [...explorerViews, ...mdimViews, ...grapherViews],
+                [1000, MAX_NON_FM_RECORD_SCORE]
+            )
+
+            // Apply post-scaling adjustments. FM source bonus is a fixed
+            // +500 to any record whose specific view is a featured metric.
+            // boostInSearch overrides the score to 9500 for editorially
```
rakyi: IMO we should avoid magic numbers in comments when they are far from the implementation. Either remove them or reference the constants where the numbers are defined.

rakyi (commented):
We might have failed to propagate it properly, but I think it was on a dev weekly where we decided the canonical name for multidimensional Grapher charts is "multi-dim" instead of "mdim". One reason was that "mdim" is too close to "mim" (most important metrics), I don't remember the rest.

I know that we currently have both in our code, but could you use "multi-dim" for the new code?

adminSiteClient/FeaturedMetricsPage.tsx
ikesau: Would be nice! I tried all the optimistic UI strategies suggested in the react-query docs, but the DOM takes too long to render so it only updates by the time the request has finished anyway.

baker/algolia/utils/mdimViews.ts
ikesau: Oh, thanks for the nudge! You're right that this comment was wrong, but in rewriting it, I realized it was possible to solve the problem I was trying to write up.

My initial thinking was that `getPublishedLinksTo` didn't have a a way to narrow down the selection to specific views (it returns all rows with target `years-of-schooling`) but we can just filter those down inside the multi-dim `getRecords` function.

In summary: fixed 😁

baker/algolia/utils/context.ts
ikesau: `chartViews` 👍

ikesau:
Okay @rakyi - I got all the redirect resolution logic working.

To test, I recommend manually editing `indexExplorerViewsMdimViewsAndChartsToAlgolia.ts`:
1. Comment out `getExplorerViewRecords` and `getChartsRecords` calls (and the downstream references)
2. Comment out the `createFeaturedMetricRecords` call (and its downstream references)
3. Edit the filter for `multiDimsWithTags` in `getMultiDimViewRecords` to be `.filter((m) => m.multiDim.slug === "academic-performance")` 

**This will allow you to run `yarn tsx --tsconfig tsconfig.tsx.json baker/algolia/indexExplorerViewsMdimViewsAndChartsToAlgolia.js --dry-run` for quick feedback loops.**

**First test**:
1. Make sure you've run `make refresh.full` recently so that your `analytics_chart_views` table is populated
2. Dry-run index script
3. Inspect `tmp/explorer-views-and-charts.json`
4. See the score for the record with chartConfigId `01980a88-356b-7606-8872-63b64cc9c848`

**Second test**:
1. Create a multi-dim redirect from a popular grapher chart to the default view for `academic-performance`
2. Dry-run index script
3. Inspect `tmp/explorer-views-and-charts.json`
4. Confirm the record now uses the `chart_views` from that old grapher

**Third test**:
1. Create a multi-dim redirect from an explorer to the default view for `academic-performance`
2. Manually edit the explorer's default view's row in the `analytics_chart_views` table to have a very high number of `views_7d`
3. Dry-run index script
4. Confirm that the record now uses the `chart_views` from that explorer

You can also test edge cases if you'd like, but I think we're at a pretty good place at this point, so hopefully no more big overhauls necessary.

## Explanation of the changes
I refactored `getMultiDimRedirects` to give us a map of `MultiDimRedirect` which makes explicit which component is from the source and which is the target (as opposed to the confusing/misleading naming from my earlier refactor)

So now we get something like

```ts
{
  "some-grapher-slug": {
    "sourcePrefix": "/grapher/",
    "sourceSlug": "some-grapher-slug",
    "targetSlug": "some-multi-dim-slug",
    "targetQueryStr": "?fruit=banana"
  },
  "some-other-grapher-slug": {
    "sourcePrefix": "/grapher/",
    "sourceSlug": "some-other-grapher-slug",
    "targetSlug": "some-multi-dim-slug",
    "targetQueryStr": "?fruit=apple"
  },
  "some-explorer-slug": {
    "sourcePrefix": "/explorers/",
    "sourceSlug": "some-explorer-slug",
    "targetSlug": "some-multi-dim-slug",
    "targetQueryStr": "?fruit=banana"
  }
}

```

`getMultiDimRedirectsByMultiDimSlug` reverses the map from `getMultiDimRedirects` and adds a lookup key to use when joining with chart views:

```ts
{
  "some-multi-dim-slug": [
    {
      "sourcePrefix": "/grapher/",
      "sourceSlug": "some-grapher-slug",
      "targetSlug": "some-multi-dim-slug",
      "targetQueryStr": "fruit=banana",
      "lookupKey": "some-grapher-slug"
    },
    {
      "sourcePrefix": "/grapher/",
      "sourceSlug": "some-other-grapher-slug",
      "targetSlug": "some-multi-dim-slug",
      "targetQueryStr": "fruit=apple",
      "lookupKey": "some-other-grapher-slug"
    },
    {
      "sourcePrefix": "/explorers/",
      "sourceSlug": "some-explorer-slug",
      "targetSlug": "some-multi-dim-slug",
      "targetQueryStr": "fruit=banana",
      "lookupKey": "some-uuid-for-the-explorer-default-view"
    }
  ]
}

```

`predecessorsByQueryStr` uses this map and groups it further by `targetQueryStr`:

```
{
  "fruit=apple": [
    {
      "sourcePrefix": "/grapher/",
      "sourceSlug": "some-other-grapher-slug",
      "targetSlug": "some-multi-dim-slug",
      "targetQueryStr": "?fruit=apple",
      "lookupKey": "some-other-grapher-slug"
    }
  ],
  "fruit=banana": [
    {
      "sourcePrefix": "/grapher/",
      "sourceSlug": "some-grapher-slug",
      "targetSlug": "some-multi-dim-slug",
      "targetQueryStr": "?fruit=banana",
      "lookupKey": "some-grapher-slug"
    },
    {
      "sourcePrefix": "/explorers/",
      "sourceSlug": "some-explorer-slug",
      "targetSlug": "some-multi-dim-slug",
      "targetQueryStr": "?fruit=banana",
      "lookupKey": "some-uuid-for-the-explorer-default-view"
    }
  ]
}

```
We then use _this_ map with `getMaxChartViews` to get the maxChartViews of each record.

# 😮‍💨

ikesau:
@codex review

adminSiteClient/FeaturedMetricsPage.tsx
rakyi: Disabling the checkbox when the request is ongoing should be really easy though, isn't it?

baker/algolia/utils/context.ts
```diff
@@ -52,44 +57,129 @@ export async function createExplorersIndexingContext(
...
+            explorerSlugByViewKey.set(`${slug}:${viewId}`, slug)
+        } catch {
+            // Explorer may have been deleted; skip silently — caller treats
+            // missing entries as "no signal".
+        }
+    }
+    if (explorerSlugByViewKey.size === 0) return result
+
+    const rows = await knex<
+        Pick<DbRawExplorerView, "explorerSlug" | "viewId" | "chartConfigId">
```
rakyi: Here and in other places where we use the knex query builder: we don't have to use `Pick` ourselves, it's what Knex does for us. We only need to use the table type as the type argument, e.g. `knex<DbRawExplorerView>`.

baker/algolia/utils/context.ts
```diff
@@ -52,44 +57,129 @@ export async function createExplorersIndexingContext(
...
+    >(ExplorerViewsTableName)
+        .select("explorerSlug", "viewId", "chartConfigId")
+        .whereIn("explorerSlug", [...new Set(explorerSlugs)])
```
rakyi: I don't think we need to do the deduplication, do we? If we do, then we should do it before we iterate over the slugs above.

packages/@ourworldindata/types/src/domainTypes/Search.ts
rakyi: It looks like this and the type below are now unused.

baker/algolia/utils/context.ts
```diff
@@ -45,13 +56,130 @@ export async function createExplorersIndexingContext(
...
+    const explorerAdminServer = new ExplorerAdminServer()
+    const explorerSlugByViewKey = new Map<string, string>()
+    for (const slug of explorerSlugs) {
+        try {
+            const program = await explorerAdminServer.getExplorerFromSlug(
+                knex,
+                slug
+            )
+            const firstChoice =
+                program.decisionMatrix.allDecisionsAsQueryParams()[0]
```
rakyi: This isn't sufficient to get the default view dimensions. We have to consider the `defaultView: true` option. We probably need to do the same thing the explorer code does for the UI when choosing the default view.

baker/algolia/utils/pageviews.ts
```diff
@@ -1,13 +1,70 @@
...
+            "chart_slug" | "view_config_id" | "views_7d"
+        >
+    >(AnalyticsChartViewsTableName).select(
+        "chart_slug",
+        "view_config_id",
+        "views_7d"
+    )
+    const map = new Map<string, number>()
+    for (const row of rows) {
+        const key = row.view_config_id || row.chart_slug
```
rakyi: I don't know if this is a data issue, but we should probably change this logic regardless. For example, I see that `chart_slug = 'plastic-pollution'` has currently rows for all three of `explorer`, `grapher_chart` and `multidim` (including an `explorer` with empty `view_config_id`).

baker/algolia/utils/mdimViewsLogic.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    interval: "yearly",
+})
+const POVERTY_VIEW_ID = dimensionsToViewId({
+    view: "poverty",
+    interval: "yearly",
+})
+const STUNTING_WEEKLY_VIEW_ID = dimensionsToViewId({
+    view: "stunting",
+    interval: "weekly",
+})
```
rakyi: I think these are semantically top level variables rather than constants, since we compute them dynamically using business logic, therefore we should make them `camelCase`.

rakyi (commented):
It looks like we lost multi-dim → multi-dim redirect support for analytics chart views. If it's not an easy fix, we could create an issue for it for later, since we currently don't have such redirects yet.

<details>
<summary>Minimal repro:</summary>

1. Create this temporary redirect:

```sql
INSERT INTO multi_dim_redirects (source, multiDimId, viewConfigId, createdAt, updatedAt)
SELECT
  '/grapher/years-of-schooling',
  id,
  NULL,
  NOW(),
  NOW()
FROM multi_dim_data_pages
WHERE slug = 'academic-performance';
```

2. Run the dry-run indexer with only `academic-performance` mdim records enabled.

3. Inspect the record:

```bash
jq '.[] | select(.chartConfigId=="01980a88-356b-7606-8872-63b64cc9c848") | {id, chartConfigId, views_7d}' tmp/martin-explorer-views-and-charts.json
```

Expected:

```json
{
  "views_7d": 1521
}
```

Actual:

```json
{
  "views_7d": 65
}
```

Why this reproduces it:

- `/grapher/years-of-schooling` is itself a multi-dim.
- Its default view has `analytics_chart_views.views_7d = 1521`.
- The target academic-performance default view has own `views_7d = 65`.
- A mdim→mdim redirect should inherit the predecessor mdim view signal, but it doesn’t.

Cleanup:

```sql
DELETE FROM multi_dim_redirects
WHERE source = '/grapher/years-of-schooling';
```
</details>

Overall, great job on a complex problem, we are getting there! Unfortunately there are more things that need fixing.

baker/algolia/utils/context.ts
rakyi, line 1: Could we move `getChartViewsMap` into the base context? We now call it in each of the specialized context builder functions.

baker/algolia/utils/pageviews.ts
```diff
@@ -1,13 +1,161 @@
...
+    DbPlainAnalyticsChartViewsRow,
+    DbPlainMultiDimDataPage,
+    DbPlainMultiDimRedirect,
+    DbRawExplorerView,
+    ExplorerViewsTableName,
+    MultiDimDataPagesTableName,
+    MultiDimRedirectsTableName,
+} from "@ourworldindata/types"
+import * as db from "../../../db/db.js"
+import { partition } from "remeda"
```
rakyi: Nit: Could we stick to the `import * as R from "remeda"` pattern and then use `R.partition`? It makes it easy to know where does the function come from by glance (and Lodash also has `partition`).

packages/@ourworldindata/types/src/dbTypes/AnalyticsChartViews.ts
```diff
@@ -0,0 +1,13 @@
+export const AnalyticsChartViewsTableName = "analytics_chart_views"
+
+export type AnalyticsChartViewsType = "grapher_chart" | "explorer" | "multidim"
+
+export interface DbPlainAnalyticsChartViewsRow {
```
rakyi: The usual naming scheme used in `dbTypes` would be `DbPlainAnalyticsChartView`.

rakyi (approved):
Excellent work, this now looks great to me!

Left a few small optional comments.

PR #6494, opened 2026-05-16
(demography) Allow wider value range in manual parameter entry

danyx23:
This PR extends the range of values you can enter via the manual text entry box and the +/- icons of that box when clicking icons. The drag range stays more limited, but the manual entry can go to pretty extreme values.

You can play with it [here](http://staging-site-demography-wider-manual-entr/population-simulation-tool)

GPT 5.5. generated overview below

## Problem

The manual text input on each control-point dot was bounded by the chart's auto-computed domain — e.g. ~110 years for South Korea's life expectancy — making it impossible to explore scenarios beyond the data-driven range. The cap wasn't intentional; it was an accidental consequence of passing `yScale.domain()` directly as `min`/`max` to the `NumberField`.

## Changes

**`helpers/parameterConfigs.ts`**
- Add `inputMin` / `inputMax` to the `ParameterConfig` interface for manual text-entry limits
- Add optional `dragMin` / `dragMax` limits for parameters where dragging should stay within a narrower range than manual entry
- Set absolute manual-entry limits per parameter:
  - Life expectancy: 10 – 150 years
  - Fertility rate: 0 – 10 births per woman
  - Net migration rate: −100 – +100‰
- Keep drag limits for net migration at −20 – +20‰

**`components/DemographyParameterEditor.tsx`**
- Pass `config.inputMin` / `config.inputMax` to react-aria's `NumberField` instead of the chart domain bounds — manual entry now accepts the configured full range; the +/− step buttons respect the same limits
- Include the current user-set control-point values in the chart domain calculation, so the chart rescales automatically when a typed value falls outside the auto-computed range
- Keep source/reference values in the chart domain calculation, so unusually high/low source migration rates remain visible
- Clamp dragging with the configured drag limits where present, falling back to manual-entry limits otherwise

## Behaviour preserved / clarified

Dragging a control point continues to clamp via `yScale.clamp: true` and the configured drag/input bounds, so dots stay within the rendered chart. For net migration, manual entry can use the wider −100 – +100‰ range, while dragging remains limited to −20 – +20‰.

The chart domain still respects parameter-specific visual caps such as `yFloor` / `yCeiling`; for example, life expectancy remains visually capped at 130 years even though manual entry accepts values up to 150. Resetting to UN WPP values returns the chart to its normal auto-scaled range.

### DIFF
--- a/bespoke/projects/demography/src/helpers/parameterConfigs.ts
+++ b/bespoke/projects/demography/src/helpers/parameterConfigs.ts
@@ -51,7 +51,7 @@ export const parameterConfigByKey: Record<ParameterKey, ParameterConfig> = {
         axisUnit: "births per woman",
         yPadding: 2,
         yFloor: 0,
-        inputMin: 0.1,
+        inputMin: 0,
         inputMax: 10,
         step: 0.1,
         decimals: 1,
--- a/bespoke/projects/demography/src/helpers/parameterConfigs.ts
+++ b/bespoke/projects/demography/src/helpers/parameterConfigs.ts
@@ -27,8 +27,6 @@ interface ParameterConfig {
     yCeiling?: number
     inputMin?: number
     inputMax?: number
-    dragMin?: number
-    dragMax?: number
     step: number
     decimals: number
     subtitle: (entityName: string) => string
@@ -184,10 +182,8 @@ export const parameterConfigByKey: Record<ParameterKey, ParameterConfig> = {
         unit: "per 1,000 population",
         axisUnit: "‰",
         yPadding: 5,
-        inputMin: -100,
-        inputMax: 100,
-        dragMin: -20,
-        dragMax: 20,
+        inputMin: -20,
+        inputMax: 20,
         step: 0.1,
         decimals: 1,
         subtitle: (entityName: string) => {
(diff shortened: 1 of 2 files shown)


bespoke/projects/demography/src/helpers/parameterConfigs.ts
```diff
@@ -176,6 +184,10 @@ export const parameterConfigByKey: Record<ParameterKey, ParameterConfig> = {
         unit: "per 1,000 population",
         axisUnit: "‰",
         yPadding: 5,
+        inputMin: -100,
+        inputMax: 100,
+        dragMin: -20,
+        dragMax: 20,
```
sophiamersmann: Why is drag min/max special-cased for net migration? Jumping to 100 via the input field and then dragging leads to an unpleasant jump

sophiamersmann (approved):

bespoke/projects/demography/src/helpers/parameterConfigs.ts
danyx23: The reason I added it is that e.g. Syria had stronger negative net migration than -20‰ - but thinking a bit more about it, these extreme values are only useful for short term spikes and it is not worth the added complexity to keep this. Removed this now.

PR #6498, opened 2026-05-18
Drop approved-by field from data insights

edomt:
## Summary
- We no longer use the ArchieML `approved-by` mechanism — approvals are tracked outside the gdocs flow, so the publish-time validation was pure friction.
- Removes the field from the type definitions, the admin UI (input field and \"approved by X\" display on the index page), the client + server validation, and the lightning-prop map.
- ArchieML accepts unknown keys, so existing data insights that still have `approved-by: …` in their front-matter keep parsing unchanged — the value just isn't surfaced anywhere.

## Test plan
- [x] `yarn typecheck` passes (verified locally).
- [x] `yarn testLintChanged` / `yarn testFormatChanged` clean (verified locally).
- [x] On staging, open a data insight with no `approved-by` front-matter: no error/warning shown, **Publish** button enabled.
- [x] Open a data insight that still has `approved-by` in its gdoc: page loads without error; the admin index no longer shows \"(approved by …)\".
- [x] Publish a data insight end-to-end and confirm the resulting page renders normally.

### DIFF
--- a/adminSiteClient/DataInsightIndexPage.tsx
+++ b/adminSiteClient/DataInsightIndexPage.tsx
@@ -158,7 +158,7 @@ function createColumns(ctx: {
             dataIndex: "authors",
             key: "authors",
             width: 150,
-            render: (authors: string[], dataInsight) => (
+            render: (authors: string[]) => (
                 <>
                     {authors.map((author, index) => (
                         <Fragment key={author}>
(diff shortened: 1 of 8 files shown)


ikesau (approved):
There's a reference to the field in `docs/agent-guidelines/gdocs-class-hierarchy.md` that still needs to be removed, but otherwise tested and LGTM!

PR #6499, opened 2026-05-18
Bump node to 22.22.3

rakyi:
Pi dropped support for =< 22.18.0 and at least for my setup the most convenient fix is to bump the node version of our project, since I install it into the same env.

Since it's just a minor version bump it should hopefully require no other changes.


marcelgerber (approved):
All good here, review comments in https://github.com/owid/ops/pull/482#pullrequestreview-4308884186

PR #6508, opened 2026-05-19
🐛 (admin) avoid resetting entity selection on switching to the Basic tab

### LINKED ISSUE #6485: Admin: Switching back to the "Basic" tab resets entity selection
## Steps to reproduce

1. Go to any chart in admin
2. In the entity selector, select a bunch of entities
3. Navigate away from the "Basic" tab
4. Go back to the "Basic" tab
5. -> The entity selection will reset to what's been saved before

## Video

https://github.com/user-attachments/assets/19ba8077-912b-432d-96da-56754edb53f4

sophiamersmann:
Fixes #6485

In the chart editor, selecting entities and then navigating back to the "Basic" tab caused the selection to be reset.

When the user edits the selection in the admin, `legacyConfigAsAuthored` is updated to match. The `selectedEntityNames` property on GrapherState, however, is only set at construction. On switching to the basic tab, `EditorBasicTab` reloaded data, the inputTable setter saw "selection == authored" and called `applyOriginalSelectionAsAuthored`, which then reset to the stale `selectedEntityNames` value rather than the live authored selection.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
+++ b/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
@@ -1680,13 +1680,21 @@ export class GrapherState
     }
 
     @action.bound private applyOriginalFocusAsAuthored(): void {
-        if (this.focusedSeriesNames?.length)
-            this.focusArray.clearAllAndAdd(...this.focusedSeriesNames)
+        const authoredSeriesNames =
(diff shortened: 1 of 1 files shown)


marcelgerber (approved):
good find! I tested this a bunch with various explorers, Mdims, chart carousels etc., and this seems to work well.

PR #6510, opened 2026-05-19
🐛 fix react-aria/PortalProvider deep import

Marigold:
## Summary

- The deep import `react-aria/PortalProvider` is not declared in `react-aria@3.45.0`'s `exports` field (only `.` and `./i18n` are exposed).
- Vite resolves it leniently, so the website builds fine — but Node's ESM resolver (used by `tsx`) refuses it with `ERR_PACKAGE_PATH_NOT_EXPORTED`, breaking every tsx-based dev script (`yarn query`, `yarn refreshExplorerViews`, etc.) on `master`.
- `UNSAFE_PortalProvider` is also re-exported from the bare `"react-aria"` entry point — same form already used in `bespoke/projects/demography/src/components/DemographyParameterEditor.tsx:9`.

Introduced in #6488.

## Test plan

- [ ] `yarn query "SELECT 1"` succeeds (currently throws `ERR_PACKAGE_PATH_NOT_EXPORTED` on `master`)
- [ ] Site dev server still builds and the popover still renders correctly

### DIFF
--- a/packages/@ourworldindata/grapher/src/controls/PortaledPopover.tsx
+++ b/packages/@ourworldindata/grapher/src/controls/PortaledPopover.tsx
@@ -1,6 +1,6 @@
 import type { ReactElement } from "react"
 import { Popover, type PopoverProps } from "react-aria-components"
-import { UNSAFE_PortalProvider } from "react-aria/PortalProvider"
+import { UNSAFE_PortalProvider } from "react-aria"
 
 export interface PortaledPopoverProps extends PopoverProps {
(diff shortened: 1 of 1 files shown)


Marigold:
Closing — my mistake. The lockfile pins react-aria@3.48.0, which exposes `./*` as a wildcard subpath, so the deep import resolves fine on a synced checkout. I was working with a stale `node_modules` still on 3.45.0 (which only exposed `.` and `./i18n`). `yarn install` resolves it.

PR #6515, opened 2026-05-20
Changes for Node 24

marcelgerber:
These are optional changes that we can now make, now that we are switching to Node 24.
Can merge this anytime after Node 24 lands, and ideally not right away in case we need to roll it back for some reason.


Marigold (approved):
Looks good, though codex knows much better than me :). I merged with master and rebuilt staging on node 24.

marcelgerber:
@codex review

PR #6518, opened 2026-05-20
🐛 fix external indicators in the entity selector

### LINKED ISSUE #6487: Entity selector: Sorting by population doesn't work (no data)
## Description

Sorting by population in the entity selector doesn't work for some charts.

## Steps to reproduce

1. Go to https://ourworldindata.org/grapher/perceived-comfort-speaking-anxiety-depression-vs-talked-to-family-friends-when-had-anxiety-depression
2. Open the entity selector and sort by 'Population'
3. Note that all entities report 'No data'

## Screenshots

Image

sophiamersmann:
Resolves #6487

The catalog data has data for the latest year for each entity. We used to apply infinite tolerance and then try to get the value for the currently selected year, expecting it to always resolve to something. But tolerance doesn’t fill blanks for years without any data, which is why we saw ‘No data’ entries for specific years.

This was buggy, but also a lot of unnecessary work, loading as an OwidTable, applying tolerance, etc. I’ve simplified this now, and externally loaded indicators should now be fixed to their latest year.

### DIFF
--- a/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
+++ b/packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
@@ -24,7 +24,6 @@ import {
     FuzzySearch,
     getUserNavigatorLanguagesNonEnglish,
     getRegionAlternativeNames,
-    convertDaysSinceEpochToDate,
     checkIsOwidIncomeGroupCode,
     checkHasMembers,
     Region,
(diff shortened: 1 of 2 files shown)


packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -1844,3 +1831,21 @@ function isExternalSortIndicatorMatch(
...
+    const column = table.get(slug)
+
+    const maxYear = column.maxTime
+    if (maxYear === undefined) return table
+
+    const rows = column.owidRows.map((row) => ({
+        entityName: row.entityName,
+        year: maxYear,
+        [slug]: row.value,
+    }))
```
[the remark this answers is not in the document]
sophiamersmann: No need to deduplicate because catalog data only contains one entry per entity

marcelgerber (changes requested):
On [charts with daily data](http://staging-site-fix-entity-selector-external/grapher/excess-deaths-cumulative-economist-single-entity), this now formats them weirdly.

Image

sophiamersmann:
yup, I had noticed that

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
```diff
@@ -1844,3 +1839,41 @@ function isExternalSortIndicatorMatch(
         })
         .exhaustive()
 }
+
+function buildOwidTableForCatalogData(
```
marcelgerber, line 1869: I found an edge case where the GDP variable is converted to daily values [in this chart](https://fix-entity-selector-external.owid.pages.dev/grapher/tests-of-covid-19-per-thousand-people-vs-gdp-per-capita), and thus the year/date is displayed wrong.

It's not important overall, since this might be the only chart ever where we use "daily" GDP data.

marcelgerber (approved):

packages/@ourworldindata/grapher/src/entitySelector/EntitySelector.tsx
sophiamersmann: Good catch! I added a fix - this is actually a non-external indicator with yearly data used in a chart with daily data

PR #6527, opened 2026-05-22
Clean up dependencies

rakyi:
Using this skill:
https://www.skills.sh/anthonyshew/dotfiles/npm-deps-cleanup

- Removed unused direct dependencies across root, workspace packages, and bespoke packages.
- Moved dev-only root tools (`@playwright/test`, `chokidar-cli`) to `devDependencies`.
- Replaced `@ourworldindata/utils` umbrella `d3` dependency with direct `d3-array`/`d3-format`.
- Added direct `d3-scale` in bespoke demography because it imports it directly.
- Updated `yarn.lock` and `bespoke/yarn.lock`.


rakyi:
@marcelgerber could you please check if there is something clearly wrong with this? I don't want us to spend too much time on it if the agent messed up and would only proceed if there are no big issues.

marcelgerber (approved):
@rakyi yup, looks good so far, happy to continue with this

rakyi:
Thanks! That's it, there is no follow-up, I was just trying to avoid making you spend a lot of your time on this.

Issue #6530, opened 2026-05-22
Cycle 2026.3: Data page enhancements


ikesau:
It's been a while since we launched data pages and we've got some ideas on how to improve them.

## Motivation

They're one of the main entry points for people onto our site. We want to make them clearer, easier to take in, and better at inviting readers to continue reading.

## Proposed scope

- Make header optional (default off)
- WYSK box
- Explorer more section

PR #6531, opened 2026-05-22
✨ Remove datapage header

ikesau:
## Context

The first and easiest part of https://github.com/owid/owid-grapher/issues/6530

image

### Before merging

### DIFF
--- a/site/DataPageContent.scss
+++ b/site/DataPageContent.scss
@@ -5,6 +5,7 @@
 }
 
 .DataPageContent {
+    // Header components shown in multidims, but not in standalone grapher datapages.
     .header__wrapper {
         padding-top: 24px;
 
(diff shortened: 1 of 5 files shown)


site/DataPageContent.scss
```diff
@@ -5,6 +5,7 @@
 }
 
 .DataPageContent {
+    // Header components shown in Multidims , but not in Datapages.
```
rakyi: AFAIK our ontology (in code) is:

- Explorer pages
- Grapher pages: old `/grapher/<slug>` pages, only standalone graphers
- Data pages: standalone graphers and multi-dims

Could you update the comment to reflect that? Currently, it assumes multi-dims are not data pages.

rakyi (commented):
I remember from the product meeting that there were cases where we thought we'll need to keep the header to distinguish between very similar charts where e.g. the chart title was the same, but the data source was different.

Did we decide that we don't need to do that after all? Or did we solve it some other way?

ikesau:
@rakyi yeah we decided to be bold and try removing them all first, to see if we strongly miss the header. Embracing YAGNI 😁

rakyi (approved):

PR #6534, opened 2026-05-24
🐛🤖 Fix EntitySelector Fast Refresh invalidation

rakyi:
Move EXTERNAL_SORT_INDICATOR_DEFINITIONS out of the React component module and into the catalog-data module. Vite Fast Refresh expects component modules to export only components, so the previous non-component export caused HMR invalidation warnings while editing EntitySelector.


sophiamersmann:
I don't think `loadCatalogData.ts` is the right place for EXTERNAL_SORT_INDICATOR_DEFINITIONS - maybe create a new file called EntitySelectorConstants.ts instead?

rakyi:
Done, I wasn't sure what's the best place to put them in.

PR #6537, opened 2026-05-25
✨ automatically expand latest page item when anchor tag present

ikesau:
When clicking on a data announcement from the homepage, I think we should automatically expand the hit.


mlbrgl (approved):
makes sense, good call!

PR #6539, opened 2026-05-26
Fix baking of data insights

rakyi:
We recently switched the baking to `getAndLoadLastPublishedDataInsights` by mistake, which only returns the last 20 data insights. Switch to a new function that returns all of them.

### DIFF
--- a/db/model/Gdoc/GdocDataInsight.ts
+++ b/db/model/Gdoc/GdocDataInsight.ts
@@ -9,7 +9,7 @@ import {
 import { GdocBase } from "./GdocBase.js"
 import * as db from "../../../db/db.js"
 import {
-    getAndLoadLastPublishedDataInsights,
+    getAndLoadPublishedDataInsights,
     getLatestDataInsights,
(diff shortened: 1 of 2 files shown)


mlbrgl (approved):
we could possibly mimic `getAndLoadPublishedGdocs` options and pass the limit through `getAndLoadPublishedDataInsights` (instead of keeping both functions)?

rakyi:
I prefer the two functions for now, since we clearly have these two use cases and don't need more flexibility.

Not sure if expected, but http://staging-site-fix-bake-all-data-insights/latest doesn't work.

mlbrgl:
I ran indexing for the unprefixed staging sites index (for those staging sites that don't have a dedicated indexed via `staging-algolia` so that pages-chronological contains the right shape of records for /latest to work. My thinking is, as with search, if anyone need to work on it then they'll flag `staging-algolia`. For the rest of PRs, latest will work but will fall out of sync. If you think that's worth discussing, let me know or add a point to today's call.

PR #6548, opened 2026-05-28
Remove unused summary gdoc field

### LINKED ISSUE #6490: Remove unused content.summary field from GdocPost
## Context

In [`GdocPost._enrichSubclassContent`](db/model/Gdoc/GdocPost.ts) (lines 88-92):

```ts
if (content.summary) {
    content.summary = content.summary.map((html: RawBlockText) =>
        htmlToEnrichedTextBlock(html.value)
    )
}
```

A query against the production DB returns no gdocs with a top-level `summary` field:

```sql
SELECT id, slug, type
FROM posts_gdocs
WHERE JSON_CONTAINS_PATH(content, 'one', '$.summary')
-- 0 rows
```

A looser `LIKE '%"summary"%'` match returns a single row (`polio`, linear-topic-page), but its top-level keys are `["toc", "body", "refs", "type", "title", "authors", "excerpt", "dateline", "subtitle", "sidebar-toc", "featured-image"]` — `summary` appears only as nested string content, not as the field this code handles.

`summary` is also not mentioned in the ArchieML documentation as a supported frontmatter field.

## Suggested action

1. Independently confirm no gdoc uses the top-level `content.summary` field.
2. If confirmed, remove the dead branch in `GdocPost._enrichSubclassContent`, drop `summary` from the `OwidGdocPostContent` type if no other code depends on it, and remove any related references.

rakyi:
No gdocs currently use the field. Removed also all helpers that were only used by the summary code.

Resolves #6490

### DIFF
--- a/adminSiteClient/GdocsIndexPage.tsx
+++ b/adminSiteClient/GdocsIndexPage.tsx
@@ -20,7 +20,6 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
 import {
     DbChartTagJoin,
     OwidGdocType,
-    spansToUnformattedPlainText,
     checkIsGdocPost,
     OwidGdocIndexItem,
 } from "@ourworldindata/utils"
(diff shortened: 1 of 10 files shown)


mlbrgl (approved):
thanks!

PR #6551, opened 2026-05-28
🎉 validate record lengths before indexing

### LINKED ISSUE #6535: Validate all record lengths before indexing
Our staging and prod algolia indices have different maximum sizes (20kB on prod, something larger on staging)

We should manually verify record size (for all indices) before attempting to index, so that we can catch errors in staging.

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/6535

Charts with IDs `5879`, `8149`, and `8150` used to trigger the old `"too many entities, skipping its entities"` warning.

The entities are AI models (`"Llama 2-70B", "Llama 2-7B"` etc) so it's not useful for them to be indexed, but I checked to see their size with the new `shrinkRecords` technique and all 3 are around ~17500 in length now.


marcelgerber (approved):
yup looks good!

PR #6571, opened 2026-06-02
🔨 safer DB metadata export

ikesau:
We've currently got a few tables in our metadata export that shouldn't be in there, and our user table has information in it that doesn't need to be.

This PR modifies the export script to be whitelist-based (so that new tables don't leak if we forget) and removes info from the users table.

### DIFF
--- a/db/exportMetadata.ts
+++ b/db/exportMetadata.ts
@@ -1,4 +1,14 @@
-// Script to export everything in the database except sensitive info
+// Script to export the public metadata database.
+//
+// The output (published as owid_metadata.sql.gz) is downloaded by anyone who
+// runs the Grapher stack locally, so it shouldn't contain anything sensitive.
+//
+// By default, only each table's structure is exported, with no rows.
+// For data to be exported, it must be explicitly included in the INCLUDE_DATA_TABLES list
+//
+// As a backstop, the script fails if it encounters a table that isn't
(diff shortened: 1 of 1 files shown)


rakyi (approved):

db/exportMetadata.ts
```diff
@@ -16,38 +26,196 @@ import { execWrapper } from "./execWrapper.js"
 const argv = parseArgs(process.argv.slice(2))
 const filePath = argv._[0] || "/tmp/owid_metadata.sql"
 
-const excludeTables = [
-    "analytics_pageviews",
-    "analytics_grapher_views",
-    "donors",
+const INCLUDE_DATA_TABLES = [
+    "admin_api_keys",
```
ikesau, line 30: @Marigold - I'm assuming this table is necessary for staging? If not, I can move it to the filter list.

Marigold: Yep, ETL uses those keys for auth.

Marigold (approved):
Whitelisting with safeguard for new tables is a nice improvement!

As for `analytics_popularity`, it was meant to be shareable (unlike `analytics_chart_views`). Popularity is a transformed score from pageviews between 0-1. It's used in etl for ranking search results and in some VSCode extensions. I'd keep it in the whitelist if possible.

PR #6579, opened 2026-06-03
✨ datapage redesign

### LINKED ISSUE #6530: Cycle 2026.3: Data page enhancements

ikesau:
Resolves https://github.com/owid/owid-grapher/issues/6530

https://github.com/user-attachments/assets/8ea9f50d-d8a6-43f9-8664-2cc71edc2daa

## Overview
- Changes the design of 10 single-indicator datapages (specified in `experiments/config.ts`)
- Puts all editorial content inside an expandable box (with further `<details>` elements inside it)
- Adds a second search bar
- Adds coviews-based related charts
- Adds indicator owners (e.g. Lucas, Pablo, Tuna, etc) from https://github.com/owid/etl/pull/6143


packages/@ourworldindata/components/src/ExpandableToggle/ExpandableToggle.tsx
```diff
@@ -10,20 +10,31 @@ export const ExpandableToggle = ({
...
 }: {
     label: string
     content?: ReactNode
     alwaysVisibleDescription?: ReactNode
     isExpandedDefault?: boolean
     isStacked?: boolean
     hasTeaser?: boolean
+    contentId?: string
+    /** Called with the new open state whenever the toggle expands or collapses. */
+    onToggle?: (isOpen: boolean) => void
```
ikesau, line 24: This is for scrollY setting purposes. The default browser behaviour is weirdly inconsistent.

packages/@ourworldindata/components/src/IndicatorKeyData/IndicatorKeyData.tsx
```diff
@@ -13,10 +13,12 @@ export const makeSource = ({
     attribution,
     owidProcessingLevel,
     isEmbeddedInADataPage,
+    processingId = DATAPAGE_SOURCES_AND_PROCESSING_SECTION_ID,
 }: {
     attribution?: string
     owidProcessingLevel?: OwidProcessingLevel
     isEmbeddedInADataPage?: boolean
+    processingId?: string
```
ikesau, line 21: This is so that "with major processing" can link to open the "How did Our World in Data process this data?" details element.

packages/@ourworldindata/grapher/src/footer/Footer.tsx
```diff
@@ -446,6 +481,43 @@ abstract class AbstractFooter<
             lineHeight: this.lineHeight,
         })
 
+        // On new-datapage-design pages, the sources line itself links to the
```
ikesau, line 484: This was annoying but the only way I found to do this reliably.

site/gdocs/components/Byline.tsx
```diff
@@ -1,22 +1,30 @@
-import { Fragment } from "react"
+import { Fragment, type ReactNode } from "react"
 import LinkedAuthor from "./LinkedAuthor.js"
 
 export const Byline = ({
     names,
     authorRoles,
+    prefix = "By ",
+    includeImage = false,
```
ikesau: Technically this should be `includeImages` but I went for singular for grep simplicity.

site/search/Autocomplete.tsx
```diff
@@ -567,9 +597,91 @@ export function Autocomplete({
         containerRef,
         allTopics,
         synonymMap,
+        recentSearchesPlugin,
         userCountryNameRef,
+        searchSource,
     ])
 
+    // Close the panel on outside click. We can't rely on autocomplete-js's
```
ikesau, line 605: This bit's all Claude - but it works well

site/DataPageV2Content.tsx
```diff
@@ -163,99 +174,178 @@ export const DataPageV2Content = ({
                                     isPreviewing={isPreviewing}
                                 />
                             )}
-
-                            <AboutThisData
+                            {!useNewDatapageDesign && (
```
ikesau, line 177: Just a lot of nesting things in `!useNewDatapageDesign &&` or `!!useNewDatapageDesign &&` in this file

packages/@ourworldindata/utils/src/experiments/config.ts
```diff
@@ -81,4 +85,50 @@ export const experiments: Experiment[] = [
...
+    new Experiment({
+        id: DATA_PAGE_METADATA_EXPERIMENT_ID,
+        expires: "2026-12-31T00:00:00.000Z",
+        arms: [
+            { id: "control", fraction: 0.0 },
+            { id: "treatment", fraction: 1.0 },
+        ],
+        paths: [
+            // single indicator data pages (multi-indicator data pages aren't supported yet)
+            "grapher/gdp-per-capita-maddison-project-database",
```
ikesau: This URLs collide with ones in `all-charts-vs-featured-v1`, but that's expired (and should be fully removed in a separate PR)

baker/DatapageHelpers.ts
```diff
@@ -14,15 +14,49 @@ import {
...
+import {
+    KnexReadonlyTransaction,
+    getImageMetadataByFilenames,
+} from "../db/db.js"
 import { parseFaqs } from "../db/model/Gdoc/rawToEnriched.js"
+import { getMinimalAuthorsByNames } from "../db/model/Gdoc/GdocBase.js"
 import { getSlugForTopicTag } from "./GrapherBakingUtils.js"
 import { getShortPageCitation } from "../site/gdocs/utils.js"
 
+export async function getLinkedAuthorsForDatasetOwners(
```
sophiamersmann: Maybe this could be generalised to `getLinkedAuthorsByNames` since that's all you need from the owners?

baker/GrapherBaker.tsx
```diff
@@ -203,13 +210,44 @@ export async function renderDataPageV2(
...
 
+    if (datapageMetadataExperimentActive) {
+        // Only show owners for x&y plotted variables (i.e. exclude color/size for scatterplots/marimekkos)
+        const ownerVariableIds = _.uniq(
+            _.compact(
+                grapher.dimensions
+                    .filter(
+                        ({ property }) =>
+                            property === DimensionProperty.y ||
+                            property === DimensionProperty.x
```
sophiamersmann: It doesn’t matter much now, but maybe we should drop the x-dimension here? It’s often GDP per capita (scatter plots) or population (Marimekkos). And if it isn’t, it usually comes from the same dataset as the y-indicator, for example in male vs female scatter plots. Since we’re only showing the owners of the first dataset, I’m thinking of failure cases where the x-dimension comes before the y-dimension and we end up showing the owners of the GDP per capita dataset, for example, even though that’s not the relevant one. Pretty far fetched but idk feels safer to me to just drop it for now

baker/GrapherBaker.tsx
```diff
@@ -203,13 +210,44 @@ export async function renderDataPageV2(
...
+        )
+        datapageData.owners = await getOwnersForVariables(
+            knex,
+            ownerVariableIds
+        )
+
+        const { linkedAuthors, imageMetadata: ownerImageMetadata } =
+            await getLinkedAuthorsForDatasetOwners(knex, datapageData.owners)
+        datapageData.linkedAuthors = linkedAuthors
+        imageMetadata = ownerImageMetadata
```
sophiamersmann: This rename confused me. Also, is the image metadata for authors used anywhere?

packages/@ourworldindata/grapher/src/footer/Footer.tsx
```diff
@@ -446,6 +481,43 @@ abstract class AbstractFooter<
...
+        // label's width in the source's wrap so long source names still wrap.
+        if (this.linkSourcesToDatapageMetadata) {
+            const label = "Data source: "
+            const labelWidth = Bounds.forText(label, {
+                fontSize: this.sourcesFontSize,
+                fontWeight: 700,
+            }).width
+            const sourceName = new MarkdownTextWrap({
+                text: this.sourcesLine,
+                maxWidth: this.sourcesMaxWidth - labelWidth,
```
sophiamersmann, line 504: This works for single-line sources, but if the source wraps into multiple lines than the second line doesn't use the full width available. I guess the visual artefact isn't a biggie but it's then possible for the actual height and the computed height to be out-of-sync, which might result in overflowing content. Using `SimpleMarkdownText` solves the problem, but then DoDs don't work... which doesn't matter now because the ten test pages don't have DoDs here, but not sure what a proper solution would look like. Probably adding an option to TextWrap/MarkdownTextWrap to offset the first line by a certain amount?

packages/@ourworldindata/grapher/src/sidePanel/SidePanel.tsx
```diff
@@ -10,6 +10,10 @@ export const SidePanel = ({
     children: React.ReactNode
 }) => {
     return (
+        // NOTE: site/IndicatorMetadataBox.scss narrows the indicator metadata
+        // box's grid when this side panel is present, via
+        // `.chart-key-info:has(.side-panel) ~ .indicator-metadata-box`. Removing
+        // or renaming this class will silently break that layout.
```
sophiamersmann: No need to say this again here, no?

site/gdocs/components/Byline.tsx
sophiamersmann: Unused though, right?

site/search/Autocomplete.tsx
```diff
@@ -437,13 +447,25 @@ export function Autocomplete({
...
+    // Optional id override. Defaults to "autocomplete" so existing call sites
+    // (topnav, homepage) keep working. Pass a unique id when more than one
+    // Autocomplete renders on the same page — the underlying library uses the
+    // container element to scope click-outside / blur detection, and duplicate
+    // ids prevent the panel from closing on outside clicks.
+    id = "autocomplete",
+    // Where this autocomplete is rendered (e.g. "topnav", "homepage",
+    // "datapage"). Attached to instant-search-click analytics so we can tell
+    // which search bar a click came from.
+    searchSource,
```
sophiamersmann, line 459: Maybe worth making this non-optional?

site/DataPageV2Content.tsx
```diff
@@ -163,99 +174,153 @@ export const DataPageV2Content = ({
...
-                                    experimentState &&
-                                    experimentState[
-                                        `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
-                                    ]?.isPageInExperiment &&
-                                    experimentState[
-                                        `${EXPERIMENT_PREFIX}-all-charts-vs-featured-v1`
-                                    ]?.arm === "featured-metrics"
-                                        ? "all-charts"
-                                        : ""
-                                }
```
sophiamersmann, line 229: Just double-checking that you wanted to get rid of this and the comment above?

site/IndicatorMetadataBox.tsx
```diff
@@ -0,0 +1,552 @@
...
+                                    <ArticleBlocks
+                                        blocks={faq.answer}
+                                        containerType="datapage"
+                                    />
+                                }
+                                onToggle={(isOpen) =>
+                                    logExpandableToggle(
+                                        // faq.question comes from the gdoc data
+                                        // (source English text), not the
+                                        // possibly-translated DOM.
```
sophiamersmann: tmi imo

site/RelatedDataCharts.tsx
```diff
@@ -0,0 +1,83 @@
...
+    chart: RelatedChart,
+    index: number
+): SearchChartHit {
+    return {
+        type: ChartRecordType.Chart,
+        slug: chart.slug,
+        title: chart.title,
+        variantName: chart.variantName ?? undefined,
+        availableEntities: [],
+        availableTabs: [],
```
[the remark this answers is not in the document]
sophiamersmann: Technically true, but only relevant for scatter-only charts where the data table references a scatter thumbnail. Scatter thumbnails are slightly taller than other chart types, so they have room for a couple more rows. So content wouldn’t overflow, we’d just have a couple too few table rows.

sophiamersmann (approved):
Looks good !!

packages/@ourworldindata/grapher/src/sidePanel/SidePanel.tsx
ikesau: Just want to make it really hard for regressions here in case you only look at one of 2 places before find-and-replacing. Shortened the comment down.

site/DataPageV2Content.tsx
ikesau: Yep, the experiment's expired so I removed the worst of the ternary gore. We should remove the whole path in a separate PR.

PR #6585, opened 2026-06-04
Index DIs in search immediately after (un)publishing

### LINKED ISSUE #6564: Index data insights to the main search (pages) index on publish/update
## Core problem

When a data insight is published/updated, it appears immediately in the `/latest` feed (`pages-chronological` index) but **not** in the main site search (`pages` index). It only shows up there after the weekly Monday reindex job (`indexPagesToAlgolia`).

This is because the publish flow only does an immediate `pages`-index update for types that pass `checkIsGdocPostExcludingFragments` (`adminSiteServer/apiRoutes/gdocs.ts`), which excludes `DataInsight` (see `checkIsGdocPostExcludingFragments` in `packages/@ourworldindata/utils/src/Util.ts`). Articles pass this check and so get indexed to search immediately; data insights don't.

## Desired behavior

On data insight publish/update (and unpublish), update the `pages` index immediately, the same way articles do via `indexIndividualGdocPost`.

## Notes

This is the data-insights analogue of #6003 ("Index charts to Algolia on publish/update"), which describes the same "wait till the Monday index job" problem for charts.

rakyi:
Fixes #6564

## Testing guidance

1. Set up Alogia indexing locally
2. Publish an unpublished data insight and check that it gets added to the pages index
3. Unpublish it and check that it gets removed

### DIFF
--- a/adminSiteServer/apiRoutes/gdocs.ts
+++ b/adminSiteServer/apiRoutes/gdocs.ts
@@ -340,14 +340,16 @@ async function indexAndBakeGdocIfNeccesary(
     const nextJson = nextGdoc.toJSON()
     const hasChanges = checkHasChanges(prevGdoc, nextGdoc)
     const action = getPublishingAction(prevJson, nextJson)
-    const isGdocPost = checkIsGdocPostExcludingFragments(nextJson)
+    const shouldIndexInPages =
+        checkIsGdocPostExcludingFragments(nextJson) ||
+        checkIsDataInsight(nextJson)
     const isProfile = checkIsProfile(nextJson)
     const isChronologicalPost = checkIsChronologicalGdoc(nextJson)
 
     await match(action)
         .with(GdocPublishingAction.SavingDraft, _.noop)
         .with(GdocPublishingAction.Publishing, async () => {
-            if (isGdocPost) {
+            if (shouldIndexInPages) {
                 await indexIndividualGdocPost(
                     nextJson,
                     trx,
@@ -365,7 +367,7 @@ async function indexAndBakeGdocIfNeccesary(
             await triggerStaticBuild(user, `${action} ${nextJson.slug}`)
         })
         .with(GdocPublishingAction.Updating, async () => {
-            if (isGdocPost) {
+            if (shouldIndexInPages) {
                 await indexIndividualGdocPost(nextJson, trx, prevGdoc.slug)
             }
             if (isProfile) {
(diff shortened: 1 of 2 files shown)


adminSiteServer/apiRoutes/gdocs.ts
```diff
@@ -588,8 +590,11 @@ export async function deleteGdoc(
         .where({ gdocId: id })
         .delete()
     if (gdoc.published) {
-        if (checkIsGdocPostExcludingFragments(gdoc)) {
-            await removeIndividualGdocPostFromIndex(gdoc)
+        if (
+            checkIsGdocPostExcludingFragments(gdoc) ||
+            gdoc.content.type === OwidGdocType.DataInsight
```
rakyi, line 600: We are not using `checkIsDataInsight` because our types are a mess and fixing them just for that is not worth it, since the normal equals check is enough.

rakyi:
@codex review

rakyi:
@codex review

adminSiteServer/apiRoutes/gdocs.ts
```diff
@@ -340,14 +340,21 @@ async function indexAndBakeGdocIfNeccesary(
...
     const isProfile = checkIsProfile(nextJson)
+    const wasChronologicalPost = checkIsChronologicalGdoc(prevJson)
     const isChronologicalPost = checkIsChronologicalGdoc(nextJson)
 
     await match(action)
         .with(GdocPublishingAction.SavingDraft, _.noop)
         .with(GdocPublishingAction.Publishing, async () => {
-            if (isGdocPost) {
+            if (shouldIndexInPages) {
                 await indexIndividualGdocPost(
```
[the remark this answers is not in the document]
rakyi: This is a real issue, we just assume that data insights and other gdoc posts will never have the same slug (currently they don't). So it should not happen in practice.

rakyi:
Codex is doing really well on the review here. Cool that we fixed several related bugs.

baker/algolia/utils/pages.ts
```diff
@@ -500,7 +500,7 @@ async function getExistingRecordsForSlug(
  *   - They're of the form `${gdoc.id}-c${chunkNumber}` but we don't know how many chunks exist
  */
 export async function indexIndividualGdocPost(
```
mlbrgl: nit: this function is now a bit of a misnomer

baker/algolia/utils/pages.ts
```diff
@@ -614,9 +619,7 @@ export async function getIndividualGdocRecords(
     )
 }
 
-export async function removeIndividualGdocPostFromIndex(
-    gdoc: OwidGdocPostInterface
-) {
+export async function removeIndividualGdocPostFromIndex(slug: string) {
```
mlbrgl: nit: same here, name is now a bit misaligned

mlbrgl (approved):
nice, thanks! Not sure if you're still working on it, but I think the collision issue is worth addressing. Not a big deal rn as you said given the absence of collisions for the current posts/insights, but it doesn't seem to be a big change to patch it.

rakyi:
The collision issue is quite a big change, I think. We'd need to either change the meaning of or rename the `slug` attribute of Algolia records. I'll create a new issue for it.

I'll change the function names, but our whole naming around gdocs sucks, e.g. the table is called `posts_gdocs`, but not all the rows are `GdocPost`.

PR #6590, opened 2026-06-04
🎉 Add catalogPath to charts (ETL identity)

pabloarosado:

Small, isolated PR into the chart-config feature branch (`etl-chart-config`), so it can be reviewed on its own.

## What

Adds a nullable, unique **`catalogPath`** column to the `charts` table, linking an ETL-authored chart to the ETL step that produced it — mirroring `multi_dim_data_pages.catalogPath`. This is the chart's **stable ETL identity** (e.g. `animal_welfare/latest/banning_of_chick_culling#banning_of_chick_culling`), distinct from the mutable, admin-managed `slug`.

Motivation: a single chart is a zero-dimension mdim, and mdims are already identified by `catalogPath` (slug is nullable/secondary there). Giving `charts` the same column lets the two be keyed identically — the foundation for unifying charts and mdims (owid/etl#6069).

## Changes

- **Migration** `1780587927881-AddCatalogPathToCharts` — `ADD COLUMN catalogPath VARCHAR(767) NULL` + `UNIQUE INDEX`. NULL allowed for hand-authored charts (multiple NULLs are fine under a MySQL unique index).
- **Type** — `catalogPath?: string \| null` on `DbInsertChart`.
- **Endpoint** — `putChartsChartIdEtlConfig` reads `?catalogPath=` and persists it via `COALESCE(?, c.catalogPath)` (so non-ETL callers never null it). This is the single chokepoint ETL hits on every push.

## Notes

- No change to `createChart`: the bootstrap row is created with `catalogPath = NULL`, then the etlConfig call fills it in within the same upsert.
- **chart-sync → prod** does not yet carry `catalogPath` (it writes `chart_configs.patch`, not the etlConfig endpoint). Out of scope here; tracked as a follow-up.

Pairs with the ETL-side PR.

### DIFF
--- a/adminSiteServer/apiRoutes/charts.ts
+++ b/adminSiteServer/apiRoutes/charts.ts
@@ -1033,6 +1033,10 @@ export async function putChartsChartIdEtlConfig(
 ) {
     const chartId = expectInt(req.params.chartId)
 
+    // ETL's stable identity for this chart (mirrors `multi_dim_data_pages.catalogPath`).
+    // Optional so other callers don't clobber it; persisted via COALESCE below.
+    const catalogPath = (req.query.catalogPath as string | undefined) ?? null
+
     let etlConfig: GrapherInterface
     try {
         etlConfig = migrateGrapherConfigToLatestVersionAndFailOnError(req.body)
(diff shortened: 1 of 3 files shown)


pabloarosado:

Superseded by #6599, which folds this catalogPath change into the row-model rework (`charts.configIdETL`). Closing this subbranch.

PR #6596, opened 2026-06-05
Improve alignment of Grapher controls

rakyi:
- Update custom icon sizes and viewBoxes to better match Font Awesome
- Add padding to text labels to improve their vertical alignment

| Before | After |
|--------|--------|
| admin-before | admin-after |
| map-line-scatter-before | map-line-scatter-after |
| dumbbell-marimekko-slope-before | dumbbell-marimekko-slope-after |

It's still not perfect, but much better IMO.

[Slack](https://owid.slack.com/archives/C46U9LXRR/p1780586846486279)


sophiamersmann (approved):
Thanks!

PR #6597, opened 2026-06-05
Fix react-aria portal provider import

Marigold:
## Summary
- import `UNSAFE_PortalProvider` from the public `react-aria` entrypoint instead of the unsupported `react-aria/PortalProvider` subpath
- fixes the Vite/Rolldown production build failure seen while building bespoke food-trade assets

## Testing
- `cd bespoke && yarn workspace @owid/bespoke-food-trade build`
- `cd bespoke && yarn build`
- `yarn testFormatChanged`
- `yarn testLintChanged`

Note: I did not run typecheck per agent instructions.


rakyi:
I randomly saw this, and I'm not sure if we have a bigger underlying root cuse here. FYI @marcelgerber since it seems vaguely related to a conversation about React Aria imports we had some time ago.

Marigold:
The real problem turned out to be outdated `node_modules/` folder. I tested this on the wrong server... Anyway, fixed it.

PR #6599, opened 2026-06-05
🔨🤖 Store ETL chart config as a separate config row (rows model) + catalogPath

pabloarosado:

Subbranch PR into `etl-chart-config`. Reworks how a chart's ETL-authored config is stored, from a **column** on `chart_configs` to a **separate config row** reached by a pointer — and folds in the `catalogPath` change (supersedes #6590).

## Why

Following Sophia's [feedback](https://app.notion.com/p/owid/Feedback-Adding-chart_configs-etlConfig-37574b71f92e805dbec2d746d4eae8d3?source=copy_link) on #6553: keep `chart_configs` a generic config store, and let each owner declare its own layers via pointers (exactly how `variables.grapherConfigIdETL` already works). This preserves "one row = one layer", reuses the existing config-row machinery (validation, R2, merge), and avoids baking a chart-specific concept into the shared `chart_configs` table.

## What changes vs the column approach

- **Schema:** drop the `chart_configs.etlConfig` column; add `charts.configIdETL char(36) UNIQUE NULL` → FK to `chart_configs(id)`, mirroring `variables.grapherConfigIdETL`. The chart's ETL config lives in its own `chart_configs` row (`patch == full`, absolute).
- **Merge (unchanged result):** the rendered `full` is still `merge(variableETL, etlConfig, patch)`. The middle layer is now read through the pointer (`LEFT JOIN chart_configs cc_etl ON cc_etl.id = c.configIdETL`) — in the admin-save path, `parent.json`, and the variable cascade (`updateAllChartsThatInheritFromIndicator`).
- **Endpoints:** `PUT /charts/:id/etlConfig` upserts the separate ETL config row and sets the pointer; `DELETE` clears the pointer then deletes the row. `deleteChart` also drops the orphaned ETL row. API request/response shapes are unchanged, so the admin editor needs no changes.
- **R2:** the ETL config row is an intermediate layer — MySQL-only, never uploaded (same as the variable-ETL rows).
- **MDIM views:** their ETL layer is dropped from the merge here; they get it via the link table in Phase 3.

## Folded in: `catalogPath` (was #6590)

Adds `charts.catalogPath` (ETL's stable identity, mirrors `multi_dim_data_pages.catalogPath`) and wires it into the ETL-config endpoint. Brought in here so the whole charts-table reshape is one reviewable PR; #6590 is superseded.

## Testing

- `yarn typecheck` passes.
- DB-backed tests (`adminSiteServer/tests/charts.test.ts`) and the migrations need a MySQL test DB — to be validated in CI (`make dbtest`). The two storage-specific assertions were updated to check the separate row via `configIdETL`. Logic closely mirrors the existing, tested variable-ETL pattern.
- ETL side (owid/etl#6174) is unaffected — the endpoint contract is identical.


pabloarosado:
@codex review

PR #6600, opened 2026-06-05
🐛 restore missing data in db export

ikesau:
https://github.com/owid/owid-grapher/pull/6571 broke staging auth by masking emails in the new DB export script, when the emails are needed for our auth middleware.

This PR reverts that change. We should come up with a better solution for this.


ikesau:
This wasn't the right fix. `githubUsername` is the field that's needed for staging auth with tailscale.

Fixed with https://github.com/owid/owid-grapher/commit/a4362927f4025a3c1b81b5f80d5562cd4fbe0b6e

PR #6603, opened 2026-06-09
Send a User-Agent when fetching indicator data from the data API

Marigold:
## What

The baker fetches `/v1/indicators/*.data.json` and `*.metadata.json` from the data API while baking, but the fetches set **no `User-Agent`**, so they go out as Node's default `node` UA. In GA4 that's **~69k requests/day to `api.ourworldindata.org/v1/indicators`** (from our Hetzner ASN 24940, referrer-less) that analytics counts as anonymous **external programmatic traffic** rather than our own infra.

## Change

Pass a descriptive `User-Agent` on the two data-API fetches in `loadVariableDataAndMetadata` (`packages/@ourworldindata/grapher/src/core/loadVariable.ts`):

```
owid-grapher/1.0 (Our World In Data; +https://ourworldindata.org)
```

The string contains "Our World In Data" so the existing `bot_classified_traffic` rule (`%our world in data%` → `OWID Internal`) classifies it without any analytics-side change.

## Safety

`loadVariableDataAndMetadata` also runs in the browser (chart views). `User-Agent` is a [forbidden header name](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) and is silently ignored by `fetch()` in the browser, so this only takes effect in Node (the baker and other server-side callers). Browser chart loads are unaffected.

## Context

Found while investigating a rising "files served" trend on the analytics Data Downloads dashboard — a large `Unknown`/external-`Programmatic` bucket turned out to be our own un-tagged infra. Companion to analytics#823 (which classifies datacenter-ASN scrapers). A sibling Python `aiohttp` service hitting the same endpoint (~121k/day) is being tracked separately.

### DIFF
--- a/db/model/Variable.ts
+++ b/db/model/Variable.ts
@@ -6,6 +6,7 @@ import {
     omitUndefinedValues,
     mergeGrapherConfigs,
     diffGrapherConfigs,
+    getOwidDataFetchUserAgent,
 } from "@ourworldindata/utils"
 import {
     getVariableDataRoute,
@@ -835,12 +836,23 @@ const createS3JsonParseError = (
     )
 }
 
+// Tag our server-side data API fetches with a descriptive User-Agent; see
+// getOwidDataFetchUserAgent. This code only ever runs in Node (the baker and
+// admin), so a value is always present, but we guard for undefined anyway.
+const dataFetchHeaders: HeadersInit | undefined = (() => {
+    const userAgent = getOwidDataFetchUserAgent()
+    return userAgent ? { "User-Agent": userAgent } : undefined
+})()
+
 export const fetchS3DataValuesByPath = async (
     dataPath: string
 ): Promise<OwidVariableMixedData> => {
     const resp = await retryPromise(
         () =>
-            fetch(dataPath, { keepalive: true }).then((response) => {
+            fetch(dataPath, {
+                keepalive: true,
+                headers: dataFetchHeaders,
+            }).then((response) => {
                 if (!response.ok) {
                     // Trigger retry
                     throw new Error(
@@ -867,7 +879,10 @@ export const fetchS3MetadataByPath = async (
 ): Promise<OwidVariableWithSourceAndDimension> => {
     const resp = await retryPromise(
         () =>
-            fetch(metadataPath, { keepalive: true }).then((response) => {
+            fetch(metadataPath, {
+                keepalive: true,
+                headers: dataFetchHeaders,
+            }).then((response) => {
                 if (!response.ok) {
                     // Trigger retry
                     throw new Error(
--- a/packages/@ourworldindata/grapher/src/core/loadVariable.ts
+++ b/packages/@ourworldindata/grapher/src/core/loadVariable.ts
@@ -11,13 +11,18 @@ import urljoin from "url-join"
 // indicator data while baking pages) so analytics can attribute them to OWID
 // rather than counting them as anonymous external traffic. Includes "Our World
 // In Data" so the existing bot-classification rule tags it as internal.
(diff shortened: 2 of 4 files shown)


Marigold:
Good catch on the Firefox case — fixed in e-latest. The header is now only attached when `typeof window === "undefined"` (Node), so the client chart path is unaffected and there's no cross-origin preflight risk. The intent really is server-side-only (the baker).

rakyi (commented):
There is also `fetchS3DataValuesByPath` and `fetchS3MetadataByPath` that are used on the server.

Even though we probably don't need it currently, completely overriding the user agent is quite dramatic, as we'd lose potentially useful information (the actual user agent). Could we use a custom header instead, e.g. `X-OWID-Request-Origin: grapher`, `etl`, etc.?

Another reason is that user agent is not precise, because we publicly recommend setting it to a custom one that contains "Our World in Data" in our data download docs to avoid Cloudflare misclassifying it to be a bot and responding with 403, see the screenshot:

Image

If you really want to keep using the user agent, could we extend the existing one instead of overriding it completely?

Of course none of this is a full solution without involving some sort of secret value that wouldn't be easily spoofed.

Marigold:
Trying to make sure I understand how the UA propagates on a download. Please correct me, I might be missing something! I think it's:

  1. User clicks download → browser requests `ourworldindata.org/grapher/<slug>.zip`. This hits our grapher Pages function and `analyticsMiddleware` logs it to GA4 with the **user's real browser UA**.
  2. To build the zip, that function calls `loadVariableDataAndMetadata`, which makes a **brand-new server-side `fetch()`** to `api.ourworldindata.org/v1/indicators/*.data.json`. This is *not* a proxied request and
  it does **not** forward the user's UA. Today it goes out with no `User-Agent` (the Workers runtime doesn't add one; the baker sends Node's `node`), which is why these show up as anonymous/`node` traffic. After
  this PR it sends `owid-grapher/1.0 (Our World In Data; …)`.
  3. That `/v1/indicators/*` request is served by a **separate worker** (`owid-api`), which proxies to R2 and logs it to GA4 via `logCfRequest` - capturing the UA from step 2 (empty/`node` today, our internal UA
  after the PR).

  So the user's UA and ours never mix: step 1 is logged with the visitor's UA, step 3 with ours. The download still counts as a real user; only the infra fetch behind it (step 3) gets tagged internal. Given that,
  I don't think we're losing a "real" UA by overriding - there isn't one on that leg to begin with.
  
Also good point with the snippet User agent, we should classify it as human in analytics. (But it can work alongside our internal UA, they're different)

And to your idea about custom header - I haven't thought of that, that would have been probably more elegant. I thought that since everyone (LLM providers) uses UA, we should use it too.

rakyi:
Yeah, your understanding is correct AFAICT. What I was trying to say is that if we override the user agent with a custom one, we lose the information of which requests come from node and which from the workers, for example (I wasn't aware that workers don't use a UA). That's why I was suggesting to at least extend it, instead of overrriding it.

Also, don't miss this part from my previous comment:

> There is also fetchS3DataValuesByPath and fetchS3MetadataByPath that are used on the server.

Marigold:
Good call. Rather than a fixed override, the UA now embeds the runtime's own identity (`navigator.userAgent` → `Node.js/<version>` for the baker, `Cloudflare-Workers` for the CF functions), so node-vs-worker
  stays distinguishable while still carrying "Our World In Data" for the bot rule. Not a literal append (fetch doesn't expose the outgoing UA, and workers don't send one), but it preserves the distinction you
  were after. Also covered `fetchS3DataValuesByPath` / `fetchS3MetadataByPath`. Could you re-review, please?

rakyi (approved):

PR #6605, opened 2026-06-11
🐛 classify related_charts in metadata export, catch missing tables in CI

Marigold:
## What

The nightly `mysql-export` Buildkite job failed:

```
Error: exportMetadata: the following tables are not classified in INCLUDE_DATA_TABLES or SCHEMA_ONLY_TABLES: related_charts.
```

The `related_charts` table (added in 165b956339) was never classified in `exportMetadata.ts`, so the script's backstop stopped the export.

- **Fix:** add `related_charts` to `INCLUDE_DATA_TABLES`. It's small and non-sensitive (chart→chart recommendation pairs with a score, written by the `etl related-charts` CLI, read by the data-page baker), so the public dump should carry its data.
- **CI guard:** move the classification lists into a side-effect-free module (`db/exportMetadataTables.ts` — `exportMetadata.ts` runs the export on import, so it can't be imported by tests) and add `db/tests/exportMetadata.test.ts`, which runs the same classification check against the freshly migrated test database. A PR that adds a `CREATE TABLE` migration without classifying the table now fails `make dbtest` in CI with the same actionable message, instead of breaking the nightly export.

## Verification

Ran `make dbtest` locally: the new test initially (and correctly) flagged `_test_db_ready`, the marker table created by the test-DB bootstrap — proving the detection works; it's now excluded as test infrastructure. Full suite passes: 71 tests in 11 files.

Caveat: the test only covers tables created by migrations. A table created directly in prod by an external system would still only surface at export time.

### DIFF
--- a/db/exportMetadata.ts
+++ b/db/exportMetadata.ts
@@ -4,11 +4,14 @@
 // runs the Grapher stack locally, so it shouldn't contain anything sensitive.
 //
 // By default, only each table's structure is exported, with no rows.
-// For data to be exported, it must be explicitly included in the INCLUDE_DATA_TABLES list
+// For data to be exported, it must be explicitly included in the
+// INCLUDE_DATA_TABLES list in exportMetadataTables.ts.
 //
 // As a backstop, the script fails if it encounters a table that isn't
-// classified in any of the lists below, forcing whoever added it to make a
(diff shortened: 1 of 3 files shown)


Marigold:
@codex review

PR #6607, opened 2026-06-11
enhance(mdim): set canonical url, include mdim views in sitemap for some mdims, and include mdim title in page title

marcelgerber:
We noticed that some Mdim views, like [`Number of newborns who have had a hepatitis B vaccine dose within the first 24 hours of delivery`](https://ourworldindata.org/grapher/vaccination-coverage-who-unicef?metric=vaccinated&antigen=hepb_bd), are not findable on Google _at all_ currently.

This PR makes three changes to Mdims, essentially:
- Set `<link rel="canonical">` in CF Functions, to ensure that the Mdim dimension choices are included in the canonical URL (so a `?metric=total` would be part of the canonical URL, but a `?tab=map` wouldn't)
- Enumerate all of these views in the sitemap
- Include the Mdim name in the page title, so we e.g. get `Share of children vaccinated, by vaccine, World | Childhood vaccination coverage - by vaccine | Our World in Data`

The sitemap change could have some negative SEO repercussions potentially, if Google finds that they don't like us indexing so many different views. But it's hard to know.

Antigravity says this:
> Including hundreds of similar page variations in a sitemap wastes crawl budget and causes keyword cannibalization, diluting the domain's overall search authority. To mitigate this, ensure all variations have self-referential canonical tags and only list the default and select, high-value variations rather than every possible permutation.

### Demo links:
- [Childhood vaccination Mdim](http://staging-site-mdim-canonical-url/grapher/vaccination-coverage-who-unicef?metric=unvaccinated&antigen=comparison)
- [Sitemap](http://staging-site-mdim-canonical-url/sitemap.xml)

TODO after merge:
- [x] Add a [Search Console annotation](https://support.google.com/webmasters/answer/16530728?hl=en)

### DIFF
--- a/baker/sitemap.ts
+++ b/baker/sitemap.ts
@@ -85,6 +85,14 @@ async function getPublishedGdocPosts(knex: db.KnexReadonlyTransaction) {
     )
 }
 
+const MDIM_INDEX_INDIVIDUAL_VIEWS: Record<
+    string,
+    boolean | Record<string, string>
+> = {
+    "vaccination-coverage-who-unicef": true,
+    "years-of-schooling": { sex: "both" }, // only index all views with sex=both
+}
+
 export const makeSitemap = async (
     explorerAdminServer: ExplorerAdminServer,
     knex: db.KnexReadonlyTransaction
@@ -175,22 +183,44 @@ export const makeSitemap = async (
 
                 const { slug, config } = multiDim
                 if (!slug) return []
-                if (!config.views || !Array.isArray(config.views)) {
+
+                const indexIndividualViews = MDIM_INDEX_INDIVIDUAL_VIEWS[slug]
+
+                if (!indexIndividualViews) {
                     return [
                         {
                             loc: urljoin(BAKED_GRAPHER_URL, slug),
                             lastmod,
                         },
                     ]
+                } else {
+                    return config.views.flatMap((view) => {
+                        if (
+                            typeof indexIndividualViews === "object" &&
+                            Object.entries(indexIndividualViews).some(
+                                ([dimension, value]) =>
+                                    view.dimensions[dimension] !== value
+                            )
+                        ) {
+                            return []
+                        }
+
+                        const searchParams = new URLSearchParams(
+                            view.dimensions
+                        )
+                        searchParams.sort()
+                        const queryStr = searchParams.toString()
+                        return [
+                            {
+                                loc: urljoin(
+                                    BAKED_GRAPHER_URL,
+                                    `${slug}?${queryStr}`
+                                ),
+                                lastmod,
+                            },
+                        ]
+                    })
                 }
-                return config.views.map((view) => {
-                    const searchParams = new URLSearchParams(view.dimensions)
-                    const queryStr = searchParams.toString()
-                    return {
-                        loc: urljoin(BAKED_GRAPHER_URL, `${slug}?${queryStr}`),
-                        lastmod,
-                    }
-                })
             })
         )
         .concat(
--- a/baker/sitemap.ts
+++ b/baker/sitemap.ts
@@ -6,10 +6,8 @@ import {
 import { dayjs, countries, getEntitiesForProfile } from "@ourworldindata/utils"
 import {
     DbPlainChart,
-    DbPlainMultiDimDataPage,
     DbPlainSlideshow,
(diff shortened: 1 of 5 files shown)


functions/_common/grapherTools.ts
```diff
@@ -389,6 +390,52 @@ export function rewriteMetaTags(
...
+                if (href && mdimDimensionsObj) {
+                    try {
+                        const searchParams = url.searchParams
+                        const newSearchParams = new URLSearchParams()
+                        // Ensure that the dimensions are set in dimension order, so we always have a consistent order.
+                        for (const [dim, defaultChoice] of Object.entries(
+                            mdimDimensionsObj
+                        )) {
+                            const value = searchParams.get(dim) ?? defaultChoice
+                            newSearchParams.set(dim, value)
```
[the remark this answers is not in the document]
marcelgerber: that's correct and unfortunate, but the only solution would be to include a comprehensive list of all available views and then filter them down.

I don't think we want to do that?

rakyi (commented):
I thought we wanted to treat a mutli-dim as the "page" and query params just as a way to interact with it and link to a specific view, similar to how search and latest work. Why do you think we need to change that? Is Ed aware of this proposed change?

Should we do the same for explorers?

baker/sitemap.ts
```diff
@@ -181,10 +170,28 @@ export const makeSitemap = async (
         )
         .concat(explorers.map(explorerToSitemapUrl))
         .concat(
-            multiDims.map((multiDim) => ({
-                loc: urljoin(BAKED_GRAPHER_URL, multiDim.slug),
-                lastmod: dayjs(multiDim.updatedAt).format("YYYY-MM-DD"),
-            }))
+            multiDims.flatMap((multiDim) => {
```
ikesau, line 181: I just checked and it seems the max sitemap size is 50k

We're currently at ~6500

We currently have ~650 multidim views

We have ~6500 explorer views

So we should be okay with this approach without having to filter down the URLs by "high distinctiveness/value only"

site/multiDim/MultiDimDataPage.tsx
```diff
@@ -81,6 +85,7 @@ export function MultiDimDataPage({
                 baseUrl={baseUrl}
                 staticAssetMap={assetMaps?.static}
                 archiveContext={archiveContext}
+                attrs={{ "data-owid-mdim-dimensions": mdimDimensions }}
```
ikesau: Could we maybe call this `-inital-view-dimensions` to be completely clear how it works?

ikesau (approved):
This makes sense to me! Tested it locally on wrangler and it all works as expected.

I say it's worth a try, but I'm open to arguments otherwise.

site/multiDim/MultiDimDataPage.tsx
rakyi: Also, it would be better if we didn't nest it into an object. I know that we should not render this part on the client (for now), but always creating a new object with the same properties causes unnecessary re-renders and is a React antipattern.

site/Head.tsx
```diff
@@ -49,11 +49,12 @@ export const Head = (props: {
     }
     staticAssetMap?: AssetMap
     archiveContext?: ArchiveContext
+    attrs?: Record<string, string | undefined>
```
rakyi, line 52: A more idiomatic solution would be to extend the `React.HTMLAttributes<HTMLHeadElement>` type, and then we should be able to pass `data-` attribute as props to the `<Head/>` component.

rakyi: A more complete example:

```tsx
 interface HeadProps extends React.HTMLAttributes<HTMLHeadElement> {
    title?: string
  }

  export function Head({ title, children, ...rest }: HeadProps) {
    return (
      <head {...rest}>
        {title && <title>{title}</title>}
        <meta charSet="utf-8" />
        {children}
      </head>
    )
  }
```

baker/sitemap.ts
marcelgerber: Thanks for checking, didn't think of that!

For reference, the new, updated commit has (all) 41 views for `vaccination-coverage-who-unicef`, and 9 views for `years-of-schooling` (filtered to `sex=both`).

marcelgerber:
@codex review

Issue #6483, opened 2026-05-13
Review CI/CD and automation security practices


rakyi:
Go through [this article by Astral](https://astral.sh/blog/open-source-security-at-astral) and apply any best practices that we are not currently doing. Use best judgment for which things apply to us, since we don't publish extremely popular packages, but do use GitHub Actions etc.

ikesau:
Putting an appetite of a day on this.

rakyi:
I added Pinact and Zizmor to grapher and etl.

PR #6612, opened 2026-06-15
Update and pin GitHub Actions and Workflows

rakyi:
Pin all actions except [our internal ones](https://github.com/owid/actions). See [this post](https://astral.sh/blog/open-source-security-at-astral) for rationale.

Part of #6483

PR #6615, opened 2026-06-15
Harden GitHub Actions workflow security with zizmor

rakyi:
### TL;DR

Harden GitHub Actions workflows against supply-chain and credential-leakage attacks by adding least-privilege permissions, disabling credential persistence on checkout, replacing `stefanzweifel/git-auto-commit-action` with inline push steps, and introducing a new `zizmor` security-scanning workflow.

### How to test?

Open a pull request and verify that the `Zizmor` check appears in the CI status list and that the `format` and `check-default-grapher-config` jobs still auto-commit formatting/config changes correctly without relying on `stefanzweifel/git-auto-commit-action`.

### Why make this change?

Unpinned or overly-privileged workflow steps are a common vector for supply-chain attacks. Persisted checkout credentials can be exfiltrated by malicious steps later in a job. Using a dedicated security scanner (`zizmor`) as a required CI check makes it easier to catch these issues automatically going forward, and the inline push approach removes a third-party action dependency while keeping the same auto-commit behaviour.

Part of #6483

### DIFF
--- a/.github/workflows/bespoke.yml
+++ b/.github/workflows/bespoke.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
 
       - uses: ./.github/actions/setup-node-yarn-deps
         with:
@@ -33,6 +35,8 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
 
       - uses: ./.github/actions/setup-node-yarn-deps
         with:
@@ -50,6 +54,8 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
 
       - uses: ./.github/actions/setup-node-yarn-deps
         with:
--- a/.github/workflows/bespoke.yml
+++ b/.github/workflows/bespoke.yml
@@ -6,6 +6,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
+
 defaults:
   run:
     working-directory: bespoke
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
     branches: [master]
 
+permissions:
+  contents: write
+
 # Runs `oxfmt` on a pull request so everything is nice and formatted!
 # If something's wrong, a new commit fixing the issues will automatically be pushed to the PR branch.
 
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -16,6 +16,7 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
+          persist-credentials: false
           ref: ${{ github.head_ref }}
 
       - uses: ./.github/actions/setup-node-yarn-deps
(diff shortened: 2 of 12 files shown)


rakyi:
* **#6615** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6612** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

rakyi:
As you should be able to see from the log of actions above, I verified that the `format.yml` workflow correctly commits and pushed the formatted code after the changes in this PR.

Edit: It seems that GitHub elides the logs, [here's the test commit](https://github.com/owid/owid-grapher/commit/19c84004a1d30c676b7c68ab33266a654bb0ea08) from the action.

.github/workflows/format.yml
```diff
@@ -25,6 +29,19 @@ jobs:
       - name: Run formatter
         run: yarn fixFormatAll
 
-      - uses: stefanzweifel/git-auto-commit-action@04702edda442b2e678b25b537cec683a1493fcb9 # v7.1.0
```
rakyi, line 28: This action relies on `persist-credentials: true`, which is insecure, so we switched to a manual inline script, which seems fine for our use case. The uv folks do something similar.

rakyi:
The Zizmor test check worked, see https://github.com/owid/owid-grapher/security/code-scanning/240

We have to configure [code scanning merge protection](https://docs.github.com/en/code-security/how-tos/find-and-fix-code-vulnerabilities/manage-your-configuration/set-code-scanning-merge-protection) for it to block merges.

rakyi:
Decided to switch Zizmor to a normal CI check rather than the advanced security mode, because it would be complicated to hook it up into our merge process to block merges on failure.

rakyi:
Zizmor [correctly fails on a test commit](https://github.com/owid/owid-grapher/actions/runs/27559297493/job/81466659339?pr=6615) and blocks the merge.

.github/workflows/bespoke.yml
```diff
@@ -6,6 +6,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
```
marcelgerber, line 10: Instead of manually specifying these all the time, should we just set the repo-wide default to read-only, and then explicitly set it to write in the few Actions that need write access? (AFAICT that's still possible)

Image

marcelgerber (commented):
Yup, overall this makes sense. But as I wrote, I'd set the default repo permissions to read-only; then we don't need to remember to set it always.

rakyi:
@codex review

.github/workflows/bespoke.yml
rakyi: Default read-only would be better, but Zizmor currently assumes the read-write default (the "default default") and [can't be configured otherwise AFAICT](https://github.com/orgs/zizmorcore/discussions/959). IMO it's better to keep using Zizmor rather than change the default permissions, since it checks for many other things as well.

It's not a question of remembering as the build will fail on Zizmor and prevent merge if we set the permissions incorrectly, similarly to type checking failure, so it should be fine.

We can discuss it on dev weekly, if you think otherwise.

marcelgerber: yup, makes sense to always have it be explicit.

but probably also doesn't hurt to set the repo-wide permission level to read-only, too. and I've just found out that [GH has changed the default in 2023](https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/), so changing it should be safe!

rakyi:
### Merge activity

* **Jun 17, 2:47 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6615).
* **Jun 17, 2:48 PM UTC**: @rakyi merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6615).

PR #6618, opened 2026-06-16
🔨 Keep hashed admin API keys out of the public metadata dump

Marigold:
## 👨 Summary

Generalize `make refresh.analytics` to `make refresh.private` and include admin keys there. Drop them from the public dump. If there's someone wanting to use grapher together with ETL locally, they can generate their own keys.

## What

`admin_api_keys` was shipped **with data** in the public `owid_metadata.sql.gz` (downloadable by anyone). That table stores SHA-256 hashes of admin API keys — authentication material that shouldn't be in a public artifact. This moves it into the **private sidecar dump**, which is generalised from the old analytics-only dump.

## How

- **`db/exportMetadataTables.ts`** — single source of truth, tables now classified into three tiers:
  - `PUBLIC_DATA_TABLES` (was `INCLUDE_DATA_TABLES`, minus `admin_api_keys`) — data ships publicly
  - `PRIVATE_DATA_TABLES` — `admin_api_keys` + `analytics_*`; structure public, **data only in the private sidecar**
  - `SCHEMA_ONLY_TABLES` — `donors`
- **`db/exportAnalyticsData.ts` → `db/exportPrivateData.ts`** — outputs `owid_private.sql`, dumps `PRIVATE_DATA_TABLES` (no more duplicated table list).
- Renamed docker scripts (`refresh-private-data.sh`, `download-grapher-private-mysql.sh`, `OWID_PRIVATE_DUMP_URL`) and the Makefile target **`refresh.analytics` → `refresh.private`**.
- Regression test: `admin_api_keys` must stay out of `PUBLIC_DATA_TABLES`.

## Why staging still works

Staging servers refresh from the same dumps and ETL authenticates against the staging admin API with a key whose hash lives in `admin_api_keys`. Keeping those rows in the **private** dump (imported via `make refresh.private`) preserves that, while removing them from the public one. External contributors get an empty `admin_api_keys` (they use cookie auth locally) — same graceful-skip behaviour as the analytics tables.

## ⚠️ Deploy ordering

Must land together with the matching **owid/ops** PR (its `mysql-export` now produces `r2:owid-private/owid_private.sql.gz` via `exportPrivateData.ts`, and staging calls `make refresh.private`). Until the ops export runs once, `owid_private.sql.gz` won't exist and `refresh.private` skips gracefully — so don't leave a gap, or staging's `admin_api_keys` ends up empty and ETL 401s. Old `owid_analytics.sql.gz` can be deleted from R2 afterwards.


Makefile
```diff
@@ -187,7 +187,7 @@ sync-cloudflare-images: node_modules
 	@echo '==> Syncing images table with Cloudflare Images'
 	@yarn syncCloudflareImages
 
-refresh.full: refresh refresh.analytics
+refresh.full: refresh refresh.private
```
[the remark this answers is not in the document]
Marigold: Good catch — fixed in fc33576. `refresh.atomic` now downloads the private sidecar and `atomic-grapher-data.sh` imports it into the replacement DB *before* the RENAME swap (both first-time and swap paths), so the live DB never exposes an empty `admin_api_keys` after a refresh. Falls back gracefully when the private dump isn't present.

Marigold:
@codex review

devTools/docker/download-grapher-private-mysql.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p $FOLDER
+
+if [[ -n "${OWID_PRIVATE_DUMP_URL:-}" ]]; then
+    echo "Downloading private dump (owid_private) from \$OWID_PRIVATE_DUMP_URL"
+    curl -fLo $FOLDER/owid_private.sql.gz "$OWID_PRIVATE_DUMP_URL"
+elif command -v rclone >/dev/null 2>&1 && rclone lsf r2:owid-private/owid_private.sql.gz >/dev/null 2>&1; then
+    echo "Downloading private dump (owid_private) from r2:owid-private"
+    rclone copyto r2:owid-private/owid_private.sql.gz $FOLDER/owid_private.sql.gz
+else
+    echo "No access to the private dump — skipping (private tables stay empty)"
```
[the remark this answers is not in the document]
Marigold: Fixed in 8fc72ab. `download-grapher-private-mysql.sh` now `rm -f`s any existing `owid_private.sql.gz` up front, so a run without access really does leave the private tables empty rather than re-importing a stale dump (which now carries `admin_api_keys`). Good find — the old analytics script had the same gap, but it matters more now that auth-key hashes ride along.

Marigold:
@codex review

devTools/docker/refresh-private-data.sh
```diff
@@ -19,8 +21,8 @@ fi
 : "${GRAPHER_DB_PORT:?Need to set GRAPHER_DB_PORT non-empty}"
 : "${DATA_FOLDER:?Need to set DATA_FOLDER non-empty}"
 
-if [ ! -f "${DATA_FOLDER}/owid_analytics.sql.gz" ]; then
-    echo "owid_analytics.sql.gz missing in ${DATA_FOLDER} — nothing to import (analytics tables stay empty)"
+if [ ! -f "${DATA_FOLDER}/owid_private.sql.gz" ]; then
+    echo "owid_private.sql.gz missing in ${DATA_FOLDER} — nothing to import (private tables stay empty)"
     exit 0
```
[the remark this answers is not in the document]
Marigold: Intentional — we're keeping graceful-skip rather than adding a required mode. Two reasons: (1) the same script path serves local devs and external contributors who legitimately have no `r2:owid-private` access and should get empty private tables (not a failed refresh), matching the prior analytics behaviour; (2) the dump is operationally guaranteed for staging — it's pre-seeded at `r2:owid-private/owid_private.sql.gz` and produced by the prod export alongside `owid_metadata.sql.gz`. So the staging/`refresh.full` path won't proceed with empty `admin_api_keys` in practice. Gating a 'required mode' on staging vs. dev would add branching to a shared script for a case the deploy guarantees away, so we're accepting absence-as-success deliberately.

devTools/docker/atomic-grapher-data.sh
```diff
@@ -51,6 +51,19 @@ import_db() {
     cat $1 | gunzip | sed s/.\*DEFINER\=\`.\*// | grep -vF GLOBAL.GTID_PURGED | _mysql $2
 }
 
+# Import the private sidecar dump (admin_api_keys + analytics_*) into $1 if it
+# was downloaded. We import it into the replacement DB *before* the atomic swap
+# so these tables (notably admin_api_keys, which gates API-key auth) are never
+# empty in the live DB after a refresh. Skips gracefully when not available.
```
rakyi: This comment is not useful and can become easily outdated.

devTools/docker/download-grapher-private-mysql.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env  bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# Downloads the PRIVATE sidecar dump: full data for the PRIVATE_DATA_TABLES
+# that the public owid_metadata.sql.gz ships schema-only — hashed admin API
+# keys and analytics_* view counts (see db/exportPrivateData.ts).
```
rakyi: I'd exclude restating the tables, otherwise we are just duplicating the same information in an actual variable and multiple different comments.

devTools/docker/refresh-private-data.sh
```diff
@@ -3,10 +3,12 @@ set -o errexit
...
 set -o nounset
 
-# Imports the private analytics dump (owid_analytics.sql.gz, downloaded by
-# download-grapher-analytics-mysql.sh) into the grapher database. The dump
-# contains DROP TABLE + CREATE TABLE + INSERTs for the analytics_* tables, so
-# importing fully replaces them.
+# Imports the private sidecar dump (owid_private.sql.gz, downloaded by
+# download-grapher-private-mysql.sh) into the grapher database. The dump
+# contains DROP TABLE + CREATE TABLE + INSERTs for the PRIVATE_DATA_TABLES
+# (hashed admin API keys + analytics_* view counts), so importing fully
```
rakyi: Let's not repeat the tables here again.

rakyi (approved):
Looks good overall, except a bit too comment heavy. Thank you!

devTools/docker/atomic-grapher-data.sh
Marigold: Done in 8204add — trimmed to a single line on why we import before the swap, no table list.

devTools/docker/download-grapher-private-mysql.sh
Marigold: Done in 8204add — dropped the table restatement, just point to db/exportPrivateData.ts now.

devTools/docker/refresh-private-data.sh
Marigold: Done in 8204add — removed the table list here too.

PR #6627, opened 2026-06-17
✨ Cache gdoc-independent queries in GDoc validate() to speed up baking

Marigold:
## Problem

`bakeGDocPosts` validates every published gdoc serially. `GdocBase.validate()` re-issued several **gdoc-independent** DB queries *per gdoc*:

- `getEnrichedStaticVizList` (global, uncached)
- `getMultiDimRedirectTargets` ×2 (grapher + explorer)

The whole bake runs in a single read-only transaction — i.e. **one connection** — so these queries can't overlap; each is a serialized round-trip. Locally (localhost DB) that's cheap, but in production the per-query RTT dominates, which is why the gdoc-posts step takes ~140s there vs ~15s locally.

## Change

Cache the gdoc-independent results for the lifetime of the transaction (the same `cachedInTransaction` pattern already used by `mapSlugsToIds`, `getDods`, etc. — safe because the bake transaction is read-only):

- `getMultiDimRedirectTargets` now caches the **full-prefix** fetch (`slugs === undefined`). `validate()` requests the full maps instead of the gdoc's own slugs. The full map is a strict superset keyed identically, so the `.get(link.target)` lookups in `validate()` are unchanged. Slug-specific callers (archival, `loadLinkedCharts`) are untouched.
- `getEnrichedStaticVizList` wrapped in `cachedInTransaction`.

## Verification

| Check | Result |
|---|---|
| Baked HTML output | **690/690 posts byte-identical** to master |
| DB queries during a local gdocPosts bake | **1,371 → 3** (681 multidim + 690 staticViz → 2 + 1) |
| Format / lint | pass |

The ~1,368 eliminated round-trips are the production win; at a typical remote RTT this should shave a substantial chunk off the gdoc-posts bake step. Local wall-clock is unchanged because localhost has no latency to recover.

Out of scope: the remaining bake cost is React rendering + in-memory attachment picking (CPU-bound), which would need parallelism rather than caching.

### DIFF
--- a/db/model/Gdoc/GdocBase.ts
+++ b/db/model/Gdoc/GdocBase.ts
@@ -1143,16 +1143,13 @@ export class GdocBase implements OwidGdocBaseInterface {
             getAllNarrativeChartNames(knex),
             getDods(knex).then((dods) => indexBy(dods, (dod) => dod.name)),
             getEnrichedStaticVizList(knex),
-            getMultiDimRedirectTargets(
-                knex,
-                this.linkedChartSlugs.grapher,
-                "/grapher/"
-            ),
-            getMultiDimRedirectTargets(
-                knex,
-                this.linkedChartSlugs.explorer,
-                "/explorers/"
-            ),
+            // Fetch the full (gdoc-independent) redirect maps so the result is
+            // cached for the whole transaction. Passing the gdoc's own slugs
+            // would issue an uncached query per gdoc during baking; the full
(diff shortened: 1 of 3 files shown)


Marigold:
@codex review

Marigold:
Closing: profiling on the staging server showed the validate-caching change addresses only ~8% of the gdoc-posts bake, and the real bottleneck (React SSR render, ~50%) has no simple fix — it needs render parallelization across cores, which is out of scope here. Findings captured separately.

PR #6639, opened 2026-06-24
Fix article excerpts in search

rakyi:
Use the excerpt as defined in the gdoc when there is no search query, i.e. empty search or topic/country filters only.

This is how it was supposed to be the whole time, and somehow we realized only now. See [Slack](https://owid.slack.com/archives/C6C0MCXU0/p1782234703301219).

PR #6640, opened 2026-06-24
Clean up search

rakyi:
- Add type parameters to Algolia methods
- Remove response type casting where possible
- Move search queries from promises to async/await syntax
- Add searchSingleForHits helper for single-query searches
- Remove explicit types where we can rely on inference
- Standardize analytics click event handlers
- Fix inconsistent position in analytics click events between the two layouts in SearchWritingResults
- Refactor a few places for readability


rakyi:
* **#6640** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6640?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6639** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6639?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

site/search/queries.ts
```diff
@@ -41,6 +35,14 @@ function makeStateForKey(state: SearchState) {
     return R.pick(state, ["query", "filters", "requireAllCountries"])
 }
 
+async function searchSingleForHits<T>(
+    liteSearchClient: LiteClient,
+    searchParams: Parameters<LiteClient["searchForHits"]>[0]
+) {
+    const response = await liteSearchClient.searchForHits<T>(searchParams)
+    return response.results[0]
+}
```
mlbrgl, line 44: nice!

site/search/evaluateSearch.ts
```diff
@@ -38,6 +38,10 @@ type ScoredQuery = {
     scores: Scores
 }
 
+type SearchEvaluationHit = {
+    slug: string
+}
```
mlbrgl: could be a Pick<PageRecord, "slug">

mlbrgl (approved):
Nice cleanup!

Do you mind updating CLAUDE.md with the rationale for not using explicit return types in this case? (the part where it says: Use type definitions for function params and return values)

While we're at it, I think we can get rid of IPageHit and IChartHit.

We can also probably get the hit fields (e.g. __position) from Hit<...> when we declare them on types, but no need to tackle that now.

Another cleanup opportunity is dropping instantsearch.js imports (and direct deps) and replacing them with algoliasearch's counterparts (since that's what we use as a client, not instantsearch).

With this we'll need to drop/update the silently broken __position attribute (which is returned by instantsearch but not algoliasearch).

rakyi:
I did most of your suggestions. I feel like the CLAUDE.md change warrants some discussion on dev weekly.

PR #6645, opened 2026-06-26
🐝 🤖 Clean up tsconfig base options

rakyi:
Switch `module` from `commonjs` to `preserve` so it matches our `bundler` module resolution and the ESM reality of our toolchain — the emitted JS is never run (tsx and Vite consume the TS source directly), so the previous CommonJS setting was both inert and inconsistent.

Drop options that are redundant with the inherited `@tsconfig/node24` base or with current TS defaults:
- `sourceMap` — emitted .js.map files are never used; nothing runs or debugs the tsc output.
- `alwaysStrict` — implied by the base's `strict: true`.
- `forceConsistentCasingInFileNames`, `allowJs` — already the defaults.

### DIFF
--- a/devTools/tsconfigs/tsconfig.base.json
+++ b/devTools/tsconfigs/tsconfig.base.json
@@ -1,9 +1,8 @@
 {
     "extends": "@tsconfig/node24/tsconfig.json",
     "compilerOptions": {
-        "module": "commonjs",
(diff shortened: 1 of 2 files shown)


rakyi:
@codex review

marcelgerber (approved):

PR #6558, opened 2026-05-29
🔨🤖 Tailored per-consumer content.toc (GdocToc data shape + migration)

mlbrgl:
Second PR in the sidebar-TOC-v2 stack — the data-shape layer.

`content.toc` becomes a tailored discriminated union, `GdocToc`, generated **only for pages that render a TOC** and shaped to exactly what that page's consumer needs:

- **sidebar** — pages with the `sidebar-toc` flag or an `ltp-toc` block (in practice, linear topic pages) → H1 sections, each carrying its chart bullets;
- **inline** — `sdg-toc` articles → a flat H2/H3 heading list;
- **otherwise** → no toc.

`generateGdocToc` is the single source of truth for this gating, so the consumers are dumb 1:1 renderers. Chart blocks gain an `anchorId` so the sidebar can deep-link into the page. A migration backfills the new shape for published pages.

#6578 swaps in the chart-aware sidebar that consumes this shape.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6557, opened 2026-05-29
🔨🤖 Disambiguate the two TableOfContents components + TOC cleanup

mlbrgl:
First PR in the sidebar-TOC stack — mostly internal prep and cleanup ahead of the H2-depth and UI work upstack. Only one user-facing change:

- **The SDG page's table of contents becomes a proper `<nav>` landmark.** It was rendering its links as an ARIA menu (`role="menu"`/`menuitem`), which told screen readers the anchor links were menu items with keyboard navigation that doesn't actually exist. It's now a `<nav>` with a plain list, matching the sidebar TOC, so assistive tech announces it as a list of links.

Everything else has no observable effect:

- **Renames the two `TableOfContents` components** so the later PRs are unambiguous: `SidebarTableOfContents` (the topic-page rail) and `SDGTableOfContents` (the `sdg-toc` `<details>` list), with matching CSS.
- **Removes dead code**: the never-invoked `runTableOfContents` hydrator, the `wrapInDiv` util it was the sole caller of, and the unused `entry-summary` gdoc block. `entry-summary` is obsolete — no published doc uses it, and any residual stored instance drops on the next re-parse.

Stacks up into: #6663 (include H2s + more block entries) → #6664 (new sidebar UI).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6664, opened 2026-06-30
✨🤖 Sidebar TOC UI on the flat TOC model

mlbrgl:
## Context

New sidebar table of contents for linear topic pages and profiles, built on the flat TOC model: each H1 is a section, with its H2s nested beneath it (replacing the chart-bullet concept).

To try it out, open a linear topic page and a profile page at the widths below.

**Desktop** (>1200px):

- It opens shortly after the page renders and stays open while you scroll (it starts collapsed on load, so there's no flicker while the page measures whether there's room). Links navigate to their section, and the row for the section you're reading highlights as you scroll — the sidebar keeps that active row visible within itself.
- It collapses out of the way while full-width content is on screen and reopens once you've scrolled past it, waiting until scrolling settles before reacting so it doesn't flip in and out mid-scroll. A toggle lets the reader collapse or expand it at any time, including over full-width content.
- On very wide screens (roughly ≥1875px) the sidebar has room to sit alongside the text column, so it stays open through full-text-width content and only steps aside for truly full-bleed blocks. At narrower desktop widths it also collapses for full-text-width content.
- Returning to the top of the page — scrolling up above the first section, or clicking "Back to top" — clears any `#section` left in the URL, so the address bar no longer points at a section the reader has scrolled away from. (Click a section link first to put `#section` in the URL, then check it's dropped.)
- "Back to top" at the top; a "See all charts on this topic" link at the bottom (omitted on archived pages, where the search it links to doesn't exist).
- Keyboard: tabbing to the collapsed sidebar's toggle and expanding it flows focus into the revealed links.

**Mobile** (≤1200px): a floating "Contents" pill opens the TOC as a drawer; choosing a link navigates and closes it. Resizing back to desktop while the drawer is open dismisses it.

**Also in this PR:** the old entry-sidebar styling is removed (`site/css/sidebar.scss` and the `.entry-sidebar` rules in `centered-article.scss`).

### Before merging


mlbrgl:
* **#6682** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6682?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6664** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6664?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6663** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6663?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6557** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6557?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6558](https://github.com/owid/owid-grapher/pull/6558) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6558?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

site/SidebarTableOfContents.tsx
```diff
@@ -1,218 +1,335 @@
...
+        <a
+            className="sidebar-toc__cta"
+            href={buildSearchHrefForCard(SearchResultType.DATA, tagName)}
+            data-track-note="toc_see_all_charts"
+        >
+            <FontAwesomeIcon icon={faMagnifyingGlass} />
+            {/* Full label by default; a container query swaps in the short one
+                in panels too narrow to fit it. */}
+            <span className="sidebar-toc__cta-label--full">
+                See all charts on this topic
```
ikesau, line 328: Image

Could we decrease the padding a tad so this doesn't wrap on when the screen is full width?

packages/@ourworldindata/types/src/domainTypes/Toc.ts
```diff
@@ -7,3 +7,10 @@ export interface TocHeading {
 export interface TocHeadingWithSupertitle extends TocHeading {
     supertitle?: string
 }
+
+// A top-level (h1) heading together with the h2 subheadings nested beneath it,
+// as displayed in the sidebar TOC.
+export interface TocSidebarSection {
```
ikesau, line 13: Love when the types for a PR are so simple 😅

site/search/searchState.ts
```diff
@@ -381,6 +382,25 @@ export function stateToSearchParams(state: SearchState): URLSearchParams {
     return params
 }
 
+/**
+ * Build a link to the site search, pre-filtered to a single topic and a given
+ * result type (Data / Writing). Used by the LTP "Sections" block and the
+ * sidebar TOC's bottom "See all charts" CTA.
+ */
+export function buildSearchHrefForCard(
```
ikesau, line 390: While we're centralizing this, could you update `AllCharts.tsx` to use this too?

site/SidebarTableOfContents.scss
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    .sidebar-toc__sidebar--collapsed {
+        border-right-color: transparent;
+    }
+}
+
+// ── Reduced motion ────────────────────────────────────────────────────────────
+
+@media (prefers-reduced-motion: reduce) {
```
ikesau, line 389: Sorry if this is a bit of a "no good deed goes unpunished" suggestion, but maybe we should also disable the auto-hiding behaviour _entirely_ if a user has a no-motion preference. (i.e. closed on pageload and only opening on direct interaction with the open button)

site/useClearHashOnReturnToTop.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Clears a lingering #section from the URL when the reader returns to the top
+ * of the page (scrolling up above the first section, or clicking "Back to
+ * top"), so it no longer points at a section they've scrolled away from.
+ *
+ * Fires only on the transition from a section to "" — `undefined` (scroll-spy
+ * hasn't measured yet) never counts, so a deep-link #section survives the
+ * initial render. replaceState avoids both a new history entry and a scroll
+ * jump.
+ */
+export const useClearHashOnReturnToTop = (
```
ikesau, line 13: Nice touch!

site/useWideBlockInView.ts
```diff
@@ -0,0 +1,88 @@
...
+        const fullBleed = new Set<Element>()
+        const wideCol = new Set<Element>()
+
+        const apply = () => {
+            setInView(
+                fullBleed.size > 0 || (!canWideColCoexist && wideCol.size > 0)
+            )
+        }
+
+        const observer = new IntersectionObserver(
```
ikesau, line 57: There's only one case of this in the wild, but the `time-use` topic page has a gray-section with only centered content in it, and I feel like that shouldn't hide the sidebar.

Not a big deal if it's a pain to fix, though.

https://github.com/user-attachments/assets/be36a427-91e9-4a39-a67f-6d5765a0c2de

site/useTocScrollSpy.ts
```diff
@@ -0,0 +1,69 @@
+import { useState, useEffect } from "react"
+
+/**
+ * Scroll-spy for the sidebar TOC. Given a document-ordered list of element ids
+ * (H1 section slugs interleaved with chart anchor ids), returns the id the
+ * reader is currently on: the last target whose top edge has scrolled above a
+ * fixed activation line near the top of the viewport. Pure function of scroll
```
ikesau, line 7: Maybe you meant it in a more colloquial sense, but it's not really a pure function is it? Side effects galore. And it doesn't take scrollY as its argument.

More tractably: I think this comment could be simplified 😅

site/useTocScrollSpy.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export function useTocScrollSpy(ids: string[]): string | undefined {
+    const [activeId, setActiveId] = useState<string | undefined>(undefined)
+
+    useEffect(() => {
+        if (ids.length === 0) return
+
+        let frame = 0
+
+        const computeActiveId = (): string => {
```
ikesau, line 28: No fix needed, but I found an edge case that breaks this >:)

Image

ikesau (approved):
Nice! tested on safari, chrome, ff, and iOS

Just a few small nits/suggestions

site/SidebarTableOfContents.scss
mlbrgl: no that makes sense, thanks. I can see how open/close on scroll can be unsettling. We could argue that they can always discard it for the current page if that's the case, but let's err on the side of caution (esp. given that they'd need to do that on every page load).

site/SidebarTableOfContents.tsx
mlbrgl: I'm surprised this pops up again, after the whole container query ceremony and testing with scrollbars on - but good catch! 👍

site/useWideBlockInView.ts
mlbrgl: yes you're right, visually that doesn't make sense. It's not impossible to fix, but it would add complexity for a scenario that is not mainstream in the current LTP model (IIRC time-use was an article before being promoted to LTP, without changing much of its structure). Also in this case, the sidebar would quickly slide out of the way anyway due to the key charts block coming right after.

site/useTocScrollSpy.ts
mlbrgl: Fair — dropped the "pure function" claim (you're right, side effects and no scrollY arg) and trimmed the comment down.

mlbrgl: also removed some of the edge cases that were only relevant when charts were i the mix (charts being in the dom but display:none depending on viewport)

site/useTocScrollSpy.ts
mlbrgl: oh yeah, the spy doesn't do 2D 🙂 Off the top of your head, do you know if we ever do that btw (side by sides with text + text vs text + chart)?

ikesau: No, I don't think we do. It's a completely unrealistic edge case 😮‍💨

mlbrgl:
### Merge activity

* **Jul 9, 7:14 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6664).
* **Jul 9, 7:19 AM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6664) rebased this pull request as part of a merge.
* **Jul 9, 7:19 AM UTC**: @mlbrgl merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6664).

PR #6680, opened 2026-07-01
Improve scroll to selected topic in /latest

rakyi:
Move the scroll to selected topic to a callback ref. This now works both for on load and on selection change.

We can use the element from the focus event as a scroll target. Also switch from "nearest" to "center", since otherwise the whole pill might not be visible.


ikesau (approved):
Much cleaner! Works well 👍

PR #6688, opened 2026-07-02
✨🤖 (time-interval) add timeInterval metadata, deprecate yearIsDay

sophiamersmann:
First step of migrating `yearIsDay` to `timeInterval` ([technical write-up](https://app.notion.com/p/owid/2026-02-16-Supporting-More-Time-Intervals-30974b71f92e801db7dcc0bc4cd79cb4?source=copy_link)): Adds support for `timeInterval` alongside `yearIsDay`.

Stacked PRs add support for more time intervals (week, month, quarter, decade), before dropping `yearIsDay`. After that, some more work is done to actually improve Grapher's presentation of non-yearly data.

---
This is **part 1 of 12 in a stack** made with GitButler:
- <kbd>&nbsp;12&nbsp;</kbd> #6752 
- <kbd>&nbsp;11&nbsp;</kbd> #6723 
- <kbd>&nbsp;10&nbsp;</kbd> #6722 
- <kbd>&nbsp;9&nbsp;</kbd> #6721 
- <kbd>&nbsp;8&nbsp;</kbd> #6699 
- <kbd>&nbsp;7&nbsp;</kbd> #6700 
- <kbd>&nbsp;6&nbsp;</kbd> #6696 
- <kbd>&nbsp;5&nbsp;</kbd> #6733 
- <kbd>&nbsp;4&nbsp;</kbd> #6751 
- <kbd>&nbsp;3&nbsp;</kbd> #6694 
- <kbd>&nbsp;2&nbsp;</kbd> #6689 
- <kbd>&nbsp;1&nbsp;</kbd> #6688 👈

PR #6689, opened 2026-07-02
🎉🤖 (time-interval) support more time intervals in metadata

sophiamersmann:
Add support for the new time intervals `week`, `month`, `quarter`, and `decade`.

This PR doesn’t use these new intervals yet (other than specifying custom formatting functions). That will happen in a later PR after the migration is complete.

All sub-yearly time intervals (day, week, month, and quarter) are represented internally by a date, allowing them to reuse the existing year/day table-joining machinery.

The given dates are snapped to the start of the interval (e.g. the first day of the month or week) because that simplifies the downstream code a bit. I’m slightly worried this might be surprising, but the chosen date shouldn’t carry any information or surface anywhere, so it should be fine.

---
This is **part 2 of 12 in a stack** made with GitButler:
- <kbd>&nbsp;12&nbsp;</kbd> #6752 
- <kbd>&nbsp;11&nbsp;</kbd> #6723 
- <kbd>&nbsp;10&nbsp;</kbd> #6722 
- <kbd>&nbsp;9&nbsp;</kbd> #6721 
- <kbd>&nbsp;8&nbsp;</kbd> #6699 
- <kbd>&nbsp;7&nbsp;</kbd> #6700 
- <kbd>&nbsp;6&nbsp;</kbd> #6696 
- <kbd>&nbsp;5&nbsp;</kbd> #6733 
- <kbd>&nbsp;4&nbsp;</kbd> #6751 
- <kbd>&nbsp;3&nbsp;</kbd> #6694 
- <kbd>&nbsp;2&nbsp;</kbd> #6689 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #6688


packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -1004,11 +1021,102 @@ class DateColumn<
...
+
+    // Fill gaps one month at a time. We space in month-space (not the raw
+    // days-since-epoch of the base class) because months are 28–31 days apart,
+    // which would make the base GCD collapse to a daily grid. This also lets the
+    // GCD respect the cadence, so e.g. bi-monthly data isn't force-filled.
+    // Observed times keep their original day; only fillers land on the 1st.
+    override getUniformlySpacedTimes(sortedTimes: number[]): number[] {
+        const dayByMonth = new Map<number, number>()
+        for (const time of sortedTimes)
+            dayByMonth.set(MonthColumn.monthsSinceEpoch(time), time)
```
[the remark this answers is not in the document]
sophiamersmann: Fixed upstream

packages/@ourworldindata/core-table/src/CoreTable.ts
```diff
@@ -488,9 +488,6 @@ export class CoreTable<
             this.columnsAsArray.find(
                 (col) => col instanceof ColumnTypeMap.Year
             ) ??
-            this.columnsAsArray.find(
-                (col) => col instanceof ColumnTypeMap.Quarter
-            ) ??
```
sophiamersmann, line 493: We replaced the existing `Quarter` column with a new one that extends ColumnTypeMap.Day which is why the extra check here is unnecessary. (The old `Quarter` column was only relevant for explorers, but no published explorer actually uses it)

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -937,108 +934,176 @@ class YearColumn<
...
+}
+
+// A sub-yearly quarter column, storing days-since-epoch
+// (a representative day per quarter)
 class QuarterColumn<
     TABLE_TYPE extends CoreTable = CoreTable,
     DEF_TYPE extends CoreColumnDef = CoreColumnDef,
-> extends TimeColumn<TABLE_TYPE, DEF_TYPE> {
-    preposition = "in"
+> extends DayColumn<TABLE_TYPE, DEF_TYPE> {
```
sophiamersmann, line 1091: QuarterColumn already existed (but with different semantics). It was only supported in Explorers, and none of our existing Explorers actually use it, so this PR repurposes it to represent the new QuarterColumn concept.

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -937,108 +934,176 @@ class YearColumn<
...
+    }
+
+    // The first year of the decade the given year belongs to, e.g. 2025 → 2020
+    private static startOfDecade(year: number): number {
+        return Math.floor(year / 10) * 10
+    }
+
+    override formatValue(value: number): string {
+        const decade = DecadeColumn.startOfDecade(value)
+        return decade < 0 ? `${Math.abs(decade)}s BCE` : `${decade}s` // "2020s"
```
sophiamersmann, line 1127: Not sure how Before Christ Decades should be formatted. I went with 2020s BCE

packages/@ourworldindata/grapher/src/core/LegacyToOwidTable.ts
```diff
@@ -716,15 +740,27 @@ const timeColumnValuesFromOwidVariable = (
...
-        : yearsRaw
+    let times = yearsNeedTransform
+        ? convertLegacyYears(years || [], display.zeroDay!)
+        : years || []
+
+    // Snap sub-yearly values (except plain days) to the start of their period,
+    // so variables that pick different representative days for the same period
+    // still align
+    if (isSubYearly(interval) && interval !== TimeInterval.Day)
+        times = times.map((time) => snapToIntervalStart(time, interval))
```
sophiamersmann, line 761: Here we’re snapping the underlying date to the start of the period (e.g. the first day of the month or week), which I can imagine is a bit surprising. This is done so that indicators with different representative days still align (although that feels like a fairly far-fetched scenario). But relying on the representative date always being the start of the interval also simplifies downstream code somewhat.

In theory, we could pick any arbitrary date since it should never be surfaced. In practice, though, it does become visible for weekly and quarterly data when manually selecting a time (via the date picker on hover), because weekly data reuses the day picker and quarterly data reuses the month picker. I don’t think that’s a big deal, so I decided to keep it for now.

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -937,108 +943,190 @@ class YearColumn<
...
+)
+// The week as a Monday–Sunday date range. Snaps to the ISO-week Monday
+// so indicators that pick a different representative day for the same
+// week still show the same label
+const memoFormatWeek = _.memoize((value: number): string => {
+    const monday = convertDaysSinceEpochToDate(value).startOf("isoWeek")
+    const sunday = monday.add(6, "days")
+    return monday.month() === sunday.month()
+        ? `${monday.format("MMM D")}–${sunday.format("D, YYYY")}` // "Jan 16–22, 2023"
+        : `${monday.format("MMM D")} – ${sunday.format("MMM D, YYYY")}` // "Jan 30 – Feb 5, 2023"
```
sophiamersmann, line 984: I thought this was nicer than 'W12 2026', but it's also quite long... I also considered 'Week of Jan 30, 2026', but that would have been even longer.

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -937,108 +943,190 @@ class YearColumn<
...
+})
+const memoFormatWeekShort = _.memoize((value: number): string => {
+    const date = convertDaysSinceEpochToDate(value)
+    return `W${date.isoWeek()} ${date.isoWeekYear()}` // "W3 2023"
```
sophiamersmann, line 988: This is only used in the timeline on smaller screens; otherwise, the start and end labels would be too long, leaving no space for the timeline itself.

packages/@ourworldindata/core-table/src/CoreTableColumns.test.ts
```diff
@@ -19,33 +18,75 @@ describe(ColumnTypeNames.Quarter, () => {
...
+        expect(col.formatTimeShort(0)).toEqual("W4 2020")
+        expect(col.formatForCsv(0)).toEqual("2020-W04")
+    })
+
+    it("formats weeks across month and year boundaries", () => {
+        // day 6 = 2020-01-27, a Monday; the week ends in February
+        expect(col.formatValue(6)).toEqual("Jan 27 – Feb 2, 2020")
+        // day -22 = Monday 2019-12-30, ISO week 1 of 2020: the range shows
+        // the week's end year, the compact format the ISO week year
+        expect(col.formatValue(-22)).toEqual("Dec 30 – Jan 5, 2020")
```
marcelgerber: should we make this even more explicit, for added clarity?
```suggestion
        expect(col.formatValue(-22)).toEqual("Dec 30, 2019 – Jan 5, 2020")
```

packages/@ourworldindata/core-table/src/CoreTableColumns.test.ts
```diff
@@ -19,33 +18,75 @@ describe(ColumnTypeNames.Quarter, () => {
...
+        expect(col.formatTimeShort(-22)).toEqual("W1 2020")
```
marcelgerber, line 89: Can we add another test for 
```ts
expect(col.formatValue(-20)).toEqual("Dec 30 – Jan 5, 2020")
```

to check that `-20` (2020-01-01) properly maps to the same week, starting in 2019?

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
marcelgerber: Yeah, it is quite long...

I think in a tooltip it's fine, but in the chart title it really is rather long.

I know it's a good bunch more effort, but maybe an option here would be to write out the short number in the chart title (W32 2020), and then provide a DoD-style tooltip on it that gives detail: "Calendar week 32 in 2020, from Mon 3 Aug to Sun 9 Aug".

marcelgerber (approved):
nice, well done!

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
sophiamersmann: Good point about the title. I was mainly thinking about the axis and the timeline where the timeline's start/end time switch to 'W1 2026' at some point because otherwise the labels would take too much space. I think using the shorter one in the title as well is a good call, but I'll hold off on the tooltip because that might interfere with the map-tolerance tooltip we might add after changing the style of the entity/time suffix.

PR #6696, opened 2026-07-03
✨🤖 (time-interval) interval-aware uniform spacing for stacked bar charts

sophiamersmann:
Stacked bar charts ensure that bars are uniformly spaced by filling in missing time points. For example, if the data contains 2010, 2011, and 2013, then 2012 is inserted automatically.

This behavior had previously been disabled for daily data because weekly and monthly indicators were incorrectly tagged as daily. Enforcing uniform spacing in those cases resulted in many empty slots being inserted between real data points.

This PR extends the same behavior to the new sub-yearly intervals. It remains disabled for daily data for now, since during the transition there are still some indicators incorrectly marked as daily. Once the migration is complete, we can remove the flag.

---
This is **part 2 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;6&nbsp;</kbd> #6752 
- <kbd>&nbsp;5&nbsp;</kbd> #6843 
- <kbd>&nbsp;4&nbsp;</kbd> #6823 
- <kbd>&nbsp;3&nbsp;</kbd> #6733 
- <kbd>&nbsp;2&nbsp;</kbd> #6696 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #6700

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreTableColumns.ts
+++ b/packages/@ourworldindata/core-table/src/CoreTableColumns.ts
@@ -6,6 +6,8 @@ import {
     formatDay,
     convertDaysSinceEpochToDate,
     convertDateToDaysSinceEpoch,
+    EPOCH_DATE,
+    withUniformSpacing,
     sortNumeric,
     omitUndefinedValues,
     isPresent,
@@ -928,6 +930,15 @@ export abstract class TimeColumn<
     override parse(val: unknown): number | ErrorValue {
         return parseInt(String(val))
     }
+
+    /**
+     * Given the sorted, de-duplicated times present in the data, returns them
+     * with gaps filled so that positions are uniformly spaced (one filler per
+     * missing step)
+     */
+    getUniformlySpacedTimes(sortedTimes: number[]): number[] {
+        return withUniformSpacing(sortedTimes)
+    }
 }
 
 class YearColumn<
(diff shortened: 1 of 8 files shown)


packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -1028,6 +1039,42 @@ class MonthColumn<
...
+    }
+
+    // Fill gaps one month at a time. We space in month-space (not the raw
+    // days-since-epoch of the base class) because months are 28–31 days apart,
+    // which would make the base GCD collapse to a daily grid. This also lets the
+    // GCD respect the cadence, so e.g. bi-monthly data isn't force-filled.
+    override getUniformlySpacedTimes(sortedTimes: number[]): number[] {
+        const dayByMonth = new Map<number, number>()
+        for (const time of sortedTimes)
+            dayByMonth.set(MonthColumn.monthsSinceEpoch(time), time)
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Claude Fable 5 — @sophiamersmann at the wheel._

This scenario can't occur through the data pipeline: all sub-yearly times (except plain days) are snapped to the start of their interval at load time — see `timeColumnValuesFromOwidVariable` in `LegacyToOwidTable.ts`, which maps every value through `snapToIntervalStart` per variable before any joining. So every observation in a given calendar month carries the same day number (the first of the month) regardless of which representative day an indicator originally stored, and `sortedTimes` can never contain two distinct days for the same month. The `dayByMonth` de-duplication is a no-op on real data, and the exact-position lookup in `withMissingValuesAsZeroes` always hits.

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -1037,6 +1048,42 @@ class MonthColumn<
...
+        const spacedMonths = withUniformSpacing([...dayByMonth.keys()])
+        return spacedMonths.map(
+            (month) =>
+                dayByMonth.get(month) ?? MonthColumn.daysAtStartOfMonth(month)
+        )
```
marcelgerber: Since we only ever use `dayByMonth.keys()`, this can just be a set instead.

And if we don't want to run `monthsSinceEpoch` a lot of times (which is probably sensible!), then we should also run `uniq` on sortedTimes.

packages/@ourworldindata/core-table/src/CoreTableColumns.ts
```diff
@@ -1081,6 +1128,43 @@ class QuarterColumn<
...
+    override getUniformlySpacedTimes(sortedTimes: number[]): number[] {
+        const dayByQuarter = new Map<number, number>()
+        for (const time of sortedTimes)
+            dayByQuarter.set(QuarterColumn.quartersSinceEpoch(time), time)
+        const spacedQuarters = withUniformSpacing([...dayByQuarter.keys()])
+        return spacedQuarters.map(
+            (quarter) =>
+                dayByQuarter.get(quarter) ??
+                QuarterColumn.daysAtStartOfQuarter(quarter)
+        )
```
marcelgerber: Same applies here.

packages/@ourworldindata/grapher/src/stackedCharts/StackedBarChartState.ts
```diff
@@ -12,9 +12,16 @@ import { ColorScaleManager } from "../color/ColorScale.js"
...
     ColorSchemeName,
+    ColumnTypeNames,
 } from "@ourworldindata/types"
 import { ChartManager } from "../chart/ChartManager.js"
 
+// Whether daily data should get uniform spacing (one filler per missing day).
+// Disabled during the transition to explicit time intervals: weekly indicators
+// are currently still marked as daily, and spacing them at daily granularity
+// would be wrong. Flip to `true` once those indicators use timeInterval "week".
+const ENABLE_DAILY_UNIFORM_SPACING = false
```
marcelgerber, line 24: I think weekly data should be fine, as there is a regular cadence of 7. It's just monthly (and quarterly) data that is off.

marcelgerber (approved):
looks good!

PR #6699, opened 2026-07-03
✨🤖 (time-interval) calendar-aware ticks & labels for monthly axes

sophiamersmann:
## What

Improves the horizontal time axis for **monthly** indicators (interval `month`, stored day-encoded).

**Tick positions** — instead of d3 picking round *day-numbers* (which land mid-month), monthly ticks now snap to calendar-nice, January-anchored boundaries via a month-step ladder (1/2/3/6 months, then 1/2/5/10/… years), chosen so the tick count stays near the target.

**Tick labels** — redundant years are dropped:
- **Continuous axes** (line/area): bare years at yearly cadence (`2020 · 2021 · 2022`); at sub-year cadence, month-only with the year on the first tick and each January (`Jan 2020 · Apr · Jul · Oct · Jan 2021`). A trailing mid-year endpoint that would read as a false year boundary is dropped.
- **Discrete axes** (stacked bars): the month is always kept, since each label identifies a specific bar and a bare `2025` would misread a monthly bar as a whole year (`Jan 2020 · Jan 2021 · …`).

## How

- New self-contained helpers in `axis/timeAxisTicks.ts` (`getMonthlyAxisTickValues`, `getMonthlyAxisTickLabels`), using only the general date utilities — no column-layer changes.
- `Axis.getTickValues` uses the generator when the axis's time column is monthly; `Axis.tickLabels` applies the label suppression (skipped for author-supplied `config.ticks`, e.g. slope/sparkline, which keep fully-qualified endpoints).
- Monthly stacked bars stop emitting a per-bar `ticks` list so they pick up the generated calendar-nice set (like stacked area already does), keeping `domainValues` for bar positioning.

All behavior is gated on the time column's interval being `Month`; every other column keeps its existing d3 tick behavior.

## Tests

Unit tests for both helpers (step selection, first-of-month invariants, continuous vs discrete labels, endpoint handling) plus axis- and stacked-bar-level integration tests.

---
This is **part 3 of 9 in a stack** made with GitButler:
- <kbd>&nbsp;9&nbsp;</kbd> #6810 
- <kbd>&nbsp;8&nbsp;</kbd> #6752 
- <kbd>&nbsp;7&nbsp;</kbd> #6733 
- <kbd>&nbsp;6&nbsp;</kbd> #6723 
- <kbd>&nbsp;5&nbsp;</kbd> #6722 
- <kbd>&nbsp;4&nbsp;</kbd> #6721 
- <kbd>&nbsp;3&nbsp;</kbd> #6699 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #6696 
- <kbd>&nbsp;1&nbsp;</kbd> #6700

### DIFF
--- a/packages/@ourworldindata/grapher/src/axis/Axis.test.ts
+++ b/packages/@ourworldindata/grapher/src/axis/Axis.test.ts
@@ -2,13 +2,23 @@ import { expect, it, describe } from "vitest"
 import * as R from "remeda"
 
 import { HorizontalAxis } from "../axis/Axis"
-import { ScaleType, AxisConfigInterface } from "@ourworldindata/types"
 import {
+    ScaleType,
+    AxisConfigInterface,
+    ColumnTypeNames,
+} from "@ourworldindata/types"
+import {
+    OwidTable,
     SynthesizeFruitTable,
     SynthesizeGDPTable,
 } from "@ourworldindata/core-table"
 import { AxisConfig } from "./AxisConfig"
-import { AxisAlign } from "@ourworldindata/utils"
+import {
+    AxisAlign,
+    dayjs,
+    convertDateToDaysSinceEpoch,
+    convertDaysSinceEpochToDate,
+} from "@ourworldindata/utils"
 
 it("can create an axis", () => {
     const axisConfig = new AxisConfig({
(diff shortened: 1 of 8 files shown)


sophiamersmann:

Superseded by #6823, which consolidates the monthly/daily/weekly/quarterly axis-ticks branches into a single branch so they can be reviewed together.

PR #6700, opened 2026-07-03
✨🤖 (time-interval) month-aware timeline manual input

sophiamersmann:
When manually editing a timeline handle, monthly charts now show a month/year date field without a day segment, rather than the full day/month/year picker.

Unfortunately, React Aria doesn’t support the granularities we need, so the month/year picker is implemented in a slightly hacky way. For now, I haven’t attempted custom solutions for weekly and quarterly data because it’s not yet clear to me what those should look like. Quarterly charts also use the month/year picker, though, since that’s still a bit better than the day picker.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

---
This is **part 1 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;6&nbsp;</kbd> #6752 
- <kbd>&nbsp;5&nbsp;</kbd> #6843 
- <kbd>&nbsp;4&nbsp;</kbd> #6823 
- <kbd>&nbsp;3&nbsp;</kbd> #6733 
- <kbd>&nbsp;2&nbsp;</kbd> #6696 
- <kbd>&nbsp;1&nbsp;</kbd> #6700 👈

### DIFF
--- a/packages/@ourworldindata/grapher/src/timeline/TimelineComponent.scss
+++ b/packages/@ourworldindata/grapher/src/timeline/TimelineComponent.scss
@@ -281,8 +281,12 @@ $timelineHeight: 32px;
     display: flex;
     align-items: center;
     justify-content: center;
-    min-width: 120px;
     cursor: text;
+    min-width: 120px;
+
+    &.GrapherTimeline__DateInput--month {
+        min-width: 84px;
+    }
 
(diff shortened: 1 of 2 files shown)


marcelgerber (approved):
nice! this works well, and I don't think it's too hacky

PR #6708, opened 2026-07-04
feat(mdim): Redirect from explorers to Mdims, incorporating source query params

marcelgerber:
Lets a single explorer/grapher source redirect to different multi-dim targets depending on the incoming query params.

- **DB:** `sourceQueryParams` column on `multi_dim_redirects`.
- **Matcher:** a serializable query-param decision tree (`QueryParamDecisionTree` in `@ourworldindata/utils`) that resolves (slug + query params) → target by rule specificity, with redundant-branch pruning.
- **Baking:** one decision tree per explorer source slug, baked to `explorers/_explorerRedirects.json`.
- **CF function:** resolves the baked tree against the request's query params, applies the target view's params, and strips explorer-only source params.
- **Admin:** optional "source query params" field on the multi-dim redirect form (shown in both listings); multiple redirects per source, differing by query params, are now allowed.

Based on `remove-wp-redirects` (not `master`) since it builds on that branch.

## Review comments

Martin, I've changed this over to single-row-per-redirect after all, unlike we discussed before.
The main reason for this is that this way, we can make proper use of the DB table's foreign keys linking out to the slug & specific Mdim view, which I find useful.
It also easily enables us to split the outgoing links from one explorer into multiple Mdims, if needed.
For performant matching, we are building up a decision tree for each explorer.

## TODO
- [ ] Check all uses of the `getMultiDimRedirectTargets` function - they don't take source query params into account
  - e.g. resolving gdocs, archiving, ...
- [ ] Should we change `_grapherRedirects.json` to the decision tree code, too?
- [x] Actually test this...!


marcelgerber:
@codex review

marcelgerber:
@codex is the decision tree approach sensible? are there any other changes you'd make, architecturally?

PR #6715, opened 2026-07-08
🎉 Search: show closest matches when a query has no exact results

Marigold:
When a section's search returns nothing, the query is retried once with Algolia's `removeWordsIfNoResults: allOptional` **as a query-time parameter** (not an index setting) and the section renders in a *closest matches* state instead of staying blank.

## How it works

This isn't Algolia's default behavior, and turning the flag on doesn't give you "the closest matches" for free either — both need explaining:

- Algolia's default (`removeWordsIfNoResults: none`) never drops query words on its own. A query with no exact match just returns 0 hits, full stop.
- Setting it to `allOptional` makes Algolia retry the query treating every word as droppable, and it returns **everything** that matches at least one word, ranked by how many words matched. That's a big, noisy list — not a curated one. Left as-is, a broadened query can come back with "186 results" where only a handful are actually close.

So this PR adds two things on top of the raw Algolia flag:

1. **Tier cut** — keep only the hits tied for the *best* match quality, drop the rest. The relaxed request sets `getRankingInfo: true`, which exposes `_rankingInfo.words` per hit (how many query words that specific hit matched). We read the top hit's word count and discard every hit that matched fewer words. That's the difference between a raw `allOptional` retry ("186 results", mostly one-word noise) and what the section actually shows (~10, all genuinely close).
2. **Distinctiveness check** — even the *best* hit can be a false signal if it only shares one word with the query and that word is common. E.g. searching `"world cup"` and matching only on "world" isn't a real rescue — "world" appears in thousands of unrelated documents. So when the top hit's tier is a single word, we also check how many documents that word matches index-wide (the relaxed query's own `nbHits`) and only treat it as a genuine rescue below 100 matches. `"malaria worldwide"` also only rescues on one shared word ("worldwide" matches nothing) — but "malaria" appears in only 43 documents, so it passes the distinctiveness check and the section shows the malaria hits as closest matches. `"world cup"` fails the same check and correctly stays on the honest empty state.

The fallback only fires when the exact search already came back empty, so every search that works today is byte-for-byte identical and pays no extra request. Applies to the Data, Research & Writing, and Data Insights sections (`queryCharts`, `queryArticles`, `queryDataInsights`).

**Not at parity with `/api/search`.** Our public search API (`functions/api/search/`) hits the same Algolia indices but has no equivalent fallback — a zero-result query there still just returns `nbHits: 0`. The technique would carry over (it's a plain Algolia REST parameter, not a client-library feature), but that endpoint is a documented, external, versioned contract with its own zero-hit special case (invalid-topic validation) to sequence against, so it's being left as a separate follow-up rather than folded into this PR. It's also missing the other half of this PR: `SearchApiResponse`/`SearchPagesApiResponse` have no field equivalent to the "No exact matches — showing the closest matches instead" banner, so even a ported fallback would need a new response field (and an OpenAPI doc update) before API consumers could tell relaxed results apart from exact ones. Flagged in a comment above `searchCharts`/`searchPages` in `searchApi.ts`.

**Fixed from Codex review:** two real bugs, both in commit 2a4cbebe82 — (1) the one-word-tier fallback reported the full relaxed `nbHits` while `nbPages` was always 1, so `useInfiniteSearchOffset` (articles/topics/profiles pagination) kept requesting further pages that could never load anything; `nbHits` now always equals what's actually returned. (2) The Research & Writing section showed the closest-matches notice and mixed in rescued articles whenever the *articles* subquery alone was empty, even when topic pages or profiles already had exact hits — now gated on the whole section (articles + topic pages + profiles) being empty, since topic pages and profiles never go through the fallback.

**Supersedes the index-setting experiments (#6676 charts, #6714 pages).** Doing it query-time makes "this result set is relaxed" **exact** (no stopword heuristics — the index-level setting relaxes silently and the frontend can't reliably tell), needs no index changes, and applies uniformly to charts / articles / data insights. #6714 can be closed in favor of this mechanism.

**Spot-check on staging** (All tab): [estimated daily excess deaths during covid, world](http://staging-site-search-closest-matches/search?resultType=all&q=estimated%20daily%20excess%20deaths%20during%20covid%2C%20world) · [malaria worldwide](http://staging-site-search-closest-matches/search?resultType=all&q=malaria%20worldwide) (distinctive one-word rescue, "malaria" matches 43 documents index-wide, below the threshold) · [world cup](http://staging-site-search-closest-matches/search?resultType=all&q=world%20cup) (stays empty by design) · [country profile](http://staging-site-search-closest-matches/search?resultType=all&q=country%20profile) (stays empty; naive shows 9,177 results)

## Evaluation

*(2026-07-08, ~950/440/440 real queries per surface, LLM-judged, volume-weighted)*

Production vs this PR's behavior, evaluated end-to-end with the owid/analytics search-demand evaluator (owid/analytics#907; the pages surfaces strip country terms from the sample first, since the country filter is compositional noise):

| | charts | writing | data insights |
|---|---|---|---|
| zero-result searches | 34.2% → **11.5%** | 34.5% → **14.2%** | 52.1% → **14.5%** |
| queries served (judge rel≥1.5) | 52.0% → **59.5%** | 39.7% → **46.8%** | 36.5% → **40.9%** |
| net demand newly served | **+7.5%** | **+6.7%** | **+4.2%** |
| regressions | 0.1% | 0% | 0% |
| NDCG@5 | 0.60 → **0.72** | — | — |

Versus the naive index-level `allOptional` (same rescues, no display logic): identical serve rates, but naive's median result count on rescued queries is ~90–1,000 ("9,177 results" for `country profile`) at 59–80% judged-irrelevant, where this PR shows a median of 4–8 with the junk tiers cut and a fifth to a third of queries kept honestly empty. Full interactive per-query reports: vibe PR owid/vibe-webapps#44.


site/search/queries.ts
```diff
@@ -43,6 +47,84 @@ async function searchSingleForHits<T>(
...
+
+    // Algolia ranks relaxed hits by matched words, so the best tier is a
+    // prefix. For a one-word tier, EVERY match shares that word, so the full
+    // count is the tier size; multi-word tiers may extend past the fetched
+    // page, where the page length is a (honest) undercount.
+    const tier = relaxed.hits.filter((hit) => words(hit) === topWords)
+    return {
+        ...relaxed,
+        hits: tier,
+        nbHits: topWords === 1 ? relaxed.nbHits : tier.length,
```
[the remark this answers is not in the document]
Marigold: Confirmed, thanks — fixed in 2a4cbebe82. `nbHits` for the one-word tier now always equals `tier.length` (matching the multi-word branch) instead of the full relaxed `nbHits`. Since `nbPages` here is unconditionally 1, advertising a larger total than what's actually returned was internally inconsistent regardless of which consumer read it — it just happened to bite `useInfiniteSearchOffset` specifically, since `useInfiniteSearch` (charts/data-insights) only checks `page < nbPages - 1` and never looks at `nbHits` to decide whether to continue.

site/search/queries.ts
```diff
@@ -292,7 +380,10 @@ export async function queryArticles(
         },
     ]
 
-    return searchSingleForHits<FlatArticleHit>(liteSearchClient, searchParams)
+    return searchSingleForHitsWithClosestMatches<FlatArticleHit>(
+        liteSearchClient,
+        searchParams
+    )
```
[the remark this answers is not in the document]
Marigold: Confirmed, thanks — fixed in 2a4cbebe82. Since `queryTopicPages`/`queryProfiles` never go through the closest-matches fallback, any of their hits are exact, so the section only counts as genuinely rescued when articles, topic pages, and profiles are *all* empty. Added `sectionHasExactHitsElsewhere` (topics or profiles > 0) in `SearchWritingResults.tsx`: when true, the rescued articles are dropped entirely (not just the notice) — they're excluded from `orderedHits`, the `MultiColumnResults` articles prop, and `totalCount`, so a query that already succeeded via topic pages/profiles renders exactly as it would if articles had simply been at parity (empty), with no closest-match content or notice mixed in.\n\nOne residual imperfection I'm leaving as-is rather than restructuring further: `topicFirstPageSize`/`noArticles` (used to size topics' own first fetch) are computed from the raw `articlesQuery.totalResults` *before* topics or profiles have resolved, so in the rare case where a rescued-articles collision happens, topics' first page may request slightly fewer items than ideal (since it reserved slots for articles that end up hidden). It's a minor first-page sizing nicety, not a correctness issue — nothing wrong renders, and pagination still works normally. Fixing it fully would mean deferring the article fallback decision until topics/profiles are known, which is a bigger sequencing change than this fix warrants.

site/search/SearchClosestMatchesNotice.tsx
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Shown when a section displays relaxed ("closest matches") results because
+ * the exact search returned nothing — see the closest-matches fallback in
+ * queries.ts. Tells the user honestly that we don't have an exact answer,
+ * while still offering the nearest content instead of a blank page.
+ */
+export function SearchClosestMatchesNotice() {
+    const { state } = useSearchContext()
+    return (
+        <p className="search-closest-matches-notice">
```
ikesau, line 12: Was this asked for by Marwa or Joe? Or just something Claude thought we needed?

I think it's far too loud personally, perhaps even unnecessary. Users only care about whether the results are relevant, not how our heuristic works.

The one exception is typos, I think (e.g. Google's "Did you mean **receive**") but that's not really what this PR has added (and is some Algolia already resolves behind the scenes, anyway)

site/search/queries.ts
```diff
@@ -43,6 +47,88 @@ async function searchSingleForHits<T>(
...
+    query?: string
+    page?: number
+    offset?: number
+}
+
+type RankedHit = { _rankingInfo?: { words?: number } }
+
+async function searchSingleForHitsWithClosestMatches<T>(
+    liteSearchClient: LiteClient,
+    searchParams: SingleSearchRequest[]
```
ikesau: Based on all the usage I can see, I don't see any benefit to requiring the argument to be in an array. 

I know the underlying API requires it, but we could do that wrapping in `searchSingleForHits` and simplify a lot of this code, no?

site/search/queries.ts
```diff
@@ -31,6 +31,10 @@ import {
 } from "./searchUtils.js"
 import { RichDataComponentVariant } from "./SearchChartHitRichDataTypes.js"
 
+// Above this many index-wide hits, a single shared word is treated as too
+// common to be a distinctive match (see rationale below).
+const DISTINCTIVE_SINGLE_WORD_MAX_HITS = 100
```
ikesau, line 40: OOC, was this qualitatively assessed? Or just feels right?

site/search/queries.ts
```diff
@@ -43,6 +47,88 @@ async function searchSingleForHits<T>(
...
+    const relaxedRequest: SingleSearchRequest = {
+        ...request,
+        removeWordsIfNoResults: "allOptional",
+        getRankingInfo: true,
+    }
+    const relaxed = await searchSingleForHits<T>(liteSearchClient, [
+        relaxedRequest,
+    ] as Parameters<LiteClient["searchForHits"]>[0])
+
+    const words = (hit: T) => (hit as RankedHit)._rankingInfo?.words ?? 0
```
ikesau: `getWords` please 🙂

site/search/queries.ts
```diff
@@ -43,6 +47,88 @@ async function searchSingleForHits<T>(
...
+        return primary
+
+    // Algolia ranks relaxed hits by matched words, so the best tier is a
+    // prefix of the fetched page. nbPages is always 1 below — closest matches
+    // are a single, non-paginated tier — so nbHits must equal tier.length,
+    // never the full relaxed.nbHits: offset/length consumers
+    // (useInfiniteSearchOffset) keep requesting further pages while
+    // nbHits > items loaded so far, and the next request's offset > 0 skips
+    // this fallback entirely and returns the original empty exact page,
+    // producing a "load more" that can never load anything.
```
ikesau: This is a confusing comment. I think it's maybe overloaded or too terse. Could you try to reexplain it?

ikesau (approved):
Really nice to get this in! Thanks for pushing it as far as you did 🙂

site/search/SearchClosestMatchesNotice.tsx
Marigold: Me and @danyx23 thought it'd be a good idea to let users know. Let's discuss it later today.

site/search/queries.ts
Marigold: I tuned it on a couple of examples. It's of course not optimal in a quantitative sense, but I found that it doesn't make difference whether it's 50, 100 or 200.

site/search/queries.ts
Marigold: Fixed

site/search/queries.ts
Marigold: Done!

site/search/queries.ts
Marigold: You're right, explained in more detail.

PR #6725, opened 2026-07-08
🔨🤖 (chart-titles) add TextWrapGroup, replacing MarkdownTextWrap.fromFragments

sophiamersmann:
## Context

Groundwork for rendering the entity/time suffix of chart titles in a smaller, muted style (see the next PR in the stack). That requires laying out pieces of text with different font styles as one wrapped unit — something we couldn't do well before.

This PR introduces **`TextWrapGroup`**: a text wrap that takes a list of text fragments, each with its own font style (family, size, weight, color), and wraps them together into lines. By default, a fragment is kept intact where possible: it continues on the current line only if it fits there in its entirety, and otherwise moves to a new line as a whole (where it may wrap internally if it's longer than a full line).

It generalizes and replaces `MarkdownTextWrap.fromFragments`, which could only distinguish fragments by boldness. Under the hood it reuses the existing `MarkdownTextWrap` machinery, so markdown, details-on-demand, and HTML/SVG rendering all work per fragment.

## New classes

- **`TextWrapGroup`** — the new public API. Takes a group-level style plus a list of `TextWrapFragment`s (text + optional style overrides + placement: `avoid-wrap` / `continue-line` / `always`, and optional inline/new-line gaps). Each fragment is converted to IR tokens (markdown-parsed or plaintext) and the merged token stream is laid out as one unit.
- **`AbstractTokenTextWrap`** — base class extracted from `MarkdownTextWrap`, now shared by both. Subclasses only supply the IR token stream; the base class implements everything downstream: line-splitting against `maxWidth`, metrics (width/height/line count), and HTML/SVG rendering incl. baseline math. `MarkdownTextWrap` itself is unchanged in behavior — it's the same pipeline, just with the token source made pluggable.
- **`IRFragment`** — a new IR element token that represents one fragment in the merged stream. It carries the fragment's style *delta* relative to the group (rendering no wrapper element at all when a fragment matches the group style, which keeps existing SVG output byte-identical) and knows about the gap preceding it, so line-breaking treats the gap as a break opportunity just like whitespace.

### DIFF
--- a/packages/@ourworldindata/components/src/MarkdownTextWrap/AbstractTokenTextWrap.tsx
+++ b/packages/@ourworldindata/components/src/MarkdownTextWrap/AbstractTokenTextWrap.tsx
@@ -5,6 +5,7 @@ import {
     imemo,
     FontFamily,
     VerticalAlign,
+    type RequiredBy,
 } from "@ourworldindata/utils"
 import * as R from "remeda"
 import {
@@ -33,19 +34,59 @@ export interface TokenTextWrapOptions {
     detailsOrderedByReference: string[]
 }
 
+export type TokenTextWrapProps = {
+    fontSize: number
+    maxWidth?: number
+    lineHeight?: number
+    fontWeight?: number
+    fontFamily?: FontFamily
+    verticalAlign?: VerticalAlign
+    style?: CSSProperties
+    detailsOrderedByReference?: string[]
+}
+
 /**
  * Base class for text wraps that lay out a stream of IR tokens.
  *
- * Subclasses only provide the token stream (via the abstract `text`, `tokens`
- * and `options` getters); this class implements everything downstream of it:
+ * Subclasses only provide the token stream (via the abstract `text` and
+ * `tokens` getters); this class implements everything downstream of it:
  * - line-splitting against `maxWidth` (`htmlLines`, `svgLines`)
  * - metrics (`width`, `height`, `lineCount`, per-line heights and gaps)
  *
  * Known subclasses: `MarkdownTextWrap` (tokens parsed from a markdown string)
  * and `TextWrapGroup` (tokens assembled from multiple fragments).
  */
-export abstract class AbstractTokenTextWrap implements ITextWrap {
-    protected abstract get options(): TokenTextWrapOptions
+export abstract class AbstractTokenTextWrap<
+    Props extends TokenTextWrapProps = TokenTextWrapProps,
+> implements ITextWrap {
+    private static readonly defaultOptions = {
+        maxWidth: Infinity,
+        lineHeight: 1.1,
+        verticalAlign: VerticalAlign.bottom,
+        detailsOrderedByReference: [] as string[],
+    } as const satisfies Partial<TokenTextWrapProps>
+
+    private readonly initialProps: Props
+    constructor(props: Props) {
+        this.initialProps = props
+    }
+
+    @imemo get props(): RequiredBy<
+        Props,
+        keyof typeof AbstractTokenTextWrap.defaultOptions
+    > {
+        return {
+            ...AbstractTokenTextWrap.defaultOptions,
+            ...this.initialProps,
+        }
+    }
+
+    protected get options(): TokenTextWrapOptions {
+        // The cast is safe since the defaults guarantee that all required
+        // options are set, but TS can't verify this for a generic Props
+        return this.props as TokenTextWrapOptions
+    }
+
     abstract get text(): string
     abstract get tokens(): IRToken[]
 
--- a/packages/@ourworldindata/components/src/MarkdownTextWrap/AbstractTokenTextWrap.tsx
+++ b/packages/@ourworldindata/components/src/MarkdownTextWrap/AbstractTokenTextWrap.tsx
@@ -6,9 +6,11 @@ import {
     FontFamily,
     VerticalAlign,
 } from "@ourworldindata/utils"
-import { match } from "ts-pattern"
 import * as R from "remeda"
-import { type ITextWrap } from "../TextWrap/TextWrap.js"
+import {
+    getTextPositionForSvgRendering,
+    type ITextWrap,
+} from "../TextWrap/TextWrap.js"
 import {
     appendReferenceNumbers,
     getLineFontSize,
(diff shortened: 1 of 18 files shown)


sophiamersmann:
* **#6905** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6905?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6887** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6743** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6726** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6726?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6725** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6725?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6725?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/components/src/MarkdownTextWrap/AbstractTokenTextWrap.tsx
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * Subclasses only provide the token stream (via the abstract `text`, `tokens`
+ * and `options` getters); this class implements everything downstream of it:
+ * - line-splitting against `maxWidth` (`htmlLines`, `svgLines`)
+ * - metrics (`width`, `height`, `lineCount`, per-line heights and gaps)
+ *
+ * Known subclasses: `MarkdownTextWrap` (tokens parsed from a markdown string)
+ * and `TextWrapGroup` (tokens assembled from multiple fragments).
+ */
+export abstract class AbstractTokenTextWrap implements ITextWrap {
```
sophiamersmann: Most of the content of this class is copy-pasted from MarkdownTextWrap

packages/@ourworldindata/components/src/MarkdownTextWrap/IRTokens.tsx
```diff
@@ -0,0 +1,752 @@
+import * as _ from "lodash-es"
```
sophiamersmann, line 1: Also just copy-pasted into its own file from MarkdownTextWrap

packages/@ourworldindata/components/src/MarkdownTextWrap/IRTokens.tsx
```diff
@@ -0,0 +1,752 @@
...
+
+export interface IRFragmentStyle {
+    fontSize?: number
+    fontFamily?: FontFamily
+    fontWeight?: number
+    color?: string
+}
+
+/** Wraps the tokens of a single TextWrapGroup fragment */
+export class IRFragment extends IRElement {
```
sophiamersmann, line 226: This is the new 'fragment' element

packages/@ourworldindata/components/src/MarkdownTextWrap/TextWrapGroup.tsx
```diff
@@ -0,0 +1,251 @@
...
+     * Horizontal gap between the preceding content and this fragment.
+     * Only applies when the fragment is placed inline (continues a line)
+     */
+    inlineGap?: number
+
+    /**
+     * Vertical gap between the preceding line and this fragment.
+     * Only applies when the fragment is placed on a new line
+     */
+    newLineGap?: number
```
sophiamersmann, line 53: These options are added because I needed them for the new chart titles

packages/@ourworldindata/components/src/MarkdownTextWrap/MarkdownTextWrap.tsx
```diff
@@ -1086,7 +436,7 @@ export function canAppendTextToLastLine({
 
     const spaceWidth = Bounds.forText(" ", { fontSize }).width
     const availableWidthInLastLine =
-        maxWidth - lastLineWidth - spaceWidth - reservedWidth - 10 // 10px wiggle room
```
sophiamersmann, line 1089: Dropping the 10px wiggle room here is what causes the differences reported by the SVG tester. I think the wiggle room was helpful when text measurements could be slightly off, but now that we use the actual character widths of the fonts we render with, I don’t think it’s necessary anymore.

packages/@ourworldindata/components/src/MarkdownTextWrap/MarkdownTextWrapComponents.tsx
```diff
@@ -21,7 +30,7 @@ function MarkdownTextWrapLine({
 export function MarkdownTextWrapHtml({
```
sophiamersmann, line 30: Wasn't sure if I should rename these components now since they're also used to render TextWrapGroup which _may_ contain Markdown, but doesn’t necessarily.

packages/@ourworldindata/components/src/MarkdownTextWrap/TextWrapGroup.tsx
```diff
@@ -0,0 +1,251 @@
...
+        super()
+        this.initialProps = props
+    }
+
+    @imemo get props(): RequiredBy<
+        TextWrapGroupProps,
+        keyof typeof TextWrapGroup.defaultOptions
+    > {
+        return { ...TextWrapGroup.defaultOptions, ...this.initialProps }
+    }
```
marcelgerber: These are shared exactly between this and MarkdownTextWrap, suggesting that we can move them to the abstract base class.

packages/@ourworldindata/components/src/MarkdownTextWrap/MarkdownTextWrapComponents.tsx
marcelgerber: Guess we can just do a sweep and rename them to `TextWrapHtml`?

marcelgerber (approved):
nice!

sophiamersmann:
### Merge activity

* **Aug 7, 7:39 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6725).
* **Aug 7, 7:40 AM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6725) rebased this pull request as part of a merge.
* **Aug 7, 7:41 AM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/6725).

PR #6732, opened 2026-07-09
🎉🤖 (site) iframe embed mode for bespoke visualizations

sophiamersmann:
Let external sites embed a bespoke visualization by iframing its article page. This isn't advertised yet for any of our existing bespoke components.

Setting `iframeEmbed: true` on a bespoke component enables the behavior:

```yaml
{.bespoke-component}
  bundle: causes-of-death
  variant: treemap
  {.config}
    urlSync: true
    iframeEmbed: true
  {}
{}
```

One limitation of iframes is that they can't automatically resize to fit their content. To work around this, the embedded page reports the visualization's rendered height to the parent window via `postMessage`. It sends an `owid-bespoke-embed-height` message on load and whenever the height changes.

Embedding sites that want the iframe to fit its content need to add a listener like this:

```html
<script>
    window.addEventListener("message", (event) => {
        if (event.origin !== "https://ourworldindata.org") return
        if (event.data?.type !== "owid-bespoke-embed-height") return
        for (const iframe of document.querySelectorAll("iframe")) {
            if (iframe.contentWindow === event.source) {
                iframe.height = event.data.height
            }
        }
    })
</script>
```

Grapher charts don't need this because they set the height to 100vw and then the rest flows from there. However, that's not how bespoke data viz are implemented, and it would be quite a bit of work to rewrite them.

---
- <kbd>&nbsp;2&nbsp;</kbd> #6732 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #6731

### DIFF
--- a/site/gdocs/OwidGdocPage.tsx
+++ b/site/gdocs/OwidGdocPage.tsx
@@ -1,5 +1,6 @@
 import * as _ from "lodash-es"
 import { Head } from "../Head.js"
+import { IFrameDetector } from "../IframeDetector.js"
 import { SiteHeader } from "../SiteHeader.js"
 import { SiteFooter } from "../SiteFooter.js"
 import { CitationMeta } from "../CitationMeta.js"
@@ -22,6 +23,7 @@ import { match, P } from "ts-pattern"
 import {
     ARCHIVED_THUMBNAIL_FILENAME,
     ArchiveContext,
+    OwidEnrichedGdocBlock,
     OwidGdocDataInsightInterface,
     OwidGdocPostInterface,
     OwidGdocProfileInterface,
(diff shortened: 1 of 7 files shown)


site/gdocs/OwidGdocPage.tsx
```diff
@@ -123,6 +125,25 @@ function isJsonLdArticlePredicate(
...
+/**
+ * A page qualifies for bespoke iframe embedding if it has exactly one
+ * top-level bespoke component with `iframeEmbed: true` in its config.
+ * When such a page is loaded inside an iframe, all site chrome and article
+ * content is hidden and only the flagged bespoke component is shown.
+ */
+function checkIsBespokeIframeEmbedPage(gdoc: OwidGdocUnionType): boolean {
+    const body = "body" in gdoc.content ? gdoc.content.body : undefined
+    if (!body) return false
+    return body.filter(isIframeEmbeddableBespokeBlock).length === 1
```
ikesau, line 144: We should probably add validation in gdocs to ensure that only one bespoke-viz component has `iframeEmbed: true`

Checkout [validateBody](https://github.com/owid/owid-grapher/blob/6fa053278f66c9fe371eabdf49e214904ace6653/adminSiteClient/gdocsValidation.ts#L73)

site/gdocs/components/BespokeComponent.scss
```diff
@@ -1,5 +1,23 @@
 @use "sass:color";
 
+html.IsInIframe body.GdocBespokeIframeEmbedPage {
```
ikesau, line 3: We might want to add

```css
html.IsInIframe:has(body.GdocBespokeIframeEmbedPage) {
    scrollbar-gutter: stable;
}
```

To fix the height oscillation issue (though it adds ~15px of margin-right to the viz on desktops and is only at [92% support](https://caniuse.com/wf-scrollbar-gutter))

ikesau (commented):
This works elegantly!

Even though this would be a silent launch, thinking ahead - are we sure this is the canonical way we want people to embed these? Supporting embeds adds URL stability/redirect resolution obligations and this would conflate our article namespace with our bespoke viz namespace.

I know Joe is thinking about having a venue for all our bespoke viz, and I wonder if, with that, we might add something like https://ourworldindata.org/viz/causes-of-death which, if so, would probably be the better place to handle embeds.

I'm not sure the extent to which you've already had these conversations though. Maybe it's something you and others have already thought through and are fine with?

sophiamersmann:
I based this on [this prototype](https://github.com/owid/owid-grapher/pull/6496) by Daniel, but I don't think there's been much discussion around this. I have now reached out to Joe! Thanks Ike :)

sophiamersmann:
We’ll have a shaping session about standalone bespoke viz pages next week, so this code will likely become obsolete. Closing this for now, thank you!

PR #6740, opened 2026-07-10
🐛🤖 Cap sub-1 sig-fig rounding, stop abbreviating thousands in compact labels

### LINKED ISSUE #6157: Sig fig rounding is odd for small numbers
## Description

For a chart like [this](https://ourworldindata.org/grapher/causes-of-death-in-children-under-5?country=~FIN), significant figures are great to round numbers for big values (World on the left), but produce some really strange values (Finland on the right). 

## Expected behaviour

Ideally we'd be able to apply sigfigs and round numbers to the nearest integer

We already use the `numDecimalPlaces` setting in the table when sig fig rounding is enabled. We could also use it for small numbers more generally. Marcel suggested to simply try both rounding strategies and pick the number with the lower precison

## Screenshots

Image
Image

sophiamersmann:

Fixes #6157 and #5172.

Two ends of the same problem: number formatting showed precision nobody reads — spuriously precise small values in sig-fig mode, and pointless `k` abbreviations on compact labels.

## Small values (#6157)

With `roundingMode: significantFigures`, big values format nicely ("1.23 million") but small values came out spuriously precise — e.g. "0.902" or "0.000826" deaths on [causes-of-death-in-children-under-5](https://ourworldindata.org/grapher/causes-of-death-in-children-under-5?country=~FIN) when the indicator is configured with 0 decimal places.

When rounding a value below 1 to significant figures would show more decimal places than `numDecimalPlaces`, the value is now rounded to `numDecimalPlaces` instead. Values of 1 or more always keep sig-fig rounding. Positive values below the cap's resolution render as e.g. "<1"; negative values that would round to zero display as zero.

| value (0 decimal places configured) | before | after | comment |
| --- | --- | --- | --- |
| 0.902 | 0.902 | <1 | sig figs would show 3 decimals but the indicator says whole numbers; below the cap's resolution → "<" |
| 0.4 | 0.400 | <1 | same cap; "<1" rather than a misleading "0" |
| −0.4 | -0.400 | 0 | capped rendering rounds to zero → shown as zero (avoids "-0") |
| −0.7 | -0.700 | -1 | negatives round normally at the configured resolution |

Negatives round at the configured resolution (only those within half a step of zero display as zero); positives use "<" instead, since rendering a nonzero value as plain zero would misrepresent it.

## Compact labels (#5172)

The "short" format (slope chart value labels, map annotations, thumbnails, entity selector, tooltip fallback) abbreviated everything from 1,000 up — but `$25.24k` is exactly as wide as `$25,240` and harder to parse. Three changes:

1. Everything below 1 million is written out in full; abbreviation starts at 1 million, same as "long". Labels of decimal-mode columns never use "k" — within one chart, thousands aren't a mix of written-out and abbreviated forms.
2. Written-out values of 1,000 or more drop their decimals — the fraction is noise at that magnitude. This only applies where the old code abbreviated: percent values were never abbreviated and keep their configured decimals.
3. Abbreviated values round to 3 significant figures instead of always carrying 2 mantissa decimals. This also applies to "long" abbreviations ("12.84 million" → "12.8 million").
4. Sig-fig columns are the exception to 1: they are written out only while every digit shown is significant. From 10^`numSignificantFigures` on (capped at 1M), writing out would pad with placeholder zeros, so they abbreviate instead — "12.3k" claims exactly the three digits a 3-sig-fig column has, where "12,300" implies two more. At the default of 3 sig figs that threshold is 1,000, so sig-fig columns end up formatted as they are today. Zeros are trimmed in compact contexts ("1M", "1.5M"). One 3-sig-fig column across magnitudes: 999 · 1.23k · 25.2k · 123k · 950k · 1M · 12.3M.

Threshold decisions are made on the value as it will be rendered, not on the raw value, so a value that rounds up across a threshold lands on the right side of it: "999,999.99" abbreviates to "1M" instead of being written out as "1,000,000". The same goes for change 2's cutoff — a value that rounds up to 1,000 drops its decimals with everything else at that magnitude, so "999.996" renders as "1,000" rather than "1,000.00". That one is only visible where trailing zeros are kept, i.e. in tooltip tables; elsewhere the trailing zeros were already trimmed. Together this also fixes two long-standing TODOs in the tests.

| value | before | after | comment |
| --- | --- | --- | --- |
| 999.99 | $999.99 | $999.99 | unchanged — below 1,000 nothing changes |
| 1,234 | $1.23k | $1,234 | written out; note the k form saved no space (change 1) |
| 25,240 | $25.24k | $25,240 | thousands written out; abbreviation now starts at 1M (change 1) |
| 25,240.99 | $25.24k | $25,241 | decimals dropped from 1,000 up; the tooltip keeps them (change 2) |
| 99,999 | $100k | $99,999 | written out; previously rounded up to $100k (change 1) |
| 123,456.78 | $123.46k | $123,457 | written out with decimals dropped (changes 1 + 2) |
| 999,999 | $1M | $999,999 | written out right up to the 1M boundary (change 1) |
| 999,999.99 | $1M | $1M | unchanged — rounds up to 1M, so it abbreviates rather than rendering "$1,000,000" |
| 999.996 (tooltip table) | 1,000.00 | 1,000 | rounds up to 1,000, so it drops its decimals like every other value at that magnitude |
| 12,840,000 | $12.84M | $12.8M | 3 sig figs instead of always 2 mantissa decimals (change 3) |
| 846,691,847 ("long") | 846.69 million | 847 million | "long" abbreviations also round to 3 sig figs (change 3) |
| 1,234.56 (% column) | 1,234.56% | 1,234.56% | unchanged — percents were never abbreviated, so their configured decimals stay |
| 999 (sig-fig column) | 999 | 999 | unchanged — written out while every digit is significant |
| 25,240 (sig-fig column) | $25.2k | $25.2k | unchanged — sig-fig columns abbreviate from 10^sf, which at 3 sig figs is 1,000 |
| 123,456 (sig-fig column) | $123k | $123k | unchanged |
| 999,999 (sig-fig column) | $1.00M | $1M | sig-fig abbreviations trim their zeros in compact contexts, so round-ups sit naturally next to k forms |
| 25,000 (axis tick) | $25k | $25k | unchanged — faceted/sparkline ticks keep abbreviating from 1,000 |
| 10,020,000 (tick on a 10.02M–10.08M axis) | 10.02 million | 10.02 million | unchanged — the axis requests enough sig figs to keep neighbouring ticks apart |

**Axis ticks.** Faceted charts and the map sparkline keep abbreviating from 1,000 (via a new internal `abbreviationThreshold` option), since ticks are round numbers where `k` genuinely saves axis width ($25k vs $25,000). Since abbreviated values now carry only 3 significant figures, ticks on a narrow domain at high magnitude (say 10.02M–10.08M) could collide into duplicate "10M" labels — so the axis computes the precision the tick spacing actually requires and passes it via a second internal option, `abbreviationSignificantFigures`, mirroring how it already computes adaptive decimal places for written-out ticks. Ordinary axes are unaffected.

Note: `numSignificantFigures` deliberately has no effect outside sig-fig mode — abbreviations in decimal-places mode use a hard-coded 3, which `abbreviationSignificantFigures` raises where the axis needs it. In sig-fig mode that option can only raise the column's own precision, never lower it.


packages/@ourworldindata/utils/src/formatValue.ts
```diff
@@ -65,47 +75,87 @@ function getType({
...
+        // stays written out as "25,200" since "25.2k" would need a decimal
+        const defaultThreshold =
+            numberAbbreviation === "short" &&
+            roundingMode === OwidVariableRoundingMode.significantFigures
+                ? Math.min(
+                      10 ** (numSignificantFigures + 2),
+                      DEFAULT_ABBREVIATION_THRESHOLD
+                  )
+                : DEFAULT_ABBREVIATION_THRESHOLD
+        const threshold = abbreviationThreshold ?? defaultThreshold
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Fable 5 — @sophiamersmann at the wheel._

Good catch — fixed in a2f580f716 by passing `abbreviationThreshold: 1e3` in `formatPopulationAxisLabelShort`, the same pattern this PR uses for Grapher's own tick-like callers (facet axes, map sparkline). The narrow pyramid axes keep rendering `1k`/`25k`.

We also swept the remaining `numberAbbreviation: "short"` call sites without an explicit threshold: all of them are compact value labels (tooltips, on-chart annotations, sankey node labels), where writing out thousands is this PR's intended behavior, so they were left unchanged.

packages/@ourworldindata/utils/src/formatValue.test.ts
```diff
@@ -177,17 +182,18 @@ describe("rounding to significant figures", () => {
...
         ["numberAbbreviation billion specific", 123_456_789_012, "123 billion", { numberAbbreviation: "long" }],
         ["numberAbbreviation long with unit", 1_000_000_000, "$1.00 billion", { numberAbbreviation: "long", unit: "$" }],
-        ["numberAbbreviation short", 1_000_000_000, "1.00B", { numberAbbreviation: "short" }],
+        ["numberAbbreviation short", 1_000_000_000, "1B", { numberAbbreviation: "short" }],
         ["numberAbbreviation %", 20_000, "20,000%", { numberAbbreviation: "short", unit: "%" }],
         ["numberAbbreviation false", 1_000_000_000, "1,000,000,000", { numberAbbreviation: false }],
-        ["numberAbbreviation false very small", 0.000000001, "0.000000001", { numberAbbreviation: false, numSignificantFigures: 1 }],
+        ["numberAbbreviation false very small", 0.000000001, "<0.01", { numberAbbreviation: false, numSignificantFigures: 1 }], // capped at the default of 2 decimal places
+        ["abbreviationSignificantFigures has no effect in sig-fig mode", 1_234_567, "1.23 million", { numberAbbreviation: "long", abbreviationSignificantFigures: 5 }],
```
marcelgerber: Does this make sense? Should we not also apply `abbreviationSignificantFigures` in this case?

packages/@ourworldindata/utils/src/formatValue.test.ts
```diff
@@ -205,3 +211,79 @@ describe("rounding to significant figures", () => {
...
+    // oxfmt-ignore
+    const cases: [string, number, string, TickFormattingOptions][] = [
+        ["author decimals below 1k", 999.99, "999.99", {}],
+        ["decimals dropped from 1k", 1234.56, "1,235", {}],
+        ["decimals dropped in the ten thousands", 25_240.99, "25,241", {}],
+        ["negative value with decimals dropped", -25_240.99, "-25,241", {}],
+        ["written out in the hundred thousands", 123_456.78, "123,457", {}],
+        ["round value written out", 950_000, "950,000", {}],
+        ["written out just below 1M", 999_999, "999,999", {}],
+        ["rounds across the threshold but stays written out", 999_999.99, "1,000,000", {}],
```
marcelgerber: I can see how this could be complicated, but I'd prefer this as `1M`, too.

marcelgerber (approved):
A for effort!

This is a really good thing, and very clearly improves the number formatting in many cases - I think the "111 billion" vs "111.11 billion" case in here is the most important and common one.

I commented on two cases where I (slightly) disagree with the behavior, but they are uncommon either way and are fine, so it really is fine if you leave them as-is and merge away.

packages/@ourworldindata/utils/src/formatValue.ts
```diff
@@ -65,47 +75,87 @@ function getType({
...
+    if (numberAbbreviation) {
+        // Compact labels of sig-fig columns abbreviate from 10^(sf + 2) —
+        // the magnitude from which abbreviations carry no decimals. At
+        // 3 sig figs that's 100k: 123,456 renders as "123k", while 25,240
+        // stays written out as "25,200" since "25.2k" would need a decimal
+        const defaultThreshold =
+            numberAbbreviation === "short" &&
+            roundingMode === OwidVariableRoundingMode.significantFigures
+                ? Math.min(
+                      10 ** (numSignificantFigures + 2),
```
marcelgerber: Ok, now that I thought about this for a few minutes longer, I feel like adding the `+ 2` here is possibly unnecessary?

Imho, in sig-fig rounding mode with 3 sigfigs, `1114` is nicer as `1.11k` rather than `1,110` (with the last digit "arbitrarily" being capped at zero). Similarly for `12345`, where I prefer `12.3k` over `12,300`. What do you think?

sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Agreed, and dropped in 2987464e10 — the threshold is now `min(10 ** numSignificantFigures, 1e6)`.

Your framing convinced me: `1,110` and `12,300` show trailing zeros that look like digits but are rounding artifacts, whereas `1.11k` / `12.3k` claim exactly the digits the column has. And the width argument that motivated writing out thousands in the first place (`$25.24k` is no shorter than `$25,240`) never applies here, because a sig-fig mantissa can never grow past the sig-fig budget: `12.3k` (5 chars) vs `12,300` (6), `123k` (4) vs `123,000` (7).

So the rule is now "written out only while every digit shown is significant; abbreviated from the magnitude where writing out would pad with zeros" — 1,000 at the default of 3 sig figs, i.e. the same threshold sig-fig columns had before this branch. Nice side effect: the PR now only changes how decimal-mode columns are formatted, which is a much easier story to tell. The 1M cap stays so high-sig-fig columns still abbreviate from 1M like everything else.

packages/@ourworldindata/utils/src/formatValue.test.ts
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

You are right, and the old behaviour was a trap. Fixed in 2987464e10: `abbreviationSignificantFigures` now also applies in sig-fig mode, as a floor rather than a replacement — `Math.max(numSignificantFigures, abbreviationSignificantFigures)`, and only for abbreviated (`"s"`) values, since written-out values keep the column's own precision.

Its purpose is "keep neighbouring abbreviated ticks apart", which is a request for *more* precision, so it should never be able to lower what the column asks for. It happened not to bite yet only because the sole producer, `Axis.getTickFormattingOptions`, hard-codes `roundingMode: decimalPlaces` for ticks — i.e. an option that silently did nothing on a path nobody exercises today.

packages/@ourworldindata/utils/src/formatValue.test.ts
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Fixed in 2987464e10 — it renders as `1M` now.

Root cause was that the threshold was tested against the raw value: 999,999.99 is below 1M, so it went down the written-out path, and only then did the decimals get dropped. The threshold (and the drop-decimals-from-1,000 cutoff) is now tested against the value *as it will be rendered*, so a value that rounds up across a threshold lands on the abbreviated side of it.

Bonus: this was the same bug behind two TODOs that had been sitting in these tests, so `999,999.999` → `1 million` and (in sig-fig mode) `999,999` → `1.00 million` now work as well, and both TODOs are gone.

sophiamersmann: done!

packages/@ourworldindata/utils/src/formatValue.ts
sophiamersmann: yeah, that makes sense to me!

PR #6766, opened 2026-07-16
Update all oxc packages and fix yarn testLint

rakyi:
I did this because tsgolint was crashing in Cursor for me (used by the Oxc plugin), and I'm hoping this might fix it.

Turns out we can get rid of a lot of the type casts, presumably because of the recently updated TS version.


adminSiteClient/DataInsightIndexPage.tsx
```diff
@@ -424,15 +424,15 @@ export function DataInsightIndexPage() {
             // Show a notification if any items failed to sync
             if (result.errored > 0) {
                 notificationApi.error({
-                    message: "Failed to refresh some data insights",
+                    title: "Failed to refresh some data insights",
```
rakyi, line 427: `message` is deprecated.

marcelgerber (approved):
nice!

PR #6772, opened 2026-07-17
Add opt-in map option to hide the tooltip tolerance notice

pabloarosado:

**Draft / for discussion** — opening this so we can decide whether it's worthwhile before polishing.

## What this does

Adds an opt-in map config boolean, **`map.hideToleranceNotice`** (defaults to `false`), that suppresses the map tooltip's tolerance notice:

> Data not available for {year}. Showing closest available data point instead

When the flag is off (the default), behavior is unchanged. When a chart author enables it via the new checkbox in the map editor's **Tooltip** section, the notice is hidden.

## Why

On some maps the time dimension isn't really a time series — it's a per-entity *attribute*. For example, a map where each country is plotted at its own target year. In that case the map defaults its selected time to the latest year present, so most countries render the "Data not available for {year}, showing closest" notice even though their value is perfectly valid for that country. The notice is confusing there. This opt-in checkbox lets a chart author suppress it for those maps, without affecting any map that doesn't set the flag.

## How it's wired

Mirrors the existing `map.tooltipUseCustomLabels` field end-to-end. The notice is produced by the `toleranceNotice` getter in `MapTooltip`, which returns `undefined` when the flag is set (so both the tooltip footer and the `subtitleFormat` "notice" styling that key off it behave). The flag is read from `manager.mapConfig`, which is already reachable inside `MapTooltip`.

Because this is an additive optional field, no schema version bump or DB migration is needed (per the schema README, non-breaking additions just edit the YAML). `defaultGrapherConfig.ts` was regenerated from the schema.

## Files changed

- `packages/@ourworldindata/types/src/grapherTypes/GrapherTypes.ts` — add `hideToleranceNotice?: boolean` to `MapConfigInterface`
- `packages/@ourworldindata/grapher/src/mapCharts/MapConfig.ts` — add the observable field + default
- `packages/@ourworldindata/grapher/src/schema/grapher-schema.010.yaml` — add the schema property
- `packages/@ourworldindata/grapher/src/schema/defaultGrapherConfig.ts` — regenerated default (`hideToleranceNotice: false`)
- `packages/@ourworldindata/grapher/src/mapCharts/MapTooltip.tsx` — gate `toleranceNotice` on the flag
- `adminSiteClient/EditorMapTab.tsx` — add the admin checkbox in the Tooltip section

## Verification

- Typecheck (`tsc -b`) passes.
- Lint (`oxlint`) and format (`oxfmt`) pass on the touched files.
- `MapTooltip`, `MapConfig`, schema migration, and grapher-config-utils unit tests pass.

### DIFF
--- a/adminSiteClient/EditorMapTab.tsx
+++ b/adminSiteClient/EditorMapTab.tsx
@@ -182,6 +182,12 @@ class TooltipSection extends Component<{ mapConfig: MapConfig }> {
             : undefined
     }
 
+    @action.bound onHideToleranceNotice(hideToleranceNotice: boolean) {
+        this.props.mapConfig.hideToleranceNotice = hideToleranceNotice
+            ? true
+            : undefined
+    }
+
     override render() {
         const { mapConfig } = this.props
         return (
@@ -192,6 +198,13 @@ class TooltipSection extends Component<{ mapConfig: MapConfig }> {
                     }
                     value={!!mapConfig.tooltipUseCustomLabels}
                     onValue={this.onTooltipUseCustomLabels}
+                />
+                <Toggle
+                    label={
+                        "Hide the notice shown when a value comes from a nearby year (tolerance)"
+                    }
+                    value={!!mapConfig.hideToleranceNotice}
+                    onValue={this.onHideToleranceNotice}
                 />
             </Section>
         )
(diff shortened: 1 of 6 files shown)


pabloarosado:
@codex review

pabloarosado:
We decided against this option (see [Sophia's feedback](https://owid.slack.com/archives/C0149NKLZAM/p1784553175628599?thread_ts=1784283461.767589&cid=C0149NKLZAM)).

PR #6776, opened 2026-07-17
✨ fix casing for headings on topic pages

ikesau:
## Context

[Slack thread](https://owid.slack.com/archives/C03NV9Z3YSV/p1784126797524169)

For now, we want all non-MTPs to use sentence casing for their headings.

This requires a fiddly amount of context-aware formatting, which is this PR.

**LTP**:
image

**LTP with acronym**:
image

**MTP**:
image

## Testing guidance

Check out 

http://staging-site-fix-casing/hiv-aids
http://staging-site-fix-casing/coronavirus
http://staging-site-fix-casing/poverty
http://staging-site-fix-casing/urbanization

And make sure it's all consistent with the style guide


ikesau:
@codex review

rakyi:
This seems to work great! However, I'm not sure why we keep using title case in MTPs. Could we change it to sentence case everywhere?

ikesau:
I've been putting off resolving this because it's not a great reason but I also don't want to bother Ed about it - the design for modular topic pages currently specifies title case. It's how they were designed and I think it looks kind of weird when they're sentence case.

I'd like to merge and we can revisit it later if Marwa wants?

rakyi (approved):
OK, then let's go for it. I just got the feeling that we were making this code change because it's "easier" than making the casing the same, while it felt like it's the other way around.

No big deal either way, we can revisit it later.

PR #6778, opened 2026-07-17
Use subpath imports for React Aria

rakyi:
This should minimize the bundle size.


rakyi:
Looks like Vite is already tree shaking React Aria even with top level imports.

PR #6796, opened 2026-07-21
Preserve selectedEntityNames when switching to Marimekko/Scatter

### LINKED ISSUE #6794: Setting Marimekko as default chart clears selectedEntityNames from config
## How to replicate

1. Pick chart with line and selected entities such as [this one](http://staging-site-docs-companion-files-lesson/admin/charts/639/edit)
2. Add Marimekko chart type, switch to it on the chart and update the chart
3. `selectedEntityNames` are gone

Marigold:
## Summary
There were two independent places in the admin that could permanently clear `selectedEntityNames` when a chart's active chart type became Scatter or Marimekko:

- `DimensionSlotView.updateDefaultSelection()` (admin-only) unconditionally called `selection.clearSelection()` whenever the chart type was Scatter or Marimekko, even if a selection (e.g. an authored `selectedEntityNames`) already existed. Every other chart-type branch in that method only sets default entities when the selection is empty; Scatter/Marimekko were the only ones actively wiping out an existing selection. Fixed by making that branch a no-op instead (new charts still get no auto-picked default selection, matching the original intent from #965).
- `GrapherState.ensureEntitySelectionIsSensibleForTab()` clears the selection whenever it still matches the authored config and the new tab is a "shows all entities" type (Scatter/Marimekko). On the live site this is a temporary display adjustment that gets restored when switching back to a normal tab. But the chart editor persists the live selection as-is on save, so simply previewing Marimekko/Scatter via the chart-type tab switcher and clicking "Update chart" silently wiped `selectedEntityNames` — this is the exact repro in #6794. Fixed by skipping the clearing when `grapherState.isEditor` is true.

**Viewer-facing behavior is unchanged**: on the live site, switching to a Marimekko/scatter tab still clears the selection unless the user has customized it — per review, that's intentional (these chart types work better without entities highlighted by default). An earlier commit on this branch that changed this for Marimekko was reverted.

Fixes #6794.

## Test plan
- [x] Added a failing-then-passing test (`adminSiteClient/EditorBasicTab.test.tsx`) reproducing the first bug: switching `chartTypes` to Marimekko previously cleared `selectedEntityNames`.
- [x] Added a failing-then-passing test (`GrapherState.test.ts`) reproducing the second bug: switching the active tab to Marimekko in the editor previously cleared `selectedEntityNames`.
- [x] `yarn test run --reporter dot packages/@ourworldindata/grapher/ adminSiteClient/` — all tests pass.
- [x] `yarn testFormatChanged` / `yarn testLintChanged` clean on changed files.
- [x] Manually reproduced against chart 639 locally (`make up.headless`): enabled Marimekko, switched to it via the chart-type tab switcher, clicked "Update chart", and confirmed via a direct DB query that `selectedEntityNames` was preserved (`["World"]`) after save.
- [x] Verified viewer-facing pages in a browser locally: scatter tab switches still clear an authored-matching selection (`air-pollution-vs-gdp-per-capita`).

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
+++ b/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
@@ -2017,19 +2017,17 @@ export class GrapherState
         const isChartTypeThatShowsAllEntities =
             this.isChartTypeThatShowsAllEntities(tab)
 
-        // On a scatter plot, a selection moves all other entities into a
-        // faded background layer, so a selection authored for another tab
-        // (e.g. a line chart) would visually filter the scatter. We therefore
-        // clear the selection unless the user has explicitly changed it.
-        // This intentionally does NOT apply to:
-        // - Marimekko charts, where a selection only bolds and prioritises
-        //   entity labels and hides nothing, and where most published charts
-        //   share their selection with sibling tabs like a line chart
-        // - the editor, which persists the live selection on save, so this
-        //   temporary display adjustment would permanently wipe an authored
-        //   selection (see #6794)
+        // If the chart show all entities (e.g. scatter plot or Marimekko chart),
+        // then we typically prefer no selection unless the user has explicitly
+        // made changes to the default selection.
+        // In the editor, this clearing is skipped: it's meant to be a
+        // temporary, display-only adjustment that gets restored when
+        // switching back to a tab that shows a subset of entities, but the
+        // editor can persist the live selection as-is when saving, which
+        // would otherwise wipe out an authored selection just by previewing
+        // a different chart type.
         if (
-            tab === GRAPHER_TAB_NAMES.ScatterPlot &&
+            isChartTypeThatShowsAllEntities &&
             !this.areSelectedEntitiesDifferentThanAuthors &&
             !this.isEditor
         ) {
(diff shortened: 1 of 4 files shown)


Marigold:
@sophiamersmann I'm not sure what the purpose of clearing the selection was. It was originally created by @marcelgerber four years ago, so I'm not sure who has better context now.

Marigold:
@codex review

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
sophiamersmann, line 2120: I think it makes sense to wrap both of these in `!this.isEditor` - to make sure we don't edit the time or entity selection while editing

Marigold:
@sophiamersmann I found that selected default entities from the line chart aren't transferred to Marimekko (e.g. https://ourworldindata.org/grapher/maternal-mortality). I guess that's expected? When I add an entity in line chart and switch to Marimekko, I see all of them highlighted, but when I switch to Marimekko without adding any entity, none of them are highlighted. I'd have expected the highlighting to happen nonetheless.

sophiamersmann:
Yeah, that’s expected. Marimekkos and scatter plots both work better without entities highlighted by default, so that’s the default behavior (no selection). However, if the user has actively changed the selected entities, we preserve their selection. It’s not something I’m particularly happy with - more of a band-aid for not being able to support per-chart default selections.

Why are you looking into this? If we were to drop this special behaviour, there'd be a few more (visual) changes I'd want to make, but I could look into it if that would make things easier for you.

Marigold:
> Why are you looking into this?

@sophiamersmann I was just surprised. Don't worry, I won't change this behavior and just fix the admin bug.

sophiamersmann:
@Marigold I'm often surprised myself.........

PR #6798, opened 2026-07-21
📜🤖 Rewrite root CLAUDE.md

Marigold:
## 👨 Summary

Recreated CLAUDE.md from scratch with `/init` and merged with the original one. I made some additional improvements based on dogfooding.

## 🤖 Summary

Rewrites the root `CLAUDE.md` from scratch with Fable. The previous version had grown into a grab-bag; this one is rebuilt around what agents actually need, while keeping the hard-won operational knowledge from the old file.

New in this version:

- **Full command reference**: dev stack (`make up` / `up.headless` with its traps), all test suites (`yarn test`, `make dbtest`, `make svgtest`, BDD, bundlemon), single-file vitest invocation, and the changed-files lint/format scripts with an explicit "oxfmt, not prettier" warning.
- **Architecture as dependency layers** (packages → settings/serverUtils → db/jobQueue → apps → Cloudflare functions), the chart-component three-layer pattern, and the `knexReadonlyTransaction`/`knexReadWriteTransaction` convention — each with pointers to the deeper `docs/agent-guidelines/` docs instead of duplicating them.

Carried over from the old file:

- Code style (MobX `makeObservable` setup, BEM SCSS without `&` concatenation, double quotes, no `any`, browser-support check)
- Branch naming rule with the staging-server rationale
- "ALWAYS read `db/docs/` first" directive and `yarn query` usage
- `internal: true` for new skills, and the cloud-sandbox section

Follow-up commits on this branch:

- **Deduplicated against auto-loaded context**: the `create-migration` skill self-triggers (and `db/CLAUDE.md` points to it), so the root mention is gone; the `db/readme.md` guidance (down migrations, past examples) moved into the skill itself. The inlined gitmoji table became a plain pointer to `commit-messages.md` to avoid drift.
- **Cloud-sandbox section extracted** to `docs/agent-guidelines/cloud-sandbox.md` (cross-linked with `claude-code-web.md`); the root file keeps a one-line pointer gated on `CLAUDE_CODE_REMOTE=true`, so local sessions don't carry ~15 unused lines in context.
- **Nested CLAUDE.mds** for `functions/` (edge runtime, `.dev.vars`, README pointer) and `packages/@ourworldindata/grapher/` (chart-components pattern, `make svgtest`), following the existing `db/CLAUDE.md` pattern, each with an `AGENTS.md` symlink.

`AGENTS.md` remains a symlink to `CLAUDE.md`, so `.github/copilot-instructions.md` keeps working.


CLAUDE.md
```diff
@@ -1,101 +1,97 @@
...
+2. **Never guess GitHub handles** — a wrong `@`-tag pings a real person. Without a confirmed handle, write the plain name and ask the user for the handle.
 
-# Database documentation
+3. **This repo is public — keep internal context out of it.** PR descriptions, commit messages, and comments must not identify people who contact us (say "a reader pointed out …" instead of names/roles/employers) and must not reference internal discussions (Slack, Notion) or who suggested what internally. Motivate changes using public facts only.
 
-Our main datastore is a mysql 8 database. The documentation for this lives in db/docs - there is README.md file which is a good overview and starting point, then one TABLE-NAME.yml file per table describing the table in more detail. ALWAYS list the directory db/docs/ to understand which tables are available and read the relevant table description files before constructing a query or writing a migration.
-When creating a new migration, make sure to read the [database README](./db/readme.md).
+## Code style
+
+- Double quotes for string literals.
```
marcelgerber, line 114: Can we add here that we use oxfmt for formatting?

I've seen it recently try to run prettier first, only to then realise that that doesn't exist.

functions/CLAUDE.md
```diff
@@ -0,0 +1,8 @@
+This directory contains Cloudflare Pages Functions — the dynamic edge routes of an otherwise static site (`/grapher/[slug]`, thumbnails, data downloads, `/api`, donations).
+
+Read [README.md](./README.md) before working here. It covers:
+
+- File-based routing plus [`_routes.json`](../_routes.json) (which paths are served dynamically); itty-router is sometimes used inside a route file for sub-routing.
+- Local development: copy `.dev.vars.example` to `.dev.vars`, then `make up.full` (whole stack) or `yarn startLocalCloudflareFunctions` (functions only). Keep compatibility dates in sync between `package.json` and the Cloudflare dashboard.
```
marcelgerber: The compatibility date goes into `wrangler.jsonc`, not package.json (and the file here is the authorative source, not the dashboard).

marcelgerber (approved):
yep sounds good, and now has more structure for sure!

CLAUDE.md
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Added to the Code style section in 1aa975b548 — the tests section already said "oxfmt, not prettier", but that's clearly not where an agent looks before formatting.

functions/CLAUDE.md
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

You're right — `package.json` doesn't carry one at all anymore. Fixed in 1aa975b548, and also fixed `functions/README.md`, which is where the wrong info came from. Both now point at `wrangler.jsonc` as authoritative, with a note that `functions/test/wrangler.e2e.jsonc` keeps its own copy.

PR #6816, opened 2026-07-22
🔨 add BAKED_BASE_URL to CF env

ikesau:
Adds `BAKED_BASE_URL` to CF Pages Functions env variables.

Doesn't set a value in the `preview` `vars` block in `wrangler.jsonc` because those are dynamic.


Marigold (approved):
Approved. I haven't tested it with real staging, but it's low risk anyway.

PR #6823, opened 2026-07-23
🎉🤖 (time-interval) calendar-aware ticks & labels for sub-yearly time axes

sophiamersmann:

Generate calendar-nice ticks and labels for time axes of month, day, week and quarter interval columns, instead of the generic d3 ticks.

- **Continuous axes** step along calendar-anchored grids: a January-anchored month grid (1/2/3/6/12/… month steps, restricted to quarter multiples for quarterly axes) and a day grid (1/2 days, weekly/biweekly on Mondays; week multiples only for weekly axes). For yearly ticks, the start and end ticks have the highest priority and are always shown.     This does not apply to sub-yearly axes, where the start and end ticks are only rendered when they coincide with a time grid value.
- **Discrete (band) axes** (stacked bars) place one tick per bar and thin the labels to the finest evenly-spaced calendar tier that fits (every value → biweekly → first-of-month/quarter/year, per interval), labeled by the column's own time format.

### DIFF
--- a/packages/@ourworldindata/grapher/src/axis/Axis.test.ts
+++ b/packages/@ourworldindata/grapher/src/axis/Axis.test.ts
@@ -2,13 +2,23 @@ import { expect, it, describe } from "vitest"
 import * as R from "remeda"
 
 import { HorizontalAxis } from "../axis/Axis"
-import { ScaleType, AxisConfigInterface } from "@ourworldindata/types"
 import {
+    ScaleType,
+    AxisConfigInterface,
+    ColumnTypeNames,
+} from "@ourworldindata/types"
+import {
+    OwidTable,
     SynthesizeFruitTable,
     SynthesizeGDPTable,
 } from "@ourworldindata/core-table"
 import { AxisConfig } from "./AxisConfig"
-import { AxisAlign } from "@ourworldindata/utils"
+import {
+    AxisAlign,
+    dayjs,
+    convertDateToDaysSinceEpoch,
+    convertDaysSinceEpochToDate,
+} from "@ourworldindata/utils"
 
 it("can create an axis", () => {
     const axisConfig = new AxisConfig({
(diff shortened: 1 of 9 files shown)


packages/@ourworldindata/grapher/src/axis/Axis.test.ts
```diff
@@ -234,3 +244,324 @@ describe("manual ticks", () => {
...
+        )
+        axis.formatColumn = table.get("month")
+        axis.hideFractionalTicks = true
+        axis.range = [0, 800]
+        return axis
+    }
+
+    it("places monthly time-axis ticks on first-of-month, January-anchored boundaries", () => {
+        const day = (date: string): number =>
+            convertDateToDaysSinceEpoch(dayjs.utc(date))
```
marcelgerber: We should pull this out as a singler helper function across the whole test file.

packages/@ourworldindata/grapher/src/axis/Axis.ts
```diff
@@ -690,19 +739,34 @@ export class HorizontalAxis extends AbstractAxis {
...
+                const placedTickLabels = option.map((tick) =>
+                    this.placeTickLabel(tick.value, tick.label)
+                )
+                if (labelsFit(placedTickLabels, { padding: 3 }))
+                    return placedTickLabels
+            }
+            // If no evenly-spaced option fits, fall through to the greedy labeling below
+        }
+
+        // Otherwise: place all ticks greedily drop individual overlaps.
```
marcelgerber:
```suggestion
        // Otherwise: place all ticks greedily, then drop individual overlaps.
```

packages/@ourworldindata/grapher/src/axis/Axis.ts
```diff
@@ -962,6 +1026,30 @@ export class DualAxis {
     }
 }
 
+/** Whether no two labels overlap, keeping at least `padding` of space between them. */
+function labelsFit(
+    tickLabels: TickLabelPlacement[],
+    { padding = 0 }: { padding?: number } = {}
+): boolean {
+    for (let i = 0; i < tickLabels.length; i++) {
+        for (let j = i + 1; j < tickLabels.length; j++) {
```
marcelgerber: If `tickLabels` are given in sorted order (which they currently are, I think), then we could just look at the pair `(i, i+1)` always, and don't need the `N²` loop.

packages/@ourworldindata/grapher/src/axis/Axis.ts
```diff
@@ -690,19 +739,34 @@ export class HorizontalAxis extends AbstractAxis {
...
-        // Get ticks with coordinates, sorted by priority
-        const tickLabels = _.sortBy(
+        // A discrete (band) time axis shows a single evenly-spaced labeling:
+        // the finest one that fits (every day, then week, month, quarter, year, …)
+        const interval = this.calendarTickInterval
+        if (interval !== undefined && this.config.bandValues) {
+            const options = getDiscreteTimeTickOptions({
+                interval,
+                bandValues: this.config.bandValues,
+            })
```
marcelgerber, line 770: I think we should have some constant number, e.g. 30, of the maximum ticks we can ever place.

If we find a spacing that generates more than this many ticks, we can just skip this option entirely as "unreasonable", and don't even need to bother with placing and fit-checking them.

marcelgerber:
* **#6850** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6850?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6823** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/grapher/src/axis/timeAxisTicks.ts
```diff
@@ -0,0 +1,570 @@
...
+    }
+}
+
+/**
+ * The reference day that anchors a day grid with the given step: a Monday for
+ * weekly steps (so ticks land on week starts), the epoch otherwise.
+ */
+function dayGridReference(step: number): number {
+    return step % 7 === 0 ? snapToIntervalStart(0, TimeInterval.Week) : 0
+}
```
marcelgerber, line 99: This code runs for each value in `bandValues`. Can we instead make it a constant, or a lazy run-once expression?

```ts
const WEEK_GRID_REFERENCE = snapToIntervalStart(0, TimeInterval.Week)
```

packages/@ourworldindata/grapher/src/axis/timeAxisTicks.ts
```diff
@@ -0,0 +1,570 @@
...
+/**
+ * One tick per domain value. The ticks carry no label, so the axis falls back
+ * to the column's own time format.
+ */
+export function buildDiscreteTimeAxisTicks({
+    bandValues,
+}: {
+    bandValues: number[]
+}): Tickmark[] {
+    return _.sortBy(_.uniq(bandValues)).map((value) => ({ value, priority: 2 }))
```
marcelgerber: this is a bit faster:
```suggestion
    return sortNumeric(_.uniq(bandValues)).map((value) => ({ value, priority: 2 }))
```

marcelgerber (approved):
This is so good!!!

As we discussed on Slack and in the call, I could well see a future where we use the discrete code more often, and don't place them based on `targetCount`. but that's for the future!

PR #6825, opened 2026-07-24
🎉 Allow addressing charts by their config UUID in the admin API

danyx23:
> _Written by Claude Fable 5 — @danyx23 at the wheel._

## Summary

Every chart already has a stable, client-generated UUID: `chart_configs.id`, linked 1:1 via the unique `charts.configId` and never changed over a chart's lifetime. This PR makes that UUID usable as an external identifier in the admin API, as groundwork for referencing charts from the ETL by a stable id instead of the auto-increment id (which collides across environments) or the slug (which changes and only exists for published charts).

- The `:chartId` param of all `/charts/...` admin API routes now accepts a chart's config UUID as an alternative to the numeric id, resolved via the new `expectChartId` helper (`SELECT id FROM charts WHERE configId = ?`, unique index).
- `saveGrapher` / `saveNewChart` accept an optional caller-supplied config UUID when creating a chart (validated, 409 on collision). Not yet exposed on any route — the follow-up PR #6826 builds on this.

No schema changes, no behavior changes for existing numeric-id callers.

**Now stacked on the chart-configs refactor** (#6971 → #6979 → #6980 → #6988 → #6987). In that world a chart has two config rows — `configId` (the rendered config) and `patchConfigId` (the editor-authored layer) — and the UUID used for external addressing is the **rendered config's id** (`charts.configId`), which keeps its value through the refactor's migration. A caller-supplied config UUID on chart creation is threaded through `insertChartConfigPair` and becomes the id of the rendered config's row, never the patch row's.

## Test plan

- New DB tests covering UUID addressing of GET/PUT routes, 404 for unknown UUIDs, 400 for malformed ids, and creation with a supplied UUID incl. the duplicate-UUID conflict — re-run green on top of the refactor stack (all 114 DB tests pass).


pabloarosado (commented):
> _Written by Claude Fable 5 — @pabloarosado at the wheel._

Read through it: clean seam, every `:chartId` route (including update and delete) goes through the shared helpers, and the tests cover the matrix I'd want. Three small suggestions inline, none blocking.

About the red checks, so you don't chase them: the bake failure was a master bug (`ER_OUT_OF_SORTMEMORY` on the multi-dim redirects query) fixed by #6828, which is why I merged master into this branch. `site-screenshots-diff` now just needs a human glance, since that merge brought a week of real visual changes. The BDD soft-fail is all in `features/search.feature` and comes from this staging server's search state; the same tests pass on #6826's server.

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -212,21 +213,57 @@ export const getReferencesByChartId = async (
...
+        (typeof chartIdOrConfigId === "string" &&
+            /^\d+$/.test(chartIdOrConfigId))
+    )
+        return expectInt(chartIdOrConfigId)
+
+    if (
+        typeof chartIdOrConfigId === "string" &&
+        uuidValidate(chartIdOrConfigId)
+    ) {
+        const chartId = await getChartIdByConfigId(knex, chartIdOrConfigId)
```
pabloarosado, line 231: > _Written by Claude Fable 5 — @pabloarosado at the wheel._

`uuidValidate` accepts uppercase, and MySQL lookups ignore case, but the ETL compares these UUIDs as Python strings: an uppercase one would stop chart-diff matching the chart with its production twin, and syncing it would create a duplicate.

```suggestion
        const chartId = await getChartIdByConfigId(
            knex,
            chartIdOrConfigId.toLowerCase()
        )
```

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -266,7 +325,7 @@ const saveNewChart = async (
     // insert patch & full configs into the chart_configs table
     // We can't quite use `saveNewChartConfigInDbAndR2` here, because
     // we need to update the chart id in the config after inserting it.
-    const chartConfigId = uuidv7() as Base64String
+    const chartConfigId = (providedChartConfigId ?? uuidv7()) as Base64String
```
pabloarosado: > _Written by Claude Fable 5 — @pabloarosado at the wheel._

Same normalisation where a caller supplies the UUID.

```suggestion
    const chartConfigId = (providedChartConfigId?.toLowerCase() ??
        uuidv7()) as Base64String
```

adminSiteServer/tests/charts.test.ts
```diff
@@ -973,3 +978,129 @@ describe("Chart slug validation", { timeout: 15000 }, () => {
...
+                        newConfig: { ...testChartConfig, slug: "other-slug" },
+                        chartConfigId,
+                    })
+                },
+                TransactionCloseMode.KeepOpen,
+                env.testKnex
+            )
+        ).rejects.toThrow(/already exists/)
+    })
+})
```
pabloarosado, line 1112: > _Written by Claude Fable 5 — @pabloarosado at the wheel._

`""` is already rejected here (`!== undefined` sends it into validation); this test keeps it that way, since a later switch to a truthiness check would let an empty string become a chart's identity and break its R2 key and by-uuid URL.

```suggestion
        ).rejects.toThrow(/already exists/)
    })

    it("rejects an empty config UUID", async () => {
        const user = await env.testKnex(UsersTableName).first()

        await expect(
            knexReadWriteTransaction(
                async (trx) => {
                    await saveGrapher(trx, {
                        user,
                        newConfig: testChartConfig,
                        chartConfigId: "",
                    })
                },
                TransactionCloseMode.KeepOpen,
                env.testKnex
            )
        ).rejects.toThrow(/Invalid config UUID/)
    })
})
```

pabloarosado (commented):
> _Written by Claude Fable 5 — @pabloarosado at the wheel._

One more, which can't be a suggestion because the line isn't in this diff: a supplied `chartConfigId` is ignored when the chart already exists. Verifying it would catch a caller (e.g. chart-sync's id+createdAt fallback) updating the wrong chart. In `saveGrapher`, right after `chartId = existingConfig.id!`:

```ts
if (providedChartConfigId !== undefined) {
    const chartIdForConfigId = await getChartIdByConfigId(
        knex,
        providedChartConfigId.toLowerCase()
    )
    if (chartIdForConfigId !== chartId)
        throw new JsonError(
            `Config UUID ${providedChartConfigId} does not belong to chart ${chartId}`,
            400
        )
}
```

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -405,6 +465,8 @@ export const saveGrapher = async (
         // if undefined, keep inheritance as is.
         // if true or false, enable or disable inheritance
         shouldInherit?: boolean
+        // only used when creating a new chart, see `saveNewChart`
```
rakyi: Nit: remove or reword this comment, don't repeat what can be seen from code.

rakyi (approved):
I don't want to postpone this feature any longer, so I guess we can live with it for now, but I think it's not a good practice in general to let the same path contain either a number or a UUID and I think it would be better to have different endpoints for this.

Not adding the UUID support for creating charts via `POST /charts` feels like a strange omission, but I guess we don't need it. It implies the API has a wrong shape though, and maybe we could clean it up to keep only the upsert PUT and remove the insert-only POST.

adminSiteServer/apiRoutes/charts.ts
danyx23: @pablorosado that seems needlessly fragile on the ETL side - UUIDs are case insensitive. Can we change this on the ETL side instead?

pabloarosado: > _Written by Claude Fable 5 — @pabloarosado at the wheel._

Agreed — done on the ETL side, in your companion PR's branch (owid/etl@9f278a91c3): chart-diff and chart-sync now match config UUIDs case-insensitively, and `etl chart-config-id lookup` writes the canonical lower-case form into the YAML (the strict YAML validation already required that form, so a config file copied from an upper-case DB value would previously have been rejected). No change needed in this PR, and no normalization expected from grapher.

pabloarosado: (Also, the wrong Pablo Rosado was tagged :-))

pabloarosado (approved):

danyx23:
> I don't want to postpone this feature any longer, so I guess we can live with it for now, but I think it's not a good practice in general to let the same path contain either a number or a UUID and I think it would be better to have different endpoints for this.
> 
> Not adding the UUID support for creating charts via `POST /charts` feels like a strange omission, but I guess we don't need it. It implies the API has a wrong shape though, and maybe we could clean it up to keep only the upsert PUT and remove the insert-only POST.

Good catch, but this is added in the PR above in the stack order, the one that actually needs this. Thanks for raising!

PR #6826, opened 2026-07-24
🎉 ETL-authored chart configs, addressed by chart config UUID

danyx23:

## Summary

Charts become first-class ETL citizens, addressed by a stable client-generated UUID instead of the auto-increment id (which collides across environments) or the slug (which changes and only exists for published charts). The substance of the etlConfig layer is @pabloarosado's work from #6553; this PR re-applies it on top of #6825 and adds config-UUID-based upsert. **#6553 will be closed as superseded when this merges.** Companion ETL PR: owid/etl#6511 (see there for the client side).

- **`PUT /charts/by-config/:chartConfigId/etlConfig`** — upsert semantics, so the ETL can address a chart by a UUID it generated itself. Three cases: chart + ETL layer exist → update the layer in place; chart exists without a layer → attach one; neither exists → create a minimal draft chart whose `configId` is the caller-supplied UUID, then attach the layer. A new chart's admin patch contains only the non-inheritable `slug`, so the ETL layer owns everything else (notably `dimensions`) from birth. The response reports `{chartId, created}` so callers get the numeric id for follow-up calls like `setTags`.
- **`DELETE /charts/:chartId/etlConfig`** (detach) accepts either a numeric id or a config UUID.
- **`POST /charts?configId=<uuid>`** — optional caller-supplied config UUID on chart creation, so chart-sync can carry a hand-authored chart's identity from staging to production and retire the id+createdAt matching heuristic over time. (This is the `POST /charts` gap @rakyi raised on #6825.)

The numeric-id `PUT /charts/:chartId/etlConfig` is **removed** in favour of the by-config route. owid/etl#6511 is already on the new route (`chart_sync/admin_api.py` calls `PUT /charts/by-config/{uuid}/etlConfig` and passes `configId` on `POST /charts`), but released ETL code is not, so the two branches must be merged together.

## Chart identity

A chart is identified by its config UUID (`charts.configId`), assigned once and never changed. Config UUIDs are normalised to lower case on the way in, so the stored identity doesn't depend on how a caller typed it.

`charts.etlConfigCatalogPath` records which ETL step currently owns a chart. It is deliberately *not* an identifier: a push carrying a different path than the one on record simply updates it, so a renamed or moved step is an ordinary update, and detaching a chart releases the path. Assigning a path that already belongs to a different chart is rejected with a 409, which turns the unique index's duplicate-key failure into an actionable error; a malformed path gets a 400, using the same shared `CATALOG_PATH_PATTERN` the multi-dim routes enforce.

The column is named `etlConfigCatalogPath` rather than plain `catalogPath` because a bare `catalogPath` on `charts` reads like a chart-level identifier; the name now says what it actually is — the step that authored the chart's ETL config layer (`patchConfigIdETL`). The route's `?catalogPath=` **query parameter keeps its name**, since the `/etlConfig` route segment already supplies that context and the ETL side therefore needs no change for the rename.

## Known limitation

With `dimensions` now inherited like any other property (see the review round below), arrays are still inherited whole: if the ETL layer specifies `dimensions` and an author changes one `display` field in the admin, the entire array lands in the chart's patch and later ETL changes to it never reach the rendered config. Raised by @sophiamersmann and agreed as out of scope here; fixing it means keying entries by variable id and merging `display` per entry.

<details>
<summary>Details</summary>

## Schema and naming

- The chart's ETL layer pointer is named **`charts.patchConfigIdETL`** (renamed from `configIdETL`), following the naming rule from the chart-configs refactor: `patchConfigId*` names an authored layer, `configId` names the config that renders. It mirrors `variables.patchConfigIdETL`, per chart instead of per indicator.
- The ETL layer is stored as a single config row. The upsert/detach endpoints read the admin patch from `charts.patchConfigId`'s row and the rendered config from `charts.configId`'s row, and write both on a rendered-config change.
- All UUID addressing resolves and mints against **`charts.configId` — the rendered config's id**, the chart's stable external identity, never a patch row's id.
- Indicator propagation merges `merge(indicator config, chart's ETL layer, chart patch)` into the rendered row.
- New columns on `charts`: `patchConfigIdETL` and `etlConfigCatalogPath`. `db/docs/charts.yml` also gained the previously-undocumented field descriptions for both.

## Review round

Addressing @sophiamersmann's and @rakyi's feedback:

- **`dimensions` is now inherited like any other property.** It used to be a "required key" that `diffGrapherConfigs` copied into a chart's patch unconditionally, so the patch always won and dimensions could never be inherited — while the ETL push path used a second diff function without that carve-out, giving two inheritance models that fought each other. `REQUIRED_KEYS` and `rediffPatchAgainstNewParentStack` are gone, indicator-level configs no longer carry a `dimensions` array, and a migration strips the existing ones.
- **The catalog path is mutable**, as described above — the previous behaviour rejected a changed path, which had identity the wrong way round.
- **`PUT /charts/:chartId/etlConfig` is removed** in favour of the by-config route.
- **The editor no longer drops the ETL config layer** when the parent indicator changes or inheritance is toggled. Because `updateLiveGrapher` resets grapherState first, the omitted layer fell back to grapher defaults and the next save wrote those defaults into the patch as explicit overrides of fields the ETL layer owns.
- `upsertEtlConfigByChartConfigId` renamed from `putChartsByConfigIdEtlConfig`.

Two of @sophiamersmann's suggestions were deliberately not taken — passing the slug on creation, and routing the ETL update path through `saveGrapher`; discussed [in this thread](https://github.com/owid/owid-grapher/pull/6826#issuecomment-5400730096) and settled.

Fixes from @pabloarosado's review:

- **A no-op ETL re-push no longer moves `charts.updatedAt`.** The no-op branch guarded its write with `patchChanged || catalogPath`, and the ETL sends `catalogPath` on every call, so the guard was always true and a byte-identical re-push still bumped the timestamp. chart-diff reads `charts.updatedAt` to decide whether a chart was edited in production after a staging server was created, so routine ETL re-runs surfaced as conflicts nobody caused on every open branch, going stale and blocking chart-sync.
- **ETL chart ownership is cleaned up safely**: catalog paths are released when a chart detaches from ETL, and catalog-path conflicts are preflighted before a UUID-addressed draft config is created and uploaded.

## Follow-up (not blocking)

The two new `charts` columns are absent from the `owid/analytics` public-export allowlist (`analytics/duckdb/public_schema.json`), so they won't appear in the public export until someone adds them there. That allowlist is opt-in, so nothing leaks and nothing breaks in the meantime; `variables.patchConfigIdETL` is already exported, if we want parity.

## Test plan

- DB tests for all three upsert cases, idempotent re-push (including `updatedAt` staying untouched across an identical re-push that carries the same `catalogPath`), malformed-UUID rejection, UUID case normalisation, creation via `POST /charts?configId=...`, catalog-path update plus the duplicate-path (409) and malformed-path (400) guards, and a lifecycle assertion covering ownership and rendered output across each attach/detach transition.
- Unit tests for the single inheritance model in `grapherConfigUtils`, and a regression test for the editor's parent stack that fails if the ETL layer is dropped again.
- The `dimensions` migration was run against a full local database (222,638 indicator configs): every rendered chart and multi-dim view config is byte-identical before and after (checksummed), no chart lost its dimensions, and the down migration restores all 222,638 arrays with the correct `variableId`.

</details>


adminSiteClient/AbstractChartEditor.ts
```diff
@@ -55,7 +55,15 @@ function isValidEditorTab(tab: string): tab is EditorTab {
 export interface AbstractChartEditorManager {
     admin: Admin
     patchConfig: GrapherInterface
+    // For the main chart editor, `parentConfig` is the indicator's grapher_config
+    // (variable.grapherConfigETL). For other editor variants (indicator/narrative)
```
danyx23: IIRC this isn't entirely accurate - I think the parent is always the variable.graphrConfigAdmin if it exists, then variable.grapherConfigETL if it exists, or null if neither exist - but worth checking, I'm not 100% sure

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -848,13 +889,18 @@ export async function createChart(
     if (req.query.forceDatapage) {
         forceDatapage = req.query.forceDatapage === "true"
     }
+    // optional caller-supplied config UUID, e.g. chart-sync carrying a chart's
+    // identity from staging to production; validated in saveNewChart
+    const chartConfigId =
```
danyx23, line 908: We should validate that this is a valid UUID here

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -915,6 +994,556 @@ export async function updateChart(
...
+
+    let chartId = await getChartIdByConfigId(trx, chartConfigId)
+    const created = chartId === undefined
+
+    if (chartId === undefined) {
+        const { chartId: newChartId } = await saveGrapher(trx, {
+            user: res.locals.user,
+            newConfig: {
+                $schema: defaultGrapherConfig.$schema,
+                slug: etlConfig.slug,
```
sophiamersmann: It might be a bit confusing that, when a chart is first created, the ETL layer passes its slug to the admin layer, but on subsequent updates the ETL layer’s slug is ignored (because slugs are never inherited). If there’s no good reason to pass the slug on during creation, I think it might be cleaner to init the new config with just the schema.

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -915,6 +994,556 @@ export async function updateChart(
...
+    const existingFull = parseChartConfig(row.full)
+
+    // Look up the chart's parent indicator (only if inheritance is enabled),
+    // resolving it from the dimensions the chart will plot *after* this push:
+    // the admin's override if the patch carries one (patch is the top merge
+    // layer, so it wins), else the incoming ETL config's. Resolving from the
+    // pre-push config would, on a dataset re-version, leave the chart
+    // inheriting the old indicator's fields (title, subtitle, note) while
+    // plotting the new one — and nothing later recomputes it. Variable configs
+    // never carry `dimensions`, so this doesn't depend on which parent we pick.
```
sophiamersmann: The opposite is true - all variable configs carry dimensions (but it still doesn't matter because the dimensions are never inherited)

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -915,6 +994,556 @@ export async function updateChart(
...
+                        c.updatedAt = ?
+                    WHERE c.id = ?
+                `,
+                [serializeChartConfig(newPatch), now, catalogPath, now, chartId]
+            )
+        }
+        return { success: true, etlConfig, patch: newPatch }
+    }
+
+    // The rendered chart changed — record it: bump version, rewrite the main
```
sophiamersmann, line 1312: Unless there's some blocker, calling `saveGrapher` when we need to update the chart, would probably get rid of a lot duplicate code (and be a bit less error-prone)

adminSiteClient/ChartEditor.ts
```diff
@@ -219,9 +219,15 @@ export class ChartEditor extends AbstractChartEditor<ChartEditorManager> {
...
-        const chartJson = { ...patchConfig }
+        const { patchConfig, etlConfig } = this
+
+        // An ETL-managed chart keeps its title/subtitle/etc. in `etlConfig`, not
+        // in `patchConfig`, so a plain `{ ...patchConfig }` copy would silently
+        // drop them. Fold the ETL layer in: the new chart is a normal chart that
+        // still inherits the indicator's config, so merging `etlConfig` under the
+        // admin `patchConfig` reproduces what the source rendered. For non-ETL
+        // charts `etlConfig` is undefined, so this is a no-op.
+        const chartJson = mergeGrapherConfigs(etlConfig ?? {}, patchConfig)
```
sophiamersmann: It took me a minute to understand what's happening here. I _think_ (though it would need a bit of testing), we can also just use `fullConfig` here since the variable-level config is diffed in the save function anyway (it would just be important to double check the `fullConfig` doesn't include defaults)

adminSiteClient/AbstractChartEditor.ts
```diff
@@ -200,9 +224,19 @@ export abstract class AbstractChartEditor<
...
+     * neither layer contributes anything.
+     */
     @computed get activeParentConfig(): GrapherInterface | undefined {
-        return this.isInheritanceEnabled ? this.parentConfig : undefined
+        const variablePart = this.isInheritanceEnabled
+            ? (this.parentConfig ?? {})
+            : {}
+        const etlPart = this.etlConfig ?? {}
+        if (_.isEmpty(variablePart) && _.isEmpty(etlPart)) return undefined
+        return mergeGrapherConfigs(variablePart, etlPart)
```
sophiamersmann, line 246: There are a couple more places where the configs are merged. What comes to mind is when the parent config is updated (e.g. as result of a dimension change) or when inheritance is disabled/enabled. If I see correctly, both actions currently silently drop the ETL layer.

packages/@ourworldindata/utils/src/grapherConfigUtils.ts
```diff
@@ -83,3 +83,30 @@ export function diffGrapherConfigs(
...
+/**
+ * Recompute a chart patch against a new parent stack. Unlike
+ * `diffGrapherConfigs`, this doesn't keep required keys unconditionally: every
+ * field falls through to the parent stack when it matches, and only real
+ * overrides survive in the patch. Descends into nested plain objects (e.g.
+ * `yAxis`) field-by-field, so a partially-adopted nested override (some of its
+ * children now match the new parent stack, others don't) drops only the
+ * matching children instead of keeping the whole object pinned in the patch.
+ */
+export function rediffPatchAgainstNewParentStack(
```
sophiamersmann: The only real difference to `diffGrapherConfigs` is how dimensions are treated, which signals to me that we should probably unify the inheritance behaviour across the stack (see my general comment below)

sophiamersmann (commented):
Makes sense to me overall!

One thing that stood out to me is that dimensions are treated differently depending on whether indicator-level config or ETL config is inherited. (Variable-level configs never inherit dimensions.) That adds quite a bit of complexity to the code, and it also means we now have two slightly different flavours of inheritance to keep in mind.

It definitely makes sense for the ETL to own the dimensions array, but instead of supporting two different inheritance models, I wonder if it would make more sense to treat dimensions as a normal property that is inherited in the same way as most other properties throughout the stack. To do that, we’d need to drop the dimensions array from all variable-level configs and update the code so that dimensions are never written into those configs (they’re currently injected when variable-level configs are stored). I think that would simplify the implementation in a few places (including removing the need for the special `rediffPatchAgainstNewParentStack` function).

One other thought regarding the dimensions array: I’m sure you’re already aware of this, but in the current inheritance model, arrays are inherited as a whole. So, for example, if the ETL layer specifies dimensions and an author updates just one field in the admin (e.g. the display name of a variable), the whole dimensions array is replaced. Any later changes to the array in the ETL layer would then never reach the final config. Until now this wasn’t really an issue because dimensions were effectively excluded from inheritance. But if we start inheriting them normally, I think it might make sense to treat this array specially. Perhaps keying entries by variable ID and property, then merging the display field normally, would make sense. But of course that could also be a follow-up rather than something to do now, but I think it’s worth keeping in mind.

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -1047,16 +1088,87 @@ export async function putChartsChartIdEtlConfig(
...
+ * config UUID exists yet, a minimal draft chart is created that carries the
+ * caller-supplied UUID as its identity, and the ETL config layer is attached
+ * to it. The new chart's admin patch starts out (almost) empty, so the ETL
+ * layer owns all rendered fields from birth and only genuine admin edits ever
+ * end up in the patch. The one exception is `slug`: identity/publishing keys
+ * are deliberately excluded from inheritance, so the slug is copied into the
+ * patch at creation (matching admin-created charts). Created charts are
+ * drafts — publishing stays an admin act via the regular chart API.
+ */
+export async function putChartsByConfigIdEtlConfig(
```
rakyi: We could probably come up with a better name for this.

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -1047,16 +1088,87 @@ export async function putChartsChartIdEtlConfig(
...
+        trx,
+        res.locals.user,
+        chartId,
+        etlConfig,
+        catalogPath
+    )
+}
+
+/**
+ * Like `putChartsChartIdEtlConfig`, but addressed by the chart's config UUID
```
rakyi: Can't we get rid of  `putChartsChartIdEtlConfig` now? I thought the idea was that we'd switch to only using the UUID for this use case.

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -1047,16 +1088,87 @@ export async function putChartsChartIdEtlConfig(
...
+    try {
+        etlConfig = migrateGrapherConfigToLatestVersionAndFailOnError(req.body)
+    } catch (err) {
+        return { success: false, error: String(err) }
+    }
+
+    let chartId = await getChartIdByConfigId(trx, chartConfigId)
+    const created = chartId === undefined
+
+    if (chartId === undefined) {
```
rakyi: Nit: use the `created` variable defined above.

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -1068,6 +1180,43 @@ export async function putChartsChartIdEtlConfig(
...
 
+    // Guard against pushing ETL content to the wrong chart. A chart's catalog
+    // path is its stable ETL identity; if the incoming push carries a
+    // different one than we've already recorded, that almost always means the
+    // config UUID used to address this chart is stale or wrong on the ETL
+    // side. Refuse rather than silently stamping unrelated content onto it —
+    // and while `charts.catalogPath` has a unique DB index, checking here
+    // first turns a raw duplicate-key failure into a clear, actionable error
+    // before any other writes happen. A chart with no catalog path yet is
+    // free to adopt one for the first time.
```
rakyi: Isn't this the other way around? The chart catalog path can change, but the UUID of the chart config should never change.

danyx23:

Thanks @sophiamersmann — the dimensions simplification is in, along with the `AbstractChartEditor` merge sites and `saveAsNewGrapher`. Two of your suggestions I'd like to push back on rather than take, and it seemed better to explain here than to leave them silently unaddressed.

**Passing the slug on creation** ([comment](https://github.com/owid/owid-grapher/pull/6826#discussion_r3675122321))

The premise doesn't quite hold: dropping the slug at creation wouldn't just make things consistent, it would mean an ETL-created chart never gets one at all. `slug` is in `KEYS_EXCLUDED_FROM_INHERITANCE`, so `mergeGrapherConfigs` strips it from every layer except the last — the ETL layer's slug therefore never reaches the rendered config. Copying it into the patch at creation is the only path by which the slug an ETL author wrote in their YAML ever reaches the chart. Without it, every ETL-created chart would be unpublishable until someone typed a slug in by hand.

The inconsistency you spotted is real — the ETL slug is honoured once at creation and ignored on every later push — but that's the intended split: the ETL seeds the slug, the admin owns it from then on, and publishing stays an admin act. That also matches how admin-created charts behave. Happy to be talked round if you think the ETL should keep owning the slug, but that's a bigger change than dropping the line.

**Calling `saveGrapher` from the ETL update path** ([comment](https://github.com/owid/owid-grapher/pull/6826#discussion_r3675205039))

The duplication is real, but the two paths do genuinely different things. `updateChart` takes a full desired config and *derives* the patch from it; the ETL push takes a new ETL layer and must *preserve* the existing admin patch, re-diffing it against the new parent stack. Routing the push through `saveGrapher` would mean either giving up the no-op detection — which deliberately skips the version bump, the revision, the R2 upload and the static build, so a routine re-push or a bulk ETL run doesn't churn a chart's history — or pushing ETL-specific branches (catalogPath, the ETL config row, no-op detection) down into the general admin path.

What's actually duplicated is about two `UPDATE` statements. The coupling looks like it would cost more than it saves, so I'd rather leave the paths separate.

**On the array-inheritance point in your top-level review**: agreed, and it's now live rather than theoretical — with `dimensions` inherited normally, one admin tweak to a `display` field replaces the whole array and masks later ETL changes to it. Not addressed here, as you suggested; worth a follow-up issue for keying entries by variable ID and merging `display`.

sophiamersmann:
Thank you @danyx23, makes sense!

danyx23:
> Thank you @danyx23, makes sense!

​Sorry for the Claude word salad, I just wanted to make sure this doesn't get silently dropped last night. Will write these myself in the future!

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -899,6 +983,541 @@ export async function updateChart(
...
+    //     admin override): `full` is identical, but if we don't store the
+    //     reduced patch, that stale entry would mask future ETL updates to a
+    //     field ETL now owns.
+    //   - A new or changed catalog path, which still needs writing onto a
+    //     chart that already has a `patchConfigIdETL` (the
+    //     `etlConfigCatalogPath` writes below this point would otherwise be
+    //     skipped).
+    if (!fullChanged) {
+        const patchChanged = !_.isEqual(newPatch, existingPatch)
+        if (patchChanged || catalogPath) {
```
pabloarosado: > _Written by Claude Fable 5 — @pabloarosado at the wheel._

This guard is always true, so the write below always runs and sets `c.updatedAt = now`, even on a completely unchanged re-push. The ETL sends `catalogPath` on every call ([`chart_upsert.py`](https://github.com/owid/etl/blob/etl-charts/etl/collection/chart_upsert.py#L86), and `catalog_path` is a required field on a collection), so the `|| catalogPath` half never rules anything out.

Measured on `staging-site-chart-diff-layers` running this branch: pushing a byte-identical ETL config left `version` at 16 and `lastEditedAt` in April (so it did take this `!fullChanged` path), but moved `charts.updatedAt` to the time of the push. The column has no `ON UPDATE CURRENT_TIMESTAMP`, so it is this explicit write that moves it.

That matters on the ETL side, which is why it isn't visible from here: chart-diff reads `charts.updatedAt` to decide whether a chart was edited in production after a staging server was created. So after production's ETL re-runs a chart's step — a routine data update whose chart comes out identical — every open branch listing that chart sees a conflict nobody caused, and approvals for it go stale and block chart-sync.

It also seems to cut against the intent stated in the comment just above, that a no-op re-push shouldn't churn the chart's history.

```suggestion
        const catalogPathChanged =
            catalogPath !== null && catalogPath !== row.etlConfigCatalogPath
        if (patchChanged || catalogPathChanged) {
```

`row.etlConfigCatalogPath` is already selected by this function's query, and the same comparison is used a few lines above for the duplicate-path check. Not compiled or CI-tested on my side.

danyx23: Good catch - thanks!

PR #6828, opened 2026-07-24
Fix ER_OUT_OF_SORTMEMORY in multi-dim redirect bake

Marigold:
> _Written by Claude Sonnet 5 — @Marigold at the wheel._

The static-site bake started failing on `staging-site-master` with:

```
Error: ER_OUT_OF_SORTMEMORY: Out of sort memory, consider increasing server sort buffer size
```

This traces back to #6708, which added an `ORDER BY` to the SQL query in `getMultiDimRedirectTargets` (and `getMultiDimRedirectRulesBySource` has the same shape). Both queries also `SELECT` the multi-dim `config` column, a JSON blob that can reach several hundred KB for a single view. MySQL's filesort has to hold whole rows in `sort_buffer_size` when there's no usable index for the `ORDER BY`, so one wide row is enough to overflow the buffer — even though the actual result set is tiny (dozens of rows).

**Fix:** drop the SQL-side `ORDER BY` in both functions and sort the already-fetched (small) result set in JS instead, using the same ordering logic. This never touches `sort_buffer_size`, and preserves the original selection/determinism behavior (unconditional rule preferred, tie-broken by `id`).

I reproduced the exact failure by running the original query directly against the `staging-site-master` database, and confirmed the rewritten query (without the SQL `ORDER BY`) runs successfully against the same data.

## How to test it

Already verified:
- `yarn typecheck` and `yarn testLintChanged` pass.
- Reproduced the original `ER_OUT_OF_SORTMEMORY` error directly against a real database by running the exact SQL from `getMultiDimRedirectTargets`, then confirmed the rewritten query (no `ORDER BY`) succeeds against the same data.

Worth doing:
- Push triggers a fresh bake on `staging-site-mdim-redirect-oom-fix` — once it's up, confirm the site baked without the redirects step failing (this branch doesn't change any redirect *behavior*, just how the ordering is computed, so `/grapher/*` and `/explorers/*` redirects should resolve exactly as before).

### DIFF
--- a/db/model/MultiDimRedirects.ts
+++ b/db/model/MultiDimRedirects.ts
@@ -87,6 +87,8 @@ export async function getMultiDimRedirectTargets(
         multiDimSlug: string
         viewConfigId: string | null
         config: string
+        id: number
+        hasSourceQueryParams: number
     }>(
         knex,
         `-- sql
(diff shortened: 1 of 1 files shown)


Marigold:
@codex review

PR #6840, opened 2026-07-24
🔨 Register Energy Institute and Ember as full region providers

pabloarosado:
> _Written by Claude Code (Opus 4.8), @pabloarosado at the wheel._

Frontend counterpart to owid/etl#6520 (merged), which added the **Energy Institute** and **Ember** world regions to the shared regions dataset. Both were previously suffix-only `AdditionalRegionDataProvider`s; now that they have member definitions, they're upgraded to full `RegionDataProvider`s, so their `(EI)` / `(Ember)` entities get entity-selector grouping and a hover tooltip with a mini-map.

## Changes

- Regenerated `regions.data.ts` from the prod catalog: adds the seven `EI_*` and seven `EMBER_*` aggregates, extending the derived `RegionDataProvider` / `RegionGroupKey` / `TooltipKey` types.
- Moved `ei` and `ember` from `ADDITIONAL_REGION_DATA_PROVIDERS` into the full-definition registries (`REGION_DATA_PROVIDER_LABELS`, `regionGroupLabels`).
- Added `RegionTooltipData` entries (description + map order) for both.
- Pinned each region's color in `ContinentColors` to the standard OWID map colors, so the hover and the admin "Continents" picker match the region-definition maps.
- Fixed the `ei` group label that read "Education International regions" (it's the Energy Institute).

`yarn typecheck` is clean.

The tooltip descriptions link to `world-region-map-definitions#energy-institute-ei` and `#ember`, so those two sections still need adding to that article for the anchors to resolve.


pabloarosado:
@codex review

packages/@ourworldindata/grapher/src/color/CustomSchemes.ts
```diff
@@ -454,6 +454,24 @@ export const ContinentColors = {
     "South and South-East Asia (WID)": OwidMapColors.SkyTurquoise,
     "Oceania (WID)": OwidMapColors.Lavendar,
 
+    // Energy Institute regions — Marwa's curated palette (semantic OWID map colors).
```
sophiamersmann: It’s ok to use the muted colours here for now since other providers now seem to be doing the same. I'll probably do a sweep to correct that at some point, and then for future updates it would be good to use stronger colours here instead (from `OwidDistinctColors`, for example), and the muted colours below.

Can you also please delete the second half of the comment (Marwa's curated palette...), here and below?

sophiamersmann (approved):

packages/@ourworldindata/grapher/src/color/CustomSchemes.ts
pabloarosado: > _Written by Claude Code (Opus 4.8) — @pabloarosado at the wheel._

Done, trimmed both comments to just the region label. Noted on the stronger-colors-for-tooltip preference for future updates.

PR #6843, opened 2026-07-27
✨🤖 (time-interval) label weekly time as readable dates, not ISO week numbers

sophiamersmann:
> _Written by Claude Opus 4.8 — @sophiamersmann at the wheel._

I did some more thinking about how to represent weeks in Grapher and landed on the following:

- I think I prefer ‘Week of Jun 12, 2026’ over ‘6-12 Jun 2026’ because, although the latter is often a bit shorter, it (i) isn't immediately obvious that it represents a week, (ii) is a bit confusing because it's written as a time range, and (iii) gets unwieldy for weeks that cross month or year boundaries.
- For single-time charts, the title now says ‘Week of …’. For time range charts, however, ‘Week of Jun 12, 2021 to Week of Jul 2, 2026’ feels too verbose, so instead we display the range as plain dates, e.g. ‘Jun 12, 2021 to Jul 8, 2026’, where the second date is the last day of the final week (Sunday).
- The timeline also uses plain dates. There's a slight discrepancy with the title because the end label on the timeline shows the first day of the week rather than the last.
- In tooltips, we use the long label (‘Week of …’), and discrete axis ticks use the same format. It's quite space-consuming, but I think that's the least confusing option.

CSV export (`2026-W23`) and the continuous time axis (already week-start dates) are unchanged.

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
+++ b/packages/@ourworldindata/grapher/src/core/GrapherState.tsx
@@ -2729,15 +2729,13 @@ export class GrapherState
 
         if (startTime === undefined || endTime === undefined) return undefined
 
-        // Dumbbell charts compare two points, so use "vs" instead of "to"
-        const separator = this.isOnDumbbellTab ? " vs. " : " to "
+        if (startTime === endTime) return formatTime(startTime)
 
-        const time =
-            startTime === endTime
-                ? formatTime(startTime)
-                : formatTime(startTime) + separator + formatTime(endTime)
+        // Dumbbell charts compare two points, so use "vs." instead of a range
+        if (this.isOnDumbbellTab)
+            return `${formatTime(startTime)} vs. ${formatTime(endTime)}`
 
-        return time
+        return timeColumn.formatTimeRange(startTime, endTime)
     }
 
     @computed get sourcesLine(): string {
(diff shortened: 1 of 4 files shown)


packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -2729,15 +2729,13 @@ export class GrapherState
...
-        const separator = this.isOnDumbbellTab ? " vs. " : " to "
+        if (startTime === endTime) return formatTime(startTime)
 
-        const time =
-            startTime === endTime
-                ? formatTime(startTime)
-                : formatTime(startTime) + separator + formatTime(endTime)
+        // Dumbbell charts compare two points, so use "vs." instead of a range
+        if (this.isOnDumbbellTab)
+            return `${formatTime(startTime)} vs. ${formatTime(endTime)}`
```
marcelgerber, line 2736: Not related to this PR, but arguably, we should use "vs" more: It very much applies for two-year map charts, too (and we could make a case for slope charts, but I don't think we should go that far).

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
```diff
@@ -2729,15 +2729,13 @@ export class GrapherState
...
 
-        return time
+        return timeColumn.formatTimeRange(startTime, endTime)
```
marcelgerber, line 2738: The table also shows a time range:

marcelgerber (approved):

packages/@ourworldindata/grapher/src/core/GrapherState.tsx
sophiamersmann: good point!

PR #6845, opened 2026-07-27
📜🤖 Move staging testing notes out of CLAUDE.md into a skill

rakyi:
> _Written by Anthropic Claude Opus 5 — @rakyi at the wheel._

Found by `/doctor`: the "Testing on staging" section of the root CLAUDE.md loads into every session, but it only applies when we are actually poking at a staging server. It moves verbatim into a new `test-on-staging` skill, leaving a pointer line behind.

Rebased onto #6798, which landed in the meantime and already covered the cloud-sandbox extraction and the codebase-overview rewrite, so this is now just the one skill. The skill also picks up the `staging-algolia` bullet that #6798 added to the section.

### DIFF
--- a/.claude/skills/test-on-staging/SKILL.md
+++ b/.claude/skills/test-on-staging/SKILL.md
@@ -1,0 +1,13 @@
+---
+name: test-on-staging
+description: Test a change on a branch's staging server without a browser — inspect or patch a chart's config remotely, pick the right preview link to share, and know why /search 404s there. Use when testing on staging-site-<branch>, reading or changing a chart config on a remote server, or pointing someone at a chart on staging.
+metadata:
+    internal: true
+---
+
+# Testing on staging
+
+- `yarn tsx devTools/callAdminApi.ts get <chartId> --branch <branch>` / `set <chartId> '<jsonPatch>' --branch <branch>` / `unset <chartId> <field> --branch <branch>`: inspect or change a chart's config on any `staging-site-<branch>` (or `--host` for local/prod) without the browser.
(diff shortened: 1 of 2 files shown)


Marigold (approved):
I've already addressed the cloud sandbox is the CLAUDE restructuring PR https://github.com/owid/owid-grapher/pull/6798. For some reason Claude went with doc rather than skill, but I don't have an opinion. Feel free to make the call. Approved!

#6798 landed a few minutes ago and overlaps with this one — same `/doctor` finding, two PRs, sorry for the collision. It moved the cloud-sandbox section to `docs/agent-guidelines/cloud-sandbox.md` (same text, behind a `CLAUDE_CODE_REMOTE`-gated pointer), and rewrote the codebase overview, so the "sort of monorepo…" sentence you delete is gone as well.

The `test-on-staging` skill is the part that isn't covered, and I think it's the better of your two extractions — it's task-shaped, so the description can actually trigger it. Could you rebase down to just that skill plus the pointer line?

On the sandbox notes we went with a doc rather than a skill because being in a sandbox is an ambient precondition, not a task: there's no user request for a skill description to match, so an env-var-gated pointer is the more reliable trigger. Happy to be talked out of it — the content is identical either way.

rakyi:
> _Written by Anthropic Claude Opus 5 — @rakyi at the wheel._

Rebased down to just the `test-on-staging` skill plus the pointer line — no collision damage, and your doc-vs-skill call for the sandbox notes stands, no need to talk you out of it. Being in a sandbox is a precondition rather than a request, so there is nothing for a description to match; the env-var-gated pointer is the right trigger there.

One thing the rebase picked up: #6798 added a fourth bullet to "Testing on staging" (the `staging-algolia` label / separate Algolia applications note), so that moved into the skill too, and its description mentions search so it triggers on that case as well.

PR #6864, opened 2026-07-29
Fix map brackets becoming stuck on hover & improve map bracket highlighting on mobile

### LINKED ISSUE #6396: Map brackets freeze in hovered state
## Description

Sometimes when hovering over the map brackets, they stay in hovered-over state even when the mouse leaves the area.

## Screenshots

https://github.com/user-attachments/assets/c22c0f1a-a7cd-4cf8-9157-256c9ec68ae2

## Additional context

https://owid.slack.com/archives/C46U9LXRR/p1776703169270349

marcelgerber:
> _Written by OpenAI GPT-5 — @marcelgerber at the wheel._

Fixes #6396.

Fixes the long-standing issue that map brackets regularly became stuck-active after hovering them.

The problem here was that React would re-create the DOM elements because they got resorted after becoming highlighted; and then `mouseleave` wouldn't necessarily fire on the new DOM element, because they were not necessarily marked as `mouseenter`-ed after the DOM reshuffle.

This also improves the interaction pattern on mobile: A tap on a bracket marks the map bracket as highlighted, and it stays highlighted until the user taps anywhere else.
Touch targets are now also vertically larger, covering the space around the map bracket labels, too.

## Open questions

- [x] What should the mobile on-hover behavior be?

### DIFF
--- a/packages/@ourworldindata/grapher/src/legend/HorizontalColorLegends.test.ts
+++ b/packages/@ourworldindata/grapher/src/legend/HorizontalColorLegends.test.ts
@@ -109,6 +109,7 @@ describe(HorizontalNumericColorLegend, () => {
         })
         const onLegendMouseOver = vi.fn()
         const onLegendMouseLeave = vi.fn()
+        const onLegendTouchSelect = vi.fn()
 
         const { container } = render(
             React.createElement(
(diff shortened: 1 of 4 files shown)


marcelgerber:
@codex review

PR #6876, opened 2026-07-31
Route empty-query Algolia searches through a caching Cloudflare function

marcelgerber:

## Context

Algolia multi-query requests where every query string is empty — the "browse" requests issued by default states like the search landing page, the empty autocomplete panel, `/latest`, and featured metrics — don't depend on user input, so they are highly cacheable. This PR routes them through a new Cloudflare Pages Function, `/api/search/cached-queries`, which proxies the request to Algolia and caches the response for one day, cutting into our Algolia quota usage and speeding up these default states. All other search requests still go to Algolia directly.

How it fits together:

- **`functions/api/search/cached-queries.ts`**: accepts the exact body the algoliasearch client POSTs to Algolia's multi-query endpoint, validates that every query string in it is empty (400 otherwise, plus body-size/batch-size limits), forwards it to Algolia, and caches successful responses for 24h in the Cloudflare cache, keyed on a SHA-256 hash of the body (the Cache API can't key on POST bodies directly). An `X-Cache: HIT/MISS` header makes the behavior observable; error responses are never cached.
- **`site/search/searchClients.ts`**: the shared `liteClient` now uses a custom requester that inspects each outgoing payload and reroutes qualifying all-empty-query batches to the proxy. If the proxy is unavailable or rejects a payload, the client transparently falls back to a direct Algolia request. `Autocomplete.tsx` now shares this client instead of creating its own, so the empty search-overlay panels benefit too.
- **`@ourworldindata/utils`**: the `isEmptyQuerySearchPayload()` predicate is shared between the client-side router and the function's validation, so the two can't drift. It handles the request shapes of algoliasearch v5, autocomplete's `params` object, and the legacy URL-encoded `params` string.
- **`settings/clientSettings.ts`**: new `ALGOLIA_CACHED_QUERIES_URL` setting; defaults to `BAKED_BASE_URL + "/api/search/cached-queries"` on staging/production and to disabled (empty string) in local development, where the functions dev server usually isn't running.

### Repercussions for Algolia analytics

- **Empty-query searches largely disappear from Algolia's search analytics.** Today, every pageview of a default state fires analytics-counted queries; after this change, Algolia only sees cache misses. The Cloudflare cache is per-data-center and evicts cold entries, so Algolia will see a small residual trickle (roughly one request per unique payload per Cloudflare colo per day) instead of one batch per pageview. Any dashboard metric that includes empty-query searches — total search volume, "searches without query" counts, no-results-rate denominators — will drop sharply. Metrics about actual typed queries ("top searches", etc.) are unaffected, since those requests still go to Algolia directly from the browser.
- **The residual trickle is mis-attributed.** Cache-miss requests originate from our Worker (Cloudflare egress IPs, no end-user IP or user agent), so unique-user and geography estimates for empty-query searches become meaningless. If we'd rather have a clean zero than a mis-attributed residue, a follow-up could inject `analytics: false` into the requests the function forwards to Algolia.
- **No click/conversion analytics break**, because we don't use them: the codebase sends no `clickAnalytics`/`queryID`, no `userToken`, and no Algolia Insights events, and we don't use Query Suggestions, A/B testing, or personalization/re-ranking (which learn from this data). Our own Google Analytics tracking runs client-side and is unaffected.

## Testing guidance

- Unit tests: `yarn test run packages/@ourworldindata/utils/src/search/emptyQuerySearchPayload.test.ts functions/api/search/cached-queries.test.ts`
- E2E: `yarn test:functions:e2e-search` runs the function in a real workerd runtime against production Algolia and verifies a cache MISS followed by a HIT, and a 400 for non-empty queries.
- On a deployed environment, probe the endpoint directly (replace `HOST`):
    ```
    curl -si -X POST HOST/api/search/cached-queries \
        -d '{"requests":[{"indexName":"pages","query":"","hitsPerPage":2}]}' | grep -i x-cache
    ```
    The first request returns `X-Cache: MISS`, subsequent identical ones `HIT`. (On a staging server the Algolia index only exists with the `staging-algolia` label; without it the endpoint passes Algolia's error through uncached and the site falls back to direct Algolia requests.)
- To test the client-side routing locally, run `yarn startLocalCloudflareFunctions` and set `ALGOLIA_CACHED_QUERIES_URL=http://localhost:8788/api/search/cached-queries` in `.env`, then load `/search` or `/latest` and check the network tab for requests to the endpoint.

- [ ] Does the change work in the archive? — n/a, archived pages don't use site search; the requester change is inert there since it only affects the shared search client.

https://claude.ai/code/session_01FZhxugZALZjPhk6npWbiaW


ikesau (approved):
Looks good! I added an exemption for the chronological pages index so that new posts don't take 0-24h to show up. We can add a cache invalidation path later (triggered by indexing) if we want to cache that index too (we probably _will_ want to - it has quite a few hits, but I think it's for the best to not cache it for now because it would break the `/latest` page)

UPDATE: changed the exemption to a shorter TTL instead. 15 minutes seems tolerable and will significantly reduce quota.

ikesau:
@codex review

PR #6885, opened 2026-08-03
Store chart deprecation notices outside Grapher config

pabloarosado:
> _Written by Codex GPT-5.6 — @pabloarosado at the wheel._

## Summary

This is a standalone alternative to #6808 so the two storage approaches can be compared directly.

- stores the notice as nullable lifecycle metadata on `charts.deprecationNotice`, not in the persistent Grapher config
- shows the banner above the chart title on interactive charts, including iframe embeds, while leaving PNG/SVG output unchanged
- injects the notice at runtime for baked pages and archives
- delivers notices to config-fetched embeds through R2 metadata and an exposed response header, keeping the Grapher v11 JSON body schema-valid
- gives notice changes their own response ETag so browser caches refresh even when the chart config itself is unchanged
- includes the notice in archive change detection without changing hashes for charts that have no notice
- keeps “Save as new” from copying a deprecation notice to the new chart

The current diff is 35 files, +785/−89, including the migration, runtime/R2 plumbing, and tests.

## Testing

- local formatting, changed-file lint, and typecheck passed
- full local unit suite: 2,111 passed, 5 skipped
- focused Chart Editor, Grapher state, archival checksum, and R2 endpoint tests passed
- the R2 endpoint test covers schema-valid JSON, cross-origin notice delivery, ETag refresh after notice changes, 304 responses, and clearing the notice
- GitHub format, oxlint, typecheck, unit, DB, bespoke build/test/typecheck, analysis, bundle, and security checks passed
- staging build and SVG tester passed with zero Grapher SVG differences
- the external site-screenshot comparison remains red with seven accumulated screenshot artifacts on the branch

## Open items

- **Handed off:** None.
- **Proposed:** Compare this storage/runtime approach with #6808 before choosing which version to merge; separately review or refresh the accumulated site-screenshot artifacts.
- **Unverified:** A signed-in manual staging embed check after the latest commit was not possible from this session; the same delivery path is covered by the R2 endpoint integration test.


pabloarosado:
> _Written by Claude Opus 5 — @pabloarosado at the wheel._

Pushed a fix for the archival checksum in 24fb60c: `MD5(COALESCE(deprecationNotice, ""))` was never null, so the new key was present for every published chart and the first archival run would have re-snapshotted the whole site. It now returns NULL and the key is omitted, so undeprecated charts hash exactly as before. Added a test for that property.

One thing left, which is a design call rather than a bug, so I left it alone: `chartConfigHelpers.ts` spreads `deprecationNotice` into the config saved to `publishedGrapherBySlug`. That config declares `$schema` v11, which has `additionalProperties: false` and no such property, so the published config no longer validates against the schema it claims (our ETL-side validator rejects it). Either declare it in the schema, or deliver it to config-fetched embeds outside the config the way `_OWID_ARCHIVE_CONTEXT` is delivered. Same block also recomputes the md5 locally, which the comment just above it warns against.

pabloarosado:
@codex review

pabloarosado:
> _Written by Codex GPT-5.6 — @pabloarosado at the wheel._

Resolved the open design point in [this comment](https://github.com/owid/owid-grapher/pull/6885#issuecomment-5167155016) in `3b77aa0a54`.

I chose delivery outside the config JSON: the R2 object keeps the schema-valid Grapher v11 body, while the notice is stored as R2 custom metadata and exposed to config-fetched embeds through `X-OWID-Deprecation-Notice`. The response ETag includes the notice, so add/update/clear operations invalidate caches even when the chart config is unchanged. This also removes the local MD5 recomputation.

The R2 endpoint integration test covers the unchanged JSON body, cross-origin header delivery, ETag refresh, 304 behavior, and clearing the notice.

pabloarosado:
After some discussion (see e.g. [this thread](https://owid.slack.com/archives/CQQUA2C2U/p1784805065474269)) we decided to not go ahead with the banner (more details in [this issue](https://github.com/owid/owid-issues/issues/2407#issuecomment-5190614788)).
The banner idea could still be valuable for e.g. deprecated charts in the archive, but it should be properly shaped and designed.

PR #6887, opened 2026-08-03
✨🤖 name the original time in the tolerance notice

sophiamersmann:

## Context

Tooltips already say when a value came from a different time than the one you asked for: _"Data not available for 2020. Showing closest available data point instead"_. That works when there's one time involved — but when a tooltip covers a time range and only one of the two values was interpolated, the subtitle ends up mixing a target time with an original time, and there's nothing to tell you which is which.

In that case the notice now names the original time, matching what the data table already does:

- _"Data not available for 2020. Showing closest available data point (2019) instead"_

Applies to the faceted map, slope chart and dumbbell time-range tooltips.

Deliberately left as is where a single parenthetical can't work: tooltips with one target time but several values that can each resolve to a different original time (stacked discrete bar, marimekko, dumbbell two-column, scatter), and the single-value map tooltip, whose subtitle is the original time already.

Also in here:

- The notice helper's parameters are renamed from `targetYear`/`closestYear` to `targetTime`/`originalTime`, since these aren't always years
- Fixes the slope chart saying the singular "data point" when both of its values were interpolated

## Screenshots

_(add before/after of a slope chart tooltip where only one endpoint was interpolated)_

## Testing guidance

- Open a slope chart whose start or end time is missing for some entities and hover an entity where only one of the two values fell back to another time — the subtitle should name that time in parentheses.
- Hover one where both values were interpolated — it should read "data points", plural, and no parenthetical.
- Facet a map tab by pulling the timeline handles apart and hover a country that only has data for one of the two times.
- Check the untouched cases still read sensibly: stacked discrete bar, marimekko, scatter, and a single-time map tooltip.

### DIFF
--- a/packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellTooltips.tsx
+++ b/packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellTooltips.tsx
@@ -71,17 +71,28 @@ export class DumbbellTimeRangeTooltip extends React.Component<DumbbellTooltipPro
         if (isStartOriginal && isEndOriginal) return undefined
 
         const formatTime = (t: number) => formatColumn.formatTime(t)
-        const targetYear =
+        const targetTime =
             !isStartOriginal && !isEndOriginal
                 ? `${formatTime(startTime)} and ${formatTime(endTime)}`
                 : !isStartOriginal
                   ? formatTime(startTime)
                   : formatTime(endTime)
 
+        // The subtitle shows the original time of both values. Naming the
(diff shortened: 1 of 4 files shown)


sophiamersmann:
* **#6887** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #6892, opened 2026-08-04
🎉🤖 display.inapplicableEntities: mark entities an indicator doesn't apply to

sophiamersmann:

Adds a `inapplicableEntities` field to an indicator's display config for indicators that include countries without meaningful data.

On the map tab, the not-applicable countries are excluded from the color scale and gets a distinct pattern instead, with matching legend and tooltip treatment. The field can be overridden per-dimension in the chart editor.

Notes:
- 'Not applicable' entities are expected not to have data, and Grapher doesn’t do any extra work to handle cases where they do. This also means that when updating the existing cases where we have ‘not applicable’ data, the sentinel/fake values should be removed as well.

Example: http://staging-site-reference-entity-debug/grapher/china-imports-as-share-of-gdp


packages/@ourworldindata/grapher/src/mapCharts/MapChartState.ts
```diff
@@ -348,13 +359,20 @@ export class MapChartState implements ChartState, ColorScaleManager {
...
 
         return mapColumn.owidRows
             .map((row) => {
                 const { entityName, value, originalTime } = row
+
+                // Entities with not-applicable values are expected to have
+                // no data, but skip any rows they might have anyway, so that
+                // they're guaranteed to never have a series
+                if (notApplicableEntityNamesSet.has(entityName))
+                    return undefined
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Valid, and it pointed at a real inconsistency: the series skipped these rows while `colorScaleColumn` didn't, so a marked entity with rows would still shift the automatic bin thresholds.

Fixed the other way round, though — the defensive skip is gone. Not-applicable entities are expected to have no data by construction, so the code now assumes that everywhere rather than half-defending against it in one place.

packages/@ourworldindata/grapher/src/mapCharts/MapTooltip.tsx
```diff
@@ -318,7 +338,11 @@ export class MapTooltip
                     <MapTooltipValue
                         mapColumn={mapColumn}
                         datum={endDatum}
-                        formattedValue={this.formattedEndValue?.label}
+                        formattedValue={
+                            this.isNotApplicableEntity
+                                ? this.notApplicableValueLabel
+                                : this.formattedEndValue?.label
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Same root cause as the `MapChartState` comment, and resolved the same way: not-applicable entities are assumed to have no rows, so there are no datums for `shouldShowValueRange` to pick up and no sparkline values to expose.

packages/@ourworldindata/grapher/src/schema/grapher-schema.011.yaml
```diff
@@ -134,6 +134,16 @@ properties:
...
+            notApplicableEntities:
+              type: array
+              items:
+                type: string
+              description: |
+                Entities for which the indicator's data is not applicable by construction,
+                e.g. China for "China's imports as a share of GDP", or any country for
+                "emigrants from that country". The map renders these in a dedicated pattern
+                with its own legend entry labeled "Not applicable", and tooltip and data
+                table show "Not applicable".
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Correct that nothing implements this — it was the schema description overpromising. Since these entities are expected to have no data at all, they don't appear in the data table to begin with, so the description now only covers the map and its tooltip.

adminSiteClient/DimensionCard.tsx
```diff
@@ -71,6 +76,20 @@ export class DimensionCard<
         this.onChange()
     }
 
+    @computed get notApplicableEntityOptions(): {
+        value: EntityName
+        label: EntityName
+    }[] {
+        return this.table.availableEntityNames
+            .toSorted()
+            .map((entityName) => ({ value: entityName, label: entityName }))
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Good catch — the picker sourced its options from the table's entity names, which excludes exactly the countries the field exists for. It now offers all of `mappableCountries`, matching the `isOnTheMap` filter that `notApplicableEntityNamesSet` applies anyway.

packages/@ourworldindata/grapher/src/color/ColorScaleBin.ts
```diff
@@ -215,4 +219,8 @@ export function isProjectedDataBin(bin: ColorScaleBin): bin is CategoricalBin {
     return isCategoricalBin(bin) && bin.value === PROJECTED_DATA_LABEL
 }
 
+export function isNotApplicableBin(bin: ColorScaleBin): bin is CategoricalBin {
+    return isCategoricalBin(bin) && bin.value === NOT_APPLICABLE_LABEL
+}
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Technically right, but leaving it. `isNoDataBin` and `isProjectedDataBin` identify their bins by value the same way, so this follows the established pattern and the same latent collision already exists for "No data". Threading an injected-bin flag through `CategoricalBin` is more machinery than a categorical indicator with a literal "Not applicable" category warrants.

packages/@ourworldindata/grapher/src/core/GrapherConstants.ts
```diff
@@ -146,10 +146,17 @@ export const isPopulationVariableETLPath = (path: string): boolean => {
 }
 
 export enum Patterns {
+    /** No-data hatch used by legends, chart elements and the globe */
     noDataPattern = "noDataPattern",
+    /** Viewport-scaled no-data hatch used by 2D maps */
     noDataPatternForMap = "noDataPatternForMap",
-    noDataPatternForGlobe = "noDataPatternForGlobe",
```
sophiamersmann, line 154: This was the same as 'noDataPattern', so I cleaned it up

packages/@ourworldindata/grapher/src/mapCharts/ChoroplethGlobe.tsx
```diff
@@ -718,34 +729,40 @@ export class ChoroplethGlobe extends React.Component<{
...
-        const patternId = Patterns.noDataPatternForGlobe
+        if (featuresWithNoData.length === 0) return
 
         return (
             <g
                 id={makeFigmaId("countries-without-data")}
                 className="noDataFeatures"
             >
-                <defs>
-                    <NoDataPattern patternId={patternId} />
```
sophiamersmann, line 731: This is included globally, it's not needed here again

adminSiteClient/DimensionCard.tsx
marcelgerber, line 1: I'm a bit torn on whether this should be a dimension-level setting, or whether this should rather go on either the chart level or even the map level of things (right now, I can't see us making use of this option anywhere outside of the map).

Is it an option on the dimension because then it can be set on etl? That's the only (major!) upside I see.

adminSiteClient/DimensionCard.tsx
```diff
@@ -71,6 +77,21 @@ export class DimensionCard<
...
+        label: EntityName
+    }[] {
+        return mappableCountries
+            .map((country) => country.name)
+            .toSorted()
+            .map((entityName) => ({ value: entityName, label: entityName }))
+    }
+
+    @action.bound onChangeInapplicableEntities(entityNames: EntityName[]) {
+        this.props.dimension.display.inapplicableEntities = entityNames
```
marcelgerber: Somehow, clearing the `<Select>` on http://staging-site-reference-entity-debug/admin/charts/8832/edit doesn't properly update the map chart.
Is there some inheritance going on, and some parent dimension metadata sets it to China? Or is it just a genuine bug?

packages/@ourworldindata/grapher/src/mapCharts/ChoroplethMap.tsx
```diff
@@ -478,26 +494,40 @@ export class ChoroplethMap extends React.Component<{
             >
                 <defs>
                     <NoDataPattern
-                        patternId={patternId}
+                        patternId={Patterns.noDataPatternForMap}
                         scale={1 / this.viewportScale} // The scale is crucial and projection specific
                     />
+                    {inapplicableEntityNamesSet &&
+                        inapplicableEntityNamesSet.size > 0 && (
```
marcelgerber:
```suggestion
                    {inapplicableEntityNamesSet?.size && (
```

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -392,18 +414,19 @@ export class MapChart
...
+        const numericBins = this.legendData
+            .filter((bin) => isNumericBin(bin))
             .map((bin) => this.maybeAddPatternRefToBin(bin))
 
-        // Move the no-data bin from the end to the start
-        return [bins[bins.length - 1], ...bins.slice(0, -1)]
+        if (this.hasCategoricalLegendData) return numericBins
+
+        // Prepend any leftover categorical bins (e.g. "No data" or the
+        // reference entity bin) to the numeric legend
```
marcelgerber: this comment is stale

packages/@ourworldindata/grapher/src/mapCharts/MapChart.tsx
```diff
@@ -689,6 +713,9 @@ export class MapChart
                         lineColorScale={this.colorScale}
                         targetTime={this.targetTime}
                         targetTimes={this.manager.highlightedTimesInTooltip}
+                        inapplicableEntityNamesSet={
+                            this.chartState.inapplicableEntityNamesSet
```
marcelgerber:
```suggestion
                            this.inapplicableEntityNamesSet
```

packages/@ourworldindata/grapher/src/mapCharts/ChoroplethGlobe.tsx
```diff
@@ -718,34 +729,40 @@ export class ChoroplethGlobe extends React.Component<{
...
 
     renderFeaturesWithNoData(): React.ReactElement | undefined {
-        if (this.featuresWithNoData.length === 0) return
+        const {
+            featuresWithNoData,
+            manager: { inapplicableEntityNamesSet },
+        } = this
 
-        const patternId = Patterns.noDataPatternForGlobe
+        if (featuresWithNoData.length === 0) return
```
marcelgerber, line 736: There are fringe cases where this conditional pattern-inclusion can go wrong (thanks Fable!).

I've reproed it on http://staging-site-reference-entity-debug/admin/charts/2028/edit, and saved the chart in that state.

Repro conditions:
- Set "French Southern Territories" as an inapplicable country (it is one of the only countries for which we have no population data)
- Change to faceting view, and set the region dropdown to e.g. Africa (which has no No-data countries)
- **This makes it so that `InapplicablePattern` is not included in the map, BUT it is needed for the legend bucket rendering**

-> I think the solution is to just always include all patterns - it doesn't hurt to do so.

marcelgerber (approved):

adminSiteClient/DimensionCard.tsx
sophiamersmann: I assigned it to the dimension because it's a property of the data, but as you said, it's unlikely we'll make use of this anywhere else. Maybe the map level is the right call then!

We'll soon have ETL-authored patch configs, so it'll be possible to set this from the ETL wherever it lives.

adminSiteClient/DimensionCard.tsx
sophiamersmann: Oh sorry, that's a debug branch that literally hard-codes `inapplicableEntities: ["China"]` for the variable, so clearing it won't work

packages/@ourworldindata/grapher/src/mapCharts/ChoroplethGlobe.tsx
sophiamersmann: Good catch!

marcelgerber (approved):

sophiamersmann:
@marcelgerber, just fyi that I ended up placing the inapplicable entities in the chart config (not the map) as `inapplicableEntityNames` because I liked the symmetry with `excludedEntityNames`/`includedEntityNames`, etc. It also felt a bit more future proof to not bind this too closely to the map (although I can't think of a real use case right now). Let me know if you'd do this differently!

PR #6895, opened 2026-08-04
🎉🤖 support picking an arbitrary variable as the map source

sophiamersmann:
The map tab could only display a variable that was already plotted on the chart, via `map.columnSlug`. Charts can now name a dedicated map dimension (`DimensionProperty.map`) instead, so the map renders an indicator of its own.

- A dedicated map dimension takes precedence over `map.columnSlug` when resolving the map column, and it's picked in the map slot on the editor's basic tab
- A map dimension is taken at face value: it's never stitched together with a historical y column
- A map dimension also isn't auto-plotted on chart tabs. A chart carrying only a map dimension reports "Missing Y axis column" rather than quietly plotting the map indicator on a line chart
- On the map tab, the default title, subtitle, sources line, timeline, values JSON and search unit all follow the map column. The timeline, values JSON and search unit share a tab-aware `primaryColumnSlugs` accessor; the title and subtitle each check for a map dimension themselves
- On the map tab the `values.json` endpoint reports the map column under the JSON's `y` key, because `{y, x}` is the published shape of `/grapher/slug.values.json` (and I didn't think this is enough of a reason to change that shape)
- I considered adding an error/warning message if an indicator is picked for the map slot that is already plotted (because we have map.columnSlug in that case), but I _think_ we probably want to refactor `map.columnSlug` away, so didn't think this was an issue
- **Fix.** The map tab's sources line credited every y column, so a map showing one of several plotted indicators named sources it doesn't draw. It now credits the map column alone, or both halves of a legacy stitched projection


sophiamersmann:
@codex review

sophiamersmann:
@marcelgerber, as discussed, maybe scroll through the changes (no need for a deep review if we don't ship) and let me know if you'd prefer not to ship, or if you think this is actually okay. I don't mind either way!

marcelgerber (commented):
I don't think it's that bad - but yeah, as we said, there's also not been so much demand for it.

Let's quickly talk about it in the viz call today?

PR #6897, opened 2026-08-04
Add git worktree setup to Makefile

marcelgerber:
I want to try this setup out for working with multiple worktrees in Orca. Not sure if I'll stick with it.

---

> _Written by Anthropic Claude Opus 5 — @marcelgerber at the wheel._

Adds `make up.worktree`: a git worktree's dev environment running alongside the one in your main checkout, with its own admin/vite ports, its own detached tmux session, and the MySQL every checkout already shares.

`make up.headless` can't fill this role — it `pkill`s any admin server and vite on the machine, whatever checkout they belong to. `up.worktree` only ever touches its own named tmux session, so it's safe to run from a worktree manager or an agent while another environment is up.

- `make setup.worktree` writes the worktree's `.env` with a random free port pair (admin `30xx` / vite `80xx`, sharing an offset). It skips ports that are listening or claimed by another worktree's `.env`, and keeps 3030/8090 reserved for the main checkout. Random rather than next-free so a deleted worktree's port isn't handed to its replacement. Works as a setup hook alongside `yarn install`.
- `make down.worktree` stops only that session, leaving MySQL up for the other checkouts.
- `up.worktree` refuses to run in the main checkout and points at `make up`.
- The MySQL start/wait logic moves out of `up-headless.sh` into a shared `ensure-mysql.sh` — lift-and-shift, no behaviour change.
- `ADMIN_SERVER_PORT` is now documented in `.env.example-grapher`, which also means adding it to the Makefile's inline-comment `strip` block.

Tested by running the full up/down/up cycle in a worktree, including from a `.env`-less state, while the main checkout's environment stayed up on 3030/8090 throughout


marcelgerber:
@codex is there a better or different way to do this?

marcelgerber:
@claude is there a better or different way to do this?

PR #6899, opened 2026-08-04
✨ Update bespoke notifications content-type preferences model

ikesau:
- Changes the `content-type` type to support the new form design.
- Use strings instead of the `OwidGdocType` enum so that we can support `"data-update"` etc
- Tying this field to the enum created a false contract. The two are similar but not identical.


baker/emailNotifications/NotificationEmail.tsx
```diff
@@ -238,7 +238,7 @@ function Item({ item }: { item: NotificationEmailItem }) {
                     {item.title}
                 </a>
             </h2>
-            {item.type === OwidGdocType.DataInsight ? (
+            {item.type === "data-insight" ? (
```
ikesau, line 241: These changes are just for TS - they get removed in the next PR

rakyi (approved):

PR #6900, opened 2026-08-04
✨ New subscribe page design

ikesau:
- Implements the subscribe/update preferences screens
- Puts all the relevant site code into a `Newsletter` directory
  - The main structure is that there is the _old_ `Subscribe` page, and 2 new pages: `EmailPreferences` and `EmailSubscribe`
  - Subscribe and Preferences share presentation components, logic, and styles where they can - but they have separate wrappers because their state works differently
- Currently deviates slightly from the Figma files as we don't currently have Mailchimp status in D1 (I'm assuming that'll be coming in rakyi's PR)
- Unsubscribe endpoint also isn't working correctly for a similar reason (i.e. assuming it'll change so no need to get it working atm)

image


site/runSiteFooterScripts.tsx
```diff
@@ -68,27 +68,27 @@ function runSearchPage() {
 }
 
 function hydrateSubscribePage() {
```
rakyi, line 70: The naming here is wonky, since we handle three different pages in this function. Should we perhaps make it `hydrateSubscriptionPage`? Or split into different functions etc?

site/Newsletter/useApiSubmit.ts
```diff
@@ -0,0 +1,34 @@
...
+    /** Runs the callback with the choreography below; never throws. */
+    submit: (callback: () => Promise<void>) => Promise<void>
+}
+
+/**
+ * The submit choreography shared by the email-notifications forms: clear any
+ * previous error, mark the request as in flight, and turn a thrown error into
+ * a user-facing message (ApiError messages verbatim, anything else generic).
+ */
+export function useApiSubmit(): ApiSubmitState {
```
rakyi: We could use a react-query mutation instead of a custom hook.

site/Newsletter/emailNotificationsApi.ts
```diff
@@ -0,0 +1,67 @@
...
+async function requestJson<T extends ApiResponse>(
+    path: string,
+    init?: RequestInit
+): Promise<{ response: Response; json: T }> {
+    const response = await fetch(
+        `${EMAIL_NOTIFICATIONS_API_BASE_URL}${path}`,
+        init
+    )
+    const json: T = await response.json()
+    return { response, json }
```
rakyi: Nit: This is not a nice shape for an API, IMO. We can call `response.json()` at the call site when we need it instead of returning both in an object.

site/Newsletter/SubscribePageLayout.tsx
```diff
@@ -0,0 +1,105 @@
...
+                ) : (
+                    <button
+                        type="button"
+                        className="subscribe-page__confirmation-action"
+                        onClick={action.onClick}
+                        disabled={action.disabled}
+                    >
+                        {action.label}
+                    </button>
+                ))}
```
rakyi: I think can we can lean more into the composability of components and pass this in together with the rest of the children.

site/Newsletter/NewsletterOption.tsx
```diff
@@ -0,0 +1,107 @@
+import * as React from "react"
+import { Checkbox } from "@ourworldindata/components"
+
+/**
+ * A newsletter with a checkbox to opt in or out of it, rendered by both the
+ * subscribe form and the magic-link preferences form. The copy lives in the
+ * wrappers below so the two forms can't drift apart.
+ */
```
rakyi: It's better for the reader to infer the usage from code than to duplicate it in a comment that itself can become outdated.

site/Newsletter/EmailNotificationsSubscribeForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    onSubscribed: (email: string) => void
+}) => {
+    const [email, setEmail] = useState("")
+    const [subscribeToOwidBrief, setSubscribeToOwidBrief] = useState(true)
+    const [followTopics, setFollowTopics] = useState(true)
+    const { isSubmitting, errorMessage, setErrorMessage, submit } =
+        useApiSubmit()
+    const preferences = useNotificationPreferences(topicTagGraph)
+
+    const onSubmit = async (event: React.SubmitEvent<HTMLFormElement>) => {
```
rakyi: A bit nicer convention is to use the "on" prefix for props and "handle" for implementation. See the [docs](https://react.dev/learn/responding-to-events#adding-event-handlers).

site/Newsletter/EmailNotificationsSubscribeForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+                <TextInput
+                    placeholder="Your email address"
+                    type="email"
+                    className="newsletter-form__email sentry-mask"
+                    name="email"
+                    value={email}
+                    onChange={(event) => setEmail(event.target.value)}
+                    required={true}
+                />
+                <button
```
rakyi: Could we use our `<Button>` component?

site/Newsletter/EmailNotificationsPreferencesForm.tsx
```diff
@@ -0,0 +1,411 @@
...
+    // saving unsubscribes from topic updates (but not the OWID Brief).
+    const [followTopics, setFollowTopics] = useState(true)
+    const { isSubmitting, errorMessage, setErrorMessage, submit } =
+        useApiSubmit()
+    const preferences = useNotificationPreferences(topicTagGraph)
+    const { setFromStorage } = preferences
+
+    useEffect(() => {
+        const urlToken = getTokenFromLocation()
+        setToken(urlToken)
```
rakyi: This doesn't have to run in an effect, we can put it e.g. into a top level module variable.

site/Newsletter/EmailNotificationsPreferencesForm.tsx
```diff
@@ -0,0 +1,411 @@
...
+                if (json.subscribedToOwidBrief !== undefined) {
+                    setSubscribedToOwidBrief(json.subscribedToOwidBrief)
+                    setBriefInitialStatus(json.subscribedToOwidBrief)
+                }
+            } catch {
+                // fail soft
+            }
+        }
+        void loadPreferences()
+        void loadBriefStatus()
```
rakyi: Using react-query to fetch the data would be nice. Also, we can combine the two requests into one. We don't really benefit from splitting them.

And we can use react-query mutations for the POST requests.

site/Newsletter/EmailNotificationsPreferencesForm.tsx
```diff
@@ -0,0 +1,411 @@
...
+    const [mode, setMode] = useState<Mode>({ name: "loading" })
+    const [token, setToken] = useState<string | null>(null)
+    const [enteredEmail, setEnteredEmail] = useState("")
+    // null = toggle hidden (Mailchimp unavailable or not yet answered)
+    const [subscribedToOwidBrief, setSubscribedToOwidBrief] = useState<
+        boolean | null
+    >(null)
+    const [briefInitialStatus, setBriefInitialStatus] = useState<
+        boolean | null
+    >(null)
```
rakyi: This seems unnecessarily complex. I think we only need null, true, false and don't need the logic to check whether the user changed something.

site/Newsletter/EmailNotificationsPreferencesForm.tsx
```diff
@@ -0,0 +1,411 @@
...
+            const { response, json } =
+                await apiPost<EmailNotificationsSubscribeResponse>(
+                    "/request-link",
+                    request
+                )
+            throwIfApiError(response, json)
+            setMode({ name: "link-sent" })
+        })
+
+    const save = async (unsubscribe: boolean) => {
```
rakyi: I think it would be simpler if we split this into two different handlers: `handleSave` and `handleUnsubscribe`.

site/Newsletter/EmailNotificationsPreferencesForm.tsx
```diff
@@ -0,0 +1,411 @@
...
+                any time.
+            </SubscribePageConfirmation>
+        ))
+        .with({ name: "expired" }, ({ token: expiredToken }) => (
+            <SubscribePageConfirmation
+                variant="error"
+                heading="This link has expired"
+                action={{
+                    label: isSubmitting ? "Sending…" : "Email me a new link",
+                    onClick: () => void requestLink({ token: expiredToken }),
```
rakyi: Not sure why `void` is here, but it's not necessary.

site/Newsletter/EmailNotificationsPreferencesForm.scss
```diff
@@ -0,0 +1,86 @@
+// The email-address block rendered inside the hero band, above the form.
```
rakyi: I'd rather remove the comment, it doesn't add much.

site/Newsletter/EmailNotificationsPreferencesForm.tsx
```diff
@@ -0,0 +1,411 @@
...
+                            type="email"
+                            className="newsletter-form__email sentry-mask"
+                            name="email"
+                            value={enteredEmail}
+                            onChange={(event) =>
+                                setEnteredEmail(event.target.value)
+                            }
+                            required={true}
+                        />
+                        <button
```
rakyi: This and all the other buttons: we should be using our `<Button>` component, and hopefully saving a few lines of CSS.

site/Newsletter/EmailNotificationsPreferenceFields.tsx
```diff
@@ -0,0 +1,253 @@
...
+import {
+    areAllTopicsSelected,
+    PreferencesValidationErrors,
+} from "./emailNotificationsValidation.js"
+
+/**
+ * Presentation for the "Show me" cards. The content types themselves come from
+ * EMAIL_NOTIFICATIONS_CONTENT_TYPES so that a type added there can't be
+ * silently missing here.
+ */
```
rakyi: This comment confuses me and I'd rather remove it.

rakyi (commented):
We keep passing the whole `topicTagGraph` around in multiple functions, but it looks like we could reduce it to only the area names, since that's the only thing we use from it.

PR #6905, opened 2026-08-05
✨🤖 compare two time points with "vs." instead of a range

sophiamersmann:
> _Written by Claude Sonnet 5 — @sophiamersmann at the wheel._

## Context

Chart titles annotate the time period shown, e.g. "GDP, 1990 to 2020". But some chart types don't show a period — they show two snapshots side by side and let you compare them: a faceted map plots 1990 and 2020 as separate panels, a dumbbell chart draws two dots, a slope chart draws two endpoints. For those, "1990 to 2020" reads as a span even though nothing in between is actually shown.

This adds `formatTimeComparison()` next to the existing `formatTimeRange()` on `CoreColumn`, and uses it for:

- the chart title's time annotation, for faceted maps, dumbbell charts, and slope charts (outside relative mode, where "Change in X, 1990 to 2020" already reads as a period, not a comparison)
- the matching tooltips (map, scatter with a time override, dumbbell, slope), so hovering agrees with the title

Relative slope charts, connected/time scatters, and everything else that genuinely shows a range keep "to".

## Screenshots

_(add before/after of a faceted map title and tooltip)_

## Testing guidance

- Facet a map (pull the timeline handles apart on the map tab) — the title and tooltip should read "1990 vs. 2020"
- Open a dumbbell or slope chart with the handles apart — same wording; switch a slope chart to relative mode and it should go back to "1990 to 2020"
- Open a scatter plot comparing the same variable across two times (x vs. y) — title and tooltip should read "1990 vs. 2020"
- Check line, stacked, and non-time-override scatter charts are unaffected

## Note for reviewer

Not yet reflected in `owid-grapher-svgs`. Running `make svgtest.full` will show diffs on faceted maps, the `tab=slope` grapher-views, and a handful of slope-default thumbnails/mdims — all just the time-annotation wording.

### DIFF
--- a/packages/@ourworldindata/core-table/src/CoreTableColumns.test.ts
+++ b/packages/@ourworldindata/core-table/src/CoreTableColumns.test.ts
@@ -47,6 +47,14 @@ describe(ColumnTypeNames.Quarter, () => {
             col.formatTimeRange(day("2025-02-15"), day("2026-11-20"))
         ).toEqual("Jan 2025 to Dec 2026")
     })
+
+    it("formats a comparison as the two quarters themselves", () => {
+        const day = (iso: string): number =>
+            convertDateToDaysSinceEpoch(dayjs.utc(iso))
+        expect(
+            col.formatTimeComparison(day("2025-02-15"), day("2026-11-20"))
+        ).toEqual("Q1 2025 vs. Q4 2026")
+    })
 })
 
 describe(ColumnTypeNames.Decade, () => {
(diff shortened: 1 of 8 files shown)


sophiamersmann:
* **#6905** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6905?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6905?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #6918, opened 2026-08-06
Turn @ourworldindata/grapher into a standalone npm package

marcelgerber:

Supersedes #6341, which went stale — this picks up that work and carries it through to a buildable, testable, documented package.

## Summary

Turns `@ourworldindata/grapher` into a standalone, publishable npm package: a public `GrapherLoader` API for embedding charts, a tsdown-based build producing an ES module library + standalone browser bundle + bundled type declarations, smoke tests that verify the built artifacts, consumer-facing docs, and CI + publishing workflows.

## The public API

`GrapherLoader` is the entry point: pick a data source via a static factory, then `mount` into a container.

```js
GrapherLoader.fromCsv({
    config: { title: "My chart", selectedEntityNames: ["France", "Germany"] },
    csvUrl: "./data.csv", // or inline: csv: "entityName,entityCode,..."
}).mount(document.getElementById("chart"))
```

- `fromTable({ config, data })` — pre-built `OwidTable`
- `fromCsv({ config, csv | csvUrl, columnDefs? })` — CSV, inline or fetched (exactly one of `csv`/`csvUrl`, enforced by the types)
- `fromApi({ config, dataApiUrl? })` — OWID indicators API; `config.dimensions` is required by the types
- The loader instance exposes `grapherState` (MobX escape hatch), `ready` (resolves when data has loaded, rejects on fetch failure without unhandled-rejection noise), `mount` (throws on double-mount), and `dispose`
- Container sizing is observed via `ResizeObserver`, and the chart renders at the correct size on first paint; rendering is lazy (in-view)
- The mounted chart never feeds back into the container's layout (`contain: size` on the mount wrapper), so CSS-driven sizing like `aspect-ratio` works in both directions — without this, a chart rendered at a larger size props the container open and it can grow but never shrink
- Charts whose data carries no source metadata (typical for external CSVs) get a graceful sources modal notice instead of a broken empty modal
- TS consumers get everything needed to build configs: `GrapherInterface`, `OwidChartDimensionInterface`, `OwidColumnDef`, and the `DimensionProperty` / `ColumnTypeNames` enums

## Build outputs (`yarn build`, via tsdown)

- `dist/grapher.js` — ES module build, react/react-dom external (peer deps)
- `dist/grapher.standalone.min.js` — self-contained, minified bundle for plain HTML pages (React included), also exposed as the `./standalone` export subpath
- `dist/grapher.css` — compiled styles
- `dist/grapher.d.ts` — bundled declarations for the public API (workspace-package types inlined)

The published package is dependency-free: everything except react/react-dom is bundled, so all runtime deps live in `devDependencies` and consumers install nothing else.

## Testing (`yarn testPackage` + CI workflow)

Smoke tests in `packageTest/` (16 tests) run against the built artifacts: import both JS builds, verify the external/bundled dependency split, mount real charts from the built code in happy-dom, and `yarn pack` the package to typecheck a simulated external consumer against the bundled d.ts (`moduleResolution: bundler` and `nodenext`), plus a full check of the declaration bundle itself and an [`@arethetypeswrong/cli`](https://github.com/arethetypeswrong/arethetypeswrong.github.io) pass over the packed tarball. A new `grapher-package.yml` workflow runs this on PRs touching the relevant paths.

These tests caught three real bugs in the published artifacts, fixed here at the source:

- `FuzzySearch` leaked fuzzysort's global ambient `Fuzzysort` namespace into the public type surface (unresolvable for consumers) — replaced with structural mirror types
- `MarkdownTextWrap.props` derived its return type from a `private static`, whose type is erased in declaration emit, degenerating the type to `string | number | symbol` for consumers — the static is now `protected`, which keeps its type in the d.ts
- react-aria's vendored CollectionBuilder imports the CJS-only `use-sync-external-store` shim, whose `require("react")` survived into the ESM build as a runtime `__require` call that throws for consumers — the shim is now aliased to react itself (React 19, our peer floor, ships `useSyncExternalStore` natively), with regression assertions that neither build contains `__require(`

## Publishing previews

The `grapher-publish-github.yml` workflow (dispatchable from any branch; the workflow file itself landed on master in #6921 so it's registered) builds, tests, and publishes the package to GitHub Packages as `@owid/grapher` under the `preview` dist-tag, versioned `0.0.0-preview.<run number>` — for testing the real install flow before the package exists on npm proper.

## Docs & demos

- `readme.md` restructured around package usage: installation, React/bundler and plain-HTML quick starts, data loading, config schema pointer, full `GrapherLoader` reference, sizing/fonts/TypeScript notes, and preview-publishing docs
- `demo.html` shows all three loading modes, served as-is via `yarn startDemoServer`

## Test plan

- [x] `yarn testPackage` (16 tests: importability, mounting, packed-types checks)
- [x] `yarn typecheck`, `yarn test`, lint & format
- [x] Demo pages against the built bundle via `yarn startDemoServer`
- [x] Published previews to GitHub Packages via the workflow


marcelgerber:
@codex review

Issue #4928, opened 2025-05-27
consider supporting view-specific tab and entity selection in MDIMs


veronikasamborska1994:
## Core problem

When working with MDIMs, there is currently no way to set a different default chart type (e.g. map vs. bar chart) for each view. This can result in awkward visualizations, especially when only one entity (like "World") is selected by default — leading to a single bar shown in a bar chart, which doesn’t make sense visually.

For example, when only one country or region is selected and the chart type is a DiscreteBar, the result is often unhelpful. In these cases, switching to a map view would be much nicer. 

## Proposed solution

We could define a default chart type (tab) per view in MDIMs, provided the user hasn't interacted with the tab selector (as otherwise, I can see how it would be confusing). 

The logic could be something like:

If the chart type is DiscreteBar and there is only one entity in `default_selection`, automatically show the map view by default (if available), provided the user has not manually changed the country or chart type selection.

Or we could allow for a different `default_selection` selection per view. 

## Context

This issue was discussed on slack [here](https://owid.slack.com/archives/C071F90MD0X/p1747997482393459). A screenrecording of why the current implementation can look quite weird below. 

https://github.com/user-attachments/assets/4e539a0f-7803-4f3f-8331-7e135529a87e

danyx23:
The two properties we try to keep consistent between view changes without a way to override them is entity selection (`.selectedEntityNames`) and tab choice (`.tab`). We don't want to change this behaviour fully (keeping these in sync across view changes) but we could track if the user actively changed each of these settings and if not then we _do_ apply the defaults from the view as given. 

This would mean that if the user views a chart view and has not changed the tab, and then they switch to a view where `.tab` is set to map then we would switch the active tab to the map (and similarly for the selection). But if the user _did_ switch the tab themselves (or change the country selection) we would keep the current behaviour.

sophiamersmann:
@pabloarosado, I'm closing this given your recent work on view-specific default selections. Please re-open if there's something else that should be done here.

PR #6922, opened 2026-08-06
Let a multi-dim view define its own default entity selection

pabloarosado:
> _Written by Claude Fable 5 — @pabloarosado at the wheel._

## Problem

A multi-dim view cannot define its own default entity selection. `upsertMultiDim` builds each view's chart config by merging the collection-wide `default_selection` **over** the view's config, so a view-level `selectedEntityNames` is erased before it ever reaches `chart_configs`. Empirically: across all 75 multi-dims in the DB (9,488 views), not a single served view config differs from its collection default.

This is the entity-selection half of #4928, with the semantics proposed there: the view default governs only the initial state; a reader's own selection still carries across views exactly as before (the frontend already only carries over user-*changed* params, which is why per-view `tab` already works on initial load).

## Fix

Apply the collection `default_selection` only to views that don't define their own `selectedEntityNames`. One conditional in `upsertMultiDim`; no frontend changes needed.

## Which multi-dims define view-level selections today

So nothing changes silently, the exhaustive list of collections whose raw config carries view-level `selectedEntityNames` and will start honoring it after their next upsert:

- **`covid/latest/covid#covid_vax_breakdowns`** (unpublished): 2 views. Currently commented out of the etl export step, so it only re-upserts if re-enabled.
- **`covid/latest/covid#covid_cases_tests`** (unpublished): one scatter view defines an empty selection (`selectedEntityNames: []`), i.e. open with no countries highlighted; the collection default currently overrides it.
- **The three energy multi-dims** arriving with owid/etl#6393 (`energy_mix`, `electricity_mix`, `fossil_fuels`; unpublished until that migration completes): 214 single-source views define multi-country comparisons, and two electricity net-imports views define selections with data (World has none there, so their chart tabs currently render empty — the concrete motivation for this fix).

No published multi-dim view defines the field, so deploying this changes nothing user-visible until those upserts run.

## Demo

The covid collection has been re-upserted on this PR's staging server, so the effect is visible against a production snapshot: the [manufacturer view on staging](http://staging-site-mdim-view-default-selection/admin/grapher/covid-vax-breakdowns?breakdown=manufacturer) opens with the view's own selection (European Union (27)), while the [same view on production](https://admin.owid.io/admin/grapher/covid-vax-breakdowns?breakdown=manufacturer) still shows the collection default (the four income groups, which have no manufacturer data).


pabloarosado:
@copilot review

pabloarosado:
@codex review

rakyi (approved):
Sorry for the late review, Pablo. It looks good to me, but I don't know that much about the subtleties of Grapher. @sophiamersmann could you double-check this, please?

sophiamersmann:
Makes sense to me as well! @pabloarosado, I'm currently refactoring chart configs throughout the codebase. Could you merge quickly so I can rebase onto this work? If there's some blocker for merging now, then it'd be best to wait until the refactor is done.

PR #6951, opened 2026-08-10
Support multi-dim links in explorer tiles

marcelgerber:
> _Written by Claude Fable 5 — @marcelgerber at the wheel._

`explorer-tiles` blocks in gdocs only accepted `/explorers/` URLs. Since multi-dim data pages are the successors to explorers but live under `/grapher/`, this PR makes the block work with mdim links too: they parse, validate (with a gdoc error if a `/grapher/` URL turns out to be a regular chart rather than an mdim), and render as tiles — including the topic icon, which mdim linked charts never got before.

One point for discussion: mdim tiles keep the red "Data Explorer" suffix after the title, consistent with how explorers that redirect to mdims already render in these tiles. Happy to change that if we want different text for mdims.

- **Parsing** (`db/model/Gdoc/rawToEnriched.ts`): `parseExplorerTiles` now accepts grapher-type URLs alongside explorer ones. Whether a `/grapher/` URL is actually an mdim requires DB access, so that check happens at validation time instead.
- **Validation** (`db/model/Gdoc/GdocBase.ts`): explorer-tiles links that resolve to a *regular* grapher chart (a chart-slug hit with no mdim redirect, mirroring the linked-chart resolution order in `loadLinkedChartsForSlugs`) produce a gdoc error in the admin.
- **Rendering** (`site/gdocs/components/ExplorerTiles.tsx`): mdim `LinkedChart`s render as tiles; plain-grapher links are hidden in production and show the error while previewing.
- `makeMultiDimLinkedChart` now populates `tags` from the mdim config's `topicTags` — it previously always set `tags: []`, so mdim tiles would never show the topic icon.
- Mdim links resolve through the existing attachments layer (`makeMultiDimLinkedChart`); explorer→mdim redirects already exercised that path in this component, so direct `/grapher/` links were the only missing piece.
- Testing: `yarn typecheck`, `yarn testLintChanged`, `yarn testFormatChanged` pass; gdoc round-trip tests (`db/gdocTests.test.ts`, 81 tests) pass.

</details>

### DIFF
--- a/db/model/Gdoc/GdocBase.ts
+++ b/db/model/Gdoc/GdocBase.ts
@@ -1285,7 +1285,7 @@ export class GdocBase implements OwidGdocBaseInterface {
                     ) {
                         linkErrors.push({
                             property: "content",
-                            message: `Grapher chart with slug "${link.target}" does not exist or is not published`,
+                            message: `Chart or multi-dim data page with slug "${link.target}" does not exist or is not published`,
                             type: OwidGdocErrorMessageType.Error,
                         })
                     } else if (
@@ -1319,7 +1319,7 @@ export class GdocBase implements OwidGdocBaseInterface {
                     ) {
                         linkErrors.push({
                             property: "content",
-                            message: `Explorer chart with slug "${link.target}" does not exist or is not published`,
+                            message: `Explorer with slug "${link.target}" does not exist or is not published`,
                             type: OwidGdocErrorMessageType.Error,
                         })
                     }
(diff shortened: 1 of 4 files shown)


ikesau (approved):
Sweet!

I found a few little misnomers in our validation verbiage while testing with grapher redirect slugs and whatnot, but I think they're out of scope / not that important to fix.

marcelgerber:
> _Written by Claude Fable 5 — @marcelgerber at the wheel._

Took a stab at the validation misnomers anyway (b241fea): the "Grapher chart with slug … does not exist" error now says "Chart or multi-dim data page", the explorer one no longer calls explorers "charts", and the client-side preview error in `useLinkedChart` got the same treatment.

PR #6959, opened 2026-08-10
🎉🤖 explain time tolerance in the chart's note

sophiamersmann:

Supersedes #6743 

Joe & Este found the info icon too subtle. The tolerance notice should be on screen and is appended to Grapher’s note in this PR.

The tolerance note is only shown when it applies to the currently active view, i.e. at least one of the currently plotted entities has a tolerance value for the currently selected year.

The main problem with this approach is that the note appears and disappears as you drag the time slider. This causes layout jumps that also affect the time slider itself, so the drag target can jump away underneath you.

What I tried to get rid of the layout shift:

* I tried moving the note above the slider (below the subtitle, in one of the chart corners, below or above the chart), but it all looked weird. The note is the best place for it.
* I tried reserving space for the tolerance notice even when it isn’t shown for the selected year, but that introduced weird whitespace that looked buggy and also unnecessarily ate into the available space for the chart.
* I also tried simply always showing the note, but that’s inaccurate in some cases even when carefully worded. We don’t want to talk about tolerance when it isn’t relevant for the current view.

So, in the end, I landed on this solution to make the layout shift less bad: the note is frozen during a timeline drag. While you’re dragging, the notice can be momentarily stale (it holds whatever was true when you grabbed the handle) and updates on release. The same applies when the timeline is animated.

Wording

When a single year is plotted:

> Where data for 2009 is unavailable, the value from the closest year between 2006 and 2010 is shown instead.

When two time points are plotted:

> Where data is unavailable, the closest value within 3 years is shown instead.

### Shape

New `packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts` holds `makeToleranceNotice({ inputTable, transformedTable, columns, timeTolerance, toleranceStrategy })`, which is what each chart state calls, over three helpers:

- `formatToleranceNotice({ timeColumn, timeTolerance, timeRange, toleranceStrategy })` — builds the sentence. `timeColumn` is the time column of the table the chart draws; `timeRange` is the first and last time in its unfiltered data, so the window doesn't collapse as the timeline moves.
- `columnsWithToleranceApplied(table, columns)` — of the given columns, those with a value that sits at a different time from the one it's shown for, by comparing the column's values against its `originalTime` column.
- `getMaxConfiguredTolerance(columns)` — the widest tolerance across the columns a chart matches values over, taken over the columns that were substituted rather than all of them.

The sentence is chosen in this order:

1. No tolerance, no time column, or no times at all → no notice.
2. The data covers a single time → no notice; tolerance can't substitute anything.
3. The tolerance spans the whole dataset → "…the closest value is shown instead", because indicators configured that way use a sentinel like `9999`, which would otherwise read as "within 9999 years".
4. The chart plots exactly one time and the data is yearly → the detailed sentence, with the window clipped to the data's first and last year and to the direction the tolerance reaches. A one-directional window at the edge of the data reaches nothing, so that case gets no notice either.
5. Otherwise → "…the closest value within 3 years is shown instead".

Each opting-in `*ChartState` exposes `toleranceNotice` (added to `ChartInterface`), which names only what is chart-specific:

```ts
@computed get toleranceNotice(): string | undefined {
    return makeToleranceNotice({
        inputTable: this.inputTable,
        transformedTable: this.transformedTable,
        columns: this.yColumns,
    })
}
```

`makeToleranceNotice` returns early when none of those columns has a tolerance configured, so those charts never pay for the row scan.

`GrapherState` merges it into the authored note as `effectiveNote`, which `Footer` now reads instead of `note`. The two are run together as sentences, terminating the authored note if the author didn't, so the notice doesn't read as part of it.

### The year comes from the times the chart plots

There is no per-chart-type "am I a single-year chart" flag: `formatToleranceNotice` reads `timeColumn.uniqTimesAsc` off the table the chart draws. One time means the chart is labelled with that year; a slope, time-range dumbbell or faceted map holds two, and a connected scatter more. This keeps the notice correct by construction — a chart type that changes what it plots (as the dumbbell did in #6961) carries the notice along with it.

### Direction

The window is stated as a range that excludes the year on screen, since that year having no data is the premise: "between 2006 and 2010" where it reaches both ways, "between 2007 and 2009" where it only reaches back, and — where it leaves a single year — "2009". That last case also catches a wide tolerance clipped by a short dataset (data 2022–2023, tolerance 3 → "the value from 2022").

`ToleranceStrategy` (`closest` | `backwards` | `forwards`) is resolved by `MapChartState` the same way `interpolateColumnWithTolerance` resolves it — `mapConfig.toleranceStrategy ?? mapColumn.toleranceStrategy` — and clamps one side of the window to the target year. The map is the only caller that passes one, since nothing populates the column-level strategy today; the parameter defaults to `closest`. The plainer sentences carry the direction as an adjective instead: "the closest earlier value within 3 years", "the closest later value".

### "On screen" is decided by the table

`columnsWithToleranceApplied` is called with a table already narrowed to what's visible. Chart types whose transform filters to the selected entities get that for free; the map narrows to the active continent itself via `transformedTableForActiveRegion` when zoomed in 2D. There's no separate entity-list parameter — the table is the single source of truth for what counts.

### Freezing during timeline interaction

`GrapherState.frozenToleranceNotice` stores the notice as of the start of an interaction. It's boxed (`{ notice: string | undefined }`) so that freezing an *absent* notice is distinguishable from not freezing at all — otherwise a notice that wasn't showing when the drag started would appear mid-drag, which is what the freeze exists to prevent. A reaction in `Grapher` drives it from `isTimelineInteractionActive`.

### Performance

`columnsWithToleranceApplied` reads the raw column store and moves on to the next column at the first deviation. Two nicer-reading versions were benchmarked and rejected: `column.owidRows` (~3x slower warm, 40–140x cold, since it materializes an object per row) and `times.some()` (~3x slower, one call per row). Worst case is a chart with a tolerance where nothing was substituted, which visits every row: ~0.05 ms on a 28k-row table.

### `OwidTable` changes

- New `timeRange` getter: `[Time, Time]`, or `undefined` if the table has no times.
- `minTime` and `maxTime` are now both `Time | undefined`. `minTime` previously cast its `undefined` away with `as Time` while `maxTime` didn't — the runtime behaviour is unchanged, but the `?? 0` / `?? 1900` / `?? 2000` fallbacks already in `DumbbellChartState`, `SlopeChartState` and `ScatterPlotChart` are now backed by the type instead of contradicting it.
- `filterByTimeRange` goes through `timeRange`, which drops the two `this.maxTime!` assertions it used to need.

### Also in this PR

Removes an unused `IRFragment` export from `@ourworldindata/components`. Leftover from an earlier approach to this feature; the class is only used inside its own package.

### Tests

`ToleranceNotice.test.ts` covers the sentence's branches directly: the detailed sentence and its clipping, the backwards-only and forwards-only windows (both from sitting at the edge of the data and from a one-directional strategy), the single-year window, a one-directional window with nothing to reach, the plainer sentence with and without a direction, a monthly indicator's tolerance stated in days, the unbounded sentinel, and the absent cases.

`GrapherState.test.ts` covers the integration on a map: the exact sentence, appearing and disappearing as the timeline moves between years with and without gaps, the freeze during a drag, gaps in unselected entities being ignored (via a slope chart), a gapless chart tab staying silent, and the four `effectiveNote` composition rules.

Not covered, if a reviewer wants more: the notice being wired up in each individual chart state (only the map and slope are exercised end to end), a faceted map falling back to the plainer sentence, and the scatter's relative-mode case described above.

</details>

### DIFF
--- a/packages/@ourworldindata/core-table/src/OwidTable.ts
+++ b/packages/@ourworldindata/core-table/src/OwidTable.ts
@@ -16,6 +16,7 @@ import {
 } from "@ourworldindata/utils"
 import {
     Time,
+    TimeRange,
     TransformType,
     CoreColumnStore,
     Color,
@@ -109,7 +110,7 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
     }
 
     /** The first and last time in the table, undefined if it has no times */
-    @imemo get timeRange(): [Time, Time] | undefined {
+    @imemo get timeRange(): TimeRange | undefined {
         const { minTime, maxTime } = this
         if (minTime === undefined || maxTime === undefined) return undefined
         return [minTime, maxTime]
--- a/packages/@ourworldindata/core-table/src/OwidTable.ts
+++ b/packages/@ourworldindata/core-table/src/OwidTable.ts
@@ -100,12 +100,19 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
         return this.entityIdColumn.slug
     }
 
-    @imemo get minTime(): Time {
-        return _.min(this.allTimes) as Time
+    @imemo get minTime(): Time | undefined {
+        return _.min(this.allTimes)
     }
 
-    @imemo get maxTime(): number | undefined {
+    @imemo get maxTime(): Time | undefined {
         return _.max(this.allTimes)
+    }
+
+    /** The first and last time in the table, undefined if it has no times */
+    @imemo get timeRange(): [Time, Time] | undefined {
+        const { minTime, maxTime } = this
+        if (minTime === undefined || maxTime === undefined) return undefined
+        return [minTime, maxTime]
     }
 
     @imemo private get allTimes(): Time[] {
@@ -207,16 +214,20 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
     filterByTimeRange(start: TimeBound, end: TimeBound): this {
         if (this.isBlank) return this
 
-        // We may want to do this time adjustment in Grapher instead of here.
-        const adjustedStart = start === Infinity ? this.maxTime! : start
-        const adjustedEnd = end === -Infinity ? this.minTime : end
-        // todo: we should set a time column onload so we don't have to worry about it again.
+        // Nothing to filter by if the table has no times
+        if (!this.timeRange) return this
+        const [minTime, maxTime] = this.timeRange
+
+        // We may want to do this time adjustment in Grapher instead of here
+        const adjustedStart = start === Infinity ? maxTime : start
+        const adjustedEnd = end === -Infinity ? minTime : end
+        // TODO: we should set a time column onload so we don't have to worry about it again
         const timeColumnSlug = this.timeColumn?.slug || OwidTableSlugs.Time
 
         const description = `Keep only rows with Time between ${adjustedStart} - ${adjustedEnd}`
 
         // perf: if the time range is greater than the table's time range, we can skip the filter
-        if (adjustedStart <= this.minTime && adjustedEnd >= this.maxTime!)
+        if (adjustedStart <= minTime && adjustedEnd >= maxTime)
             return this.noopTransform(description).sortedByTime
 
         return this.columnFilter(
--- a/packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
+++ b/packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
@@ -7,12 +7,47 @@ import {
     ColumnSlug,
     Time,
     TimeInterval,
+    TimeRange,
     ToleranceStrategy,
 } from "@ourworldindata/types"
 import { isSubYearly } from "@ourworldindata/utils"
 
-/** The sentence explaining the chart's tolerance, if it has one worth stating */
+/**
+ * The notice a chart should show, if it has a tolerance worth stating and any
+ * value it plots is actually filled in from another time.
+ */
 export function makeToleranceNotice({
+    inputTable,
+    transformedTable,
+    columns,
+    timeTolerance,
+    toleranceStrategy,
+}: {
+    inputTable: OwidTable
+    transformedTable: OwidTable
+    columns: CoreColumn[]
+    timeTolerance?: number
+    toleranceStrategy?: ToleranceStrategy
+}): string | undefined {
+    const tolerance =
+        timeTolerance ??
+        Math.max(0, ...columns.map((column) => column.tolerance))
+    if (!tolerance) return undefined
+
+    const notice = formatToleranceNotice({
+        timeColumn: transformedTable.timeColumn,
+        timeRange: inputTable.timeRange,
+        timeTolerance: tolerance,
+        toleranceStrategy,
+    })
+    // Skip the scan below when there's no notice for it to caption
+    if (!notice) return undefined
+
+    const slugs = columns.map((column) => column.slug)
+    return hasToleranceApplied(transformedTable, slugs) ? notice : undefined
+}
+
+export function formatToleranceNotice({
     timeColumn,
     timeTolerance,
     timeRange,
@@ -20,7 +55,7 @@ export function makeToleranceNotice({
 }: {
     timeColumn: CoreColumn
     timeTolerance: number
-    timeRange: [Time, Time] | undefined
+    timeRange: TimeRange | undefined
     toleranceStrategy?: ToleranceStrategy
 }): string | undefined {
     if (!timeTolerance || timeColumn.isMissing) return undefined
@@ -51,12 +86,21 @@ export function makeToleranceNotice({
             toleranceStrategy === ToleranceStrategy.backwards
                 ? targetTime
                 : Math.min(targetTime + timeTolerance, lastTime)
+
+        // The target year has no data, so a window ending on it stops a year short
+        const start = from === targetTime ? from + 1 : from
+        const end = to === targetTime ? to - 1 : to
+
         // A one-directional window at the edge of the data reaches nothing
-        if (from === targetTime && to === targetTime) return undefined
+        if (start > end) return undefined
 
-        const window = formatToleranceWindow(timeColumn, targetTime, [from, to])
+        const format = (time: Time): string => timeColumn.formatTime(time)
+        const window =
+            start === end
+                ? format(start)
+                : `the closest year between ${format(start)} and ${format(end)}`
 
-        return `Where data for ${timeColumn.formatTime(targetTime)} is unavailable, the value from ${window} is shown instead.`
+        return `Where data for ${format(targetTime)} is unavailable, the value from ${window} is shown instead.`
     } else {
         // "closest", plus the direction where the tolerance only looks one way
         const closest =
@@ -84,9 +128,9 @@ export function makeToleranceNotice({
 
 /**
  * Whether any value in the given columns is filled in from a different time
- * than the one it's shown for. Typically called with a `tranformedTable`.
+ * than the one it's shown for. Typically called with a `transformedTable`.
  */
-export function hasToleranceApplied(
+function hasToleranceApplied(
     table: OwidTable,
     columnSlugs: ColumnSlug[]
 ): boolean {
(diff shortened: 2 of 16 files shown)


sophiamersmann:
* **#6959** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6959?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6959?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6961** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6961?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

packages/@ourworldindata/grapher/src/mapCharts/MapChartState.ts
```diff
@@ -317,6 +336,35 @@ export class MapChartState implements ChartState, ColorScaleManager {
...
+    }
+
+    /** Whether any value shown right now is filled in from another time */
+    @computed private get isToleranceApplied(): boolean {
+        // Skip the scan below when there's no notice for it to caption
+        if (!this.toleranceNoticeIfApplied) return false
+
+        return hasToleranceApplied(this.transformedTableForActiveRegion, [
+            this.mapColumnSlug,
+        ])
```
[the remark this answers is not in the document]
sophiamersmann: Addressed in a separate PR

packages/@ourworldindata/grapher/src/dumbbellCharts/DumbbellChartState.ts
```diff
@@ -354,6 +359,29 @@ export class DumbbellChartState implements ChartState {
...
+            timeTolerance: getMaxConfiguredTolerance(this.yColumns),
+        })
+    }
+
+    /** Whether any value shown right now is filled in from another time */
+    @computed private get isToleranceApplied(): boolean {
+        // Skip the scan below when there's no notice for it to caption
+        if (!this.toleranceNoticeIfApplied) return false
+
+        return hasToleranceApplied(this.transformedTable, this.yColumnSlugs)
```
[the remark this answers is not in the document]
sophiamersmann: Addressed in a separate pr

packages/@ourworldindata/core-table/src/OwidTable.ts
```diff
@@ -207,16 +214,20 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
...
-        const adjustedEnd = end === -Infinity ? this.minTime : end
-        // todo: we should set a time column onload so we don't have to worry about it again.
+        // Nothing to filter by if the table has no times
+        if (!this.timeRange) return this
+        const [minTime, maxTime] = this.timeRange
+
+        // We may want to do this time adjustment in Grapher instead of here
+        const adjustedStart = start === Infinity ? maxTime : start
+        const adjustedEnd = end === -Infinity ? minTime : end
+        // TODO: we should set a time column onload so we don't have to worry about it again
```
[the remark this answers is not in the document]
sophiamersmann: I just upper-cased it

packages/@ourworldindata/core-table/src/OwidTable.ts
```diff
@@ -100,14 +100,21 @@ export class OwidTable extends CoreTable<OwidRow, OwidColumnDef> {
         return this.entityIdColumn.slug
     }
 
-    @imemo get minTime(): Time {
-        return _.min(this.allTimes) as Time
+    @imemo get minTime(): Time | undefined {
+        return _.min(this.allTimes)
```
sophiamersmann, line 105: Drive-by change; I don't think the assertion was safe (also inconsistent to do this for minTime, but not for maxTime)

packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
```diff
@@ -0,0 +1,151 @@
...
+            : undefined
+
+    const [firstTime, lastTime] = timeRange
+
+    // Tolerance can't be applied if the chart only has a single time
+    if (firstTime === lastTime) return undefined
+
+    // Detailed notice for a chart with yearly data that plots a single year,
+    // all other cases (time range plotted, sub-yearly data) use a simpler notice
+    if (targetTime !== undefined && !isSubYearly(timeColumn.timeInterval)) {
```
sophiamersmann, line 84: Sub-yearly data gets a simpler note because spelling out the window is overly specific for daily data (I think)

packages/@ourworldindata/grapher/src/scatterCharts/ScatterPlotChartState.ts
```diff
@@ -432,6 +437,47 @@ export class ScatterPlotChartState implements ChartState, ColorScaleManager {
...
+    /** The notice itself, regardless of whether it currently applies */
+    @computed private get toleranceNoticeIfApplied(): string | undefined {
+        // Time scatters plot time itself on the x axis, so there's no target
+        // time a value could be missing for
+        if (this.isTimeScatter) return undefined
+
+        return makeToleranceNotice({
+            timeColumn: this.transformedTable.timeColumn,
+            timeRange: this.inputTable.timeRange,
+            timeTolerance: getMaxConfiguredTolerance(this.toleranceColumns),
```
sophiamersmann: Just grabbing the bigger tolerance for practicality

packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
```diff
@@ -0,0 +1,151 @@
...
+    // its data, which is the norm: charts default to the latest year
+    const isLookingAhead = to > targetTime
+    const bound = isLookingAhead ? to : from
+
+    // Name the year itself where the window leaves just the one
+    if (Math.abs(bound - targetTime) === 1) return format(bound)
+
+    return isLookingAhead
+        ? `the closest year up to ${format(to)}`
+        : `the closest year back to ${format(from)}`
```
sophiamersmann: I wasn't sure how to word this when the tolerance is applied one-sided - please help 🙏🏻

ikesau: Yeah this is tricky because it needs to be so terse. I would check with Joe and Este.

One longer format, not sure if it's worth it / better:
- `earliest year between 1500 and 1510 where data is available`
- `latest year between  2020 and 2023 where data is available`

ikesau: These notices themselves could have tooltips in interactive mode to further explain how tolerance works if necessary, I suppose, or we could write a small article about it.

packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
```diff
@@ -0,0 +1,151 @@
...
+import {
+    ColumnSlug,
+    Time,
+    TimeInterval,
+    ToleranceStrategy,
+} from "@ourworldindata/types"
+import { isSubYearly } from "@ourworldindata/utils"
+
+/** The sentence explaining the chart's tolerance, if it has one worth stating */
+export function makeToleranceNotice({
```
ikesau, line 18: There's a discrepancy here on "available value" vs "value"

Image

Image

Maybe there's a reason that they're different that I'm not realizing, but if you wanted to be consistent, you could say `Where data for 1822 is unavailable, the closest available value from between 1812 and 1832 is shown instead.` instead?

Or maybe, better yet, just check this stuff with Joe and Este. 😅

ikesau (approved):
Nice work. I sat and thought about ways to cut this gordian knot but I think you're right that (minimized) reflow is a necessary sacrifice.

I QA'd each of the cases the code alluded to, but didn't comb through the code with a fine-toothed comb. Lmk if you want a more detailed double-check

packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
sophiamersmann: That's a good catch - I didn't mean to drop the 'available' in those cases.

The most common sentence `Where data for {year} is unavailable, the value from the closest year between {year_min} and {year_max} is shown instead.` was suggested by Ed, the edge cases around are free-styled.

packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
sophiamersmann: Yeah, tooltips have been suggested to me elsewhere as well. I experimented with it, but didn’t really see the point — I think the note as-is explains what happens quite well? Or, at least it should, I think. Anyway, I gave up on that because I wanted to focus on getting the note itself over the finish line.

sophiamersmann: Ok, I'll ask Ed about these ones (he proposed the wording) because this one is a pretty common case

ikesau: Yeah I doubt anyone could make the case that tooltips or articles are _necessary_. I just love [inside baseball](https://en.wikipedia.org/wiki/Inside_baseball_(metaphor)). 😅

packages/@ourworldindata/grapher/src/chart/ToleranceNotice.ts
ikesau: Yeah... in a sense "available" is redundant. How could it be used if it isn't available? But somehow I think it communicates more clearly when present.

PR #6966, opened 2026-08-11
🐝 Upgrade MySQL to 8.4 LTS (dev + staging)

Marigold:
Pins the dev database to `mysql:8.4`.

The `mysql:8` tag **already resolves to 8.4.11** — Docker Hub moved it when 8.4 became the newest 8.x line. So freshly pulled dev environments have silently been running 8.4 while production runs 8.0.46, and which one you get depends on when you last pulled. This pins it explicitly so the version stops drifting in either direction.

It also matches where the servers are going: MySQL 8.0 left extended support on 30 Apr 2026 and 8.0.46 was its final upstream release. The fleet side is owid/ops#600.

Nothing else changes. `mysql2` already speaks `caching_sha2_password`, and the DB init scripts create users with a bare `IDENTIFIED BY`, so they inherit the server default rather than the auth plugin 8.4 no longer loads.

## How to test it

```
docker compose -f docker-compose.grapher.yml down
docker compose -f docker-compose.grapher.yml up
make dbtest
```

8.4 upgrades an existing 8.0 datadir in place on first start. If that fails, drop the volume and `make refresh.atomic` — a 3 GB download and nothing else lost.

Verified locally: `docker compose -f docker-compose.dbtests.yml up --abort-on-container-exit --exit-code-from db-load-data` brings up **8.4.11** and `db-load-data` exits **0**, which covers database creation, user creation and the migrations mount.

`staging-site-upgrade-mysql` is running 8.4.11 for this branch.

### DIFF
--- a/docker-compose.dbtests.yml
+++ b/docker-compose.dbtests.yml
@@ -3,7 +3,7 @@
 services:
   # Stock mysql database. Root password is hardcoded for now
   db:
-    image: mysql:8
+    image: mysql:8.4
     command: --log-bin-trust-function-creators=ON
     restart: always
     volumes:
@@ -15,9 +15,9 @@ services:
       MYSQL_ROOT_PASSWORD: weeniest-stretch-contaminate-gnarl
       MYSQL_ROOT_HOST: "%"
 
-  # mysql:8 container for running the DB creation scripts
+  # mysql:8.4 container for running the DB creation scripts
   db-load-data:
-    image: mysql:8
+    image: mysql:8.4
     command: "/app/create-test-db.sh"
     volumes:
       - ./devTools/docker:/app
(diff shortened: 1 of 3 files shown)


Marigold:
Verified locally now (I'd flagged this as untested when Docker wasn't running):

```
docker compose -f docker-compose.dbtests.yml up --abort-on-container-exit --exit-code-from db-load-data
```

`db-1` came up on **8.4.11** and `db-load-data` — which runs `create-test-db.sh`, so it covers database creation, user creation and the migrations mount — exited **0**. `--log-bin-trust-function-creators=ON` is still a valid option in 8.4, and the bare `IDENTIFIED BY` in the init scripts correctly inherits `caching_sha2_password` rather than the plugin 8.4 no longer loads.

PR #6977, opened 2026-08-12
🧪🤖 (chart-configs) cover chart config lifecycles before the patch/full refactor

sophiamersmann:

Test-only groundwork for replacing `chart_configs.patch` + `chart_configs.full` with a single `config` column. No production code changes.

Several of the paths that will have to change had no db-test coverage at all:
- chart deletion
- narrative charts lifecycle
- multi-dim view removal
- `republishCharts`

This adds it now, against current behaviour, so the tests are a safety net during the refactor rather than a description of it. `make dbtest` goes from 76 to 86 tests.

### Commits

| commit | what |
| --- | --- |
| `145bc0d` | `insertTestChartConfig` / `insertTestChart` helpers, `seedDatasetAndVariables` fixture, `env.userId`, `TABLES_IN_USE` additions |
| `49e7713` | chart deletion |
| `6dcbd38` | narrative chart lifecycle (new file, 5 cases) |
| `9a5f6b1` | multi-dim view lifecycle (new file, 3 cases) |
| `a7335ec` | dataset republish version bumps |

### Why the helpers exist

`insertTestChartConfig` is the single place the suite writes down what a `chart_configs` row looks like. Four hand-rolled `{patch, full}` inserts encoded it before (`basic.test.ts` ×2, `pageviews.test.ts`, `slug-validation.test.ts`); the refactor now has one place to update, which also matters because `patchConfigId NOT NULL` makes raw inserts fail.

`TABLES_IN_USE` gains `narrative_charts` (before `charts`, `multi_dim_x_chart_configs` and `chart_configs` — it has foreign keys to all three) plus `multi_dim_view_dimensions` and `explorer_view_dimensions`, which have no foreign key and were leaking rows across test files.

### What each new test asserts

**`narrative-charts.test.ts`** — create from a parent chart (the authored layer stores the override and the `NARRATIVE_CHART_PROPS_TO_PERSIST` props even when identical to the parent's, but not inherited props or `id`/`isPublished`/`slug`/`version`; served config equals `mergeGrapherConfigs(parent, patch)`); create from a multi-dim view; update reuses the same `chartConfigId`; no propagation from the parent; delete leaves only the parent's config row. Every assertion goes through `/narrative-charts/:id.config.json`, which returns `configFull`, `configPatch` and `parentConfigFull` — so there are no column reads in this file at all.

**`mdim-views.test.ts`** — a config row and a dimensions row per view; removing a view drops its config and `multi_dim_x_chart_configs` rows while the surviving view keeps its config id, and the `multi_dim_view_dimensions` log keeps both rows; an ETL indicator config set *after* the views exist reaches them through `updateAllMultiDimViewsThatInheritFromIndicator` (only the admin route was covered before), leaves the other view's indicator untouched, and updates rows in place.

**`charts.test.ts`** — deletion, and republish bumping `$.version` in both the served config and the authored layer, with the indicator given an ETL config first so the two are genuinely different objects. Also asserts `charts.updatedAt === chart_configs.updatedAt` and that the indicator's own config row is untouched, which pins the `JOIN charts` scoping.

### Test plan

`make dbtest` green at 14 files / 86 tests. `yarn typecheck`, `yarn testLintChanged`, `yarn testFormatChanged` clean.

Each new test was mutation-checked — the covered behaviour was broken in the production code, the suite run, and the failure confirmed to be that test and only that test:

| mutation | fails |
| --- | --- |
| drop `DELETE FROM chart_configs` from `deleteChart` | chart deletion |
| drop the `NARRATIVE_CHART_PROPS_TO_PERSIST` spread from narrative patch construction | narrative create |
| drop the `cleanUpOrphanedChartConfigs` call from `upsertMultiDim` | view removal |
| drop `cc.patch = JSON_SET(...)` from `republishCharts` | republish |

All mutations reverted; no production file is modified by this PR.

### Note for whoever rebases

`env.userId` in `testEnv.ts` is byte-identical to what #6971 (`chart-config-owner-joins`) adds, and the `slug-validation.test.ts` change here subsumes that PR's edit to the same test. Trivial conflict in one direction or the other.

### Deliberately not included

- A generic "no orphaned config rows" assertion. It was written and cut: in these tests the DB starts empty and the operation is controlled, so a plain row count is both shorter and a stronger assertion.
- Coverage for the ~14 files that reference `full`/`patch` only inside raw SQL strings. There is no baker db-test harness, and building one for a column rename is out of proportion — the acceptance grep plus one local bake is the right check there.
- `export BAKE_ON_CHANGE=false` in `run-db-tests.sh`, next to the existing `ALGOLIA_INDEXING=false`. Worth one line at some point: it is false today only because nobody's `.env` sets it, and the republish test exercises a path that enqueues a deploy when it is true.

</details>

### DIFF
--- a/adminSiteServer/tests/mdim-views.test.ts
+++ b/adminSiteServer/tests/mdim-views.test.ts
@@ -1,0 +1,147 @@
+import { describe, it, expect, beforeEach } from "vitest"
+import { getAdminTestEnv } from "./testEnv.js"
+import {
+    ChartConfigsTableName,
+    GrapherInterface,
+    MultiDimViewDimensionsTableName,
+    MultiDimXChartConfigsTableName,
+} from "@ourworldindata/types"
+import { latestGrapherConfigSchema } from "@ourworldindata/grapher"
+import {
+    otherVariableId,
+    seedDatasetAndVariables,
+    variableId,
+} from "./fixtures.js"
+
+const env = getAdminTestEnv()
+
+describe("Multi-dim views", { timeout: 20000 }, () => {
+    const catalogPath = "test/catalog#path"
+
+    const totalView = {
+        config: { title: "Total energy use" },
+        dimensions: { metric: "total" },
+        indicators: { y: variableId },
+    }
+    const perCapitaView = {
+        config: { title: "Energy use per capita" },
+        dimensions: { metric: "per_capita" },
+        indicators: { y: otherVariableId },
+    }
+
+    function multiDimConfig(views: (typeof totalView)[]): object {
+        return {
+            grapherConfigSchema: latestGrapherConfigSchema,
+            title: { title: "Energy use", titleVariant: "by energy source" },
+            views,
+            dimensions: [
+                {
+                    name: "Metric",
+                    slug: "metric",
+                    choices: views.map((view) => ({
+                        name: view.dimensions.metric,
+                        slug: view.dimensions.metric,
+                    })),
+                },
+            ],
+        }
+    }
+
+    async function upsertMultiDim(views: (typeof totalView)[]): Promise<void> {
+        await env.request({
+            method: "PUT",
+            path: `/multi-dims/${encodeURIComponent(catalogPath)}`,
+            body: JSON.stringify({ config: multiDimConfig(views) }),
+        })
+    }
+
+    /** Maps each view's id (e.g. "metric=total") to its chart config id. */
+    async function getViewConfigIds(): Promise<Record<string, string>> {
+        const rows = await env
+            .testKnex(MultiDimXChartConfigsTableName)
+            .select("viewId", "chartConfigId")
+        return Object.fromEntries(
+            rows.map((row) => [row.viewId, row.chartConfigId])
+        )
+    }
+
+    async function getConfig(chartConfigId: string): Promise<GrapherInterface> {
+        const row = await env
+            .testKnex(ChartConfigsTableName)
+            .where({ id: chartConfigId })
+            .first()
+        return JSON.parse(row.full)
+    }
+
+    beforeEach(async () => {
+        await seedDatasetAndVariables(env)
+    })
+
+    it("creates a config row and a dimensions row per view", async () => {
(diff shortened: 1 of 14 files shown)


adminSiteServer/tests/narrative-charts.test.ts
```diff
@@ -0,0 +1,216 @@
...
+        for (const prop of ["id", "isPublished", "slug", "version"]) {
+            expect(narrativeChart.configPatch).not.toHaveProperty(prop)
+        }
+
+        // The served config is the authored layer merged over the parent
+        expect(narrativeChart.configFull).toEqual(
+            mergeGrapherConfigs(parentConfig, narrativeChart.configPatch)
+        )
+        expect(narrativeChart.configFull.subtitle).toBe("Parent subtitle")
+    })
```
rakyi: IMO it would be better to compare against an inline object with the expected properties instead of calling `mergeGrapherConfigs` since that's what we are implicitly testing. Also in the test below.

adminSiteServer/tests/mdim-views.test.ts
```diff
@@ -0,0 +1,147 @@
...
+        dimensions: { metric: "total" },
+        indicators: { y: variableId },
+    }
+    const perCapitaView = {
+        config: { title: "Energy use per capita" },
+        dimensions: { metric: "per_capita" },
+        indicators: { y: otherVariableId },
+    }
+
+    function multiDimConfig(views: (typeof totalView)[]): object {
```
rakyi: Let's use the type as defined, i.e. `View<IndicatorsBeforePreProcessing>[]` here and below, pls.

rakyi (commented):
Could we rename the multi-dim files you touched/created to use the word "multi-dim" (the new terminology) instead of "mdim" and also use "multi-dim" in comments (and `multiDim` fo JS identifiers)?

Jut a few nitpicks, looks good otherwise!

rakyi (approved):

sophiamersmann:
Thank you for the quick review, @rakyi!

PR #6979, opened 2026-08-12
🔨🤖 (chart-configs) replace chart_configs.patch + full with a single config column

sophiamersmann:

Replaces `chart_configs.patch` + `chart_configs.full` with a single `config` column. A `chart_configs` row now holds one config, and which layer it is follows from the pointer that names it: `charts.configId` for the final config, and the new `charts.patchConfigId` for the patch layer. Same goes for mdim views and narrative charts. `variables.grapherConfigId{ETL,Admin}` are renamed to `patchConfigId{ETL,Admin}` for consistency.

**Tables that gained a patch row**

* Standalone charts: `charts.configId` points to the final config, and the new `charts.patchConfigId` points to what used to be the `patch` config.
* Mdim views: same split, with `multi_dim_x_chart_configs.chartConfigId` for the final config and the new `patchConfigId` for the YAML-authored layer.
* Narrative charts: same again, where `narrative_charts.chartConfigId` is the final config and the new `patchConfigId` points to the patch config.

**Indicator layers**

Indicator layers drop the full config and reconstruct it in code.

* ETL-authored indicator configs: `variables.grapherConfigIdETL` becomes `patchConfigIdETL`, a rename only. The row's patch and full configs were identical.
* Admin-authored indicator configs: `variables.grapherConfigIdAdmin` becomes `patchConfigIdAdmin`. The full config is dropped, and `patchConfigIdAdmin` points to the `patch` config. The final config (etl+admin) is constructed in code.

**Untouched**

* Explorer views: unchanged, since `patch` and `full` were identical.
* Mdim redirects: unchanged, since `patch` and `full` were identical.

## The migration

- Add a new column `patchConfigId` to `charts`, `multi_dim_x_chart_configs`, and `narrative_charts`
- Copy each owner's `patch` config into a row of its own and point the owner at it, keeping the source row's timestamps so chart-sync doesn't read it as a change
- Take `charts_x_parents` off `patch` (the reference to `patch` was dead anyway)
- Run `SET full = patch` on the admin indicator rows, so `config` keeps the patch rather than the merged etl+admin config
- Drop `patch` alongside the three generated columns (`slug`, `chartType`, `fullMd5`)
- Rename `full` to `config`, then put the generated columns and the slug index back
- Rename the `variables` pointers to `patchConfigId{ETL,Admin}`

## Worth a reviewer's attention

Annotated in code comments as well.

* **Behaviour change: the R2 sync script only uploads configs that are served.** It used to upload every `chart_configs.full` config (including indicator configs, which are never read but make up the vast majority, ~200.000). This reduces the configs in the PR from ~250.000 to ~25.000. Drafts are uploaded too, and explorer views as well because `refreshExplorerViews.ts` writes them there too.
* **Behaviour change: `GET /admin/api/chart-configs/:id.config.json` returns the config, not the whole row.**
* **Bug fix: an indicator config change triggered a static build even when nothing published had changed.** A query didn't check `isPublished` correctly, so drafts counted as published.
* **Bug fix: the by-slug config in R2 went stale.** An indicator config propagating into a published chart rewrote only `config/by-uuid/`, not `config/by-slug-published/<slug>.json`, so the latter stayed stale until the next sweep.

## Sibling PRs

- ETL: https://github.com/owid/etl/pull/6667
- Analytics: https://github.com/owid/analytics/pull/1003
- Automation: https://github.com/owid/automation/pull/18

### DIFF
--- a/adminSiteClient/CreateNarrativeChartEditorPage.tsx
+++ b/adminSiteClient/CreateNarrativeChartEditorPage.tsx
@@ -88,7 +88,7 @@ class CreateNarrativeChartEditorPageInternal
         const chartConfig = await this.context.admin.getJSON(
             `/api/chart-configs/${this.props.chartConfigId}.config.json`
         )
-        this.parentConfig = chartConfig.full
+        this.parentConfig = chartConfig
     }
 
     @computed get admin(): Admin {
--- a/adminSiteServer/apiRoutes/charts.ts
+++ b/adminSiteServer/apiRoutes/charts.ts
@@ -8,8 +8,6 @@ import {
     DbPlainChart,
     R2GrapherConfigDirectory,
     DbInsertChartRevision,
-    DbRawChartConfig,
-    ChartConfigsTableName,
     DbChartTagJoin,
     ContentGraphLinkType,
     StaticVizTableName,
@@ -45,7 +43,6 @@ import {
     getWordpressPostReferencesByChartId,
     getGdocsPostReferencesByChartId,
 } from "../../db/model/Post.js"
-import { UpdatedChartInheritanceRecord } from "../../db/model/Variable.js"
 import { enqueueExplorerRefreshJobsForDependencies } from "../../db/model/Explorer.js"
 import { expectInt } from "../../serverUtils/serverUtil.js"
 import {
@@ -54,10 +51,7 @@ import {
     retrieveChartConfigFromDbAndSaveToR2,
     updateChartConfigPairInDbAndR2,
 } from "../chartConfigHelpers.js"
-import {
-    deleteGrapherConfigFromR2,
-    saveGrapherConfigToR2ByUUID,
-} from "../../serverUtils/r2/chartConfigR2Helpers.js"
+import { deleteGrapherConfigFromR2 } from "../../serverUtils/r2/chartConfigR2Helpers.js"
 import { triggerStaticBuild } from "../../baker/GrapherBakingUtils.js"
 import * as db from "../../db/db.js"
 import { getLogsByChartId } from "../getLogsByChartId.js"
@@ -566,23 +560,6 @@ export const saveGrapher = async (
     return {
         chartId,
         savedPatch: patchConfig,
-    }
-}
-
-export async function updateGrapherConfigsInR2(
-    knex: db.KnexReadonlyTransaction,
-    updatedCharts: UpdatedChartInheritanceRecord[],
-    updatedMultiDimViews: { chartConfigId: string; isPublished: boolean }[]
-) {
-    const idsToUpdate = [
-        ...updatedCharts.filter(({ isPublished }) => isPublished),
-        ...updatedMultiDimViews,
-    ].map(({ chartConfigId }) => chartConfigId)
-    const builder = knex<DbRawChartConfig>(ChartConfigsTableName)
-        .select("id", "config", "configMd5")
-        .whereIn("id", idsToUpdate)
-    for await (const { id, config, configMd5 } of builder.stream()) {
-        await saveGrapherConfigToR2ByUUID(id, config, configMd5)
     }
 }
 
--- a/adminSiteServer/apiRoutes/charts.ts
+++ b/adminSiteServer/apiRoutes/charts.ts
@@ -4,7 +4,6 @@ import {
     GrapherInterface,
     JsonError,
     DbPlainUser,
-    Base64String,
     serializeChartConfig,
     DbPlainChart,
     R2GrapherConfigDirectory,
@@ -23,7 +22,6 @@ import {
     parseIntOrUndefined,
     omitUndefinedValues,
 } from "@ourworldindata/utils"
-import { v7 as uuidv7 } from "uuid"
 import {
     References,
     StaticVizReference,
@@ -51,12 +49,13 @@ import { UpdatedChartInheritanceRecord } from "../../db/model/Variable.js"
 import { enqueueExplorerRefreshJobsForDependencies } from "../../db/model/Explorer.js"
 import { expectInt } from "../../serverUtils/serverUtil.js"
 import {
+    deleteChartConfigPairFromDbAndR2,
+    insertChartConfigPair,
     retrieveChartConfigFromDbAndSaveToR2,
-    updateChartConfigInDbAndR2,
+    updateChartConfigPairInDbAndR2,
 } from "../chartConfigHelpers.js"
 import {
     deleteGrapherConfigFromR2,
-    deleteGrapherConfigFromR2ByUUID,
     saveGrapherConfigToR2ByUUID,
 } from "../../serverUtils/r2/chartConfigR2Helpers.js"
 import { triggerStaticBuild } from "../../baker/GrapherBakingUtils.js"
@@ -94,7 +93,7 @@ export const getReferencesByChartId = async (
     const narrativeChartsPromise = db.knexRaw<NarrativeChartMinimalInformation>(
         knex,
         `-- sql
-        SELECT nc.id, nc.name, cc.full ->> "$.title" AS title
+        SELECT nc.id, nc.name, cc.config ->> "$.title" AS title
         FROM narrative_charts nc
         JOIN chart_configs cc ON cc.id = nc.chartConfigId
         WHERE nc.parentChartId = ?`,
@@ -248,7 +247,7 @@ const saveNewChart = async (
         shouldInherit?: boolean
     }
 ): Promise<{
-    chartConfigId: Base64String
+    chartConfigId: string
     patchConfig: GrapherInterface
     fullConfig: GrapherInterface
 }> => {
@@ -263,23 +262,12 @@ const saveNewChart = async (
 
     const now = new Date()
 
-    // insert patch & full configs into the chart_configs table
-    // We can't quite use `saveNewChartConfigInDbAndR2` here, because
-    // we need to update the chart id in the config after inserting it.
-    const chartConfigId = uuidv7() as Base64String
-    await db.knexRaw(
+    // Insert without publishing to R2 yet, because we need to update the chart
+    // id in the config after inserting it.
+    const { chartConfigId, patchConfigId } = await insertChartConfigPair(
         knex,
-        `-- sql
-            INSERT INTO chart_configs (id, patch, full, createdAt, updatedAt)
-            VALUES (?, ?, ?, ?, ?)
-        `,
-        [
-            chartConfigId,
-            serializeChartConfig(patchConfig),
-            serializeChartConfig(fullConfig),
-            now,
-            now,
-        ]
+        { config: fullConfig, patchConfig },
+        now
     )
 
     // add a new chart to the charts table
@@ -288,6 +276,7 @@ const saveNewChart = async (
         `-- sql
             INSERT INTO charts (
                 configId,
+                patchConfigId,
                 isInheritanceEnabled,
                 forceDatapage,
                 createdAt,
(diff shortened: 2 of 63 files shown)


sophiamersmann:
@codex review

db/migration/1786454321825-ReplaceChartConfigPatchAndFullWithConfig.ts
```diff
@@ -0,0 +1,156 @@
...
+        `)
+
+        // Drop the patch column and all three generated columns that depend on `full`
+        // Dropping slug also drops idx_chart_configs_slug
+        await queryRunner.query(`-- sql
+            ALTER TABLE chart_configs
+            DROP COLUMN slug,
+            DROP COLUMN chartType,
+            DROP COLUMN fullMd5,
+            DROP COLUMN patch
```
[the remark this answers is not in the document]
sophiamersmann: yup, that's on my mind

sophiamersmann:
@codex review

db/migration/1786454321825-ReplaceChartConfigPatchAndFullWithConfig.ts
```diff
@@ -0,0 +1,156 @@
...
+                JOIN ${table} o ON o.${resolvedPointer} = resolved.id
+                JOIN chart_configs patch ON patch.id = o.patchConfigId
+                SET resolved.patch = patch.config
+            `)
+        }
+
+        // Everything else — indicator configs, explorer views, and the authored
+        // rows themselves — held patch identical to full.
+        await queryRunner.query(`-- sql
+            UPDATE chart_configs SET patch = config WHERE patch IS NULL
```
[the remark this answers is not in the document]
sophiamersmann: I don't think this is worth fixing since we have only have a few admin-authored indicator configs that we plan to remove anyway.

adminSiteClient/CreateNarrativeChartEditorPage.tsx
```diff
@@ -88,7 +88,7 @@ class CreateNarrativeChartEditorPageInternal
         const chartConfig = await this.context.admin.getJSON(
             `/api/chart-configs/${this.props.chartConfigId}.config.json`
         )
-        this.parentConfig = chartConfig.full
+        this.parentConfig = chartConfig
```
sophiamersmann, line 91: `/chart-configs/<uuid>.config.json` used to return the full row, not it only returns the config (which I think is a bit more natural for an endpoint ending in `config.json`)

adminSiteServer/apiRoutes/charts.ts
```diff
@@ -569,23 +563,6 @@ export const saveGrapher = async (
     }
 }
 
-export async function updateGrapherConfigsInR2(
```
sophiamersmann, line 572: Moved to the file where it's used

adminSiteServer/apiRoutes/variables.ts
```diff
@@ -296,11 +301,12 @@ export async function getVariablesVariableIdJson(
...
         variableId
     )
-    const grapherConfigETL = variableWithConfigs?.etl?.patchConfig
-    const grapherConfigAdmin = variableWithConfigs?.admin?.patchConfig
-    const mergedGrapherConfig =
-        variableWithConfigs?.admin?.fullConfig ??
-        variableWithConfigs?.etl?.fullConfig
+    const grapherConfigETL = variableWithConfigs?.etl?.config
+    const grapherConfigAdmin = variableWithConfigs?.admin?.config
+    const mergedGrapherConfig = mergeVariableChartConfigs({
```
sophiamersmann, line 306: The full config is derived in code now, but this'll change in the next PR when we remove the admin layer

adminSiteServer/multiDim.ts
```diff
@@ -164,13 +165,21 @@ async function getViewIdToChartConfigIdMap(
...
         `-- sql
-        SELECT viewId, chartConfigId
+        SELECT viewId, chartConfigId, patchConfigId
         FROM multi_dim_x_chart_configs mdxcc
         JOIN multi_dim_data_pages mddp ON mddp.id = mdxcc.multiDimId
         WHERE mddp.catalogPath = ?`,
         [catalogPath]
     )
-    return new Map(rows.map((row) => [row.viewId, row.chartConfigId]))
+    return new Map(
```
sophiamersmann, line 174: Will be simplified again when we drop the patch later (this also applies to other code changes in this file)

baker/siteRenderers.tsx
```diff
@@ -634,19 +634,21 @@ export const renderExplorerPage = async (
...
-                  })
-                : {}
-            const etlConfig = row.grapherConfigETL
-                ? parseGrapherConfigFromRow({
-                      id: row.id,
-                      config: row.grapherConfigETL,
-                  })
-                : {}
-            const mergedConfig = mergeGrapherConfigs(etlConfig, adminConfig)
+            const mergedConfig =
```
sophiamersmann, line 637: Will be much simpler after dropping the admin layer in the next PR

db/model/Variable.ts
```diff
@@ -241,18 +232,19 @@ async function findAllChartsThatInheritFromIndicator(
...
     }>(
         trx,
         `-- sql
             SELECT
                 c.id as chartId,
                 cc.id as chartConfigId,
-                cc.patch as patchConfig,
-                cc.full ->> "$.isPublished" as isPublished
+                cc_patch.config as patchConfig,
+                cc.config ->> "$.isPublished" = "true" as isPublished
```
sophiamersmann, line 244: The `= "true"` was missing here, which was probably an oversight

devTools/syncGraphersToR2/syncGraphersToR2.ts
```diff
@@ -277,16 +277,29 @@ async function main(parsedArgs: parseArgs.ParsedArgs, dryRun: boolean) {
...
+        >(
+            trx,
+            `-- sql
+             select cc.configMd5, cc.id
+             from chart_configs cc
+             where cc.id in (
+                 select configId from charts
+                 union select chartConfigId from multi_dim_x_chart_configs
+                 union select chartConfigId from narrative_charts
+                 union select chartConfigId from explorer_views
```
sophiamersmann, line 290: We don't need to upload the explorer views and ideally I'd skip it, but the `refreshExplorerViews` script currently also uploads views to R2, so for consistency they're included here too (otherwise the two scripts would fight each other). I figured it doesn't matter much since we'll soon get rid of explorers.

devTools/syncGraphersToR2/syncGraphersToR2.ts
```diff
@@ -277,16 +277,29 @@ async function main(parsedArgs: parseArgs.ParsedArgs, dryRun: boolean) {
...
-            Pick<DbRawChartConfig, "fullMd5" | "id">
-        >(trx, `select fullMd5, id from chart_configs`)
+            Pick<DbRawChartConfig, "configMd5" | "id">
+        >(
+            trx,
+            `-- sql
+             select cc.configMd5, cc.id
+             from chart_configs cc
+             where cc.id in (
+                 select configId from charts
```
sophiamersmann, line 287: We now only upload the final chart configs (published or not)

db/migration/1786454321825-ReplaceChartConfigPatchAndFullWithConfig.ts
```diff
@@ -0,0 +1,142 @@
...
+                AFTER config,
+            ADD COLUMN chartType varchar(255)
+                GENERATED ALWAYS AS ((case
+                    when (json_unquote(json_extract(config, '$.chartTypes')) is null)
+                    then 'LineChart'
+                    else json_unquote(json_extract(config, '$.chartTypes[0]'))
+                end)) STORED
+                AFTER slug,
+            ADD COLUMN configMd5 char(24)
+                GENERATED ALWAYS AS (to_base64(unhex(md5(config)))) STORED NOT NULL,
```
rakyi, line 83: Is the md5 the same as before the migration? If not, it might trigger a lot of archiving, which is not necessarily too bad, but it would be nice if we didn't have to do it.

db/model/ChartConfigs.ts
```diff
@@ -1,121 +1,72 @@
...
-    parseChartConfigsRow,
     serializeChartConfig,
 } from "@ourworldindata/types"
 
+import { v7 as uuidv7 } from "uuid"
+
 import * as db from "../db.js"
 
-export async function getChartConfigById(
+export async function getChartConfigByUUID(
```
rakyi, line 14: Nit: I know we don't have established guidelines for this, but I think it's slightly better to camel-case acronyms than not, to avoid potential pitfalls like [`XMLHttpRequest`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest).

packages/@ourworldindata/types/src/dbTypes/ChartConfigs.ts
```diff
@@ -7,19 +7,17 @@ import {
 export const ChartConfigsTableName = "chart_configs"
 export interface DbInsertChartConfig {
     id: string
-    patch: JsonString
-    full: JsonString
-    fullMd5?: Base64String
+    config: JsonString
+    configMd5?: Base64String
     slug?: string | null
     chartType?: GrapherChartType | null
```
rakyi: Since these three are generated by the DB, we shouldn't allow inserting them directly and therefore remove them from this type, right? And we can add them to the `DbRawChartConfig`.

db/model/Variable.ts
```diff
@@ -151,41 +157,26 @@ export async function getMergedGrapherConfigsForVariables(
 }
 
 export async function insertNewGrapherConfigForVariable(
-    knex: db.KnexReadonlyTransaction,
```
rakyi, line 154: I'm confused. Shouldn't this raise a type/runtime error when we write in a read-only transaction? Otherwise why have different types of transactions?

db/model/Variable.ts
rakyi: Hmm, shouldn't we update all charts that inherit from an indicator, not just the published ones?

rakyi: I suppose we could make it a generated column as well, which would convert it to 0/1 for us, instead of having to do it in each query, right? Not necesarilly in this PR though.

rakyi (approved):
Nice and clean!

As I wrote in one of the comments, it would be good to avoid triggering an archive run if we don't need to. So, e.g., if the archive is up-to-date and running the archive script before your changes doesn't produce any new page to archive, it should be the same after your changes. Similarly (using a dry run), if it were to detect something to archive, it should be the same after your changes.

Could you please test it if you haven't already?

db/migration/1786454321825-ReplaceChartConfigPatchAndFullWithConfig.ts
sophiamersmann: It wasn't on my mind, but I just checked and the md5s don't change and no archiving should be triggered

db/model/ChartConfigs.ts
sophiamersmann: I'll do a sweep in a separate PR

packages/@ourworldindata/types/src/dbTypes/ChartConfigs.ts
sophiamersmann: Good point! Changed it

db/model/Variable.ts
sophiamersmann: I didn't notice this, but I just looked into it. This went unnoticed because each caller passes in a read-write transaction. In other words, if you'd actually pass in a read-only transaction, then the insert statement wouldn't go through.

db/model/Variable.ts
sophiamersmann: Yes, that would be sensible. I think the next refactors I'll do will be to move `slug` and `isPublished` out of the chart config. Making `isPublished` into a generated column is one sensible step along the way, but that's for another day.

db/model/Variable.ts
rakyi: Good to know, but also it's less useful than I thought.

db/model/Variable.ts
rakyi: Sounds good. Maybe it shouldn't be in the config either? But we don't have to solve that now.

sophiamersmann: That‘s exactly what I‘m talking about

PR #6980, opened 2026-08-12
🔨🤖 (chart-configs) drop the admin-authored indicator chart config layer

sophiamersmann:

Indicator-level chart config becomes ETL-only, the admin layer is dropped.

Two configs are deleted with the column. Neither was reachable from any rendered chart.

These admin features are deleted as well

| surface | what it did |
| --- | --- |
| `/admin/variables/:id/config` | authored the admin layer in the full chart editor UI |
| `/admin/variable-annotations` | bulk-edited the same layer across many indicators |
| the admin-config panel on `/admin/variables/:id` | showed the layer as YAML, linked to the editor, deleted it |

**Other behaviour changes**

- `GET /api/variables/mergedGrapherConfig/:id.json` and `GET /api/variables/grapherConfigETL/:id.patchConfig.json` collapse into `GET /api/variables/:id.config.json`
- `GET /api/variables/:id.json` no longer returns `grapherConfig`. It duplicated `grapherConfigETL` with the indicator's display metadata injected into its own dimension, which isn’t necessary


sophiamersmann:
@codex review

adminSiteClient/BulkGrapherConfigEditor.tsx
```diff
@@ -117,7 +116,6 @@ const bulkChartEditorColumnSets: ColumnSet[] = [
 ]
 
 const config: GrapherConfigGridEditorConfig = {
-    source: GrapherConfigGridEditorSource.SourceVariableAnnotation,
```
sophiamersmann, line 120: Unused

adminSiteClient/GrapherConfigGridEditor.tsx
```diff
@@ -354,19 +352,7 @@ export class GrapherConfigGridEditor extends React.Component<GrapherConfigGridEd
...
 
-        // Get the grapherConfig of the currently selected row and then
-        // merge it with the necessary partial information (e.g. variableId field)
-        // to get a config that actually works in all cases
-        const grapherConfig = selectedRowContent.config
-        const finalConfigLayer = this.config.finalVariableLayerModificationFn(
-            selectedRowContent.id
-        )
-
-        const mergedConfig = mergeGrapherConfigs(
```
sophiamersmann, line 382: This is to make a variable config renderable, so we don't need it anymore

adminSiteClient/VariableEditPage.tsx
```diff
@@ -461,27 +405,24 @@ class VariableEditor extends Component<{
 
     @computed private get grapherConfig(): GrapherInterface {
         const { variable } = this.props
-        const grapherConfig = variable.grapherConfig
-        if (grapherConfig)
-            return {
-                ...grapherConfig,
-                hasMapTab: true,
```
sophiamersmann, line 468: I don't think it makes sense to add a map tab by default it we have a valid config

adminSiteClient/VariableEditPage.tsx
```diff
@@ -461,27 +405,24 @@ class VariableEditor extends Component<{
...
-            return {
-                yAxis: { min: 0 },
-                map: { columnSlug: this.props.variable.id.toString() },
-                tab: GRAPHER_TAB_CONFIG_OPTIONS.map,
-                hasMapTab: true,
-                dimensions: [
-                    {
-                        property: DimensionProperty.y,
-                        variableId: this.props.variable.id,
-                        display: _.clone(variable.display),
```
sophiamersmann, line 481: Not sure why this was added here because Grapher grabs this from the metadata typically (this is probably from a time where the metadata was editable on this page)

adminSiteServer/apiRoutes/variables.ts
```diff
@@ -297,49 +268,21 @@ export async function getVariablesVariableIdJson(
...
-    })
-
-    // add the variable's display field to the merged grapher config
-    if (mergedGrapherConfig) {
-        const [varDims, otherDims] = _.partition(
-            mergedGrapherConfig.dimensions ?? [],
-            (dim) => dim.variableId === variableId
-        )
-        const varDimsWithDisplay = varDims.map((dim) => ({
-            display: variable.display,
```
sophiamersmann, line 318: I don't think we need to add in the display here - Grapher grabs this from the metadata when rendering the preview

baker/GrapherBaker.tsx
rakyi: We should be able to drop `display` for the same reason we dropped it [here](https://github.com/owid/owid-grapher/pull/6980#discussion_r3811767669).

db/model/Variable.ts
```diff
@@ -41,130 +40,72 @@ import {
...
     knex: db.KnexReadonlyTransaction,
     variableId: number
 ): Promise<GrapherInterface | undefined> {
-    const variable = await getGrapherConfigsForVariable(knex, variableId)
-    if (!variable) return undefined
-    return mergeVariableChartConfigs({
-        etl: variable.etl?.config,
-        admin: variable.admin?.config,
-    })
+    const configs = await getIndicatorChartConfigs(knex, [variableId])
```
rakyi: Nit: Probably doesn't matter much, but we call this function a lot during a full bake, so writing a separate query for the single config case would be a little faster and nicer and is easy to do.

rakyi (approved):
Nice!

PR #6718, opened 2026-07-08
WIP: Email notifications

rakyi:

PR #6982, opened 2026-08-13
Sync email subscription state with Postmark

rakyi:
Add authenticated, retry-safe Postmark webhooks for delivery and subscription changes, mirror Broadcast suppressions in D1, and reactivate eligible recipients after explicit resubscription.

Separate subscription intent from deliverability, strengthen sender recovery and configuration validation, and expand local tooling, documentation, and end-to-end coverage.


rakyi:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6982?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6982** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6982?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6718** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6718?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6797](https://github.com/owid/owid-grapher/pull/6797) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/6797?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

d1/email-notifications/migrations/0001_create_email_notifications_tables.sql
```diff
@@ -51,22 +40,47 @@ CREATE TABLE notification_preferences (
...
+    -- particular sent message.
+    messageId TEXT,
+    processedAt TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
+);
+
+-- Mirror of Postmark's broadcast-stream suppression list. Subscription-change
+-- webhooks add normalized addresses when Postmark suppresses them and remove
+-- them on reactivation.
+CREATE TABLE suppressed_addresses (
+    email TEXT PRIMARY KEY,
```
rakyi: Can we use a FK for the users table?

baker/emailNotifications/sendEmailNotifications.ts
```diff
@@ -200,108 +205,65 @@ async function buildNotificationItems(
...
+    const response = await client.sendEmail({
+        From: EMAIL_NOTIFICATIONS_FROM_ADDRESS,
+        To: email.to,
+        Subject: email.subject,
+        HtmlBody: email.htmlBody,
+        // Bulk emails must go through a Postmark broadcast stream (not the
+        // transactional "outbound" stream). Postmark adds the unsubscribe
+        // headers automatically.
+        MessageStream: "broadcast",
+        Tag: "email-notifications",
```
ikesau: Maybe worth a comment explaining that this is optional - just for analysis if we choose to add further email types in the future

baker/emailNotifications/sendEmailNotifications.ts
```diff
@@ -200,108 +205,65 @@ async function buildNotificationItems(
...
-    postmarkMessageId: string | null,
+    postmarkMessageId: string,
     sentAt: Date
 ): Promise<void> {
     const sentAtIso = sentAt.toISOString()
     await d1.query(
         `UPDATE notification_preferences
-         SET last_sent_at = ?1, updated_at = ?1
-         WHERE user_id = ?2`,
+         SET lastSentAt = ?1, updatedAt = ?1
```
ikesau, line 267: One other option instead of having to always remember to update `updatedAt` columns in D1 is [adding a trigger](https://www.sqlite.org/lang_createtrigger.html).

```sql
CREATE TRIGGER notification_preferences_set_updatedAt
AFTER UPDATE ON notification_preferences
FOR EACH ROW
WHEN NEW.updatedAt IS OLD.updatedAt
BEGIN
    UPDATE notification_preferences
    SET updatedAt = strftime('%Y-%m-%dT%H:%M:%fZ', 'now')
    WHERE id = NEW.id;
END;
```

We'd have to remember to write this any time we create a new table/column though, so maybe it's about the same. Just putting it here as food for thought.

baker/emailNotifications/sendEmailNotifications.ts
```diff
@@ -346,118 +314,102 @@ async function sendEmailNotifications(options: {
...
+
+            // We only send to subscribers absent from suppressed_addresses, so
+            // Postmark refusing one means a subscription-change webhook was
+            // missed. Leave it visible for manual reconciliation rather than
+            // guessing why it was suppressed.
+            console.error(
+                `Email send refused userId=${subscriber.userId} reason=inactiveRecipient`
+            )
+            Sentry.captureException(error)
+            failures.push({ userId: subscriber.userId, error })
```
ikesau, line 392: So here, we'll see the console.error in buildkite and the error itself in Sentry, right? That seems good

devTools/postmarkCatcher/postmarkCatcher.ts
```diff
@@ -75,13 +70,15 @@ function acceptEmail(payload: Record<string, unknown>) {
 function renderIndexPage(): string {
     const rows = emails
         .map((email, index) => {
+            // Coerced with String(): the payload is typed for convenience but
+            // comes off the wire, so a sender bug can put anything in it.
```
ikesau: Not sure this comment is adding much value

functions/api/email-notifications/subscribe.ts
```diff
@@ -191,7 +204,7 @@ async function createSubscribedUser(
     await db
         .prepare(
             `INSERT INTO notification_preferences
-                 (user_id, topic_tags, content_types, frequency)
+                 (userId, topicTags, contentTypes, frequency)
```
ikesau: Should we also clear the suppressions table here? Perhaps it's not possible, but in my testing (where I was manually editing my local DB) I manually added a suppression and then tried to subscribe the address. 

I think the scenarios where this could happen on prod are pretty contrived, but it's also only one line to fix here.

d1/email-notifications/migrations/0001_create_email_notifications_tables.sql
rakyi: We can, but it wouldn't be very useful. It's better to record a suppression even if we don't have a user with that email than to reject the event or create dummy user rows.

rakyi:
@codex review

baker/emailNotifications/sendEmailNotifications.ts
ikesau: In practice, putting a line of documentation somewhere for LLMs to always remember to check that `updatedAt` columns always get updated would also be sufficient.

functions/api/email-notifications/preferences.ts
```diff
@@ -117,38 +126,65 @@ export const onRequestPost: PagesFunction<Env> = async ({ request, env }) => {
...
                      SET status = 'unsubscribed',
-                         updated_at = strftime('%Y-%m-%dT%H:%M:%fZ', 'now')
+                         updatedAt = strftime('%Y-%m-%dT%H:%M:%fZ', 'now')
                      WHERE id = ?1`
                 )
                 .bind(userId)
                 .run()
         } else if (data.preferences) {
+            if (user.activeSuppressionChangedAt) {
+                await reactivatePostmarkRecipient(env, user.email)
```
ikesau, line 165: Can the transactional stream get suppressed? e.g. with a `HardBounce` or `SpamComplaint` - and if so, should we also check for / delete the suppression here also?

If a user flags a transactional email as spam, they probably don't want to receive any emails from us at all, but if they got a `HardBounce` (e.g. from a full inbox) then maybe they do.

package.json
```diff
@@ -47,7 +47,7 @@
         "testFormatAll": "yarn oxfmt --check",
         "testFormatChanged": "bash -c 'files=$((git diff --name-only --diff-filter=ACMRTUXB && git diff --name-only --cached --diff-filter=ACMRTUXB) | sort -u | grep -E \"\\.(tsx?|jsx?|ts|js|json|md|html|css|scss|yml)$\" | tr \"\\n\" \" \"); if [ -n \"$files\" ]; then yarn oxfmt --check $files; else echo \"No uncommitted files to check formatting\"; fi'",
         "test": "vitest",
-        "test:functions:e2e-r2": "vitest functions/test/grapher-config-r2.e2e.test.ts",
+        "test:functions:e2e": "vitest functions/test/*.e2e.test.ts",
```
ikesau, line 54: Is this change meant to be here?

functions/_common/postmarkWebhook.ts
```diff
@@ -42,53 +40,228 @@ export function checkPostmarkWebhookAuthorization(
...
+    if (
+        event.RecordType === "Delivery" &&
+        event.MessageStream === POSTMARK_BROADCAST_MESSAGE_STREAM
+    ) {
+        // Sandbox delivery can beat the send API response, which means the
+        // sender has not inserted its messages row yet. Do not acknowledge or
+        // record the webhook in that case: Postmark will retry after the send
+        // has been recorded.
+        const message = await db
+            .prepare("SELECT 1 FROM messages WHERE messageId = ?1")
```
ikesau: Wrong column name here

```suggestion
            .prepare("SELECT 1 FROM messages WHERE id = ?1")
```

functions/api/email-notifications/postmark-webhook.ts
```diff
@@ -1,79 +1,49 @@
...
-            Sentry.captureMessage(
-                `Unexpected Postmark webhook payload: ${z.prettifyError(error)}`
-            )
-            return okResponse()
-        }
+    let event: PostmarkWebhookEvent
+    try {
+        event = await request.json<PostmarkWebhookEvent>()
+    } catch (error) {
+        Sentry.captureException(error)
```
ikesau: Could we wrap these in a handler that also logs? Would make local debugging easier

ikesau (approved):
Whoops, that was meant to be an approval.

Looks good! I spent a lot of time simulating queries with curl and realizing I was slightly misunderstanding how things worked, so that was time well spent.

I didn't end up going as far as setting up ngrok, but hopefully between curl and the catcher, it's sufficient.

functions/api/email-notifications/preferences.ts
rakyi: Yes, the transactional stream has its own suppressions.

- A full inbox is a soft bounce, not a hard bounce, but this flow requires clicking on a recent magic link, which means the inbox should not be full (and didn't hard bounce)
- Spam complaints can't be deleted via the API regardless of the stream

How Postmark models it is that streams are separate logically, and one reason is that they want to prevent the deliverability of one from affecting the other. That's also why we should use different subdomains for them. So to me it makes sense to treat them separately and not try to be clever.

If you remember, we talked about this during the sync review: we said that we won't record suppressions of the transactional stream for now and will monitor the dashboard and add them later if handling them automatically seems necessary (i.e., manually in Postmark wouldn't be enough).

package.json
rakyi: Yes, we added a new test file and renamed the old one.

baker/emailNotifications/sendEmailNotifications.ts
rakyi: Added a comment.

baker/emailNotifications/sendEmailNotifications.ts
rakyi: I went with docs instead. The trigger would technically work, but it counts as an additional query, and in this case the `lastSentAt` and `updatedAt` would not be the same time like now, where we supply the same value for each.

baker/emailNotifications/sendEmailNotifications.ts
rakyi: Yes!

functions/api/email-notifications/postmark-webhook.ts
rakyi: Done.

functions/_common/mailchimp.ts
```diff
@@ -45,35 +45,47 @@ function makeAuthHeader(env: Env): string {
...
-        interests: { [env.MAILCHIMP_OWID_BRIEF_INTEREST_ID]: enabled },
-    }
+    const member = shouldSubscribe
+        ? {
+              email_address: email,
+              status_if_new: "subscribed",
+              // status_if_new only applies when creating a member. An
+              // explicit Brief opt-in also re-subscribes an existing globally
+              // unsubscribed member.
+              status: "subscribed",
```
ikesau, line 68: Have you confirmed that this works? Claude thinks there's a chance Mailchimp will reject resubscribes if not done via a form it hosts.

ikesau (approved):

functions/_common/mailchimp.ts
rakyi: Good catch! We have to change how this works. I'll do it in the next PR to keep this one from becoming even bigger.

rakyi:
Folded into `email-notifications`.

PR #6983, opened 2026-08-13
Gdoc migration

ikesau:
This PR adds a gdoc migration system for structural migrations over archie content in actual google documents and the parsed JSON in `posts_gdocs.content`.

It allows us to do things like changing the name of an attribute in an archie component without having to manually edit and republish every single gdoc.

It does this in 2 stages:

1. During grapher deploy, typeorm migration files call `applyGdocMigrationToDb` which applies the migration's `dbTransform` to `posts_gdocs.content`. This ensures that the DB is only ever updated in lockstep with new code that accepts the new shape
2. Afterwards, via CLI, a developer calls `yarn gdocMigration plan|apply|verify|status` to update the gdocs themselves.

## The CLI

The intended sequence on the prod server is `plan` → review the report → `apply` → `verify`, with `status` available any time to see where a run stands.

<details><summary>Commands</summary>
<ol>
<li><code>plan</code> — the dry run. Resolves the candidate doc list, fetches each doc from Google, computes the edits, and prints the grouped report: docs grouped by several heuristics to bucket simple and edge cases, followed by flagged and failed docs. It writes the journal but never writes to Google or the database.</li>
<li><code>apply</code> — the real thing, gdoc-side. For each doc: fetch fresh, re-plan against that fresh fetch, snapshot the pre-edit doc JSON to disk, then send the <code>batchUpdate</code>. Can be re-run if interrupted and already-processed gdocs will be skipped.</li>
<li><code>verify</code> — a standalone re-check, run later. Re-fetches docs and asserts the migration is now a no-op everywhere — no edits, no flags. Useful as the final confirmation pass after apply, or after manually fixing flagged docs.</li>
<li><code>status</code> — offline; just reads the journal and prints counts by state (planned / verified / flagged / failed / no-match), plus the details of every doc that still needs attention.</li>
</ol>
</details> 

<details><summary>Flags</summary>
<ol>
<li><code>--migration <name></code>(required) — which migration to run. </li>
<li><code>--id <docId></code>— target specific Google Doc id(s) instead of running the migration's SQL discover query; useful for testing a migration against a personal gdoc before the real run</li>
<li><code>--published-only</code></li>
<li><code>--concurrency <n></code>(default 4) </li>
<li><code>--journal-dir <path></code>— where the per-migration journal JSON and the pre-edit doc snapshots are written.</li>
<li><code>--force</code>— only meaningful for <code>apply</code>. Normally <code>apply</code> is resumable: any doc whose journal status is already verified is skipped, so a crashed or interrupted run can just be re-run. <code>--force</code> re-processes those docs anyway.</li>
</ol>
</details> 

## The migration engine

Migrations come in two modes: 
1. component (a `transform` for the raw gdoc block + a separate `dbTransform` for the enriched stored block)
2. frontmatter (`renameKey` and `mapValue`).

### How it works

Per gdoc:
1. Fetch the live doc, record its `revisionId`.
2. Parse with `gdocToSourceMappedLines` to record which character range each archie line came from
7. Scan the archie and find the keys/blocks we want to migrate
8. Parse them to JSON, run the migration's transform
9. Calculate the diff with `planDocumentPatch`
10. Turn each diff into the smallest possible set of `batchUpdate` requests that will preserve surrounding formatting, chips, refs, etc
11. Order requests by position in the gdoc so that the rewrites to the bottom of the page are done first
12. Send the requests with `writeControl.requiredRevisionId` (so a concurrent author edit fails the write (retried once with a re-plan, then flagged)).

To verify a migration has worked, the script refetches and checks each line in each gdoc, comparing each of its lines to the lines it's expected to have, and that re-planning the migrated doc is a no-op.

## An example migration

`2026-07-prominent-link-gdoc-urls.ts` rewrites prominent-link URLs from `https://ourworldindata.org/<slug>` to the gdoc's docs.google.com URL so the content graph resolves metadata automatically, dropping the now-redundant title/thumbnail/filename properties

It would run on ~322 blocks across 123 docs.

## A new skill
I wrote a new agent skill to help someone create and test a migration. It's written to be a walkthrough, giving you the commands to run, to make sure you understand what's happening and have multiple chances to notice if assumptions you've made are incorrect.

## Reverting
Reverting isn't simple, unfortunately, as there's no API to set a gdoc's current revisionID (it's a UI-only feature.)

Here's a plan Claude came up with to allow us to do reverts, but it would be another large batch of complexity, so I'm tempted to leave it out for now.

<details><summary>Can we just do reverts with revision IDs?</summary>
<p>Not by revision id, but the same outcome is buildable from what apply already saves.</p><p>Why revision ids can't do it</p><ul><li>The Docs API has no restore endpoint. revisionId is read-only and exists only for writeControl, which lets a write say "fail unless the doc is still at revision X". There is no "set the doc back to revision X".</li><li>The Drive API can list a Google Doc's revisions and export one as docx or text, but not restore it. Re-importing an export replaces the whole body, which loses comments, pending suggestions, images, and anything else the engine deliberately avoids touching. That is the whole-doc rewrite the design rejects.</li><li>Version-history restore exists only in the Docs UI, per doc, by hand.</li></ul><p>What would work: replay the snapshots in reverse</p><p>apply already writes the exact pre-edit doc JSON per doc and journals the before and after revision ids. A revert command could work like this, per doc:</p><ol><li>Load the snapshot and the edits that were applied. The journal today stores only human-readable summaries, so apply would start recording the structured edits too.</li><li>Invert each edit. Every edit kind has a clean inverse: a rename becomes the opposite rename, a set-value restores the old value, a delete-property re-inserts the old line, an insert-property deletes it. The snapshot lines carry the original span markup, so re-inserted values get their bold and links back.</li><li>Fetch the live doc and check the migrated lines are still exactly what the migration produced, using the same expected-lines comparison apply uses for verification. If an author has since edited inside a migrated block, flag the doc rather than touching it. Edits elsewhere in the doc don't matter.</li><li>Send the inverse batchUpdate guarded by the current revision id, re-fetch, and confirm the migrated lines now match the snapshot. Journal the doc as reverted.</li></ol><p>Usage would be exactly what you sketched:</p><p>yarn gdocMigration revert --migration <name> [--id <docid>]</docid></name></p><p>Same resumability, concurrency, and flags as apply, and it never needs a reverse migration file, so it covers lossy migrations too. The prominent-link one, which deletes title and thumbnail, would get those lines back verbatim.</p><p>Limits to be aware of</p><ul><li>It depends on the snapshots and journal from the machine that ran apply. If prod runs happen on the server, runs/ needs to be kept or copied off afterwards.</li><li>Deleted blocks are the one edit kind that is harder to reverse in place, since the inverse must re-insert several lines at a known anchor. Doable, but it needs its own test coverage.</li><li>It reverts the gdoc side only. The DB side still goes through the down migration and a code revert, in that order, and docs republished in between will have been parsed through the alias, which the down migration handles.</li></ul><p>I'd estimate this at roughly the size of the db-plan plus create-test-doc work: journal changes, an invertEdits function, a runRevert in the runner, and tests built on the existing simulateRequests fixtures. Want me to build it on gdoc-migration?</p>
</details>


pabloarosado:
We noticed that migrating explorers to mdims has a pain point, which is to relink gdocs. So this PR could be very useful to save us time on that task. Related: https://github.com/owid/etl/issues/6514

PR #6985, opened 2026-08-14
Replace LatestDataInsights carousel with CSS scroll-snap

marcelgerber:
This replaces Embla Carousel with [native CSS scroll snap](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Scroll_snap).

CSS scroll snap is supported in all our supported browsers by now.

Behavioral changes compared to Embla:
- The carousal (almost) fully works even without JS.
  - The only things that don't work are the left/right buttons and the dots at the bottom - they are hidden in this case.
  - Instead, we show a thin scrollbar in this case.
- On desktop, scrolling with a mouse drag doesn't work any more (including the overscroll behavior on the left).
- On desktop, you can also scroll with a horizontal scroll wheel, and using the arrow keys while the scroll container is keyboard-focused.

Other than that, both work very similarly (on desktop and mobile), and there are no visual differences at all.
Also, it uses pretty much the same amount of code as before, just with two fewer dependencies.

I've tested this a good bunch on both desktop and mobile, and also in older Safari versions (macOS & iOS) on Browserstack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,8 +109,6 @@
         "d3": "^7.9.0",
         "dayjs": "^1.11.21",
         "dotenv": "^17.4.2",
-        "embla-carousel-class-names": "^8.6.0",
-        "embla-carousel-react": "^8.6.0",
         "entities": "^7.0.1",
         "express": "^5.2.1",
         "figma-api": "^2.2.0-beta",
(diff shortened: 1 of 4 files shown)


rakyi (approved):
Nice!

PR #6988, opened 2026-08-14
🔨🤖 (chart-configs) derive the mdim view config layer instead of storing it

sophiamersmann:

Drops `multi_dim_x_chart_configs.patchConfigId` and the `chart_configs` row it pointed at. A view's authored layer is now rebuilt from `multi_dim_data_pages.config` by `buildMdimViewPatchConfig`, which is shared by `upsertMultiDim` and the indicator-propagation path.

**Worth a reviewer's attention**

- Updating an indicator config, plus all the mdim views that inherit from it, is now slower
- View configs with no schema version are now reported to Sentry, once per upsert (right now about half of the mdim view configs declare no `$schema`). We could also turn this into a hard error if we'd prefer

### DIFF
--- a/adminSiteServer/multiDim.ts
+++ b/adminSiteServer/multiDim.ts
@@ -1,10 +1,6 @@
 import * as _ from "lodash-es"
 
 import {
-    defaultGrapherConfig,
-    migrateGrapherConfigToLatestVersionAndFailOnError,
-} from "@ourworldindata/grapher"
-import {
     ChartConfigsTableName,
     DbEnrichedMultiDimDataPage,
     DbPlainMultiDimDataPage,
@@ -25,13 +21,13 @@ import {
     R2GrapherConfigDirectory,
     View,
 } from "@ourworldindata/types"
-import {
-    mergeGrapherConfigs,
-    MultiDimDataPageConfig,
-    dimensionsToViewId,
-} from "@ourworldindata/utils"
+import { mergeGrapherConfigs, dimensionsToViewId } from "@ourworldindata/utils"
+import * as Sentry from "@sentry/node"
 import * as db from "../db/db.js"
-import { upsertMultiDimDataPage } from "../db/model/MultiDimDataPage.js"
+import {
+    buildMdimViewPatchConfig,
+    upsertMultiDimDataPage,
+} from "../db/model/MultiDimDataPage.js"
 import { upsertMultiDimXChartConfigs } from "../db/model/MultiDimXChartConfigs.js"
 import {
     getIndicatorChartConfigs,
@@ -43,9 +39,8 @@ import {
     saveMultiDimConfigToR2,
 } from "../serverUtils/r2/chartConfigR2Helpers.js"
 import {
-    saveNewChartConfigPairInDbAndR2,
+    saveNewChartConfigInDbAndR2,
     updateChartConfigInDbAndR2,
-    updateChartConfigPairInDbAndR2,
 } from "./chartConfigHelpers.js"
 
 function catalogPathFromIndicatorEntry(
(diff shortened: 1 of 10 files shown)


sophiamersmann:
#codex review

sophiamersmann:
@codex review

db/model/Variable.ts
```diff
@@ -250,28 +255,33 @@ async function findAllMultiDimViewsThatInheritFromIndicator(
...
-                mdxcc.chartConfigId as chartConfigId,
-                cc.config as patchConfig,
-                md.published as isPublished
-            FROM multi_dim_data_pages md
-                JOIN multi_dim_x_chart_configs mdxcc ON mdxcc.multiDimId = md.id
-                JOIN chart_configs cc ON cc.id = mdxcc.patchConfigId
-            WHERE mdxcc.variableId = ?
-        `,
-        [variableId]
+    const rows = await trx<DbPlainMultiDimXChartConfig>(
```
sophiamersmann, line 268: This is now slower than it used to be

adminSiteServer/multiDim.ts
```diff
@@ -272,147 +250,111 @@ export async function upsertMultiDim(
...
-            mdxcc.patchConfigId,
-            cc.config,
-            cc_patch.config AS patchConfig
-        FROM chart_configs cc
-        LEFT JOIN multi_dim_x_chart_configs mdxcc ON mdxcc.chartConfigId = cc.id
-        LEFT JOIN chart_configs cc_patch ON cc_patch.id = mdxcc.patchConfigId
-        WHERE cc.id IN (?)`,
-        [chartConfigIds]
+/** Warns if any view config does not declare a schema version */
+function reportViewConfigsWithoutSchema(
```
sophiamersmann: We could also turn this into a hard error to force mdims to declare config schemas the next time they update

sophiamersmann: I just chatted with Lucas, who did some work on the ETL, and every multi-dim now comes with a schema, so we could/should be stricter here now

adminSiteServer/multiDim.ts
```diff
@@ -272,154 +250,110 @@ export async function upsertMultiDim(
...
-                config: parseChartConfig(row.config),
-                patchConfig: row.patchConfig
-                    ? parseChartConfig(row.patchConfig)
-                    : undefined,
-            },
-        ])
+    // Static message, so that all affected pages group into one Sentry issue
+    Sentry.captureMessage(
+        "Multi-dim view configs without a schema version, assumed latest",
+        { level: "warning", extra: { catalogPath, viewIds } }
```
rakyi: JSYK we currently only alert on high priority issues in Slack, which warnings are not. But they are of course visible in Sentry directly.

db/model/Variable.ts
rakyi: Out of curiosity, do you know how much slower?

rakyi (approved):
LGTM!

adminSiteServer/multiDim.ts
sophiamersmann: I think now that all multi-dim views come with a schema version, we can just throw. I made it a warning back when I thought we'd still have thousands of configs without a version.

rakyi: Right, throwing seems fine then.

db/model/Variable.ts
sophiamersmann: This is what Claude measured on my machine:

| case                                                | before | after |
| --------------------------------------------------- | ------------------------------ | -------------------------------------------- |
| `causes_of_death`, 1.7 MB / 5034 views (2 inherit) | 0.4 ms                         | 13.9 ms                                      |
| max fan-out, 7 views across 5 mdims                 | 0.3 ms                         | 2.3 ms                                       |
| small pages, 855 B + 3.2 KB, 3 views                | 0.2 ms                         | 1.4 ms                                       |

rakyi: That's difficult to interpret without more context. 😄 But nevermind, it should be fine.

PR #6996, opened 2026-08-18
🎉🤖 Add ghost-variable cleanup endpoint for ETL

Marigold:

Adds `POST /variables/delete`, so ETL stops deleting variables with its own SQL. Grapher had no variable-delete endpoint at all before this — not bulk, not single — so this is the general primitive rather than one shaped around ETL. Paired with owid/etl#6694 — same branch name, so both share `staging-site-ghost-variables`.

This is the admin-API version of owid/etl#6672, which @sophiamersmann closed after talking to @rakyi so the fix would land here rather than in ETL. The orphaned-`chart_configs` fix from that PR is included, both legs of it.

## What it does

Given a list of variables to remove, it:

- reports any that a chart still uses as `blocked`, and never deletes them;
- deletes the others, their link rows in the five tables that carry a `variableId`, and the `chart_configs` rows they own — the variable's own ETL/admin configs, and the mdim view configs whose link rows go with them — from MySQL and R2.

Taking the variables to delete rather than a dataset plus everything to keep was Sophia's suggestion on review, and it inverts the failure direction: an incomplete caller-side view now leaves a ghost behind rather than destroying a variable the caller didn't know about.

It deliberately doesn't decide whether a blocked variable should fail the caller's run. ETL's answer depends on which environment it's in and on a chart-diff against production, so the endpoint hands back facts and lets it choose.

## Why the endpoint rather than leaving it in ETL

The list of tables that have to be cleared before a variable row can go is Grapher's schema knowledge, and ETL had no way to know when it changed. #6672 is what that costs: every foreign key into `chart_configs` points outward with `ON DELETE RESTRICT`, so ETL's `DELETE FROM variables` succeeded and stranded the config row — about 1,600 times in production. ETL also can't clean up R2, because it doesn't know the config layout.

## Still open

- **The ~1,600 existing orphans in production are not cleaned up by this.** Wants a one-off migration deleting `chart_configs` rows unreferenced by `charts`, `variables`, `narrative_charts`, `multi_dim_x_chart_configs`, `explorer_views` and `multi_dim_redirects` — carried over from #6672, still unwritten.
- **No load testing.** A full ETL deploy will hit this endpoint once per grapher step, concurrently. Nobody has checked how the admin server handles that.

### Shape

```
POST /variables/delete
  { variableIds: number[] }
→ { success: true,
    deleted: number[],
    blocked: [{ variableId, variableName, chartId, chartSlug }] }
```

Array in, partial results out — chosen with the bulk upserts discussed on the Aug 17 call in mind, so the same shape can carry over. `blocked` is one row per (variable, chart) pair and carries names and slugs so the caller can print something a person can act on.

### Delete order

Every FK into `chart_configs` is `ON DELETE RESTRICT`, so the referencing rows go first: link tables, then `variables`, then `chart_configs`, then R2 (outside the transaction, matching `deleteChartById`).

The mdim leg skips any view config referenced by `multi_dim_redirects` — that FK is `RESTRICT` too, and leaking one row is better than failing a grapher step over a rare case. Same guard #6672 used.

### Files

- `db/model/Variable.ts` — `deleteVariables`, plus the `VARIABLE_LINK_TABLES` list that is the reason this lives here.
- `adminSiteServer/apiRoutes/variables.ts` — route handler, input validation, R2 deletes.
- `adminSiteServer/apiRouter.ts` — registration.
- `adminSiteServer/tests/variable-delete.test.ts` — five DB tests: the plain delete, delete-everything, no-op, blocked-by-chart, the leftover indicator config, and the mdim view config.

### Checked

`yarn typecheck`, `yarn testLintChanged`, `yarn fixFormatChanged`, and `make dbtest` (92 passed) on this branch.

</details>


sophiamersmann:
@Marigold, do you mind if this waits until the chart configs refactor has merged? I'm happy to carry it through then.

Marigold:
Yeah, there's no rush.

sophiamersmann:
Closing in favour of https://github.com/owid/owid-grapher/pull/7030 which I stacked on top of the refactor

PR #6999, opened 2026-08-18
Exclude timeout errors from Sentry

rakyi:
This happens a lot and it's not really useful to log the errors.


sophiamersmann (approved):
Thanks!

PR #7004, opened 2026-08-19
🐛🤖 keep entity selection when saving a narrative chart

sophiamersmann:

Saving a chart as a narrative chart wiped its entity selection whenever the selection was inherited from the indicator config.

Narrative charts persist a few props explicitly (entity selection, time, focus) so a later parent edit can't move them. That snapshot was taken on the server, from the submitted patch merged with the grapher defaults — but the chart editor submits the chart's *patch*, so anything inherited isn't in it and got snapshotted as the default. The snapshot now happens in the editor, off the live config.

To check it: [chart 189](https://staging-site-narrative-chart-selection.owid.io/admin/charts/189/edit) has inheritance enabled and inherits its entity selection from the indicator config. Switch to the line tab and save it as a narrative chart — before, the narrative chart came out with the default selection; now it keeps all entities.

- New `adminSiteClient/narrativeChartConfig.ts`: `makeNarrativeChartPatchConfig(liveConfigWithDefaults, parentConfigWithDefaults?)` returns the diff to the parent plus an explicit `_.pick` of `NARRATIVE_CHART_PROPS_TO_PERSIST` off the live config.
- `ChartEditor.saveAsNarrativeChart` now goes through it — that path sent a bare diff before, which is where the bug was. `NarrativeChartEditor.patchConfig` already did the same pick inline and now delegates, so both create paths are identical.
- The fallback for "no parent layer applied" lives in the helper, so both editors diff against `defaultGrapherConfig`. `NarrativeChartEditor` previously diffed against `{}`, which inlines every default into the patch; only reachable in the window before the parent config loads.
- The server is unchanged: it still snapshots from `mergeGrapherConfigs(defaultGrapherConfig, config)`, which is a pass-through for props the editors now always send explicitly.

</details>

### DIFF
--- a/adminSiteClient/ChartEditor.ts
+++ b/adminSiteClient/ChartEditor.ts
@@ -13,7 +13,6 @@ import {
     GrapherInterface,
     getParentIndicatorIdFromChartConfig,
     mergeGrapherConfigs,
-    NARRATIVE_CHART_PROPS_TO_OMIT,
 } from "@ourworldindata/utils"
 import { DbChartTagJoin } from "@ourworldindata/types"
 import { action, computed, observable, runInAction, makeObservable } from "mobx"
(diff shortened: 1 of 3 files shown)


sophiamersmann:
@codex review

marcelgerber (approved):

PR #7005, opened 2026-08-19
🤖 Update regions [2026-08-19]

owidbot:

TODOs:
- [x] Check if there were any country renames, and if so:
    - [ ] Take extra care :)

---

No country renames in this update — the regions diff is insertions-only (no removed or changed entries), so no redirects are needed.

What's new is the **IHME Global Burden of Disease (GBD)** groupings as region data providers: `ihme_gbd_1` (7 super-regions) and `ihme_gbd_2` (the more detailed regions). This broke typecheck on the `Record`s keyed by `RegionDataProvider`, so a follow-up commit wires the new providers up, following the Ember/EI/ILO precedent.

- `RegionGroups.ts` / `EntityPresets.ts`: entity-selector group labels and admin preset labels for both levels; added an `ihmegbd` suffix handle to `ADDITIONAL_REGION_DATA_PROVIDERS` so entities suffixed "(IHME GBD)" are recognized by `parseLabel` (which lowercases and strips spaces from the suffix, so `ihme_gbd_1`/`ihme_gbd_2` would never match) — same pattern as `unm49`/`ilo`.
- `RegionTooltipData.ts`: tooltip descriptions and west-to-east display orders for both levels.
- `regionsUtils.ts`: back-filled "South Asia (IHME GBD)" and "North Africa and Middle East (IHME GBD)" into the `ihme_gbd_2` tier. In GBD's hierarchy these are both super-regions *and* regions, but each carries a single `definedBy: "ihme_gbd_1"` in the data — without the back-fill, the level-2 tier misses 26 countries (verified: exactly the South Asia + MENA members). With it, level 2 is a complete partition of all 21 official GBD regions.
</details>


marcelgerber:
@codex fix this PR.

PR #7007, opened 2026-08-19
Add post attribution only with human accounts

rakyi:
To avoid stuff like [this](https://github.com/owid/owid-grapher/pull/6982#discussion_r3777048116):

> Written by OpenAI GPT-5.6 Sol — @codex at the wheel.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,15 +68,21 @@ Use the `test-on-staging` skill when checking a change on `staging-site-<branch>
 
 ## Team
 
-Everything you post to GitHub or Slack goes out under a **human's identity**.
+When AI-authored text is posted through a **human's account**, it must be clearly attributed so readers do not mistake it for the human's own words.
 
-1. **Attribute the work.** Any prose you author for GitHub or Slack — PR bodies, issue and review comments (including replies to Codex/Copilot/reviewers), Slack messages or drafts — must start with this blockquote:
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -68,15 +68,21 @@ Use the `test-on-staging` skill when checking a change on `staging-site-<branch>
...
 ## Team
 
-Everything you post to GitHub or Slack goes out under a **human's identity**.
+When AI-authored text is posted through a **human's account**, it must be clearly attributed so readers do not mistake it for the human's own words.
 
-1. **Attribute the work.** Any prose you author for GitHub or Slack — PR bodies, issue and review comments (including replies to Codex/Copilot/reviewers), Slack messages or drafts — must start with this blockquote:
+1. **Attribute posts made under a human identity.** Any prose that will appear under a human user's GitHub or Slack account — PR bodies, issue and review comments (including replies to Codex/Copilot/reviewers), Slack messages, or drafts intended for a human to post — must start with this blockquote:
 
     ```
     > _Written by <model provider> <model name> — @<handle> at the wheel._
```
rakyi, line 76: BTW, why do we tag the user? Wouldn't it be enough to say that the message is written by an agent? The human account is visible in the normal UI, isn't it?

marcelgerber: usually it is, but there are also cases where the PR author is Claude or similar, e.g. when using Claude Tag.
In that case, it's still helpful to know who prompted Tag.

Marigold (approved):

CLAUDE.md
rakyi: OK, let's see if it works correctly after this change.

PR #7008, opened 2026-08-19
🐝🤖 include untracked files in changed-file checks

rakyi:
It could happen that an agent would create new files and change existing files, then report that all checks pass, but there would be e.g. lint errors in the new files, which previously were not checked by the scripts.


rakyi:
I decided it's not worth it.

PR #7020, opened 2026-08-20
🐛 track which graphers are loading, not how many events fired

Marigold:

`window._OWID_HAVE_ALL_GRAPHERS_LOADED` — the flag the site-screenshots tool waits on before taking a screenshot — is set immediately on nearly every page, before a single chart has drawn. `runAllGraphersLoadedListener` counts `[data-grapher-src]` embeds to know how many charts to wait for, and the baker no longer emits that attribute for charts in articles, data pages or topic pages; those are baked as `figure.chart` fallbacks and hydrated by React. Finding zero embeds, the listener declares everything loaded straight away. This replaces the count-the-embeds scheme with a live count of graphers that are loading.

*Deep on the semantics, skim on the diff — the flag's meaning changes, and its only consumer lives in another repo.*

## Why the count can't be taken up front

A chart is baked as a static fallback and only becomes a grapher when something hydrates it, and some fallbacks never do: on the homepage 8 chart figures produce 1 grapher, on `/life-expectancy` 5 produce 2, because a key indicator collection's thumbnails stay static until the reader interacts with them. So no selector can predict how many charts a page will end up rendering. What is knowable is how many are loading at this moment.

`window._OWID_GRAPHERS_LOADING` is that number. `_OWID_HAVE_ALL_GRAPHERS_LOADED` and the `grapherLoaded` event that fed it are deleted (per review): one consumer does not need two signals, and neither is read anywhere in the org.

A caller still has to scroll the page and then wait for the count to stay at zero — zero reads the same before a chart starts and after it finishes.

Grapher dispatches `grapherLoadingStateChanged` carrying `{ grapher, isLoading }`, and the site tracks it. Three details:

- `fireImmediately: true` on the `isReady` reaction, so a grapher whose config and data are already at hand — ready before it mounts — is counted at all. Today it announces nothing.
- A disposer fires `isLoading: false`, so a chart torn down mid-load doesn't leave the page looking busy forever.
- The tracker keys a `Set` on the grapher rather than incrementing a counter, so the re-renders that the `ResizeObserver` in `GrapherUseHelpers` triggers whenever it re-measures a container can't inflate the total. The old code counted events, so one chart rendering twice could satisfy the wait for a chart that hadn't rendered at all.

`site/runAllGraphersLoadedListener.ts` becomes `site/runGrapherLoadingTracker.ts` — it no longer listens for a "loaded" event.

## Verified

Measured on production before the change, with an init script counting selectors at `DOMContentLoaded` and watching when the flag flips (7 pages, the set that site-screenshots captures):

| page | `[data-grapher-src]` at DCL | flag set at | first chart drawn at |
| --- | --- | --- | --- |
| `/grapher/tomato-production` | 0 | 575ms | 703ms |
| `/life-expectancy` | 0 | 641ms | — |
| `/births-and-deaths` | 0 | 534ms | 4 charts, all later |
| `/from-1-90-to-2-15-a-day…` | 0 | 528ms | 8 charts, all later |
| `/energy` | 0 | 600ms | 3 charts, all later |
| homepage | 0 | 786ms | 1 chart, later |
| `/explorers/migration` | 1 | 1492ms | 890ms — the one page where the flag worked |

Then on `staging-site-graphers-loaded`, with a Playwright probe sampling the globals every 50ms. Because the charts on a healthy page load in single-digit milliseconds, I also ran it with a 1.5s delay injected into every data request, which is what makes the difference visible — the last column is the state at the moment the screenshot tool's wait returns:

| page | `_OWID_GRAPHERS_LOADING` peak | time above zero | spinners left when the wait ended |
| --- | --- | --- | --- |
| `/grapher/tomato-production` | 1 | 2.5s | 0 |
| `/births-and-deaths` | 4 | 39s | 0 |
| `/life-expectancy` | 2 | 15s | 0 |

Same thing again through agent-browser in a real Chrome, using a scroll instead of throttling to make the charts load: the counter hit 4 with 4 spinners on screen mid-scroll, then settled to 0 with all four charts at 813x575 and no spinners; the data page's chart came out at 1232x540, not the 130px-wide render sitting in the current reference screenshot.

The same probe against production, same 1.5s delay, is the before picture: no counter, the old flag true throughout, and the wait returning with **2 spinners still on screen** on `/births-and-deaths` and **1** on `/life-expectancy`.

Also confirmed the count returns to 0 on every page and that a grapher re-rendered by the ResizeObserver (visible as a `true`/`false`/`true`/`false` sequence of events on `/births-and-deaths`) moves the count 1 → 0 → 1 → 0 rather than inflating it.

The numbers above were measured on the first commit, before the flag was deleted; the second commit only removes the flag and the event, which nothing reads.

After the second commit deployed, confirmed on staging that `_OWID_HAVE_ALL_GRAPHERS_LOADED` is absent from `window` (not merely falsy) while the counter still tracks and settles at 0, with no spinners left, on both a data page and an article.

Downstream effect, with owid/site-screenshots#6 driving both sides: production and this staging server now produce **pixel-identical** screenshots for all 7 captured pages, top- and bottom-aligned. The current code differs on 6 of those 7.

`yarn typecheck` and `yarn testLintChanged` are clean.

## Downstream

**owid/site-screenshots#6 has to merge before this one.** It is the only consumer, and it must stop waiting on the deleted flag first — otherwise every shot waits for a flag that never appears and times out. That PR now waits on `_OWID_GRAPHERS_LOADING`, falling back to sampling chart geometry on pages baked before this lands (which is every reference screenshot until it deploys), and separately normalizes the URLs and archive links that made every staging diff a whole-page diff.

This branch's own screenshot job runs against the `graphers-loaded` branch of site-screenshots, which carries that new config, so it exercises the new wait rather than master's.


sophiamersmann (approved):
Makes sense to me!

I can't imagine you'd need both, and I'm fairly sure the screenshotting tool was always the only consumer, so if possible, could you delete `window._OWID_HAVE_ALL_GRAPHERS_LOADED` now and everything that comes with it (the `grapherLoaded` event, etc)

PR #7030, opened 2026-08-21
🎉🤖 (chart-configs) delete indicators the ETL no longer produces

sophiamersmann:

`POST /api/variables/delete` takes a list of indicator ids and deletes them along with everything that hangs off them, so the ETL can drop an indicator without stranding rows behind ([etl#6672](https://github.com/owid/etl/pull/6672)).

An indicator that a chart (including drafts), a published explorer or a live multi-dim view still uses is reported back as `blocked` instead of being deleted, and doesn't fail the request.

**Open question:** Only _published_ explorers and multi-dims are checked. Variables that are used by _draft_ explorers and multi-dims can be deleted, and the explorer/multi-dim would then complain the next time it's updated. I made the distinction because I don't want to annoyingly refuse to delete something that's not currently live, but I'm also wondering if it's worth the complexity.

Per deleted indicator it clears:

- the `variables` row itself
- the `chart_configs` row holding the indicator's own ETL-authored config
- `explorer_variables` — only for draft explorers; a published one blocks instead
- `multi_dim_x_chart_configs` — the materialized view rows, and the `chart_configs` row each one owns, in the DB and in R2 (also only for draft mdims)
- `origins_variables`, `tags_variables_topic_tags` and `posts_gdocs_variables_faqs` — ETL projections rebuilt from the indicator's metadata
- any `origins` and `sources` rows left unreachable afterwards, since `origins_variables` and `variables.sourceId` are the only ways to reach them

It also carries a migration for the orphans already in the table: 7,267 `chart_configs`, 8,280 `origins` and 3,038 `sources` rows on the current snapshot.

@Marigold's #6996, rebased onto this stack and adapted to it.

### DIFF
--- a/adminSiteServer/apiRoutes/variables.ts
+++ b/adminSiteServer/apiRoutes/variables.ts
@@ -48,6 +48,7 @@ import {
 } from "../../serverUtils/r2/chartConfigR2Helpers.js"
 import { Request } from "../authentication.js"
 import { HandlerResponse } from "../FunctionalRouter.js"
+import * as z from "zod"
 
 export async function getEditorVariablesJson(
     req: Request,
@@ -432,6 +433,10 @@ async function updateGrapherConfigsInR2(
     }
 }
 
+const deleteVariablesSchema = z.object({
+    variableIds: z.array(z.number().int()),
+})
+
 /**
  * Delete a set of indicators.
  *
@@ -444,19 +449,15 @@ export async function postVariablesDelete(
     _res: HandlerResponse,
     trx: db.KnexReadWriteTransaction
 ) {
-    const { variableIds } = req.body ?? {}
-
-    if (
-        !Array.isArray(variableIds) ||
-        !variableIds.every((id) => Number.isInteger(id))
-    ) {
-        throw new JsonError(
-            "`variableIds` must be an array of indicator ids",
-            400
-        )
+    const parseResult = deleteVariablesSchema.safeParse(req.body)
+    if (!parseResult.success) {
+        throw new JsonError(`Invalid request: ${parseResult.error}`, 400)
     }
 
-    const { deleted, blocked } = await deleteIndicators(trx, variableIds)
+    const { deleted, blocked } = await deleteIndicators(
+        trx,
+        parseResult.data.variableIds
+    )
 
     return { success: true, deleted, blocked }
 }
--- a/adminSiteServer/apiRoutes/variables.ts
+++ b/adminSiteServer/apiRoutes/variables.ts
@@ -26,6 +26,7 @@ import {
     updateAllChartsThatInheritFromIndicator,
     updateAllMultiDimViewsThatInheritFromIndicator,
     updateIndicatorChartConfig,
+    deleteIndicators,
 } from "../../db/model/Variable.js"
 import { enqueueExplorerRefreshJobsForDependencies } from "../../db/model/Explorer.js"
 import { DATA_API_URL } from "../../settings/clientSettings.js"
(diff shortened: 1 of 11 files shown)


sophiamersmann:
@codex review

db/migration/1787229552132-DeleteOrphanedConfigsOriginsAndSources.ts
```diff
@@ -0,0 +1,67 @@
...
+                    SELECT 1 FROM variables v WHERE v.patchConfigIdETL = cc.id
+                )
+                AND NOT EXISTS (
+                    SELECT 1 FROM multi_dim_redirects mdr
+                    WHERE mdr.viewConfigId = cc.id
+                )
+                AND NOT EXISTS (
+                    SELECT 1 FROM multi_dim_view_dimensions mdvd
+                    WHERE mdvd.chartConfigId = cc.id
+                )
```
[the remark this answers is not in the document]
sophiamersmann: @rakyi, is this true? Should I drop those from the migration?

db/model/Explorer.ts
```diff
@@ -172,10 +173,10 @@ async function validateVariableIds(
...
     const missing = proposed.filter((id) => !found.includes(id))
     if (missing.length > 0) {
-        console.error("missing variables in db", {
-            missing_variables: missing,
-            isPublished,
-        })
+        throw new JsonError(
+            `${missing.length} indicator(s) named by explorer ${slug} don't exist: ${missing.join(", ")}.`,
+            400
+        )
```
sophiamersmann, line 179: Shouldn't we be a bit stricter about explorers with missing variables?

db/tests/testHelpers.ts
```diff
@@ -97,11 +103,3 @@ export async function insertTestChart(
     const [chartId] = await knexInstance(ChartsTableName).insert(row)
     return { chartId, configId, patchConfigId }
 }
-
-export function sleep(time: number, value: unknown): Promise<any> {
```
sophiamersmann, line 101: Unused

db/model/Explorer.ts
rakyi: Yeah, I think so.

db/migration/1787229552132-DeleteOrphanedConfigsOriginsAndSources.ts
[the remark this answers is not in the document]
rakyi: That's true. They are used in analytics. In theory we could add some sort of garbage-collecting process and clean up the rows when we know we no longer need them, but that seems unnecessary for now.

adminSiteServer/apiRoutes/variables.ts
```diff
@@ -430,3 +431,32 @@ async function updateGrapherConfigsInR2(
...
+    req: Request,
+    _res: HandlerResponse,
+    trx: db.KnexReadWriteTransaction
+) {
+    const { variableIds } = req.body ?? {}
+
+    if (
+        !Array.isArray(variableIds) ||
+        !variableIds.every((id) => Number.isInteger(id))
+    ) {
```
rakyi: This isn't too bad but is complicated enough that we could use Zod for it; see e.g. `createNarrativeChartSchema`. In general that's the new pattern for server validation we started using, albeit sparingly for now.

rakyi (commented):
> Open question: Only published explorers and multi-dims are checked. Variables that are used by draft explorers and multi-dims can be deleted, and the explorer/multi-dim would then complain the next time it's updated. I made the distinction because I don't want to annoyingly refuse to delete something that's not currently live, but I'm also wondering if it's worth the complexity.

Best to ask the D&R team, I think. Depends how they want it to work.

Looks good overall, except we have to keep the old rows in the two tables Codex pointed out.

rakyi (approved):

db/migration/1787229552132-DeleteOrphanedConfigsOriginsAndSources.ts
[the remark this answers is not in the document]
sophiamersmann: Ok thanks! I'll drop both from the migration then

adminSiteServer/apiRoutes/variables.ts
sophiamersmann: Will do!

Marigold (approved):

PR #7032, opened 2026-08-21
🔨🤖 Derive region publisher keys from the regions file

sophiamersmann:

Entity names carry a `(Publisher)` suffix — "Africa (FAO)", "Andean Latin America (IHME GBD)" — and the keys we matched those suffixes against were hand-listed in `GrapherConstants`. The regions updater now derives a `publisher` for each region from its own name, so the keys come from the data instead.

The visible effect: a publisher's region sets (`fao_1`, `fao_2`, `fao_sdg`) collapse into one group in the entity selector and on the data tab. The admin still lists them separately, one entity preset per set, since telling the sets apart is the point there.

**Needs a look:** group keys changed, so a `?tableFilter=` link carrying a region set key (`fao_sdg`, `ilo_2`, `un_m49_1`) or a separator-less spelling (`unm49`) now falls back to the unfiltered table without saying so. There is no alias for the old keys — deliberate, but such links exist in the wild.

Also worth knowing: no published chart uses `(IHME GBD)` entities, so that publisher is covered only by unit tests and `/admin/test-region-maps`.

### Charts to review

| Chart | What it exercises |
| --- | --- |
| [global-change-over-the-last-50-years](https://ourworldindata.org/grapher/global-change-over-the-last-50-years) | Eight publishers at once: `ei`, `fao_1/2/sdg`, `maddison`, `un`, `wb` from the regions file, plus `gcp` and eight FAO names it doesn't define |
| [share-of-population-urban](https://ourworldindata.org/grapher/share-of-population-urban) | Three UN M49 sets collapsing into one group, alongside `un` and `un_sdg` |
| [female-labor-force-by-age](https://ourworldindata.org/grapher/female-labor-force-by-age) | `ilo_1` + `ilo_2` collapsing, including the "Arab States (ILO)" backfill |
| [share-of-fish-stocks-overexploited](https://ourworldindata.org/grapher/share-of-fish-stocks-overexploited) | 15 FAO entities, none in the regions file — the suffix-only case |
| [oil-production-by-region](https://ourworldindata.org/grapher/oil-production-by-region) | The plain case: one publisher, one set, complete partition |

### How the keys are derived

`devTools/regionsUpdater/update.ts` writes a `publisher` alongside `defined_by` for every region an institution defines, taken from the last parenthetical of the region's own name (`"Andean Latin America (IHME GBD)"` → `ihme_gbd`). `RegionPublisher` and `RegionSet` are then literal unions derived from the generated `regions.data.ts`, so adding a set or a publisher upstream is picked up by the type system rather than by hand-editing a list.

The updater warns rather than fails when a region has a `defined_by` but no suffix — it gets left out of its region set, and the message says so.

### One implementation of the suffix rule

`parseRegionNameSuffix` in `utils/regions` owns the regex and the canonicalisation. It previously existed in three places (the updater, `RegionGroups.ts`, and a test that re-implemented it to check the other two), which meant tightening any one of them would silently stop every publisher key from matching. The updater and `parseLabel` now both call it.

Matching is tolerant in two ways the data requires: lookups compare publisher keys with separators dropped, so `(IHMEGBD)` and `(IHME GBD)` both reach `ihme_gbd`, and names are trimmed first. A test asserts no two publisher keys collapse to the same lookup key, since one of them would otherwise silently never match.

### Region sets vs publishers

- `RegionGroupKey` is the four OWID groupings plus publishers. Region sets are deliberately not group keys: an entity name only ever identifies the publisher.
- `regionSetLabels` (admin) is the one place naming sets individually; `descriptions` and `customRegionDisplayOrder` in `RegionTooltipData.ts` also need an entry per set.
- `REGION_SET_BACKFILLS` mirrors what the ETL does for a region belonging to two of a publisher's sets at once. Its keys are typed `SuffixedRegionName`, a union of the published region names, so a typo or an upstream rename is a typecheck error rather than a set quietly ceasing to be a complete partition.

### Admin entity presets

`getAvailablePresets` used to build its suffix presets from `ADDITIONAL_REGION_PUBLISHERS`. It now groups the chart's own entity names by the publisher they name, which means:

- FAO/ILO regions the regions file doesn't define are offered again — `share-of-fish-stocks-overexploited` has 15 such entities and no per-set preset covers any of them
- the presets agree with how the entity selector groups the same entities
- a suffix preset is skipped when the per-set presets already offer every one of its entities, so single-set publishers like `wb` don't gain a duplicate

`ADDITIONAL_REGION_PUBLISHERS` now holds only publishers the regions file doesn't define, and a test fails if one it does define is listed there.

### Smaller things

- `getRegionsForKey` is memoized, so all callers share one array; it returns `readonly TooltipRegion[]` and the tooltip props follow, since a sort or push in one caller would corrupt every chart tooltip.
- The guard behind `AggregateWithPublisher` checked only `definedBy` while asserting `publisher` too.
- `CustomSchemes.test.ts` compared `Object.keys` arrays with `toEqual`, asserting declaration order rather than the membership invariant it documents.

### Test plan

- `yarn typecheck`, `yarn testLintChanged`, `yarn testFormatChanged` clean
- `yarn test run` — full suite passes
- New tests in `RegionGroups.test.ts` (suffix spelling tolerance, lookup-key collisions, stale additional publishers) and `regions.test.ts` (one key per publisher, not per set)
- No rendering logic changed, so `make svgtest` was not run

</details>

---
This is **part 6 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;6&nbsp;</kbd> #7032 👈 
- <kbd>&nbsp;5&nbsp;</kbd> #7029
- <kbd>&nbsp;4&nbsp;</kbd> #7002
- <kbd>&nbsp;3&nbsp;</kbd> #6988
- <kbd>&nbsp;2&nbsp;</kbd> #6980
- <kbd>&nbsp;1&nbsp;</kbd> #6979


sophiamersmann:

Opened against the wrong branch by mistake — this description belongs to `regions-derive-publishers`. Closing; no changes were made to this branch.

PR #7038, opened 2026-08-21
Improve animations

rakyi:
I used skills from https://emilkowal.ski/skill to improve our animations. I also added the skills to the project for future use. You don't have to review their content unless you really want to. I think we wouldn't update them and would rather remove those we don't like.

Please check mostly just the changes in `.ts{x}` files. If you disagree with some of the animation changes, please let me know. I found all of them to be an improvement, but I know it's subjective.


sophiamersmann (approved):
Looks good to me as well!

rakyi:
@codex review

rakyi:
I now also moved SlideInDrawer to React Aria. Could you check it, pls?

packages/@ourworldindata/grapher/src/slideInDrawer/SlideInDrawer.tsx
```diff
@@ -1,150 +1,41 @@
...
+        <UNSAFE_PortalProvider getContainer={() => grapherRef.current}>
+            <ModalOverlay
+                className="drawer"
+                isOpen={isOpen}
+                onOpenChange={onOpenChange}
+                isDismissable
+            >
+                <Modal className="drawer-contents">
+                    <Dialog
+                        aria-label="Entity selector"
```
sophiamersmann: I know the entity selector is the only consumer right now, but it'd still be nice to keep the SlideInDrawer agnostic of that

sophiamersmann (approved):

packages/@ourworldindata/grapher/src/slideInDrawer/SlideInDrawer.tsx
rakyi: Good point, changed.

PR #7043, opened 2026-08-24
🔨🤖 (featured-viz) share the cite and license sections between gdoc pages

sophiamersmann:

`GdocPost` and `Profile` each carried a near-copy of the "Cite this work" and "Reuse this work freely" sections at the foot of the page, plus their own copy of the citation and BibTeX strings that feed them. A third page type is coming (the `featured-viz` gdoc) and would have been a third copy, so I pulled `GdocPost`'s version out into shared `CitationSection` and `LicenseSection` components and a `buildGdocCitation` helper.

- **Fix.** The "our charts can be embedded" link on profile pages pointed at `https://github.com/owid`, which says nothing about embedding. I read that as an old copy-paste. It now points at the embedding FAQ, where every other page already sent it, and goes through `IS_ARCHIVE` so it resolves on archived pages too.
- **Behaviour change.** The Creative Commons license link on profile pages no longer opens in a new tab. Profile was the only page setting `target="_blank"` on it, so I made the minority conform rather than add the attribute to every article, topic and about page.

### DIFF
--- a/site/gdocs/components/LicenseSection.tsx
+++ b/site/gdocs/components/LicenseSection.tsx
@@ -1,0 +1,60 @@
+import { LICENSE_ID } from "@ourworldindata/utils"
+import { IS_ARCHIVE } from "../../../settings/clientSettings.js"
+import { PROD_URL } from "../../SiteConstants.js"
+
+const BASE_URL = IS_ARCHIVE ? PROD_URL : ""
+
+/** The "Reuse this work freely" section at the foot of a gdoc page */
+export function LicenseSection({ isDeprecated }: { isDeprecated?: boolean }) {
+    return (
+        <section
+            id={LICENSE_ID}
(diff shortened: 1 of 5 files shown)


site/gdocs/components/LicenseSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+                            height={57}
+                        />
+                        <h3>Reuse this work freely</h3>
+                    </>
+                )}
+
+                <p>
+                    All visualizations, data, and articles produced by Our World
+                    in Data are completely open access under the{" "}
+                    <a href="https://creativecommons.org/licenses/by/4.0/">
```
sophiamersmann, line 31: Should this be opened in a new tab or not?

ikesau: I think it's fine - users can make the decision about a new tab.

ikesau (approved):

PR #7058, opened 2026-08-25
🎉🤖 (featured-viz) featured-viz gdoc type and page

sophiamersmann:

Until now a bespoke data viz could only appear as a figure inside an article. This adds a `featured-viz` gdoc type, published at `/featured-viz/<slug>`, that gives one a page of its own.

The page splits its body around the first top-level `{.bespoke-component}` block: everything before it renders as an intro, the block itself renders on a full-bleed blue band at the width its `size` asks for, and everything after renders normally.

The final design isn't settled yet, so we might change this in the future. For the sake of this review though, just take the code at face value.

Notes:
- Validation rejects a page with no top-level bespoke block, and one with no title or excerpt. A page with several bespoke blocks is accepted, and only the first one gets the band.
- The viz on the page doesn't sync its state to the URL yet, and the page isn't embeddable. Both are in the branch on top.

Example: [Admin](http://staging-site-featured-viz/admin/gdocs/10h9cofxK2x961gzOPnIqidPS_mlsb6DAAehdRUWLd4Y/preview) / [Baked](http://staging-site-featured-viz/featured-viz/test-featured-viz-article)


adminSiteClient/gdocsValidation.ts
```diff
@@ -341,12 +341,12 @@ export const getErrors = (gdoc: OwidGdoc): OwidGdocErrorMessage[] => {
     validateTitle(gdoc, errors)
     validateSlug(gdoc, errors)
     validateBody(gdoc, errors)
+    validateRefs(gdoc, errors)
```
sophiamersmann, line 353: Refs used to be validated only for gdoc posts, but now they're also validated for other gdoc types (including featured-viz and profiles, for example). I checked with Claude whether there are any existing errors this would now produce, but that doesn't seem to be the case.

adminSiteClient/gdocsValidation.ts
```diff
@@ -355,6 +355,8 @@ export const getErrors = (gdoc: OwidGdoc): OwidGdocErrorMessage[] => {
         validateDataInsightImage(gdoc, errors)
     } else if (checkIsAuthor(gdoc)) {
         validateSocials(gdoc, errors)
+    } else if (checkIsFeaturedViz(gdoc)) {
```
ikesau, line 367: I think we should require `featured-image` to be set for these as well.

packages/@ourworldindata/components/src/GdocsUtils.ts
```diff
@@ -108,6 +108,12 @@ export function getPrefixedGdocPath(
             },
             () => `${prefix}/data-insights/${gdoc.slug}`
         )
+        .with(
+            {
+                content: { type: OwidGdocType.FeaturedViz },
+            },
+            () => `${prefix}/featured-viz/${gdoc.slug}`
```
ikesau, line 115: Are we confident on this namespace? I checked Slack to see if this had been discussed/decided somewhere or was just your call, but couldn't find anything.

site/gdocs/pages/FeaturedViz.tsx
```diff
@@ -0,0 +1,159 @@
...
+                <hr
+                    className={cx(
+                        "featured-viz-divider",
+                        getLayout("horizontal-rule")
+                    )}
+                />
+            )}
+            <ArticleBlocks blocks={before} />
+            {hero && <FeaturedVizHero block={hero} />}
+            <ArticleBlocks blocks={after} />
```
ikesau, line 63: Image

The last `p` is looking a little close to the footer here

ikesau (approved):
Nice! Worked in all the ways I tested it.

A few questions:
- Is there a plan to incorporate these on the latest page, or will they just be done as announcements?
- Do you think we could add no-JS support for bespoke viz at some point in the future? At its best, dynamic thumbnails a la Cloudflare, or more feasibly: `fallbackImageFilename` as a config value in the archie.

packages/@ourworldindata/components/src/GdocsUtils.ts
sophiamersmann: The 'featured-viz' name for the gdoc (and more generally as the user-facing name) was decided in a shaping call (we couldn't think of anything better tbh, so do say if you have sth else in mind). Publishing this under `/featured-viz/` is my call - I should probably clarify that now

sophiamersmann:
Thanks for the review, @ikesau!

> Is there a plan to incorporate these on the latest page, or will they just be done as announcements?

That's still up in the air.

> Do you think we could add no-JS support for bespoke viz at some point in the future? At its best, dynamic thumbnails a la Cloudflare, or more feasibly: fallbackImageFilename as a config value in the archie.

Good point! Dynamic thumbnails a la Cloudflare are far away, but setting an image filename in Archie is a good idea. I'll look into this and add it in a separate PR!

adminSiteClient/gdocsValidation.ts
sophiamersmann: done!

site/gdocs/pages/FeaturedViz.tsx
sophiamersmann: yes!

PR #7064, opened 2026-08-25
🔨🤖 drop the dead cssUrl path from bespoke components

sophiamersmann:

A bespoke component's registry entry could name a stylesheet URL, which the mount chain would load into the shadow root before importing the module. We're not using it. Styles get inside the shadow root by a different route: `vite-plugin-css-position` inlines them into the ES module.

I am removing it now because I am about to add a `metadataUrl` to the same interface for the featured viz methods block.


marcelgerber (approved):
great

PR #7066, opened 2026-08-25
🔨🤖 (colour-legends) split colour legends into state and render components

sophiamersmann:

The colour legends took a `manager` object that did two unrelated jobs. It was the legend component's props, and it was also the bag a chart publishes as `ChartInterface.externalLegend` so a faceted view can build one legend across its facets. That second job is why every chart had to declare `legendX`, `numericLegendY` and friends as public members nothing outside its own layout read, and why the horizontal and vertical interfaces drifted into spelling the same concept two ways (`legendMaxWidth` against `maxLegendWidth`, `legendY` against `numericLegendY` plus `categoryLegendY`).

Legends now use the state and render split that `VerticalLabels` already uses: a `*State` class owning layout, plus a function component taking `{ state, x, y, ...handlers }`. `externalLegend` gets its own data-only type. Since `x` and `y` are render props, positioning leaves the legends entirely, and public legend-related members across the ten chart classes that own a legend drop from 112 to 36.

- **Reference SVGs need regenerating, and `make svgtest` will hide real diffs until they are.** Two unrelated causes. 596 files differ in float literals only, largest delta 2.3e-13: the old numeric legend baked `legendX` into its running offset before accumulating bin widths, and it now accumulates relative to its own origin with the render component adding `x` last, so `(x + a) + b` became `x + (a + b)`. The categorical legend already worked the relative way, which is why only the numeric one moved. Separately, static maps drop the numeric legend's touch hit-area rects, which is a structural diff rather than a rounding one. I counted the 596 before that fix landed, so read it as a floor.
- Beyond that the output is unchanged. I exported the whole `graphers` suite before and after three times: as-is, with `facet=entity` to reach `FacetChart`, and with the map tab on a time range to reach `FacetMap`. Neither faceted path is exercised by any config in the suite on its own, so both would otherwise have gone untested. No structural differences in any of the three passes, though that comparison predates the hit-area fix below, which does change static map output.
- Dead weight went with it. `legendHeight` and `legendCursor` were unread on both ends, and so were an unused ref, the abstract `HorizontalColorLegend` base, `FacetChart`'s `LegendClass`, the categorical legend's unreachable `containerWidth`, and `StackedBarChart`'s `legendWidth` branch that measured the horizontal legend off the vertical one.
- Each horizontal legend takes one required width option instead of two optional ones. Numeric takes `maxWidth` because it shrinks to its ideal width and aligns inside that budget; categorical takes `width` because it always fills what it is given. No caller ever set both, so the old `maxWidth ?? width` against `width ?? maxWidth` split was arbitrating nothing.
- One layout cycle can no longer come back. `MapChart.numericLegendY` read the legend's height while the legend read `numericLegendY` back off the same manager, terminating only because `height` happened not to depend on Y. `FacetMap` had the same shape.

### Charts to look at

One per legend path, on `staging-site-legend-state`. Every one should be pixel-identical to the same chart on production.

| Legend | Chart | Link |
| --- | --- | --- |
| Horizontal numeric | Human Development Index, a map on a continuous scale | [human-development-index](http://staging-site-legend-state/grapher/human-development-index) |
| Numeric and categorical together, the only chart type that draws both | Age of full democracy, whose map has a categorical band alongside the strip | [age-of-full-democracy](http://staging-site-legend-state/grapher/age-of-full-democracy) |
| Horizontal numeric on a line chart, which only three charts in the suite do | Global temperature anomalies by El Niño and La Niña | [global-temperature-anomalies-by-el-nino-la-nina](http://staging-site-legend-state/grapher/global-temperature-anomalies-by-el-nino-la-nina) |
| Horizontal categorical, wrapping onto several lines | Animal protein consumption, a stacked discrete bar | [animal-protein-consumption](http://staging-site-legend-state/grapher/animal-protein-consumption) |
| Horizontal categorical on a Marimekko | Life expectancy marimekko | [life-expectancy-marimekko](http://staging-site-legend-state/grapher/life-expectancy-marimekko) |
| Vertical | Life expectancy vs GDP per capita, a scatter plot | [life-expectancy-vs-gdp-per-capita](http://staging-site-legend-state/grapher/life-expectancy-vs-gdp-per-capita) |
| Vertical in a narrow sidebar | Child deaths by world region, a stacked bar | [child-deaths-by-world-region-un-data](http://staging-site-legend-state/grapher/child-deaths-by-world-region-un-data) |
| `FacetChart` building one legend across facets | Broad causes of death, faceted by default | [broad-causes-of-death](http://staging-site-legend-state/grapher/broad-causes-of-death) |
| `FacetMap` building one legend across facets | Human Development Index again, on the map tab over a time range, which is what splits a map into facets | [human-development-index?tab=map&time=1990..2020](http://staging-site-legend-state/grapher/human-development-index?tab=map&time=1990..2020) |

Two pre-existing bugs turned up along the way, and both are fixed here:

- The categorical legend's `marks` mutated the objects `markLines` produced, so a change to `align` against a cached `markLines` would have shifted them twice. Unreachable today, because the state is rebuilt on every change and its caches are never stale, which makes it latent rather than observable. `marks` now copies, and the output is identical.
- **Fix.** The numeric legend never consulted `isStatic`, so a static export emitted the touch hit-area layer whenever a chart passed a touch handler. It is now gated on the same `interactive` prop the other two legends already used, so static exports stop carrying rects nothing can hit. `MapChart` is the only chart that wires up a touch handler, so non-faceted static maps with a numeric legend are the only output affected.


sophiamersmann:
@codex review

PR #7076, opened 2026-08-26
feat(mdim): search field in dropdown menus

marcelgerber:
> _Written by Claude Fable 5 — @marcelgerber at the wheel._

Stacked on #7073. Implements the search field in mdim dropdown menus from the [Explorers-to-MDims designs](https://www.figma.com/design/FM9GL3twDj7pRxyAaeM67s/Explorers-to-MDims?node-id=275-2231): dropdowns with **more than ten choices** get a fuzzy search field at the top of their menu. The searchable dropdown is a new `presentation.type` value, `"dropdown-with-search"` — the heuristic picks it automatically for >10 choices, and authors can force either dropdown variant by setting the type explicitly (the ETL schema follow-up should include the new enum value). When the search shows, it replaces the menu's name/description header.

- Matching uses `FuzzySearch` over both choice names **and their group names**, so searching for a group shows all of its choices; sections without matches are hidden, and an empty state shows when nothing matches.
- The field autofocuses when the menu opens — except on touch devices (`TOUCH_DEVICE_MEDIA_QUERY`), where the virtual keyboard would cover the menu.

Searching for the group name "Aggregates" on `energy-mix`:

- **Config**: `DimensionPresentationType` gains `"dropdown-with-search"`; the heuristic in `resolveDimensionPresentationType` (`packages/@ourworldindata/utils/src/MultiDimDataPageConfig.ts`) resolves omitted types to radio (≤2 ungrouped choices) → dropdown-with-search (>10 choices) → dropdown, and an explicit `presentation.type` always wins. It still resolves against the dimension's full choice list (in `MultiDimControls`), so the control doesn't flip as other selections change. Unit tests in `MultiDimDataPageConfig.test.ts`.
- **UI**: react-aria `Autocomplete` wrapping a `SearchField` + the existing `ListBox` inside the Select popover — this wires up virtual focus, so arrow keys move through the filtered list while typing and Enter selects. Styled per the Figma search variant (32px, gray-20 border, magnifying-glass icon, 14px text, 4px insets; webkit's built-in clear button is hidden).
- **Filtering**: `Autocomplete`'s `filter` is called per menu item, so the fuzzy results (`FuzzySearch.withKeyArray` keyed on name + group, deduped by slug) are cached per query in a ref.
- **Verified locally**: `vaccination-coverage-who-unicef` antigen dropdown (15 choices) — search shown, header hidden, autofocused, "polio" filters 15 → 3; its metric dropdown (3 choices) — no search, header kept; `energy-mix` — "aggregat" matches the group and shows all its choices with the other section hidden; touch emulation — search present but not focused.

</details>


packages/@ourworldindata/utils/src/MultiDimDataPageConfig.ts
```diff
@@ -262,7 +263,8 @@ export function resolveDimensionPresentationType(
 ): DimensionPresentationType {
     if (dimension.presentation?.type) return dimension.presentation.type
     const hasGroups = dimension.choices.some((choice) => choice.group)
-    return dimension.choices.length <= 2 && !hasGroups ? "radio" : "dropdown"
+    if (dimension.choices.length <= 2 && !hasGroups) return "radio"
+    return dimension.choices.length > 10 ? "dropdown-with-search" : "dropdown"
```
marcelgerber: This heuristic here can go wrong:
For example, for the [Covid explorer](http://staging-site-mdim-dropdown-search/admin/grapher/covid%2Flatest%2Fcovid%23covid_explorer?metric=xm&interval=cum&relative=relative)'s Interval, it shows a search bar even though in all cases there's only a few choices available - but there are 11 overall, distinct choices.

Probably fine, since this can easily be overridden using the `presentation.type` prop.

I've since raised the threshold value up to 20, and that should alleviate this issue to a good degree. In any case, it's very easy to provide an explicit value here if the heuristic ever goes wrong.

PR #7077, opened 2026-08-26
🐛🤖 stop reporting unchecked charts as visible changes

sophiamersmann:

The SVG tester report called a chart a visible change before it had looked at it. A chart the pixel check hadn't reached fell back to `"changed"`, and the card's status badge only rendered for a status other than `"changed"`, so "I found a real difference" and "I haven't started yet" were the same thing on screen. A thumbnail run with 11 differences read as 10 visible changes, when 9 of them were pure element reorders nobody had checked yet.

- **Fix.** "Not checked yet" is a value in the type now, rather than a `??` fallback in two places. `VisualVerdict` keeps the three outcomes the pixel check can actually conclude, and a new `VisualStatus` adds `pending` on top for what the report knows. The split only goes one way on purpose, so nothing can write a `pending` into storage or return one from a comparison.
- **Behaviour change.** Every card carries a status tag now, including "Visible change", which used to be the unlabelled default. Bold rather than coloured, because red and green already mean reference and current in the swipe and overlay views.
- **Behaviour change.** The filter default follows the check instead of sitting at "changed". The unchecked charts while it runs, the changed ones once it finishes, and your own pick outranks both. Leaving it fixed at "changed" would collapse the list to a single card the moment the first real change came back, with hundreds still unchecked.
- I haven't opened the report in a browser, so the thing worth confirming is that the "Visible change" count starts at zero and grows, rather than starting at the total and shrinking.
- The first commit is a pure move of the bucketing out of the 1100 line component and into `svgVisualDiff.ts`, where a test can reach it. Reading it on its own makes the second commit the behaviour change and nothing else.

### DIFF
--- a/adminSiteClient/SvgTesterSuitePage.scss
+++ b/adminSiteClient/SvgTesterSuitePage.scss
@@ -158,8 +158,15 @@ $svg-tester-error-frame: #dda4a0;
     }
 }
 
-.SvgTesterSuitePage__chart-type {
+.SvgTesterSuitePage__chart-type,
+.SvgTesterSuitePage__status {
     margin-left: 8px;
+}
+
+// The chart type beside it is neutral, so the one status that is a finding
+// needs the weight the headline gives its numbers
+.SvgTesterSuitePage__status.is-changed {
+    font-weight: bold;
 }
 
 .SvgTesterQueryParams {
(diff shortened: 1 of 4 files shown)


sophiamersmann:
* **#7077** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7077?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7077?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #7081, opened 2026-08-26
Fix Algolia caching

rakyi:
I realized we are doing a bunch of unnecessary  work and also have some potential correctness issues.

See each commit.

### DIFF
--- a/site/search/Autocomplete.tsx
+++ b/site/search/Autocomplete.tsx
@@ -26,7 +26,7 @@ import {
     BAKED_GRAPHER_URL,
 } from "../../settings/clientSettings.js"
 import { DEFAULT_SEARCH_PLACEHOLDER } from "./searchClient.js"
-import { getLiteSearchClient } from "./searchClients.js"
+import { getAutocompleteSearchClient } from "./searchClients.js"
 import {
     PAGES_INDEX,
     CHARTS_INDEX,
@@ -89,7 +89,7 @@ const buildRecentSearchesPlugin = () =>
 
 let liteSearchClient: LiteClient | null
 if (ALGOLIA_ID && ALGOLIA_SEARCH_KEY) {
-    liteSearchClient = getLiteSearchClient()
+    liteSearchClient = getAutocompleteSearchClient()
 } else {
     liteSearchClient = null
     console.warn("Algolia credentials are not set")
--- a/site/search/Autocomplete.tsx
+++ b/site/search/Autocomplete.tsx
@@ -262,6 +262,16 @@ const algoliaGetItemUrl: AutocompleteSource<BaseItem>["getItemUrl"] = ({
     item,
 }) => prependSubdirectoryToAlgoliaItemUrl(item)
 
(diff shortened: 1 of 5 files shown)


site/search/Autocomplete.tsx
```diff
@@ -262,6 +262,11 @@ const algoliaGetItemUrl: AutocompleteSource<BaseItem>["getItemUrl"] = ({
     item,
 }) => prependSubdirectoryToAlgoliaItemUrl(item)
 
+const ALGOLIA_ITEM_PARAMS = {
+    attributesToRetrieve: ["title", "slug", "type", "queryParams"],
```
mlbrgl, line 271: not a big deal, but since `queryParams` are only used by the charts query, I wouldn't "retrieve" them in the other ones (not because it makes a difference in execution, but for accuracy in the absence of stronger typing)

site/search/searchClients.ts
```diff
@@ -52,11 +57,41 @@ const createCachingRequester = (): Requester => {
...
 
+/**
+ * Algolia's response cache is disabled here so React Query is the only cache in
+ * play. Left on, it would keep a second copy of every response with no TTL and
+ * no size bound, and — sitting above the requester — it would win whenever both
+ * caches have an entry, silently defeating React Query refetches: a lower
+ * `staleTime`, `invalidateQueries` or a refetch on window focus would all
+ * appear to succeed while returning the byte-identical cached response.
+ * `requestsCache` (in-flight deduplication of identical requests) is a separate
+ * cache and stays on.
```
mlbrgl: I wonder if this isn't revealing too much of Algolia's caching architecture which is only tangentially relevant here

site/search/searchClients.ts
```diff
@@ -52,11 +57,41 @@ const createCachingRequester = (): Requester => {
...
+
+/**
+ * The search client for the autocomplete, which keeps Algolia's response cache
+ * (on by default) because it reaches Algolia through @algolia/autocomplete-js
+ * rather than React Query and would otherwise have no cache at all —
+ * backspacing or retyping a query would re-request every intermediate state.
+ *
+ * This deliberately doesn't share a client with `getLiteSearchClient`, and no
+ * caching is lost by keeping them apart: the two never issue an identical
+ * request, so a shared cache could never serve one from the other's entry.
```
mlbrgl: this explain why it doesn't hurt to have them separate, but it is rather light on the "deliberately" and doesn't really justify why splitting more desirable than alternatives (e.g. keeping a single client for both, and wire autocomplete through react query, which is not necessarily desirable)

mlbrgl (approved):
Nice that you simplified the cache layering here! 👍

Tested by forcing react query's to refire through `refetchOnWindowFocus: "always"` and could witness:
- before: no new queries sent to algolia
- after: new queries sent to algolia (which then means that once RQ's cache isn't artificially bypassed, it is the only cache in the game)

rakyi:
All good points, addressed. Thanks!

PR #7086, opened 2026-08-26
🐛🤖 Prevent flashes in delayed loading states

rakyi:
In the previous PR I added a nice little package for this, but turns out we have to vendor it because it is UMD only. And I think vendoring is the simplest solution.

### DIFF
--- a/packages/@ourworldindata/utils/src/useSpinDelay.ts
+++ b/packages/@ourworldindata/utils/src/useSpinDelay.ts
@@ -1,0 +1,69 @@
+import { useEffect, useRef, useState } from "react"
+import { useIsClient } from "usehooks-ts"
+
+// Vendored from spin-delay v2.0.1 (MIT):
+// https://github.com/smeijer/spin-delay/blob/v2.0.1/src/index.ts
+// We keep an ESM copy here because the package only publishes a UMD build,
+// which leaves a runtime `require("react")` in Grapher's published ESM bundle.
+
+type SpinDelayOptions = {
+    delay?: number
+    minDuration?: number
+    ssr?: boolean
+}
+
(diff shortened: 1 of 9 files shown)


yarn.lock
```diff
@@ -2347,6 +2347,7 @@ __metadata:
     react-dom: "npm:^19.2.4"
     react-markdown: "npm:^10.1.0"
     remeda: "npm:^2.42.0"
+    spin-delay: "npm:^2.0.1"
```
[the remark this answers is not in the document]
rakyi: It's already there.

rakyi:
@codex review

packages/@ourworldindata/utils/src/useSpinDelay.ts
```diff
@@ -0,0 +1,74 @@
...
+    useEffect(() => setIsSsr(false), [])
+    return isSsr
+}
+
+export function useSpinDelay(
+    loading: boolean,
+    options: SpinDelayOptions = {}
+): boolean {
+    const resolvedOptions = { ...defaultOptions, ...options }
+    const isSsr = useIsSsr() && resolvedOptions.ssr
```
marcelgerber: very slight preference to use `useIsClient` from `usehooks-ts` here instead, but I can also see the other side that you want to keep the code as close to original as possible.
```suggestion
    const isSsr = !useIsClient() && resolvedOptions.ssr
```

marcelgerber (approved):
yup makes sense!

packages/@ourworldindata/utils/src/useSpinDelay.ts
rakyi: Yeah, I think we can do that.

PR #7092, opened 2026-08-27
🎉🤖 render a fallback for bespoke viz without JavaScript

sophiamersmann:
A bespoke component is client-side JavaScript, so with JavaScript unavailable a bespoke embed rendered a spinner that never resolved. It now shows a fallback image if given, and the site's "JavaScript needs to be enabled" notice where it doesn't. 

The new prop on the `bespoke-component` archie is `fallbackImageFilename`

You can use this page for testing: http://staging-site-bespoke-nojs-fallback/sophias-test-article. The first embed has `fallbackImageFilename` set, the second doesn't.

### DIFF
--- a/site/gdocs/components/Image.tsx
+++ b/site/gdocs/components/Image.tsx
@@ -34,6 +34,7 @@ const gridSpan5 = generateResponsiveSizes(5)
 const gridSpan6 = generateResponsiveSizes(6)
 const gridSpan7 = generateResponsiveSizes(7)
 const gridSpan8 = generateResponsiveSizes(8)
+const gridSpan12 = generateResponsiveSizes(12)
 
 export type ImageParentContainer =
     | Exclude<Container, "sticky-right-left-heading-column">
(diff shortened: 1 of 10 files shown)


site/gdocs/components/Image.tsx
```diff
@@ -68,6 +72,9 @@ const containerSizes: Record<ImageParentContainer, string> = {
     ["latest-announcement"]: gridSpan6,
     ["chart-rows"]: gridSpan3Sm,
     ["pull-chart"]: gridSpan3Sm,
+    ["bespoke-component--narrow"]: gridSpan6,
+    ["bespoke-component--wide"]: gridSpan8,
+    ["bespoke-component--widest"]: gridSpan12,
```
sophiamersmann, line 77: Mirrors bespoke component sizing

ikesau (approved):
Thank you! :)

Authors will probably forget to add this. We should make sure it's documented in the gdocs documentation and advertised / checked by one of us for the first few pages we publish.

PR #7095, opened 2026-08-27
🔨🤖 rewrite the grapher modal with react-aria-components

sophiamersmann:

Rewrites Grapher's modal using React Aria

- **Behaviour change.** `useModalOverlay` locks page scroll and marks everything outside the modal `inert`, and offers no way to opt out. While a chart's Download or Sources modal is open, a reader cannot scroll the article or click anything else on the page.
- **Behaviour change.** A click on a link or button outside the chart now dismisses an open modal. The exemption came from #2811, for the chart revision tool whose table was dropped in `1748446282000-DropSuggestedChartRevisionsTable`. Let's try to remove it and only add it back when someone complains.
- `isOwidDropdownOpen` and its `data-owid-dropdown` marker are gone. React Aria stacks open overlays and only lets the topmost answer an outside press, and our `Dropdown` is a RAC `Popover`, so it sits on that stack. Narrower than the old check, which suppressed dismissal for any open dropdown anywhere on the page.

### DIFF
--- a/packages/@ourworldindata/grapher/src/chart/ChartUtils.tsx
+++ b/packages/@ourworldindata/grapher/src/chart/ChartUtils.tsx
@@ -28,9 +28,6 @@ import { LineChartSeries } from "../lineCharts/LineChartConstants"
 import { SelectionArray } from "../selection/SelectionArray"
 import { ChartManager } from "./ChartManager"
 import {
-    GRAPHER_SIDE_PANEL_CLASS,
-    GRAPHER_TIMELINE_CLASS,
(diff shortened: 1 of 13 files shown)


rakyi (approved):
Nice! Not sure, should we maybe make the overlay style the same for the slide-in drawer and the modal? Modal overlay seems to be slightly darker.

And the slide-in drawer overlay animates. In fact, we could use the animation skill to animate the modals as well, but up to you if you want to spend time on it now.

sophiamersmann:
Both good points, will do!

PR #7101, opened 2026-08-28
🐛🤖 test metadataHelpers and fix what the tests found

sophiamersmann:

> [!NOTE]
> Waiting on Pablo's feedback 

Report: https://claude.ai/code/artifact/acb80de5-cc1d-48df-87f8-9ac2bd61ee62

`metadataHelpers` builds the attribution and citation text on every data page, chart sources modal and data download, and had no test coverage. Writing it (98 cases across the module's 17 exports) turned up a row of wrong strings readers have been seeing, so this fixes those and puts the duplicated citation and attribution logic behind single entry points. The module is down to 13 exports, the three citation builders are private behind `getIndicatorCitations`, and the rule that hides our processing phrase lives in one function rather than three.

The fixes, roughly by how many readers saw them:

- **Fix.** A chart's sources modal and the download readme separated credits with ", " while citations and a data page's Source row used "; ". Producer names routinely contain commas and never semicolons, so the comma form hid where one credit ended: of the fourteen behind `forest-area-as-share-of-land-area`, "Kleinn, C., Corrales, L., & Morales, D. (2002)" read as three sources. All four surfaces now use "; ".
- **Fix.** 78 indicators cited themselves as "Our World in Data – with minor processing by Our World in Data", across 851 chart-variable pairs, almost all of them the region colour dimension of our scatter plots. The Source row above the citation already dropped that phrase, so the same page contradicted itself.
- **Fix.** "Last updated" in the sources modal was a day early for readers west of UTC, because the date was parsed as UTC midnight and then formatted in local time. Baked pages and the Cloudflare functions run at UTC, so it only ever hit the browser, on the 868 chart variables whose catalog version is not a full date.
- **Fix.** The admin's three "Open the metadata.yaml file" links were broken for every indicator: `getETLPathComponents` split the catalog path only on "/", leaving the indicator glued to the table name.
- **Fix.** An unparseable archival date reached the reader as "(archived on Invalid Date)." at the end of a citation. It now says "(archived)." and leaves out the date.
- **Fix.** The download package credited two different source sets in the same zip: the readme's Source row and metadata.json's citations named only the origin producers, while the readme's own in-line citation named the source too. `getAttributionFragmentsFromVariable` reads `variable.source`, which a column def flattens into `sourceName`, so the call sites that handed it the def unchanged never saw it. 70 indicators carry both — though none of them is used by a chart, explorer or mdim, so no reader can currently download a zip that shows the mismatch.
- **Fix.** The download readme's Source row spelled our name "Our World In Data". Routing it through the same composer as the citations fixes the casing.
- **Fix.** A string no current indicator produces, but one bad ETL record would: an origin with no producer printing the literal "undefined", as in "undefined (2019)" in a chart's sources line.
- **Fix.** An mdim previewed without a slug cited itself as "Retrieved from  [online resource]". Those cannot be published, so it is admin-preview only.
- **Fix.** Announcement and data insight bylines lost the Oxford comma that the site's eight other author lists carry, so three or more authors read "A, B and C" in those two places and "A, B, and C" everywhere else.
- The download modal now leaves out origins it cannot name at all. They rendered as empty, unlinkable list items, two of which collided on React's `key=""`.

An indicator with nothing to credit at all still opens its Source row and citations with a dash and nothing in front of it, as it does today. Nothing reaches that case — it needs no attribution, no legacy source and no origins at once, and none of the 788,530 variables qualifies.

### How the strings were checked

Master's helpers, and the call-site logic that lived in each surface, were copied verbatim and run beside this branch's over the real indicator metadata for the 210 indicators behind the 100 most-viewed charts. Both implementations saw identical inputs in one process, so every difference is the code rather than the data. "Last updated" was compared with the clock set to `America/Los_Angeles`, since that change is only observable west of UTC.

What changes for those 100 charts:

| Fix | Indicators | Charts |
| --- | --- | --- |
| `; ` instead of `, ` between credits | 66 / 210 | 28 |
| No processing phrase when we are the sole credit | 1 | 17 |
| "Last updated" west of UTC | 17 | 8 |
| "Our World in Data" casing in the readme | 209 | 100 |
| Admin `metadata.yaml` link | 210 | 100 |

Byte-identical across all 210 indicators: the sources line under every chart, the data page's Source row and its "How to cite this page" citation, the labels in "Sources and processing", and the download modal's source links. Two-author bylines and the eight author lists that already went through `formatAuthors` are unchanged too.

### What is only reachable in principle

- Every origin in the snapshot has a producer, and none carries an empty attribution string, so the `undefined` credit needs a bad ETL record.
- No variable has an empty attribution list — that needs no attribution, no legacy source and no origins at once — so the dangling dash cannot be reached.
- None of the 70 indicators carrying both a legacy source and origins is used by a chart, explorer or mdim.

### Checks

`yarn test run` passes (2,478 unit tests), `yarn typecheck` is clean, and lint and format pass on the changed files.

</details>

### DIFF
--- a/functions/_common/metadataTools.ts
+++ b/functions/_common/metadataTools.ts
@@ -1,6 +1,6 @@
 import * as _ from "lodash-es"
 import { GrapherState } from "@ourworldindata/grapher"
-import { OwidTableSlugs, OwidOrigin } from "@ourworldindata/types"
+import { OwidTableSlugs } from "@ourworldindata/types"
 import {
     getLastUpdatedFromVariable,
     getNextUpdateFromVariable,
@@ -72,55 +72,11 @@ export function assembleMetadata(
             timespan,
             tolerance,
             type,
-            origins,
-            sourceLink,
-            sourceName,
             owidVariableId,
             shortName,
         } = col.def
         const lastUpdated = getLastUpdatedFromVariable(col.def)
         const nextUpdate = getNextUpdateFromVariable(col.def)
-
-        let condensedOrigins:
-            | Partial<
-                  Pick<
-                      OwidOrigin,
-                      | "attribution"
-                      | "attributionShort"
-                      | "description"
-                      | "urlDownload"
-                      | "urlMain"
-                  >
-              >[]
-            | undefined = origins?.map((origin) => {
-            const {
-                attribution,
-                attributionShort,
-                description,
-                citationFull,
-                urlDownload,
-                urlMain,
-                dateAccessed,
-            } = origin
-            return {
-                attribution,
-                attributionShort,
-                description,
-                urlDownload,
-                urlMain,
-                dateAccessed,
-                citationFull,
-            }
-        })
-
-        if (!condensedOrigins || condensedOrigins.length === 0) {
-            condensedOrigins = [
-                {
-                    attribution: sourceName,
-                    urlMain: sourceLink,
-                },
-            ]
-        }
 
         const def = col.def
 
--- a/functions/_common/metadataTools.ts
+++ b/functions/_common/metadataTools.ts
@@ -4,9 +4,8 @@ import { OwidTableSlugs, OwidOrigin } from "@ourworldindata/types"
 import {
     getLastUpdatedFromVariable,
     getNextUpdateFromVariable,
-    getCitationShort,
     getAttributionFragmentsFromVariable,
-    getCitationLong,
+    getIndicatorCitations,
     stripDetailOnDemandLinks,
 } from "@ourworldindata/utils"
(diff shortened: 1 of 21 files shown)


sophiamersmann:
* **#7101** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7101?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7101?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #7104, opened 2026-08-28
Move the Worker e2e suites to createTestHarness

rakyi:
> _Written by Anthropic Claude Opus 5 — @rakyi at the wheel._

Cloudflare now recommends [`createTestHarness()`](https://developers.cloudflare.com/workers/testing/test-harness/) over `unstable_startWorker()` for integration tests, so `setUpTestWorker` becomes `setUpTestHarness` and the e2e suites dispatch through `server.fetch()`. The helper is now lifecycle only: create at file scope, `beforeAll` starts, `afterAll` closes — the documented shape.

**This PR also contains a small production change**, and it is the more interesting half. `/api/search/cached-queries` re-wrapped Algolia's response and passed its headers on wholesale. The runtime decodes a compressed subrequest body but leaves `Content-Encoding` and `Content-Length` describing the encoded one, so those headers no longer matched the body we handed on — and `caches.default` stored the mismatch. It now drops both headers where it re-wraps. This is [workerd#6835](https://github.com/cloudflare/workerd/pull/6835), open since June and stalled on a CLA; our two lines can go when it ships.

Production is not affected and never was: `POST /api/search/cached-queries` against ourworldindata.org decodes to JSON on both a MISS and a HIT, because the edge cache handles the encoding correctly. It is the local Cache API emulation that returns such a body one gzip layer deeper than its headers claim, which is how the migration surfaced this at all. The fix is first in the branch so the suites are green at each commit.

Two things deliberately not done, both measured rather than assumed: the documented `afterEach` → `server.reset()` is not adopted (it recreates the server session, ~1s per test — 1.48s → 6.48s across five tests), and the old abandoned-startup guard is gone (the harness owns the runtime from creation, so `close()` disposes it even when the hook is abandoned).

### Commits

- `🔨🤖 stop the search proxy passing on stale encoding headers` — `functions/api/search/cached-queries.ts`, two `headers.delete` calls plus a comment.
- `✅🤖 move the e2e suites to createTestHarness` — adds `functions/test/setUpTestHarness.ts`, deletes `setUpTestWorker.ts`, updates `search.e2e.test.ts` and `grapher-config-r2.e2e.test.ts`.

### What the helper keeps and drops

Keeps the explicit 60s/30s hook budgets from #7099 — vitest's inherited 10s default was never sized for booting workerd (0.7–1.3s idle, over 10s under load).

Drops the promise-holding guard that fix also needed. That existed because `unstable_startWorker` assigned the module-level `worker` only after its await, so an abandoned hook left teardown with nothing to dispose. Verified with a forced 300ms startup budget (confirmed `Hook timed out in 300ms`), three runs each of the guarded and naive variants: zero leaked `.wrangler/tmp` dirs and zero stray workerd processes either way. The same experiment leaked two dirs per run on the old helper.

Drops the `workerFetch` wrapper and its `http://example.com` prefix, which dates to the suite's original commit: `dispatchFetch` required an absolute URL, while `server.fetch` resolves relative ones. Worker-visible URLs are now localhost with a per-run port. The only thing deriving from them is the cached-queries cache key (`url.origin + url.pathname`), which stays stable within a run.

### On the encoding fix

`encodeResponseBody: "manual"` — keeping the body encoded rather than letting the runtime decode it — was tried and does not help: the MISS came back clean but the cached HIT was still a layer deep (`1f 8b`, 6046 bytes vs 60653 decoded), because the local cache re-encodes on the way out whenever the response declares an encoding. It would also serve gzip to callers that never negotiated it.

Ruled out along the way: dispatch path (`server.fetch` vs direct worker dispatch both fail), stale local cache (fails with `.wrangler/state/v3/cache` cleared), and a pre-existing failure (master's helper passes 7/7 on the same code).

### Verification

`yarn typecheck`, `yarn oxlint --deny-warnings functions/`, `functions/api/search/cached-queries.test.ts` 10/10, and all three e2e suites 33/33 with the local cache cleared first. No leaked `.wrangler/tmp` dirs or workerd processes after runs.

### Not in this PR

The Postmark webhook e2e suite also moves onto this helper, but that file only exists on the `email-notifications` stack, so it rides there as a separate commit.

</details>

---
- <kbd>&nbsp;2&nbsp;</kbd> #6718
- <kbd>&nbsp;1&nbsp;</kbd> #7104 👈


rakyi:
I separated this work out of the email notifications changes, since I was getting conflicts etc. I thought the description written by Claude was pretty good this time, so I let it in. What do you think?

rakyi:
Merging this now to unblock my up-stack PR. Happy to get a review post-hoc.

PR #7110, opened 2026-08-28
🎉🤖 (featured-viz) show methods and sources on featured viz pages

sophiamersmann:
Featured viz pages get a metadata box inside the blue hero band, directly under the viz. The metadata comes from a JSON file that a bespoke bundle names through a new `metadataUrl`. It's fetched and validated server-side when the gdoc's attachments load, then handed to the page through `AttachmentsContext`.

The same `metadataUrl` is used inside the bespoke projects as well, so that we can later render a metadata modal inside a bespoke viz.

Example: http://staging-site-bespoke-metadata-debug/featured-viz/test-featured-viz-article

- `BespokeMetadata` describes a shared metadata schema for bespoke projects that closely follows indicator metadata.
- All metadata fields in `BespokeMetadata` are optional, so the metadata box can be as detailed or lightweight as makes sense.
- A file without a `title` gets "About this data" as its heading, none of the key data rows, and no "How to cite this data" toggle.
- The metadata box should be as wide as the viz, but the `widest` size clamps the box to the `wide` column, since `widest` is hard to read for text.
- **Fix.** `isStacked` is gone from `ExpandableToggle` and `IndicatorSources`. A run of toggles now shares one border through a `:has(+ …)` adjacency rule, so data pages and the grapher sources modal get their borders from CSS rather than from a prop the caller has to index correctly.


db/model/Gdoc/GdocFeaturedViz.ts
```diff
@@ -32,6 +44,62 @@ export class GdocFeaturedViz
         )
     }
 
+    override _loadSubclassAttachments = async (): Promise<void> => {
```
ikesau, line 47: There's a comment in `GdocBase.loadState` that should be updated for this (or removed, if we're no longer trying to exhaustively list its users)

packages/@ourworldindata/components/src/ExpandableToggle/ExpandableToggle.scss
```diff
@@ -83,7 +83,8 @@ details.ExpandableToggle {
     }
 }
 
-details.ExpandableToggle--stacked {
+// Rows in a run share one border between them
+details.ExpandableToggle:has(+ .ExpandableToggle, + .NonExpandable) {
```
ikesau, line 87: weird thing: via the vite dev server, this is showing up as 2 rules 🤷 

Image

packages/@ourworldindata/types/src/domainTypes/Various.ts
```diff
@@ -1,5 +1,15 @@
+import type { z } from "zod/mini"
 import { Nominal } from "../NominalType"
 
+/**
+ * Pins a zod schema to a hand-written interface. A schema declared
+ * `satisfies SchemaFor<T>` fails to compile if it misses a field of `T`, adds
+ * one, or gives one the wrong type.
+ */
+export type SchemaFor<T> = {
```
ikesau, line 9: I guess this is to not force zod to be used everywhere that `OwidLicense` and `OwidOrigin` is used?

ikesau (approved):

packages/@ourworldindata/types/src/domainTypes/Various.ts
sophiamersmann: yeah... `OwidOrigin` is such an important type that I felt bad converting it into the less readable zod schema (then I did the same for `OwidLicense` for consistency)

db/model/Gdoc/GdocFeaturedViz.ts
sophiamersmann: I deleted the comment. Let me know if you prefer to keep it

packages/@ourworldindata/components/src/ExpandableToggle/ExpandableToggle.scss
sophiamersmann: Apparently, `owid.entry.ts` loads both `grapher.scss` and `owid.scss`, and they both import the same set of SCSS files from the `components` package. In prod, the final CSS is de-duplicated, so this only happens in dev.

Fixing it would probably mean deleting the duplicated imports in `owid.scss`, but the catch is the cascade order.

Issue #7009, opened 2026-08-19
Make data updates standard, shareable content on /latest


mlbrgl:
**TLDR; we're closing the gap between data insights and data updates**

Implements points 2 and 4 of #6477 (richer announcement card design on /latest, and standalone announcement pages as sharing targets). Those two have to be scoped together: today the card layout *is* the page layout, so any card redesign propagates straight into the page.

## How it is today: motivation

- Data updates sit in the /latest feed as small text blocks: a title, a byline, a couple of lines of text, and a "Read more" toggle that expands in place. No image, no click-through, no page to link to.
- If a data update has a chart or image, it's buried inside the text and only appears once you expand it.
- There's no URL you can share. The announcement does get its own page, but nothing links to it and it's really just a preview for editors.
- The result: data updates read as background noise. They're a big share of what we publish to /latest, and they're not far from data insights in substance — but visually they look like footnotes next to them. It also makes the feed look a bit broken, rather than deliberately varied.

## What changes

<img width="1042" height="676" alt="Image" src="https://github.com/user-attachments/assets/a38d3034-767f-412e-b51e-c41ef8487d74" />

[figma](https://www.figma.com/design/WIPdAo70ntbEwHjCjhu7vI/Latest?node-id=656-1071&t=mIGCigtPQpFNJr9f-0)

- `card design tbc` Data updates get a normal card, like the rest of the feed: an image up front, a clear footprint, and a click-through to a page.
- `needs design` Their page becomes the real destination rather than an editor preview — so it needs its own design (similar to data insight pages), decoupled from the feed card.
- Each data update gets a shareable URL, with a decent preview when it's pasted into Slack, Bluesky, or anywhere else.

## What's untouched

- Authoring stays the same: it's still an announcement doc marked as a data update. No new content type to learn.
- The /latest content-type filter is unchanged.
- Data insights and articles are unaffected.
- Plain announcements and website upgrades keep today's look and behavior (click to expand in-place).

## What this costs us

- **The feed gets more uniform.** The small cards were breaking up a long run of near-identical image-and-text cards. Turning most of them into full cards removes most of that breathing room. That's a deliberate trade: core editorial work should carry proper weight.
- **Every data update now needs a usable image.** What's optional today becomes part of the card, so we need a plan for updates that don't have an obvious chart to show
  - proposed resolution: card is shown without image in the feed 
[...]

Issue #6477, opened 2026-05-11
Tracking: Announcements — open product/design questions


mlbrgl:
## Summary

The announcement gdoc content type predates the /latest feed (#6259), but has been substantially reworked as part of bringing it into that feed. Along the way, a handful of open design and product questions have surfaced. This issue tracks four related threads so we can scope and sequence them together rather than addressing each in isolation.

## 1. CTA vs non-CTA announcements

Announcements currently come in two shapes: **CTA** (a top-level `{.cta}` block with an empty body, rendering as excerpt + a single CTA button) and **non-CTA** (body-driven, often containing an inline `{.cta}` block). The shape is being revisited in #6464. Two paths:

- **Keep both shapes.** Status quo. Authors choose between a CTA-only card and a body-driven one; the rendering branches and the indexer branches accordingly.
- **[PREFERRED] Collapse to a single shape.** Pick one model and migrate. Removes an authoring choice, a render branch, and an indexer branch. Current CTA-announcement usage and the anticipated usage confirmed in #6464 point this way.

Tradeoff is authoring expressivity (path 1) vs. removing an unnecessary variation and its maintenance overhead (path 2). Worth resolving before the card design work in point 2 — no point designing two variants if one is on the way out.

## 2. Richer announcement card design on /latest

The current `LatestAnnouncementHit` is intentionally minimal. In an unfiltered feed, announcements (no thumbnail, condensed layout) break up what would otherwise be a wall of near-identical thumbnail-plus-text cards. Data updates currently inherit this minimal layout, which contributes to that "breathing room" effect — today they make up ~61% of announcements in the feed. The counter-pressure is that data updates are core editorial work and, content-wise, aren't far from data insights; making them visually recede may diminish them too much. Discussed in this Slack thread: https://owid.slack.com/archives/C0AKTASPHUM/p1774953697787739. Two paths:

- **One minimal card for all announcement subtypes.** Status quo. Preserves feed variety; data updates stay quiet.
- **[PREFERRED] Differentiate by subtype.** E.g. data updates get a richer card (chart thumbnail, larger footprint), generic announcements stay minimal. Gives core content more visual weight; reduces feed variety as the share of "rich" cards grows.

Tradeoff is feed variety (path 1) vs. giving core content appropriate visual weight (path 2).

## 3. Kicker field: free-form text vs constrained vocabulary

[...]

PR #7117, opened 2026-08-31
🎉 Give data updates their own card on /latest

mlbrgl:
> _Written by Claude Opus 5 and Claude Fable 5 — @mlbrgl at the wheel._

Part of #7009 (points 2 and 4 of #6477).

Data updates were rendered as plain expandable text in the /latest feed. They now get a proper card — chart thumbnail, title, byline, body, clickable as a whole — in the same family as the article and data insight cards, but on the page background rather than a blue fill.

The card has two states, driven by whether we know the reader is after data updates specifically:

- **Collapsed** (general feed): body clipped with a fade, links to the announcement page.
- **Expanded** (`?type=data-update`, or a direct `#slug` link — e.g. from the homepage announcements list or a newsletter): full body, and the card links straight to the data the update is about, saving a click.

That second trigger is the one worth a look in review: arriving from a link to a specific update expands it *in place in the general feed*, so the reader gets the whole update without losing the surrounding feed.

One change reaches beyond data updates: **hover is now the same across article, data insight and data update cards** — collapsed cards shift their background, titles never underline, and the bottom CTA on article and expanded data update cards underlines on hover — the article card's on the card's hover, the data update card's on its own (`latest-card-cta` takes an optional card selector).

### Files

- `site/latest/LatestDataUpdateHit.tsx` / `.scss` — new card. Modelled on `LatestDataInsightHit`: `LatestHitMetadata` row, then a `grid grid-cols-8` card with a `span-cols-3` thumbnail and `span-cols-5` content. Geometry measured identical to the data insight card (318px card, 278px image, 286px clipped content). Kept as its own component rather than merged with the data insight card: the data insight card is about to be redesigned with a different layout (full-width chart under the byline), so the two only look alike for now.
- `site/latest/LatestArticleHit.scss` — hover harmonisation: card background instead of title underline; stretched link now stops at the card rather than covering the metadata row; "Read the article" arrow moved into normal inline flow to sit at the same baseline offset as the data update CTA's.
- `site/latest/LatestHit.tsx` — dispatches `{ type: Announcement, latestType: "data-update" }` to the new card *before* the general announcement case. Topic updates, website upgrades and plain announcements are untouched and still render via `AnnouncementContent` / `ExpandableText`.
- `site/latest/LatestSearch.tsx` — `isExpanded = latestType === "data-update" || slug === autoExpandedSlug`.
- `site/latest/latestUtils.ts` — two new helpers, both tested:
  - `findCtaUrl` — the expanded card's destination. Every published data update ends on a `cta` block (`text × N, cta, image`, verified across all 14 in the DB), so this reads that block's url. Deliberately **no** inline-link fallback: with no CTA it would sooner pick up a source citation than a real destination. Callers fall back to the announcement page instead, since nothing enforces the CTA at publish time.
  - `findThumbnailImageBlock` — replaces three duplicated copies of "find the first image block" (this card, `LatestDataInsightHit`, `LatestDataInsights` carousel).

### Deliberate choices

- **Desktop chart render.** The thumbnail passes `filename`, never `smallFilename`, matching `LatestDataInsightHit` — it's the homepage carousel and `getThumbnailUrl` that prefer the square small render. No data update carries a small variant today, but the card would ignore one if it appeared: the chart stands in for the update rather than being the point of the card.
- **`useIsLikelyBaked` now applies to data updates.** They were never links before, so they didn't need it; now a freshly-published one is HEAD-probed before its card shows, same as articles and data insights.
- **The CTA keeps its arrow but is not an `<a>`.** `shouldRenderLinks={false}` avoids nesting an anchor inside the card's anchor; the `.cta` class supplies the arrow regardless, so it still reads as the affordance it is. The card overrides the global `.cta` red with `$blue-90`.
- **Hover swaps the background to `$blue-5`** on collapsed cards, matching the pattern on the other cards; the title never underlines. Expanded cards have no hover background.
- **A 1px `$blue-10` stroke** gives the thumbnail its bounds, since the white card has no fill to do that.
- **The CTA is `body-3-bold` with no underline at rest**, in place of the global `.cta`'s `body-2-semibold`, a size larger; it picks up the underline on hover. These rules are scoped to `.latest-data-update-hit__blocks .cta` because data insight bodies carry legacy CTAs of their own that shouldn't pick up the treatment.
- **The CTA arrow sits in normal inline flow**, not a flex container: CTA labels are long enough to wrap, and a flex container would centre the arrow against the whole text block instead of trailing the last word.

### Gradient mixin

`fade-out-overlay($color, $start: 50%)` in `packages/@ourworldindata/components/src/styles/mixins.scss` replaces five hand-rolled copies of the same fade: the data insight card and its hover, the homepage carousel and its hover, and the homepage announcements list. Rendering is unchanged at all five call sites, though the compiled CSS text differs: the transparent endpoint is now the bare `transparent` keyword rather than a colour-matched `rgba(…, 0)` literal. Those literals were a workaround for pre-2014 browsers that interpolated gradients non-premultiplied; every browser we support has interpolated premultiplied for a decade, and `LatestTopicFacets.scss` already fades to bare `transparent` on this very page. One of the five had drifted to an rgba that didn't match the colour it faded into — invisible, but the kind of thing the mixin makes impossible.

### Testing

- `yarn typecheck` clean, `yarn testLintChanged` clean, `yarn testFormatChanged` clean, full unit suite passing (2376).
- Hover states measured in the browser with real pointer moves: data update card `rgb(240,244,250)` with matching gradient and no title underline; article card `rgb(219,229,240)`; metadata row hover leaves the article card untinted.
- `site/latest/latestUtils.test.ts` covers `findCtaUrl`: the real body shape, multiple CTAs, a body with only inline links (expects `undefined`), and an empty body.
- Verified against the running dev server that expanded cards resolve to their CTA targets (a data page, a dataset search, a chart collection) and collapsed cards to `/<slug>`.

</details>


mlbrgl:
* **#7175** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7215](https://github.com/owid/owid-grapher/pull/7215) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7217](https://github.com/owid/owid-grapher/pull/7217) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7117** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7117?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7117?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

site/latest/LatestDataUpdateHit.tsx
```diff
@@ -0,0 +1,114 @@
...
+    selectedTopic?: string
+    position: number
+    isExpanded: boolean
+}) => {
+    const { analytics } = useLatestContext()
+    const announcementHref = getPrefixedGdocPath("", {
+        slug: hit.slug,
+        content: { type: OwidGdocType.Announcement },
+    })
+    const isLikelyBaked = useIsLikelyBaked(announcementHref, hit.date)
```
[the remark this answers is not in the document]
mlbrgl: > _Written by Anthropic Claude Fable 5 — @mlbrgl at the wheel._

Good catch — the effect indeed never re-ran once the gated card mounted. Fixed in d8514eba10, though at the root rather than by retrying: the probes made the feed's composition unknowable at first render, so any one-shot hash handling could fire too early (and a card resolving late could still displace an already-scrolled target). The feed now counts as loading until the first page's fresh probes have settled (`useAreFreshProbesSettled`, a `useQueries` over the same query keys the cards read, so the fetches dedupe). It then renders in one commit with its composition final, and the existing hash effect always finds its anchor. First page only, deliberately — the feed is chronological, so fresh hits are always among the newest, and "load more" must never re-trigger the skeleton.

site/latest/LatestSearch.tsx
```diff
@@ -195,7 +215,7 @@ export const LatestSearch = ({
                             hit={hit}
                             selectedTopic={topics[0]}
                             position={i + 3}
-                            shouldAutoExpand={hit.slug === autoExpandedSlug}
+                            isExpanded={isExpanded(hit.slug)}
```
ikesau, line 215: Nit pick: I kind of prefer `shouldAutoExpand` as it keeps it sounding like a suggestion (which it is, because articles can't expand)

site/latest/latestHooks.ts
```diff
@@ -54,11 +64,83 @@ export function useIsLikelyBaked(
...
+ * Whether every fresh hit on the given page has a settled bake probe — the
+ * signal `LatestSearch` extends its skeleton on (see "Bake probes" above).
+ * Runs the probes itself so they start before any card mounts; the cards'
+ * own `useIsLikelyBaked` calls then read them from the cache.
+ *
+ * Pass the first page only. The feed is chronological, so fresh hits are
+ * always among the newest — and "load more" pages must never re-trigger the
+ * skeleton.
+ */
+export function useAreFreshProbesSettled(
```
ikesau, line 133: This is definitely quite an exhaustive and fiddly way of doing it. I think if it were me, I'd just bake the current timestamp and any URLs published within the last 20 minutes into an object in the footer.

Then, if Algolia returns a record with a recent `publishedAt`, check if it's in the `recentlyBakedUrls` list, and filter it if not (because we'd then know it must be a post from the _upcoming_ bake)

Probably not worth a rewrite at this point, unless that idea hadn't occurred to you and you like the simplicity of it also. 🙂

site/latest/LatestSearch.tsx
```diff
@@ -139,6 +152,13 @@ export const LatestSearch = ({
         // needlessly.
     }, [isLoading, hits.length])
 
+    // Announcements render expanded when we know the reader is after this
+    // content in particular: they filtered for data updates, or followed a
+    // link straight to one card. It's a hard override, not a default — the
+    // card renders without a Read more toggle and can't be collapsed.
+    const isExpanded = (slug: string) =>
+        latestType === "data-update" || slug === autoExpandedSlug
```
ikesau: Theoretically:
1. A new data update is published
2. It gets indexed
3. latest page would filter it out until it's baked
4. But the auto-expanded item _could_ render on the latest page when the feed is filtered or the item is deep-linked to

Point is, that the probe system stops an item from rendering when it would actually be fine if it did.

Though, if the user then turned off the content-type filter, data updates would go back to not being auto-expanded and we'd thus have to filter it out.

So I guess the way we have it is fine 🙂

packages/@ourworldindata/components/src/styles/mixins.scss
```diff
@@ -579,3 +579,56 @@
         border-radius: 4px;
     }
 }
+
+// Vertical fade used to soften the bottom edge of clipped card content
+// (a data insight or data update card whose body overflows). Apply to an
+// absolutely positioned overlay covering the clipped area; $color must match
+// the background it fades into, including on hover.
+@mixin fade-out-overlay($color, $start: 50%) {
```
ikesau, line 587: Nice 👌

packages/@ourworldindata/components/src/styles/mixins.scss
```diff
@@ -579,3 +579,56 @@
...
+// behavior.
+@mixin above-stretched-link {
+    position: relative;
+    z-index: 2;
+}
+
+// Underline on CTA hover, or on the supplied card selector's hover.
+// Keep the arrow inline so it follows the last word when the label wraps.
+// Colour and font weight come from the caller.
+@mixin latest-card-cta($card-selector: null) {
```
ikesau, line 612: I don't quite understand this one. Can you explain its utility to me?

ikesau (approved):
LGTM! just some clarifications and comments mostly.

packages/@ourworldindata/components/src/styles/mixins.scss
mlbrgl: it's to harmonize CTA-like links that are found at the bottom of data updates, articles (and later in the stack data insights cards)

site/latest/LatestSearch.tsx
mlbrgl: yea that "should" got me confused through the build when it stopped representing something the component can object to and became a "hard override". So in that sense "should" doesn't apply, but it's true it's got a bit of a "isExpandedWhenApplicable" flavour

site/latest/LatestSearch.tsx
mlbrgl: > But the auto-expanded item could render on the latest page when the feed is filtered

I agree! And it does currently (isExpanded is true when latestType === data-update). I also though it'd be a shame to hide a perfectly fine data update that has no business with its standalone page.

What I didn't catch though is the second case: when a DU is deep-linked (on the unfiltered feed) and still hasn't finished baking, then technically you'd land at the top of the feed. Conditions for it to happen are limited: it would require someone to get hold of that anchored link during the baking window. The only scenario in which that makes sense is sending a newsletter advertising such a link (before the bake finishes). I wanted to cover that use case initially, no matter how unlikely, so thanks for pointing it out!

mlbrgl: And yes, going back and forth using the filters removes the hash and would hide a transient DU. But that's yet another "if" in the tree, which is not worth breaking the "URL is driving the state" for.

PR #7119, opened 2026-08-31
🔨🤖 (featured-viz) a metadata modal for bespoke vizzes

sophiamersmann:

Adds a "About this data" modal for bespoke vizzes. It reuses the shared `BespokeMetadata*` components, so the content matches the in-article metadata box; the styling follows grapher's sources modal, since the modal sits inside a grapher-styled frame rather than on a site page.

A featured viz page shows the same metadata in a box under its hero, so the page switches the modal off there. It stamps `hideMetadataModal` onto the hero block's config next to `urlSync`, and `BespokeMetadataProvider` reads that flag from the embed config and withholds the metadata from the context when it is on.

Example: http://staging-site-bespoke-metadata-debug/causes-of-death (find the Causes of death viz, and click on 'Learn more about this data' in the footer)

<details><summary>Screenshots</summary>

image
image

</details>


bespoke/components/MetadataModal/MetadataModal.scss
```diff
@@ -0,0 +1,251 @@
+.metadata-modal {
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Two parts to this. The stylesheet was genuinely unimported on this branch, though not for the reason given. `OverlayHeader` and `ExpandableToggle` are both imported, along with the modal's other composed partials, in `bespoke/projects/causes-of-death/src/index.scss`, on the branch above this one, together with the commit that wires the provider up. Nothing renders unstyled here either. No project mounts the modal on this branch, because `causes-of-death` has no `metadataUrl` in the registry yet, so the consumer waits on the branch above until the real metadata file is published.

The underlying point was still worth acting on. The modal composes ten package partials, and asking every project to list them and keep that list in step is a trap for the next author. So is the order. The modal's citation and visited-link overrides win on source order rather than specificity, so the packages have to come first. `MetadataModal.scss` now imports them itself, which gets a project down to one import and makes the order structural. The emitted CSS is unchanged, same 25751 rules, with every package rule still ahead of the overrides.

site/gdocs/pages/FeaturedViz.tsx
```diff
@@ -26,6 +26,11 @@ import { buildGdocCitation } from "../utils.js"
 const CITATION_DESCRIPTION =
     "Our articles and data visualizations rely on work from many different people and organizations. When citing this page, please also cite the underlying data sources. This page can be cited as:"
 
+const HERO_CONFIG_OVERRIDES = {
+    urlSync: "true",
+    hideMetadataModal: "true",
```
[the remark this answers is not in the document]
sophiamersmann: > _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Fixed. The gap is a bit wider than a bake-time failure. The page also omits the box whenever the bundle has no `metadataUrl` in the registry, which is the routine case rather than an error path, and on this branch that is every entry. The page now stamps `hideMetadataModal` only when it renders the box.

.claude/skills/create-bespoke-viz/SKILL.md
```diff
@@ -19,7 +19,7 @@ Two things the readme doesn't cover:
...
... Every project has a `src/core/config.ts` exporting its config interface plus a `parseConfig(raw: Record<string, string>)` (see `food-trade/src/core/config.ts`), built from the shared value parsers in `bespoke/helpers/config.ts` — `parseBoolean`, `parseNumber`, `parseInteger`, `parseEnum(value, allowed)`, and the `VariantProps<Config>` type. Don't re-roll those; do keep project-specific parsers local. Readers author these values by hand, so a malformed one degrades to `undefined` rather than throwing. For an enum, declare the runtime list and derive the type from it (`const FLOWS = [...] as const; type Flow = (typeof FLOWS)[number]`) so `parseEnum` and the type can't drift apart. Support the conventional keys where they make sense: `title`, `subtitle` (override the generated ones), `hideControls`, and an entity default (`country` or `region`) that accepts the sentinel `"userLocation"` (resolve it with the `useResolveUserLocation` hook). Don't parse `urlSync` or `hideMetadataModal`: they describe the embedding page rather than the viz, so the mount reads them with `parseEmbedConfig` and the `EmbedConfigProvider` in the variant's provider stack hands them to the code that needs them.
```
sophiamersmann, line 22: No need to review this - I rewrote/decluttered the skill in the [next PR](https://github.com/owid/owid-grapher/pull/7252)

bespoke/projects/causes-of-death/src/core/config.ts
```diff
@@ -6,7 +6,6 @@ export interface CausesOfDeathConfig {
     ageGroup?: string
     year?: number
     hideControls?: boolean
-    urlSync?: boolean
```
sophiamersmann, line 9: Clean up after a refactor I did earlier that moved `urlSync` out of the project-specific config

bespoke/readme.md
```diff
@@ -143,17 +143,22 @@ Note that every slide of a key insights block is in the DOM from page load, not
 
 A bespoke component can also be the subject of its own page, rather than one
 figure inside an article. Those are gdocs of type `featured-viz`, published at
-`/featured-viz/<slug>`, and they behave differently in two ways:
+`/featured-viz/<slug>`, and they behave differently in three ways:
```
sophiamersmann, line 146: The readme is also decluttered in the [next PR](https://github.com/owid/owid-grapher/pull/7252)

site/gdocs/pages/FeaturedViz.tsx
```diff
@@ -174,8 +178,14 @@ function splitFeaturedVizBody(
...
     return {
         before: body.slice(0, heroIndex),
-        // The page URL tracks the featured viz, ignoring the authored config
-        hero: { ...hero, config: { ...hero.config, urlSync: "true" } },
+        hero: {
+            ...hero,
+            config: {
+                ...hero.config,
+                urlSync: "true",
+                ...(hasMetadataBox && { hideMetadataModal: "true" }),
```
ikesau, line 186: Small smell here. There could be a typo here and we wouldn't catch it.

Could we have something like this?

```tsx
function serializeEmbedConfig(
      flags: Partial<EmbedConfig>
  ): Record<string, string> {
      return Object.fromEntries(
          Object.entries(flags).map(([key, value]) => [key, String(value)])
      )
  }

// ...

config: {
      ...hero.config,
      ...serializeEmbedConfig({ urlSync: true, hideMetadataModal: hasMetadataBox }),
  }
```

ikesau (approved):

site/gdocs/pages/FeaturedViz.tsx
sophiamersmann: Good idea!

PR #7126, opened 2026-08-31
🐝🤖 make Vite configs native-loader compatible

rakyi:
This fixes Vite warnings in the console.

Vite plans to make native config loading the default, but this config
graph relied on bundled transforms for CommonJS-classified TypeScript,
TypeScript enums, emitted .js import resolution, and __dirname.

Move the config dependencies and Vitest config to explicit ESM, preserve
compatibility wrappers for application imports, and rewrite emitted
extensions so native source loading and compiled output both work.


rakyi:
@codex review

settings/clientSettings.ts
marcelgerber, line 1: I don't think we'll need any of these "placeholder" `.ts` files that just re-export everything, since all our imports are coming from our own monorepo and never from outside.

Just get rid of it, and if any imports break (in Vite or in Node), just update them. That'll also make the GH diff cleaner.

marcelgerber (approved):
Yup, can do this.

At first, I was thinking that it might just be better to switch our main package.json over to `type: module`, which would also resolve this warning. But I know that I always ran into unexpected issues when I did just that - I think it was mostly to do with packages like Algolia, which don't provide ESM exports in the correct way.
Think we should still do that switchover at some point, but that also shouldn't hold us off from resolving this warning _now_. So, feel free to merge!

settings/clientSettings.ts
rakyi: I pushed the change in a new commit. It's a lot of files. Is that what you meant?

rakyi:
Yeah, I'd love if we could fully move to ESM, but I also remembered that it's not easy at all.

settings/clientSettings.ts
marcelgerber: Ah right, I only thought about `findBaseDir` and `loadDotenv`, which are imported only rarely, and not about all of the imports of `clientSettings`. anyhow, I still think this is better, yes!

rakyi:
@codex review

Issue #7011, opened 2026-08-19
Remove CTA announcements from the CMS


mlbrgl:
Implementation follow-up to #6464, where the decision was made: remove the "CTA announcement" shape from the CMS. Also point 1 of the tracking issue #6477.

## Decision

Announcements come in two shapes today, distinguished by whether the gdoc has a **top-level** `{.cta}` block:

- **Non-CTA** — body-driven, with a `{.cta}` block placed *inside* the body if a call-to-action is needed. 54 published.
- **CTA** — empty body, a top-level `{.cta}` block, rendering as excerpt + a single button. 1 published, pointing at an external partner site.

The CTA shape goes. Any call-to-action lives inside the body from now on. Sign-off: [Charlie](https://github.com/owid/owid-grapher/issues/6464#issuecomment-4404900012) ("I honestly forgot the 'CTA outside the body' type even existed — I don't really use it") and [Edouard](https://github.com/owid/owid-grapher/issues/6464#issuecomment-4440280648).

## Scope — what is being removed

Only the **top-level** `content.cta` field on announcement gdocs. The body-level `{.cta}` block (block `type: "cta"`, parsed in `rawToEnriched.ts`) is unaffected and stays — it's how every announcement will express a call-to-action.

Branches that collapse once the shape is gone:

- `GdocAnnouncement._validateSubclass` — the mutual-exclusion rule ("a top-level `{.cta}` block must have an empty body") and the CTA url/text validation.
- `GdocAnnouncement.typeSpecificUrls` — the top-level CTA url no longer needs to be surfaced as a gdoc link.
- `GdocHomepage.loadAnnouncementLinks` — exists solely to resolve linked charts/documents from announcement CTA urls, since those links aren't caught by the normal `loadLinkedDocuments`/`loadLinkedCharts` path.
- `HomepageIntro.HomepageAnnouncement` — the `useLinkedChart`/`useLinkedDocument` resolution of `announcement.cta?.url`, and the CTA-vs-"Read more" link label.
- `AnnouncementContent` — the excerpt+button branch, leaving only the body rendering path.
- `pagesChronological` (Algolia indexer) — the CTA branch that skips loading attachments and image filenames.
- `OwidGdocAnnouncementContent.cta` and the homepage announcement interface's `cta` field.

## Migration

- The single published CTA announcement needs rewriting as a regular announcement with the CTA inside the body. This has to happen before the field is removed, otherwise the announcement renders as a title and excerpt with no way through to the partner site.
[...]

Issue #6464, opened 2026-05-08
Evaluate removing CTA announcements from the CMS (low usage)


mlbrgl:
## Context

Announcements (`type = 'announcement'` in `posts_gdocs`) come in two shapes, distinguished by whether the gdoc has a top-level `{.cta}` block. The two are mutually exclusive — validated in [GdocAnnouncement._validateSubclass](db/model/Gdoc/GdocAnnouncement.ts#L27-L48):

> An announcement with a top-level `{.cta}` block must have an empty body.

### Reminder — the two shapes (anonymized real examples)

<table>
<tr><th>Non-CTA announcement</th><th>CTA announcement</th></tr>
<tr><td>

```
title: We've updated our chart on consumer price trends
kicker: Data update
excerpt: Explore the latest figures from our updated dataset.

[+body]
A short paragraph describing what's new in the data.

Another paragraph with context and comparisons.

{.cta}
text: Explore the data
url: https://example.org/our-chart
{}
[]
```

</td><td>

```
title: A new partnership with an external research programme
kicker: Announcement
excerpt: A short paragraph explaining the news and the partner.

{.cta}
text: Learn more on the partner website
url: https://example.org/partner-programme
{}

[+body]
[]
```

</td></tr>
<tr><td>

- Body carries the article. A `{.cta}` block can be placed inside the body if needed.
- Standalone `/latest/<slug>` page renders the full body.
- On the homepage, clicking the announcement leads to the **`/latest` page** ([HomepageIntro.tsx:160-166](site/gdocs/components/HomepageIntro.tsx#L160-L166)).

</td><td>

- Body is empty; a top-level `{.cta}` block lives outside it.
- Standalone `/latest/<slug>` page renders excerpt + a single CTA button.
- On the homepage, clicking the announcement leads **directly to the CTA target** ([HomepageIntro.tsx:160-166](site/gdocs/components/HomepageIntro.tsx#L160-L166)).

</td></tr>
</table>

## Usage data (production DB, May 2026)

Since the feature launched in October 2025:

| | CTA | Non-CTA |
|---|---:|---:|
| Published announcements | 1 | 54 |

[The single CTA announcement in production](https://admin.owid.io/admin/gdocs/1qQqvByR-2TRuV_h6rIYisHGF96I0VFq5RRrPvVpngOI/preview) points to an external partner website.

## Proposal

[...]

PR #7135, opened 2026-08-31
🎉 Give data updates a real standalone page

### LINKED ISSUE #6477: Tracking: Announcements — open product/design questions

### LINKED ISSUE #7009: Make data updates standard, shareable content on /latest

mlbrgl:
> _Written by Claude Opus 5 — @mlbrgl at the wheel._

Closes #7009. Closes #6477 — this is the last of that tracking issue's four threads: point 1 was settled in #6464 / #7011, point 3 was settled on the constrained `kicker`, and point 2 lands in #7117, below this PR in the stack.

Announcement pages were a placeholder: the /latest feed card on a white slab, no way back to the feed, no copy link for readers. They now use the data insight page's chrome and design — breadcrumb, dateline, big title, author byline, footer with related topics and a copy link, and an "Our latest …" carousel of announcements of the same kind.

Two things worth a reviewer's attention:

- **The carousel deliberately buckets kickers in JS, not SQL.** Kickers are only canonical slugs (`data-update`) since `GdocAnnouncement` started validating them. Today 38 of 56 data updates and 6 of 7 website upgrades still carry prose variants (`Data update`). `WHERE content->>'$.kicker' = …` would silently skip those, leaving a website upgrade page with a one-card carousel and no error to notice. So we read every published announcement's id + kicker (77 rows, no bodies), bucket through `deriveAnnouncementLatestType`, then fetch content for the seven we keep.
- **URLs are unchanged** — announcements still bake to `/<slug>`. Per-kind namespaces (`/data-updates/<slug>`, …) are a follow-up PR: they change `getPrefixedGdocPath`'s inputs, need the kicker in the pages Algolia index, and make a kicker edit a URL change that `createRedirectForSlugChangeIfNeeded` currently wouldn't catch.

### Data flow

- `getLatestAnnouncements(knex, latestType)` in `GdocFactory.ts` mirrors `getLatestDataInsights`, in two steps (ids+kickers, then content+image metadata for the kept seven). Both now share `LATEST_CAROUSEL_SIZE`.
- Attached via `GdocBase.latestAnnouncements` → `GdocAnnouncement._loadSubclassAttachments` (admin preview) and `SiteBaker.bakeGDocPosts` (memoized per kind, since one fetch serves every page of that kind) → `AttachmentsContext` → the page.
- `extractGdocPageData` whitelists which attachments reach the client. Announcements weren't in it, so the carousel rendered server-side and then **vanished on hydration**; caught by screenshotting the live page. Added a `checkIsAnnouncement` branch.
- `deriveAnnouncementLatestType` moves from `site/latest/latestUtils.ts` to `@ourworldindata/utils` so `db/` can use it without importing `site/` (layer violation). It lands in `Util.ts` beside the `checkIsAnnouncement` / `checkIsDataInsight` family and the `slugify` it calls; its unit tests move into `Util.test.ts`.

### Shared components (extracted from the data insight page, not duplicated)

`StandalonePostBody` (+ `STANDALONE_POST_GRID_CLASSES`), `LatestBreadcrumb` (label from `latestTypeLabelPlural`, href from `buildLatestPagePath`), `CopyLinkButton` (took a site-relative path instead of hardcoding `/data-insights/`), `RelatedTopicsList`, `LatestCarouselSection`, and `LatestCarousel` — generalized from `LatestDataInsights` over a content-agnostic `LatestCarouselItem`, with mappers in `latestCarouselItems.ts`. SCSS follows, with the `latest-data-insights` prefix renamed to `latest-carousel` and full class names written out instead of `&--prev` concatenation.

### Verification

- Rendered the data insight page before and after and diffed the HTML: identical apart from the class renames. That diff caught the `<h1>` losing `display-3-semibold`, now restored. One intended copy change remains: the carousel's small-screen "see all" card on the DI page reads "See all Data Insights" rather than "See all our Data Insights", since the section now passes one string to both the header link and the card.
- Baked one page of each announcement kind through `yarn bakeGdocPosts` and screenshotted them plus the DI page in headless Chrome. All three get a 6-card carousel — the website upgrade page included, which is the legacy-kicker folding paying off.
- No DB test for `getLatestAnnouncements`: the kicker-bucketing rationale is documented in the function's docblock instead. `deriveAnnouncementLatestType` itself is unit-tested in `Util.test.ts`.
- `yarn typecheck`, `yarn test run` (2373 passed), `make dbtest`, lint and format all pass.

### Dead code removed

`DataInsightBody`'s `anchor` / `shouldLinkTitle` / `shouldHideYearInDateline` props (no callers), `.data-insight-heading-link`, `.data-insight-body--has-tags`, `.data-insight-body-container`, `.latest-data-insights__card-date` (no styles or no markup), `LatestDataInsight.index` (never read — the carousel keys off array position, and its only writer was the map this PR replaces), and `AnnouncementContent`'s `isStandalone` prop — that component is now feed-only, so the admin preview shows the reader page rather than a feed card. `id="copy-link-button"` is preserved in case external analytics hangs off it.

</details>


mlbrgl:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7135?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7175** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7215](https://github.com/owid/owid-grapher/pull/7215) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7217](https://github.com/owid/owid-grapher/pull/7217) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7135** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7135?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

site/gdocs/components/StandalonePostBody.tsx
```diff
@@ -0,0 +1,75 @@
...
+                <DataInsightDateline
+                    className="standalone-post-body__dateline"
+                    publishedAt={publishedAt ? new Date(publishedAt) : null}
+                    formatOptions={{
+                        year: "numeric",
+                        month: "long",
+                        day: "2-digit",
+                    }}
+                />
+                <h1 className="body-1-bold">{title}</h1>
```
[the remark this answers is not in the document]
mlbrgl: > _Written by Claude Opus 5 — @mlbrgl at the wheel._

This is intentional. The smaller title was requested in design review: the standalone page is meant to read like the expanded feed card rather than a display-headed article page. Data insight and announcement pages both follow that design, which is why the shared component keeps `body-1-bold`.

site/gdocs/components/LatestCarousel.scss
```diff
@@ -86,24 +86,48 @@ html:not(.js-enabled) .latest-data-insights__card,
...
+    .latest-carousel__card--see-all,
+    .latest-carousel__dot--see-all {
         display: none;
     }
 }
 
 // Sizing flex child picture elements is finnicky on Safari.
 // Make sure you test all 3 browsers if you're making changes here.
-.latest-data-insights__card-left {
+.latest-carousel__card-left {
```
ikesau, line 130: I've got a scrollbar showing up on firefox, narrow screen.

Image

packages/@ourworldindata/utils/src/Util.ts
```diff
@@ -2243,6 +2252,29 @@ export function checkIsDataInsight(
...
+
+/**
+ * Resolve an Announcement's LatestType from its kicker. Slugifies so
+ * case/spacing variants ("Data Update", "Data update") still map to the
+ * canonical slug. Anything else — missing, blank, or unrecognized — falls
+ * back to "announcement". The save-side validator
+ * (GdocAnnouncement._validateSubclass) only accepts the canonical slugs,
+ * so any edit forces legacy values into shape.
+ */
+export function deriveAnnouncementLatestType(
```
ikesau, line 2338: It will be really nice to fix this with a gdoc migration.

site/gdocs/components/LatestCarousel.tsx
```diff
@@ -229,39 +231,30 @@ const DataInsightCard = memo(function DataInsightCard({
...
                     </p>
-                    <div className="latest-data-insights__card-body">
+                    {/* Clipped under a fade with no "Continue reading" — the
+                        card is one link, the way the feed's condensed cards
+                        are; the fade is the affordance. */}
+                    <div className="latest-carousel__card-body">
                         <ArticleBlocks
                             blocks={otherBlocks}
                             containerType="data-insight"
                             shouldRenderLinks={false}
```
ikesau, line 255: This param isn't making its way into lists, so `we-built-four-interactive-tools-that-give-you-new-ways-to-explore-key-datasets` is breaking the carousel hydration e.g. http://staging-site-data-update-standalone-page/its-now-easier-to-see-which-countries-make-up-world-regions-such-as-east-asia-and-pacific-in-different-datasets

baker/SiteBaker.tsx
```diff
@@ -786,6 +797,30 @@ export class SiteBaker {
...
                     await publishedGdoc.loadAndClearLinkedCallouts(knex)
 
+                    if (publishedGdoc instanceof GdocAnnouncement) {
+                        const latestType = deriveAnnouncementLatestType(
+                            publishedGdoc.content.kicker
+                        )
+                        let latestOfType =
+                            latestAnnouncementsByType.get(latestType)
+                        if (!latestOfType) {
+                            latestOfType = await getLatestAnnouncements(
```
ikesau, line 807: Was there a reason you did this outside of the prefetched attachments system? I could imagine something like:

```ts
const latestOfType =
      this._prefetchedAttachmentsCache.latestAnnouncements[latestType]

  publishedGdoc.latestAnnouncements = latestOfType.announcements
```

Your code isn't wrong from what I can tell, it's just a different way of doing it in a system that we've tried to standardize.

ikesau (approved):
Nice! Just a few small comments/suggestions

PR #7136, opened 2026-09-01
Rework the data-download readme

Marigold:
## 👨 Summary

Rework data download readme following pablo's suggestions. The best way to see what's going on is to check side-by-side comparison [here](https://claude.ai/code/artifact/772f26df-08f3-40cf-8285-277ffd7f434c).

The biggest difference is that we show **much more information in Sources section**. It's at the bottom and could be useful for LLMs, but we can revert it.

I haven't changed `.metadata.json` or added Excel format (created a separate issue from it).

**Before/after review of the readme, rendered:** [https://claude.ai/code/artifact/772f26df-08f3-40cf-8285-277ffd7f434c](https://claude.ai/code/artifact/772f26df-08f3-40cf-8285-277ffd7f434c)

Brings the chart data-download `readme.md` up to the same format as the MDIM complete-dataset package (owid/etl#6769), following [Pablo's review](https://github.com/owid/etl/issues/6668#issuecomment-5397836871). Both files are the same document produced by two codebases; the ETL side has already landed these changes, so this is the half that reaches every chart download.

What a reader sees change:

- **Per-indicator headings sit a level deeper.** An indicator was `##` — a sibling of the `## Detailed information …` that introduces it — so the second half of the readme was flat.
- **The full citation is gone.** It restated producer names the Sources section gives in far more detail; on `per-capita-energy-stacked` it was a fifth of the file. `metadata.json` still carries `citationLong`, and each source's own requested citation is now in the readme verbatim.
- **Sources are listed once for the document, at the end** instead of repeated inside every indicator's block — 24 blocks describing 3 sources on that chart — and each entry now carries producer, publication date, direct download link, license and citation alongside the retrieval date and URL it was limited to. The list sits after the indicators: it is reference material a reader arrives at from a citation, not something to read past on the way to the column they opened the file for.
- **`Source:` moves up** with `Last updated` / `Date range` / `Unit`. It used to print immediately after the full citation, reading as a trailing fragment of it; with the full citation gone it would have sat next to the short one looking like a duplicate.
- **"Next update" → "Next expected update"**, since the date is the producer's stated period added to our last update.
- **"How is this data described by its producer - HYDE (2023); Gapminder (2022); UN WPP (2024); UN FAO (2024)?" → "How this data is described by its producers".**
- **"About the data" → "How we process data at Our World in Data"** — what the section actually says — and the sub-heading of that name inside it is gone.
- **The CSV columns are a bulleted list.**

Net effect on size cuts both ways, and that's expected: a multi-column download shrinks (`per-capita-energy-stacked` 21 kB → 13 kB) because the repetition goes, while a single-column one grows (`population-density` 6 kB → 12 kB) because six sources gain descriptions, licenses and citations they never had.

Worth a look in review: **the two readme templates are merged into one.** They were separately maintained copies of the same text, which is how only the multi-column one came to tell readers how we process data. They now differ only where they genuinely differ — whether a download date is stated, how the data columns are described, and the tolerance-column wording.

### Files

- `functions/_common/readmeTools.ts`
    - `getCitationLines` — `#### How to cite this data` holding only the short citation; the `#### In-line citation` / `#### Full citation` sub-headings and `citationLong` are dropped.
    - `getDescriptionLines` — `####` headings; the `attribution` parameter is gone with the producer names it fed into the heading.
    - `getKeyDataLines` — "Next expected update".
    - `getSources(def)` (per-indicator) replaced by `collectUniqueSources(columns)` + `getSourcesSection(columns)`, emitted after the per-indicator sections. Dedupe key is `[label, dataPublishedBy, retrievedOn, retrievedFrom]`, not the label alone — the same producer retrieved twice on different dates is two rows a reader may need.
    - `columnReadmeText` — `###` heading, `Source:` moved above the citation, per-indicator sources dropped.
    - `constructReadme` — one template instead of two; `isSingleColumn` now only selects the download-date sentence, the data-column bullet, the tolerance paragraph and the `Detailed information about …` heading.
- `packages/@ourworldindata/utils/src/metadataHelpers.ts` — `prepareOriginForDisplay` passes `producer`, `datePublished`, `urlDownload`, `license` through.
- `packages/@ourworldindata/types/src/gdocTypes/Datapage.ts` — those four as optional fields on `DisplaySource`. The Sources UIs (`IndicatorSources`, `SourcesModal`, `MetadataSection`, `IndicatorMetadataBox`) read named fields, so they ignore them; nothing else consumes the shape.

### Testing

`yarn typecheck`, `yarn testLintChanged`, `yarn testFormatChanged` clean. `yarn test run` — the four `prepareSourcesForDisplay` cases that assert an exact object needed the new `producer` field added to their expectations; the rest of the 2,492 tests pass untouched.

One existing assertion was passing vacuously and is fixed: it searched for `## Sources` in a readme whose single deduplicated source is titled `## Source`, so `indexOf` returned -1 and the ordering check held whatever the order was. It now matches the heading lines.

`functions/_common/readmeTools.test.ts` gains three tests: the full heading list of a single-indicator readme, the dropped full citation plus the softened update date plus the `Source:` ordering, and one source rendered once with every field when two indicators share it. `metadataHelpers.test.ts` gains one for the passed-through fields.

Checked against production rather than only synthetic tables: a throwaway script fed the real `population-density` and `per-capita-energy-stacked` grapher configs through `fetchInputTableForConfig` + `constructReadme`, reproduced today's published `.readme.md` for both **byte for byte** on master, then regenerated them on this branch for a side-by-side diff. The script is not committed.

### Not done

- The Excel variant Pablo suggested (Data / Metadata / Sources / Readme tabs) is on hold until we know whether anyone would use it.
- `metadata.json` stays in the download; we decided against dropping it.

</details>

### DIFF
--- a/functions/_common/readmeTools.test.ts
+++ b/functions/_common/readmeTools.test.ts
@@ -1,8 +1,12 @@
+import * as _ from "lodash-es"
 import { expect, it, describe } from "vitest"
 
-import { SynthesizeNonCountryTable } from "@ourworldindata/core-table"
+import {
+    type CoreColumn,
+    SynthesizeNonCountryTable,
+} from "@ourworldindata/core-table"
 import { GrapherState } from "@ourworldindata/grapher"
-import { ColumnTypeNames } from "@ourworldindata/types"
+import { ColumnTypeNames, type OwidColumnDef } from "@ourworldindata/types"
 import { getRandomNumberGenerator } from "@ourworldindata/utils"
 import { constructReadme } from "./readmeTools.js"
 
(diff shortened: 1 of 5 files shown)


danyx23 (approved):
Looks good - thanks!

PR #7137, opened 2026-09-01
🐛🤖 stop the admin's css breaking grapher's modals

sophiamersmann:

Grapher's modals were broken in the chart editor three ways. Dead controls, a grey panel overflowing its frame, and Embed doing nothing at all. Two of them are the modal shell using a name the admin already owns.

- **Class name.** The react-aria rewrite put `modal-dialog` on the `Dialog`, which is bootstrap's own, and the admin imports bootstrap in full. Bootstrap declares `.modal-dialog { pointer-events: none }` and grapher declares no `pointer-events`, so nothing competed and every control in every modal went dead. `.modal-content` collided too. They are now `.modal-panel` and `.modal-panel-dialog`.
- **Element name.** React Aria renders a `Dialog` as a `<section>`, and the admin styles bare `section` as its own form cards, hence the grey background, the padding, and a 40px top margin that pushed `height: 100%` past the panel bottom. Those rules are now scoped to sections outside an embedded grapher.
Worth knowing before you read the diff:

- `:where()` keeps the `section` rules at the specificity they had. `.ImportPage section:not(:first-child)` overrides the `.AdminApp` one on source order alone, and a stronger selector breaks it.
- Cascade layers would not have fixed the class collision, so the CSS boundary work does not make this redundant. Layers decide ties, and there is no tie when only one side declares the property.
- The rendering to look at is [chart 64 in the editor](http://staging-site-modal-classes/admin/charts/64/edit), with Sources, Embed and the country selector each opened.

### DIFF
--- a/adminSiteClient/admin.scss
+++ b/adminSiteClient/admin.scss
@@ -3,6 +3,11 @@
 @import "../public/fonts.css";
 @import "bootstrap/scss/bootstrap";
 
+// Bare `section` also matches the react-aria dialogs grapher renders inside an
+// embedded chart. `:where` keeps the exclusion at zero specificity, so a page-level
+// `section` rule still wins on source order alone.
+$admin-section: "section:not(:where(.GrapherComponent *))";
+
 @mixin draft-background {
     background: repeating-linear-gradient(
         135deg,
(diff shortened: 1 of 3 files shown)


sophiamersmann:
@codex review

PR #7146, opened 2026-09-02
🔨🤖 keep email schemas out of site bundle

rakyi:
Infer request contracts from shared schemas exposed through a server-only entrypoint. Trust D1 subscriber rows and update the JavaScript and CSS budgets for the new newsletter UI.


ikesau (approved):

rakyi:
Folded into `email-notifications`.

PR #7175, opened 2026-09-03
🎉 Read data insights in place in the filtered /latest feed

mlbrgl:

## Context

With the data-insight type filter on, each insight now renders whole — title, avatar byline, the chart at full card width, then the text — as a six-column reading card, instead of the condensed teaser linking out to the DI page. A **View: Expanded / Compact** toggle in the sidebar switches back to exactly the condensed card the unfiltered feed shows, so one card design means one thing wherever it appears. The unfiltered feed is unchanged.

Along the way: expanded cards (data insights and data updates) get a copy-link button, and the linked-author-with-photo byline becomes one shared `AvatarByline` used by the standalone pages and both expanded cards — the expanded data update card gains photos with it.

Worth reviewer attention:

- DI Algolia records now carry `linkedAuthors` (byline) and `linkedCharts` / `linkedDocuments` (without them a DI's CTA renders as plain text). **Both need a reindex to show up.**
- Also fixes a filter-switching flash that predates this PR: with `keepPreviousData` the previous results stay on screen while the next page loads, but cards were judged by the URL's type filter, which runs ahead of the data — so e.g. switching to the data-updates filter briefly rendered the general feed's announcements expanded. Cards are now judged by the type filter recorded on the displayed results.
- The view toggle selection deliberately never resets and is shared by all type filters that offer the toggle (today: data insights only). Only the reader's own click changes it, so it can't change under cards that are still on screen mid-fetch.

## Testing guidance

- `/latest?type=data-insight`: full reading cards; toggle to Compact and back — Compact is the same eight-column teaser as the unfiltered feed, linking out to the insight's page.
- Switch between the unfiltered feed and the data-updates / data-insights filters and back: no card should flash expanded or collapsed during the transition.
- Author links in the byline stay clickable in both views; the copy-link button on an expanded card copies that insight's URL.

**Reading card** (`LatestDataInsightExpanded`): the full insight in a six-column card, with no expand/collapse affordance of its own — navigation is through the body and author links, plus the copy-link button. It carries the insight's own `id={slug}` anchor like every other card. Unlike the condensed card it doesn't gate on a bake probe: it renders the insight's content rather than linking to a page that may not be baked yet.

**View toggle**: which type filters offer it is a list (`LATEST_TYPES_WITH_VIEW_TOGGLE`); data updates can join by adding the type and having their hit read `view`. The toggle is always rendered and disabled for filters that don't offer it, so the sidebar never shifts. Default is Expanded.

**Flash fix mechanism**: `LatestPagesResult` records the `latestType` it was fetched with; `LatestSearch` derives `displayedLatestType` from the displayed page and uses it for `isExpanded` and the cards' `view` prop, while the filter controls keep following the URL. Only the type filter needs this — it's the one filter that changes how a card renders; topics only change which hits come back. The toggle state itself is a single never-reset `useState`: the earlier reset-on-filter-change was the other source of mid-fetch flashes (it landed while the previous type's cards were still displayed).

**Deep links**: `LatestSearch` still scrolls to the `#<slug>` anchor after the first load and records that slug as auto-expanded. For announcements and data updates that's the existing behaviour, unchanged. For data insights the slug only counts inside the data-insight filter, and since the view state defaults to Expanded on every fresh load, the card a deep link lands on is already expanded — the clause only shows up if the reader then switches to Compact, where that one card stays expanded. Kept as-is rather than special-cased; it isn't worth a separate test.

**Indexing**: `pagesChronological.ts` adds `linkedAuthors` / `linkedCharts` / `linkedDocuments` to DI records; the attachments context for card rendering is built from these (`makeAttachments`).

</details>


mlbrgl:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7175** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7215](https://github.com/owid/owid-grapher/pull/7215) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7217](https://github.com/owid/owid-grapher/pull/7217) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

site/latest/AnnouncementContent.tsx
```diff
@@ -37,17 +37,13 @@ export const AnnouncementContent = ({
...
-                />
-            ))}
-        </div>
+    // Keeps its class: the SCSS uses it as a hook to reorder the byline among
+    // the expanded announcement's trailing blocks.
+    const authorByline = (
+        <AvatarByline
+            className="announcement-content__authors"
+            authors={authors}
+        />
```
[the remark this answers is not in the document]
mlbrgl: > _Written by Claude Opus 5 — @mlbrgl at the wheel._

Good catch on the mechanism: `<AvatarByline …>` is a truthy element object whatever it later renders, so `ExpandableText`'s `trailingBlocks.length > 0 || children` can't see that the byline will come back empty. Restored the `authors.length > 0 &&` guard in the parent.

Not actually reachable today, for the record: `parseAuthors` defaults to `"Our World in Data team"` and always returns at least one entry, so `content.authors` is never empty for a published gdoc — the `?? []` in the chronological indexer is defensive. Keeping the guard anyway, since it's the parent's job to decide whether there is anything worth expanding to.

site/latest/LatestSearch.tsx
```diff
@@ -57,6 +64,14 @@ export const LatestSearch = ({
     )
     const { topics, latestType } = state
 
+    // Expanded/Compact for type filters that offer the toggle. Local, not in
+    // the URL, and shared by all such filters. Deliberately never reset:
+    // only the reader's own click changes it, so it can't change under the
+    // cards that stay on screen while the next results load
+    // (keepPreviousData) the way a reset-on-filter-change would.
+    const [view, setView] = useState<LatestFeedView>(DEFAULT_LATEST_FEED_VIEW)
+    const showViewToggle = hasViewToggle(latestType)
```
ikesau: nit: can this be `shouldShowViewToggle` - it sounds like a function name otherwise

Moreover, this level of ceremony (and its tests) feels a bit YAGNI. It could just be `latestType === "data-insight"` (plus one update below, as well)

site/latest/LatestSearch.tsx
```diff
@@ -119,10 +134,8 @@ export const LatestSearch = ({
     }, [latestType, latestTypeFacetCounts])
 
     // Disable topics that would yield 0 results given the current filters.
-    // Never disable a topic that is already selected (so the user can deselect
-    // it). When topics are selected the facet counts are narrowed by Algolia's
-    // conjunctive filtering, so the counts reflect co-occurrence with the
-    // current selection — topics with 0 count genuinely add no results.
+    // Keep selected topics enabled so they can be deselected. Counts exclude
+    // the topic filter, reflecting what each replacement selection would show.
```
ikesau: "Counts exclude the topic filter, reflecting what each replacement selection would show"

What does this mean?

site/latest/LatestSearch.tsx
```diff
@@ -152,12 +165,26 @@ export const LatestSearch = ({
...
+    // stay on screen while the next page loads (keepPreviousData), and the
+    // incoming type would flash them expanded/collapsed. Only the type
+    // filter affects how a card renders; topics only change which hits
+    // come back.
+    const displayedLatestType = data?.pages[0]?.latestType ?? null
+    const activeView = hasViewToggle(displayedLatestType) ? view : undefined
+
+    // Insights expand only in their type-filtered feed. Other announcements
+    // expand for a deep link; data updates also expand under their type filter.
+    const isExpanded = (hit: PageChronologicalRecord): boolean => {
```
ikesau, line 177: Can people even get a deep link to a DI in the feed? 🤔 

The carousel links to the DI page directly. The copy link button does as well.

site/latest/LatestSearch.tsx
```diff
@@ -174,6 +201,12 @@ export const LatestSearch = ({
                 />
             </div>
             <hr className={LATEST_FILTERS_DIVIDER_CLASSES} />
+            {/* Tied to the type filter alone, so it can't mount or unmount
+                while a feed loads: gating on the hits as well would flash it
+                in and out on a feed that turns out to be empty. */}
```
ikesau: When can this happen? We have DIs for every area, no?

site/latest/README.md
```diff
@@ -45,13 +45,18 @@ A single Algolia call ([`queryLatestPages`](../search/queries.ts)) issues three
...
+
+**Data insights** vary by _where_ they render. In the unfiltered feed they're a condensed teaser linking to their page. With the data-insight type filter on, the feed offers a **View: Expanded / Compact** toggle ([`LatestViewToggle`](./LatestViewToggle.tsx)): Expanded shows each insight whole, read in place ([`LatestDataInsightExpanded`](./LatestDataInsightExpanded.tsx)); Compact is the very same teaser as the unfiltered feed — one card design, one behaviour, wherever it appears. The expanded card needs the authors' avatars, which is why the indexer loads `linkedAuthors` for data insights.
```
ikesau: This kind of documentation seems likely to drift and possibly not help with performance given agents always read the code anyway. Maybe part of a bigger strategic discussion we should have?

site/latest/LatestDataInsightExpanded.tsx
```diff
@@ -0,0 +1,74 @@
...
+                            shouldLightbox={false}
+                        />
+                    )}
+                    <div className="latest-data-insight-expanded__blocks">
+                        <ArticleBlocks
+                            blocks={otherBlocks}
+                            containerType="data-insight"
+                        />
+                    </div>
+                    <CopyLinkButton
```
ikesau, line 61: This could link to a page that hasn't baked yet on a feed that's filtered to data insights only, after a DI has just been published (because it bypasses the `isLikelyBaked` check).

Probably unlikely to matter too much.

ikesau (approved):

site/latest/LatestSearch.tsx
mlbrgl: Not directly from the site, but would like to push newsletter updates to the DIs in-situ (vs their standalone page), to make further browsing a simple scroll. That being said, the default is "expanded" for the DI filtered view so a deep-link already hits the expanded version ("view" is not exposed in the URL) and we anyway don't want to mix expanded and collapsed DIs in the same feed - they don't run on the same number of columns, and it ends up looking somewhat broken (as opposed to DUs where collapsed and expanded versions compliment each other). Conditional now simplified accordingly.

site/latest/LatestSearch.tsx
mlbrgl: yep part of the "build story" and a bet on my end seeing how through the cycle DUs and DIs kept converging. But surprisingly, that convergence stopped at the view filter 😁 Simplifying while keeping the function for clarity and keeping those two places in sync.

site/latest/LatestSearch.tsx
mlbrgl: that could happen on a cold load, not as a result of switching filters since filter combinations yielding no results are deactivated (and you're right we have DIs for all areas) and results are kept mounted while loading the next state. But there is anyway no reason to gate the filter rendering on the presence/absence of results (you'd deactivate it at best). Comment removed.

site/latest/LatestDataInsightExpanded.tsx
mlbrgl: yes correct 🦅 and that's by design IIRC. We could gate the button, but that gives no chance of that (temporary broken) link being copied. Say someone shares a link to a data insight that hasn't finished baking yet. If the recipient opens it right away, then true, they 404. But if by any chance they delay opening that notification for a few minutes, they will end up seeing it. The risk is the 404 in some cases, but that seems like a good bet.

site/latest/README.md
mlbrgl: yep let's pick that up in a call 🙂 JSYK I'm keeping these in check in terms of granularity and length, but happy to talk about the right balance. (EDIT: slightly trimmed)

site/latest/LatestSearch.tsx
mlbrgl: it's more or less a repeat of this from the README (and the queries). Harmonized.

> Each facet-count query intentionally drops its own axis so the counts answer "what would happen if I picked a different value here?" rather than self-narrowing to the current selection. That's what disables zero-match options in the type dropdown and the topic pills

PR #7177, opened 2026-09-04
Recommend topic pages from the tags of matching charts

Marigold:
## 👨 Summary

Our topic recommendations for search queries are not great. That's a shame, given that users who click on them show higher engagement.

Currently, we send the search query to Algolia and ask for topic pages. This naive keyword search often returns unrelated topics, for example:
- "gni" -> Burden of Disease
- "ai" -> Air Pollution
- "gdp" -> CO₂ and Greenhouse Gas Emissions

This PR derives the recommended topic pages from the **returned charts** instead.

I did both a qualitative and a quantitative analysis, and the new method outperforms the current one.

> _Written by Anthropic Claude Fable 5.1 — @Marigold at the wheel._

Searching "gdp" recommended the CO₂ topic page, and never the Economic Growth one. The topic-page cards were a full-text Algolia search over topic pages, but topic pages are mostly charts and key indicators whose text isn't indexed: the Economic Growth page's records never contain "GDP" at all, while long-form pages match on passing mentions, and since all topic pages have the same `importance` the order among them was just last week's pageviews. This PR derives the recommended topic pages from the charts matching the query instead: the topic tags of the best-ranked chart hits, weighted by rank. The same logic backs the public `/api/search` when a `type=pages` search is restricted to topic pages. On the site, behaviour changes for every search with a query; empty-query browsing by topic/area is unchanged.

_Review depth: skim, except the `/api/search` dispatch rule (which requests count as "topic pages only"), which is a public API contract. Two Algolia requests replace one; the ranking and orchestration live in one shared, unit-tested function._

## How it works

`searchTopicPagesOfMatchingCharts` in `@ourworldindata/utils` (shared by the site and the Cloudflare function, like the closest-matches fallback) fetches the top 50 chart hits for the query with the same facet filters as the chart results, so country and topic filters carry over. Each hit votes for its topic tags with weight 1/rank (`rankTopicsOfChartHits`); areas are skipped (they have no page of their own), and so are tags without a topic page in the tag graph. The topic-page records are then fetched in one request via the `path` filter and re-ordered client-side; pagination for "Show more" is done locally on that list.

Two things that a first version got wrong and the tests now pin down:

- **Facet counts over all matches are not the answer.** Large explorer and multi-dim view families dominate them regardless of rank: "population" matches 1514 Migration views and recommended Migration and Natural Disasters. Rank-weighted votes from the top hits give Population Growth.
- **Query words are matched as whole words first** (`queryType: prefixNone`), falling back to Algolia's default prefix matching only when that finds nothing (an unfinished word like "popul"). Otherwise "ai" is the prefix of "air", "aid" and "aids", and those charts outvote the AI ones. Algolia synonyms still apply, which is how "ai" reaches Artificial Intelligence.

**API.** `GET /api/search?type=pages&pageTypes=topic-page,linear-topic-page&q=gdp` (either type alone works too) now returns the recommendation; `countries`, `topics` and `requireAllCountries` apply to the underlying chart search. Mixed page types (`topic-page,article`) and empty queries keep the plain text search, and so does a query no chart matches. The tag graph comes from the baked `/topicTagGraph.json` via `env.ASSETS`, the same way redirects are read. Response shape is unchanged (`SearchPagesApiResponse`); `closestMatches` is never set on a recommendation. Documented in `docs/search-api.openapi.yaml`.

Rejected alternatives: indexing chart titles into topic-page records (needs a reindex and bigger records), and an Algolia synonym `gdp ↔ economic growth` (fixes one query). The text search is kept as a fallback for empty queries and for queries no chart matches, so author-name searches still find topic pages the way they do today.

## Verified

Ran the new `queryTopicPages` against the production Algolia app with the production tag graph:

"Before" is the current text search over topic pages, "after" the new recommendation, both against production Algolia (first two cards each):

| Query | Before | After |
|---|---|---|
| gdp | CO₂ and Greenhouse Gas Emissions, Poverty | Economic Growth, Poverty |
| population | Population Growth, Life Expectancy | Population Growth, Energy |
| ai | Artificial Intelligence, Air Pollution | Artificial Intelligence, Technological Change |
| life expectancy | Life Expectancy, Medicine and Biotechnology | Life Expectancy, Global Health |
| child mortality | Child and Infant Mortality, Diarrheal Diseases | Child and Infant Mortality, Global Health |
| temperature | Climate Change, CO₂ and Greenhouse Gas Emissions | Climate Change, CO₂ and Greenhouse Gas Emissions |
| guns | Bonheur et satisfaction | Homicides, War and Peace |

Also checked: co2, internet, meat, migration, energy, vaccination, democracy, obesity all lead with the matching topic page; "popul" and "child mort" (unfinished words) still resolve via the prefix fallback; "max roser" (no chart matches) falls back to the text search and finds his topic pages as before.

Note the production Algolia app (`ASCB5XMYF2`) is not the one in a dev `.env`, which points at the staging app with different synonyms and pageviews; the numbers above are from production.

The `/api/search` path has integration tests against production Algolia (`searchApi.integration.test.ts`: Economic Growth leads for "gdp", pagination is consistent) and dispatch tests with mocked `ASSETS` (`index.test.ts`).

Staging (`staging-algolia` label set): http://staging-site-better-topic-pages-recommend/search?q=gdp and http://staging-site-better-topic-pages-recommend/api/search?type=pages&pageTypes=topic-page,linear-topic-page&q=gdp

## Quantitative evaluation

Measured with the search-demand evaluator's new `topic-pages` surface (owid/analytics#1029), production vs this branch's staging deploy, both through `/api/search`. Demand comes from GA4's `owid.site_search` event over 30 days, volume-weighted, 1,171 queries, relevance judged 0–3 by Gemini.

| | prod (text) | this PR |
|---|---|---|
| top result relevant (≥1.5) | 60.7% | **70.2%** |
| MRR | 0.607 | **0.693** |
| Precision@5 | 0.342 | **0.408** |
| NDCG@5 | 0.708 | **0.743** |
| searches with no result | 7.7% | **4.7%** |

**12.7% of demand improved against 3.4% regressed**; strictly better on 26.5% of demand, strictly worse on 11.3%.

The evaluation of the first commit is what surfaced the two ranking bugs fixed in `30214a0c`. Raw facet counts gave a clean negative gradient by how many charts a query matched — positive below 200 matching charts, negative above, no improved query at all above 1,000 — and reciprocal-rank voting plus whole-words-first flipped the sign in every bin.

The remaining 3.4% is 47 low-volume queries forming one class: abstract nouns no chart is about (`safety` → Poverty, `sports` → Metals and Minerals, `carrying capacity` → Transport). Production isn't really better there so much as luckier — it returns *Human Height* for `universities` — so that argues for a relevance floor rather than for keeping the text search. Separately, 13.0% of demand is served by neither engine because no relevant topic page exists; that's a content gap and the ceiling on any ranking change.

One caveat on magnitude: an earlier run reported +14.6pp rather than +9.5pp because it weighted demand by Algolia's search-analytics table, which counts every API hit — a monitoring script had `banana production` at 7,225 "searches" (3 real ones), outweighing `gdp`. Fixed in owid/analytics#1030; direction never changed, only size.

📊 **[Full evaluation report](https://claude.ai/code/artifact/8f562a0d-cf64-40dd-8111-93e685428874)** (OWID-internal link)

## Decisions worth a second look

The chart results themselves still rank air pollution and foreign aid charts first for "ai", because the charts index uses prefix matching and `exactOnSingleWordQuery: none`. The topic card will say Artificial Intelligence next to charts that don't; the same whole-words-first approach could fix the chart results, but that's a separate change to how the data results rank.

- `rankTopicsOfChartHits` and `searchTopicPagesOfMatchingCharts` live in `packages/@ourworldindata/utils/src/search/searchTopicPages.ts` (moved there from `site/search` in the second commit); `"path"` was added to the shared `SearchFacetAttribute` union for the pages lookup.
- The topic-voting chart query (50 hits, `tags` only) costs one extra Algolia request per search (two when the whole-word query is empty) on top of the existing chart results query (they aren't shared because the Writing tab renders without the data results component).
- `rankTopicsByChartTagCounts` uses the flattened tag graph from `SearchContext`; nested topics (e.g. Climate Change under CO₂) are found by walking the whole tree.
- Debug probe used for the table above: a small `tsx` script calling `queryTopicPages` with `algoliasearch/lite` and `/topicTagGraph.json`.

</details>


functions/api/search/searchApi.ts
```diff
@@ -280,8 +282,86 @@ export async function searchPages(
...
+            attributesToRetrieve: PAGE_ATTRIBUTES,
+            offset,
+            length,
+        }
+    )
+    if (!result)
+        return searchPages(
+            config,
+            state.query,
+            offset,
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Fable 5.1 — @Marigold at the wheel._

Fixed: the fallback now passes the selected topics through to `searchPages` as a `tags` facet filter, matching the site's `queryTopicPagesByText`. Covered by an integration test ("keeps the topic filter when falling back to the text search").

packages/@ourworldindata/utils/src/search/searchTopicPages.ts
```diff
@@ -0,0 +1,156 @@
...
+    // The topic list is short (a few dozen at most), so fetch every page in
+    // one request and paginate locally. Algolia returns them in its own
+    // order; the facet order is what we want.
+    const pageTypes = params.pageTypes ?? TOPIC_PAGE_TYPES
+    const pagesResponse = await searchSingleForHits<THit>(liteSearchClient, {
+        indexName: params.pagesIndexName,
+        query: "",
+        filters: pageTypes.map((type) => `type:${type}`).join(" OR "),
+        facetFilters: formatDisjunctiveFacetFilters(
+            new Set(topics.map((topic) => `/${topic.slug}`)),
```
mlbrgl, line 138: let's try to use `getPrefixedGdocPath` here instead? (it'll probably cost moving the util some place else but worth the guarantee)

packages/@ourworldindata/utils/src/search/searchTopicPages.ts
```diff
@@ -0,0 +1,156 @@
...
+    // the beginning of "air" or "aid". Algolia's default (prefixLast) treats
+    // the last word as a prefix, which lets the many charts about air
+    // pollution and foreign aid outvote the AI charts. So match whole words
+    // first (Algolia's synonyms, e.g. ai → artificial intelligence, still
+    // apply) and only fall back to prefix matching when that finds nothing,
+    // i.e. when the word really is unfinished ("popul").
+    let chartsResponse = await searchCharts("prefixNone")
+    if (chartsResponse.hits.length === 0)
+        chartsResponse = await searchCharts("prefixLast")
+    const topics = rankTopicsOfChartHits(chartsResponse.hits, params.tagGraph)
```
mlbrgl, line 126: nit: `rankedTopics` or `orderedTopics`? Makes `orderedHits` below more natural to read.

functions/api/search/index.ts
```diff
@@ -174,17 +193,36 @@ export const onRequestGet: PagesFunction<Env> = async (context) => {
...
-                      baseUrl
-                  )
+                ? wantsTopicPagesOnly
+                    ? await searchTopicPages(
+                          algoliaConfig,
+                          searchState,
+                          await fetchTagGraph(env, baseUrl),
+                          page * hitsPerPage, // Convert page to offset
+                          hitsPerPage,
+                          pageTypes as OwidGdocType[],
```
mlbrgl, line 215: right now it is possible to request only a subset of topic pages (e.g. passing "linear-topic-page" here). However, this distinction between topic pages and linear topic pages is not a semantic boundary we should necessarily expose. It is merely a presentation concern, and the API should absorb that dichotomy. What I'm suggesting here is two-fold:
- whenever one of the topic page subtype is passed, expand to both
- searchTopicPages doesn't get a pageTypes argument, it always request both whenever one of the subtype is asked for. Effectively, this means a pageTypes=linear-topic-page request could hit searchTopicPages and return both topic pages and LTPs but the only documented/supported case should be pageTypes=topic-page (which also returns both)

mlbrgl:
* **#7303** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7303?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7177** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7177?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7177?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/@ourworldindata/utils/src/search/searchTopicPages.ts
```diff
@@ -0,0 +1,156 @@
...
+    if (topics.length === 0) return undefined
+
+    // The topic list is short (a few dozen at most), so fetch every page in
+    // one request and paginate locally. Algolia returns them in its own
+    // order; the facet order is what we want.
+    const pageTypes = params.pageTypes ?? TOPIC_PAGE_TYPES
+    const pagesResponse = await searchSingleForHits<THit>(liteSearchClient, {
+        indexName: params.pagesIndexName,
+        query: "",
```
mlbrgl, line 135: this means we're now hitting api/search/cached-queries here, which isn't obvious. If we want to double down on using the cache here (as opposed to merely riding along the side effect), we might want to sort the facetFilters attribute to make the cache key in cached-queries agnostic to the order of topics

mlbrgl (approved):
nice, thanks! 

The one thing that might require a bit more attention/discussion is the atomicity of topic pages as a concept, which I think is worth preserving.

mlbrgl:
FYI as a follow-up to cement the shared search architecture: #7303

mlbrgl:
Also not sure if that was part of your scope, but we technically have the same relevance issue in the HP/navbar autocomplete, although the context and flow is there different enough that it'd warrant a separate investigation.

PR #7193, opened 2026-09-05
Organize download tests around their contracts

danyx23:
> _Written by OpenAI GPT-6 — @danyx23 at the wheel._

Give download-table behavior a focused test file organized by columns, row eligibility, original observation times, and chart/tab boundaries. All 15 moved scenarios retain their complete inputs and assertions.

Stacked on #7192. Assertion corrections and stronger evidence follow separately.

<details><summary>Details</summary>

Move `prepareTableForDownload` from `GrapherState.test.ts` to `GrapherDownload.test.ts`, with four short contract explanations. Fixtures remain local and explicit; no new fixture DSL or production changes.

| Existing cases | New group |
| --- | --- |
| Line columns, scatter roles, annotations | columns and annotations |
| Missing x/y rows, entity-code backfill, empty table, CSV error values | row eligibility and entity codes |
| Single-y time, multiple-y source columns, empty source columns, duplicate-time placeholders | original observation times |
| Discrete bar, slope, table, map | chart and tab boundaries |

Preservation: compared complete `it()` calls across the old file and both new files using a temporary AST-based check. Every body is identical ignoring whitespace; only group placement/order changes. No cases or assertions removed. The single-y test's misleading name and weak CSV assertion intentionally remain for the next PR.

Validation: focused Grapher state/download suites, repository typecheck, formatting, changed-file lint, and whitespace checks passed. This test-only move cannot affect rendering, so SVG reference regeneration is not needed.

</details>

PR #7194, opened 2026-09-05
Assert downloaded observations and source times explicitly

danyx23:
> _Written by OpenAI GPT-6 — @danyx23 at the wheel._

Make download tests verify the actual observations and provenance, rather than only counts or nonempty output. Correct the single-y scenario's misleading name: its fixture never interpolates data, so it cannot establish source-time renaming.

Stacked on #7193, the structural move. All fixtures and production behavior remain unchanged.

| Old property | Replacement / stronger evidence |
| --- | --- |
| Two surviving rows | Exact year/GDP pairs for 2020 and 2022; still exactly two rows |
| Some UK rows, all with GBR | Both 2020 and 2021 UK observations with GBR; catches losing the row that needed backfill |
| Nonempty CSV | Exact small CSV, including blank source-time cells and borrowed 2021 provenance; preserves the nonempty-row assertion |
| Some years and no source-time column | Exact observed years 2019/2021 plus both existing column assertions; corrects the name without removing its actual protection |
| Placeholders at positions 0/2 and any non-placeholder at 1 | Exact three-element array with source year 2021 in the middle |

Targeted defect checks, each against the affected old/new named scenario:
- Shift output years by 100: old passed, new failed on 2119/2121 versus 2019/2021.
- Shift borrowed source time by 1: old passed, new failed on 2022 versus 2021.
- Write zero instead of an empty source-time placeholder: old CSV test passed, new failed on nonblank CSV cells.

These checks establish additional evidence in these scenarios, not that every defect escaped the whole old suite. All mutations restored. All 15 download tests pass, as do formatting, typecheck, changed-file lint, and whitespace checks.

</details>

### DIFF
--- a/packages/@ourworldindata/grapher/src/core/GrapherDownload.test.ts
+++ b/packages/@ourworldindata/grapher/src/core/GrapherDownload.test.ts
@@ -113,7 +113,12 @@ describe("prepareTableForDownload", () => {
             })
 
             const downloadTable = grapher.filteredTableForDownload
-            expect(downloadTable.numRows).toBe(2) // Only 2020 and 2022
+            expect(
+                downloadTable.rows.map(({ year, gdp }) => [year, gdp])
+            ).toEqual([
+                [2020, 2500],
+                [2022, 2700],
+            ])
         })
 
         it("attempts to fill in missing entity codes from inputTable", () => {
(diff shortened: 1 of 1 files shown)


danyx23:

Closing alongside #7193, which this PR depends on: #7193 creates `GrapherDownload.test.ts` and this one edits it, so the two cannot be rebased independently. See #7193 for the details.

PR #7210, opened 2026-09-06
Test search response contracts deterministically and strip ranking metadata

### LINKED ISSUE #7203: Separate deterministic search behavior tests from narrow live service contracts
> _Written by OpenAI GPT-6 — @danyx23 at the wheel._

**Prerequisite:** Implement this on top of https://github.com/owid/owid-grapher/pull/7190, which improves the testing instructions, including how to preserve existing guarantees when rewriting tests.

## Problem
`functions/api/search/searchApi.integration.test.ts` exercises a live search index, but some assertions provide weak evidence: country/topic cases only require positive hit counts, and a URL-type case asserts only if the response happens to contain that type. Changing index contents can determine which guarantees are exercised.

There is already useful deterministic filter/matching coverage in `searchFacetFilters.test.ts` and `searchClosestMatches.test.ts`; the opportunity is the service response boundary, not duplicating those unit tests.

## Proposed change
Exercise the real search API implementation against controlled transport responses for:
- Response shaping and stripping internal fields.
- Type-specific URLs and query parameters for the supported result types, with every claimed type explicitly present in the fixture.
- Pagination behavior.
- Invalid topic input and valid topics with empty results.

Use exact relevant output expectations. Stub the external transport rather than the search functions under test. Keep a narrow live Algolia/index contract and real Workers compatibility coverage where they provide distinct evidence; `functions/test/search.e2e.test.ts` currently covers that runtime/service boundary.

## Acceptance and review
Map each existing live assertion to retained deterministic or live evidence before removing it. No test should pass simply because the result type it claims to check is absent. Document which commands and CI jobs run deterministic versus external-service tests, and reconcile the testing strategy's opt-in description with the actual default Vitest inventory.

Do not remove deterministic local Workers/R2 tests as part of classifying external-service tests. This is a lower-priority improvement in semantic specificity and reproducibility, not a response to observed search test failures or measured performance problems.

danyx23:
> _Written by OpenAI GPT-6 — @danyx23 at the wheel._

Move search response-shaping assertions into deterministic transport-level tests while retaining live evidence for index filtering, pagination, and Workers compatibility. The new tests expose and fix `_rankingInfo` leaking into page search responses.

Closes #7203. Based on #7190 (`test-strategy`). Existing test execution policy is unchanged; documentation now correctly identifies live tests included in default Vitest discovery and CI.

- Run the real search implementation and Algolia client with controlled HTTP responses. Require all three chart result types and eight page types, exact URLs and public fields, pagination forwarding, empty responses, and invalid-versus-empty topic behavior.
- Keep seven existing live tests unchanged for country/all-country/topic filtering, rare valid topics, chart/page pagination, and page-type filtering. Workers search and deterministic local R2 suites remain unchanged.
- Map removed assertions to retained evidence and document commands/policy in `functions/api/search/TESTING.md`.
- Validation: all 19 original live tests passed before the rewrite. The final focused run passed 32 tests across four files (13 deterministic API tests, 7 retained live tests, and 12 Workers/search/R2 tests).
- Regression evidence: adding ranking metadata to fixtures failed all eight page-type cases before the response cleanup fix; all pass afterward.
- Repository typecheck, changed-file lint, formatting, and diff whitespace checks passed.

</details>

### DIFF
--- a/docs/testing-strategy.md
+++ b/docs/testing-strategy.md
@@ -220,11 +220,13 @@ tests and repository typechecking cannot prove.
 
 ### Runtime-specific and external-service tests
 
-The `functions/test` area includes Node-level integration tests and opt-in E2E
-tests that run handlers in a real Workers runtime or contact services such as
-Algolia and R2. They protect compatibility and integration assumptions that a
-mock cannot establish, but external state makes them slower and less
-deterministic. They should be narrowly scoped, clearly labelled, and kept out
+The `functions/test` area includes tests that run handlers in a real Workers
+runtime, including local R2 tests and live Algolia search tests. The Node-level
+`functions/api/search/searchApi.integration.test.ts` also contacts live Algolia.
+These search and local R2 files are currently included in default Vitest discovery
+and the main CI `test` job; dedicated scripts do not make them opt-in. They protect
+compatibility and integration assumptions that a mock cannot establish, but
+external state makes them slower and less deterministic. They should be narrowly scoped, clearly labelled, and kept out
 of the fast suite unless their environment can be made reliable.
 
 ### Bespoke project tests and builds
--- a/functions/api/search/TESTING.md
+++ b/functions/api/search/TESTING.md
@@ -1,0 +1,52 @@
+# Search response contracts
+
+`yarn test run functions/api/search/searchApi.test.ts` runs deterministic API
+contracts. The real search implementation and Algolia lite client run against a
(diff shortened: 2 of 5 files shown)


danyx23:

Closing as part of bringing this stack up to date with master.

The `docs/testing-strategy.md` section this branch edits has since been rewritten on the trunk, and taking this version would revert that rewrite. The test changes here are independent of that and can be redone against current master.

PR #7215, opened 2026-09-07
🧪 Experiment: sticky filters on /latest

mlbrgl:
> _Written by Anthropic Claude Fable 5.1 — @mlbrgl at the wheel._ (Testing guidance revised by Claude Opus 5 after #7217 was closed.)

Sets up the A/B/C experiment on whether the `/latest` filters should stay in reach while scrolling. Three arms, equal thirds, on `/latest` only:

- **not-sticky** — control, status quo
- **reveal-on-scroll-up** — filters scroll away with the page; scrolling back up pins them and slides them in, scrolling down again slides them off
- **fully-sticky** — filters stay pinned to the top of the viewport

On mobile both treatment arms follow the design: the content-type dropdown moves above the topic pills, the pills get a "Filter by topic" label, and only the pills pin. On desktop the whole filter row pins. Changing a filter from the pinned bar scrolls back so the new results start right below it.

Two things worth knowing:

- Every arm's layout is CSS keyed off the body class the edge middleware adds, so the baked skeleton and the mounted app render identically and nothing flashes. There's no middleware locally, which is why testing needs a forced arm.
- The topic pills are ordered most-popular-first in **every** arm, so the order isn't a confound. The ranking is static for the experiment's duration and only indicative — see the details block for where it comes from.

## Testing guidance

Staging assigns an arm at random on first load. To pick one deliberately, set the cookie and reload — the middleware only assigns when the cookie is absent, so the body class is in the HTML as served:

```js
// devtools console, on the staging origin
document.cookie = "exp-latest-sticky-filters-v1=fully-sticky; path=/" // or reveal-on-scroll-up, not-sticky
```

Locally under `make up` there's no edge middleware, so also add the class by hand: `document.body.classList.add("exp-latest-sticky-filters-v1--fully-sticky")`. A friendlier arm switcher is being picked up separately, from the admin.

Check desktop and a phone viewport: on mobile only the labelled pills should pin, in the reveal arm the bar should hide on scroll down and reappear on scroll up without sliding over the header, and changing a filter from the pinned bar should scroll back so results start right below it.

- [x] Changes to CSS/HTML were checked on Desktop and Mobile at all three breakpoints (local dev, body class forced)

### Topic pill order

"Popular" means how often each topic filter is actually set on `/latest`: the GA `owid.site_latest` event's `latest_topics` param, as a share of sessions that arrived unfiltered, from the /latest redesign retrospective (owid/analytics#994, Figure 12; window 2026-06-05 to 2026-07-30). Resulting order: Energy & Environment, Poverty & Economic Development, Violence & War, Innovation, Food & Agriculture, Health, Living Conditions, Population, Education, Human Rights & Democracy. Usage is spread thinly (0.4–0.7% of sessions per topic) and was measured under the old order, so the ranking is indicative. It's deliberately static so the order can't shift mid-experiment; a live order would be a dbt export in owid/analytics synced to an `analytics_*` MySQL table the baker reads — a natural follow-up once the experiment concludes.

### Implementation notes

- `packages/@ourworldindata/utils/src/experiments/config.ts` — experiment registration; `constants.ts` exports the id and arm ids so the site code and analytics share them. Expires 2026-11-30.
- `site/latest/LatestSearch.scss` — all arm styling. The sticky element is the facets **grid item**, not the facets block: a sticky element can only move within its parent's box, and the facets' own wrappers are no taller than the facets (first attempt did nothing for exactly this reason). On mobile a negative `top` equal to the dropdown row (32px trigger + 16px gap, both fixed in `LatestTopicFacets.scss`) lets that row scroll out so only the pills stay pinned. Desktop padding when stuck is compensated with negative margins so the normal-position layout is unchanged vs control.
- `site/latest/LatestTopicFacets.tsx` / `LatestPageSkeleton.tsx` — pills wrapped in a `latest-topic-facets__topics` block with a `latest-topic-facets__topics-label`, rendered in every arm and shown by CSS on mobile in the treatment arms. The skeleton mirrors the markup so the pre-mount page matches.
- `site/latest/latestHooks.ts` — `useLatestStickyFiltersArm` reads the arm once on mount via `getExperimentState()`. `useRevealOnScrollUp` is a three-state machine (`flow` / `revealed` / `hiding`) driven by the site's existing `useScrollDirection` plus a "past the pin point" flag: in the flow by default (so the first scroll down is just normal scrolling), pinned + slide-in keyframe animation on scroll up past the pin point, pinned + transform transition off on scroll down, back to flow on `transitionend` (immediately under reduced motion, where there is no transition). Above the pin point it's always in the flow, so it can never be translated off-screen at its layout position. `getStickyLayout` derives the layout position from the previous grid row (see below) and reads the pin offset from the `--latest-sticky-filters-top` custom property, which is defined whether or not the container is sticky right now.
- `site/latest/LatestSearch.tsx` — `scrollToTopOfPinnedElement` runs on every filter change, only when the container is currently sticky and pinned. A pinned element reports its *pinned* position through both `getBoundingClientRect` and `offsetTop`, so its layout position is derived from the previous grid row (the page header's margin box bottom, plus the container's top margin), and the target is that minus the pin offset (0 on desktop, -48px on mobile so the pills land pinned). Smooth unless reduced motion. Don't be tempted by a zero-height marker element: any extra grid item adds a row, and the newsletter block is pinned to an explicit `grid-row`.
- `site/latest/LatestTopicFacets.tsx` — `aria-labelledby` from the pill group to the "Filter by topic" label, `aria-label="Filter by type"` on the Select.
- `site/SiteAnalytics.ts` + `analyticsTypes.ts` — `experimentArm?: string` on `SiteLatestParams`, `SiteLatestResultClickParams`, `SiteLatestExpandParams`, populated from the cookie.
- `site/latest/latestUtils.ts` — `LATEST_TOPIC_AREAS_BY_POPULARITY` + `sortTopicAreasByPopularity` (stable sort; unknown areas last in tag-graph order), applied to the live pills in `LatestSearch.tsx` and the baked skeleton. Unit-tested in `latestUtils.test.ts`.
- `site/latest/README.md` — section 8 documents the above.

### Local testing

Use the forced-arm setup above. Analytics events don't log to the console (the debug flag in `GrapherAnalytics` is hardcoded off); inspect `window.dataLayer` instead, e.g. `dataLayer.filter(e => e.event === "owid.site_latest")`.

### Checks

`yarn typecheck`, oxlint with `--deny-warnings` on changed files, oxfmt check, and the latest unit tests all pass on the restacked branch.

</details>


mlbrgl:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7215** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7175** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7217](https://github.com/owid/owid-grapher/pull/7217) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

ikesau (approved):
Just one thing - we should change `scroll-margin-top` on elements with IDs when `always-sticky` is on, otherwise the bar covers the cards when deeplinked.

PR #7217, opened 2026-09-07
🎉 Switch experiment arms from /exp

mlbrgl:

Adds `/exp`, a page for picking an A/B experiment arm by hand, for QA and for local dev.

This replaces the `?exp-<id>=<arm>` URL override this PR originally proposed. That approach answered the open question the previous revision of this description left dangling — "no guard against sharing an override link publicly; a visitor following one would be forced into that arm and counted in it" — the wrong way round: it put QA code in front of every visitor, as a query-string parse on every HTML request at the edge plus ~1 kB in the site bundle, and it let anyone with the link skew a live experiment.

**Forcing an arm never needed any of that.** `experimentsMiddleware` already honours an existing `exp-<id>` cookie and only rolls the dice when there isn't one, so "force an arm" is just "set the cookie". `/exp` lists the active experiments with a button per arm and sets it — which is also the only thing that works on a phone, where devtools isn't available and browser extensions effectively aren't either (Chrome on Android supports none; iOS Safari extensions have to ship through the App Store). Each experiment also lists its enrolled pages as links, so a phone needs no typing beyond `/exp`.

Two consequences worth understanding:

- **The middleware's parameter handling comes back out.** Because the cookie is set *before* you navigate to the page under test, that page is rendered in the chosen arm from the first byte. The edge-side parsing existed only to beat the flash the client-side override caused; there's no flash to beat any more.
- **The switcher isn't served in production.** The request is handed to the baked assets instead, so a shared link or a crawler can't force an arm, and the route stays transparent to anything ever baked at that path. The flip side is the deliberate trade: you can't force an arm on live production. Staging bakes the same content, and devtools still covers the rare desktop-on-prod case.

Under plain `make up` there's no middleware at all, so the client-side stand-in stays for local dev, now behind `ENV === "development"` — a build-time constant, so it's compiled out of the staging and production bundles rather than merely inert.

**Needs a follow-up:** #7215 is stacked on this branch and its testing guidance says to force an arm with `/latest?exp-latest-sticky-filters-v1=<arm>` "locally or on staging". That still works locally but no longer on staging; there it becomes `/exp?from=/latest`. Happy to update that description.

**Also worth knowing:** `experimentsMiddleware` still runs for `/exp`, as it does for every page, so loading the switcher can itself assign you an arm of an experiment whose `paths` cover `/` (`isUrlInPaths` treats `/` as a prefix matching everything). Harmless, and Clear is right there, but the alternative was putting a QA-specific path check back into the middleware, which is what this PR is removing.

## Testing guidance

On the staging server, open `/exp`. Each active experiment should show its current arm and a button per arm.

1. Tap an arm, then open one of the pages listed under that experiment. `<body>` should carry `exp-<id>--<arm>` **in the HTML as served** (check view-source or curl, not just devtools) with no other arm's class, and no visible swap on load.
2. `/exp?from=/grapher/life-expectancy` should return you straight to that page after picking an arm.
3. Tap Clear, reload the page under test: the middleware should assign an arm afresh and set a new cookie.
4. `curl -sI https://ourworldindata.org/exp` on production should not return the switcher.

Please also give the page a look **on an actual phone** — it's the case it exists for, and it's the one thing I couldn't check from the sandbox (I couldn't reach the staging host).

Locally under `make up`, `?exp-<id>=<arm>` still works as before: the param vanishes from the URL, the cookie is set, and the body class follows.

## Checklist

- [x] Does the change work in the archive? — unaffected; the archive has its own `runSiteFooterScriptsForArchive`, which never called this
- [ ] Changes to CSS/HTML were checked on Desktop and Mobile Safari at all three breakpoints — **not done**, see above
- [ ] Changes to HTML were checked for accessibility concerns — the page is plain HTML with no JS: real `<button>`s in a `<form>`, `aria-pressed` marking the current arm, 44px tap targets, `color-scheme: light dark`. Not screen-reader tested.
- [x] Google Analytics events were adapted to fit the changes in this PR — no GA changes needed, and the analytics risk the previous approach carried is gone: with no way to force an arm in production, forced sessions can't reach live experiment data

### Files

- **`functions/exp/index.ts`** (new) — the switcher. `handleExperimentSwitcher(request, env, configuredExperiments?)` behind an `onRequest` export; the third argument is only for tests, mirroring `experimentsMiddleware`.
    - `GET` renders the page: current arm per experiment (read from the request's cookies, flagging a cookie value that isn't an arm of that experiment), a button per arm with its fraction, Clear per experiment, Clear all, and the experiment's `paths` as links in a `<details>`.
    - `POST` sets or clears the cookie (same expiry and `path=/` as the middleware) and returns `303` to the sanitised `from` or back to `/exp`. A form POST rather than a `GET` link so a prefetch can't set a cookie.
    - Unknown experiment or unknown arm → `400`, no cookie. An empty arm value means clear, so the next request gets a fresh assignment.
    - `sanitizeReturnPath` keeps `from` to a same-origin path: rejects full URLs and the protocol-relative `//host` and `/\host` forms (open redirect), and control characters, which `Location` can't carry and which would otherwise take the route down with a 500.
    - `ENV === "production"` → `env.ASSETS.fetch(request)`. This started as a bare 404 and was changed deliberately: a 404 would have shadowed any page ever baked at `/exp`, silently and only in production.
    - `no-store` on both the page and the redirect; `noindex, nofollow` belt-and-braces.
- **`functions/_common/experiments.ts`** — the `parseExperimentOverrides` loop and import removed; back to reading cookies and assigning. The doc comment now spells out that an existing cookie is honoured and never reassigned, since that contract is what `/exp` rests on.
- **`site/runSiteFooterScripts.tsx`** — `applyExperimentOverrides()` is now `if (ENV === "development") applyExperimentOverrides()`.
- **`packages/@ourworldindata/utils/src/experiments/overrides.ts`** — unchanged behaviour; header rewritten to say it's development-only and to point at `/exp` for everywhere else.
- **`packages/@ourworldindata/utils/src/index.ts`** — barrel trimmed to `applyExperimentOverrides`, the only export consumed outside the module now that the middleware no longer needs `parseExperimentOverrides`. These were added in this PR and never released, so nothing external breaks.
- **`functions/README.md`** — a `/exp` section alongside the other routes.

### Verifying the bundle really loses the module

`ENV` is `define`d as a literal by `vite.config-common.mts`, so the guard should be dead-code eliminated and the module tree-shaken. Measured rather than assumed, by building `owid.mjs` twice with the guard inverted:

| | raw | gzip -9 | `history.state` occurrences |
| --- | --- | --- | --- |
| guard inverted (module kept) | 3,890,563 | 1,047,249 | 1 |
| guard as committed | 3,889,511 | 1,046,971 | 0 |

1,052 bytes of minified JS gone, and `history.replaceState(window.history.state, …)` — performed only by `applyExperimentOverrides` — is absent from the built bundle.

### Tests

The workerd e2e suite (`functions/test/experiments.e2e.*`) now covers both halves. The fixture worker dispatches `/exp` to the switcher and everything else to the middleware, and serves `/exp-in-production` as the same handler with `ENV` forced to `production` (with an `ASSETS` stub) so the environment gate is tested without booting a second workerd.

- Middleware: assigns and stamps `<body>` with no cookie; **honours an existing cookie without reassigning** — the property `/exp` depends on.
- Switcher: lists experiments and arms; marks the cookie's arm as current; sets the cookie and `303`s to the page under test; **the arm it set is what the next request renders**; clears one cookie; clears all; rejects an unknown arm and an unknown experiment without setting anything; refuses to redirect off-origin (`//host`, `/\host`, absolute URL, CRLF injection); does none of it in production, and sets no cookie there.

12 tests, all passing. Full unit suite green (2530 passed). `yarn typecheck`, `yarn testLintChanged`, `yarn testFormatChanged` clean.

### Alternatives considered

- **Keep the URL param, edge-only** (drop the client module, let the param stay in the URL). Zero browser bytes, but it doesn't remove QA code from production — it relocates it into the edge worker, where `ENV` is a runtime var from `wrangler.jsonc` rather than a build-time define, so the best available is a dormant branch on every request. And it still leaves a shareable arm-forcing link on production.
- **Edge-only plus a 302 to strip the param.** Same objection, plus a redirect on every QA request.
- **A browser extension.** Cross-environment is the easy part (`chrome.cookies` with host permissions for the three origins); mobile is what kills it, per above.
- **A bookmarklet.** Works everywhere the cookie does, but `javascript:` bookmarks are painful on iOS Safari and impractical on Chrome Android, and it can't set the cookie before the page under test is rendered, so the arm still flashes.
- **A page in the admin.** Cookies are per-origin, so it only works where admin and site share a host.

</details>

https://claude.ai/code/session_012gWKgoWCnucpArsHFDewPF


mlbrgl:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7217** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7217?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7175** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7175?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7215](https://github.com/owid/owid-grapher/pull/7215) <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7215?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

mlbrgl:

Closing this in favour of the direction we settled on at the site meeting on 8 September: an admin page that reads the currently active experiments and renders a form to set the arm cookies, with the form's state in the query string so a specific configuration can be shared as a link.

The point that decided it is one this PR only half satisfies. `/exp` keeps the switcher out of production by serving the baked assets instead when `ENV === "production"` — but the route still ships in the Cloudflare functions bundle, and the guarantee rests on a runtime environment variable being right. In the admin there is no such gate to get wrong: experiment-testing UI never leaves the admin in the first place, and it also has somewhere natural to live next to the rest of the experiment tooling.

What this PR got right and worth carrying over to the admin page:

- Forcing an arm is just setting the `exp-<id>` cookie. `experimentsMiddleware` honours an existing cookie and only assigns when there isn't one, so nothing in the middleware needs to change — the parameter parsing this PR originally added is not needed either way.
- Setting the cookie *before* navigating to the page under test means it renders in the chosen arm from the first byte, with no flash to paper over. Any client-side override has that flaw.
- A page with real buttons is the only thing that works on a phone, where devtools isn't available and extensions effectively aren't. Listing each experiment's enrolled pages as links keeps the typing down to the one URL.
- The caveat on cookie-setting: use a form POST, not a link, so a prefetch can't set an arm; and sanitise any return path to a same-origin path (`//host`, `/\host` and absolute URLs are open redirects).

The one thing the admin page cannot cover is that cookies are per-origin, so it only forces an arm where the admin and the site share a host. That holds on staging, which is where the testing happens.

The branch is kept and has been rebased onto `master`, so the implementation and its workerd e2e tests are there to lift from if useful.

Follow-up: #7215's `site/latest/README.md` documents forcing a sticky-filters arm via `/exp`, and now points at something that doesn't exist. That paragraph needs rewriting once the admin page lands.

PR #7225, opened 2026-09-08
🎉🤖 config-only chart editor, and the admin's editors moved onto it

Marigold:
## 👨 Summary

Wrap the grapher chart editor into its own component (and eventually its own npm package, #7235). The host passes in a config and a data store, the editor hands the edited config back:

```tsx
<GrapherEditor
    config={config}                            // grapher config, in the store's own form
    store={store}                              // where indicator data & metadata come from
    indicators={adminIndicatorCatalog(admin)}  // what "Add indicator" offers; OWID admin only
    // tabs: defaults to all that apply to the chart type
    onChange={(config) => setLiveConfig(config)}  // every edit
    onSave={(config) => saveSomewhere(config)}    // "Save config" button
/>
```

I weighed two options for the interface:

1. **Heavyweight** – an `EditorBackend` object the editor drives itself: `loadChart(id)`, `saveChart(id)`, and optional methods for logs, references, tags…
    - Closest to how the admin works today
    - But it bakes chart ids and our DB row types into the package API, and is awkward anywhere charts aren't rows in our database
2. **Lightweight** – the editor only manipulates a grapher config; the host loads it and writes it back wherever it lives (our DB, a YAML file in ETL, a JSON file) in onSave
    - Lets ETL, Wizard or a VS Code extension embed the editor without a backend
    - Same pattern as other embeddable editors (Plotly's react-chart-editor, amCharts Editor)

I went with 2. The remaining OWID-specific piece is where indicator data comes from, so that's a pluggable IndicatorStore: our Data API (R2) for OWID, a table or CSV for everyone else. A host's own columns are named the same way OWID's indicators are, by `dimensions[].slug` instead of `dimensions[].variableId` — which is what #7248 adds, and why this PR sits on top of it.

And then the admin runs on it too: both the chart editor page and the two narrative-chart pages, so the thing we'd hand to CoreEcon or ETL is the same code our own editors use every day, not a fork of it.

**Walkthrough for reviewers:** [Config in, config out](https://claude.ai/code/artifact/5895dfbc-daa7-43e0-8b70-7c12e02fb1e3) — the seam, the two decisions that shape the API, what the admin loses, and which six files to actually read.

**Try it:** [staging-site-narrative-on-grapher-editor/admin/editor-playground](http://staging-site-narrative-on-grapher-editor/admin/editor-playground) (Tailnet). Switch **Data store** to "Pasted CSV" for a chart with no OWID indicators, **Show code** for the props being passed, **Output config** for what a host gets back.

---

> _Written by Anthropic Claude (Fable 5.1, later edits Opus 5) — @Marigold at the wheel._

Sits on #7248 (`dimensions[].slug`), which lands first. This was three stacked PRs (#7212, #7220, #7225) and is now one, because none of the three was independently landable: the first alone adds an editor class only a playground uses, the second alone leaves the admin half-migrated. It is one refactor in three steps, below. The npm package on top of it stays separate in #7235.

The branch is a single commit on top of #7248, plus the fixes from review and one commit dropping the translation layer that #7248 makes unnecessary, so read it by file rather than by commit; the walkthrough and this description are organised by step instead.

_Review depth: deep on `ChartEditorPage.tsx` (the save/publish/delete actions moved there from the deleted `ChartEditor.ts`, worth diffing against it) and on `ConfigEditor.ts`; skim the four other new files and the playground; the tab components only lose code._

## What changed

**`ConfigEditor` + `IndicatorStore` + `GrapherEditor` (step 1).** The editor only knew how to edit a chart row in the admin database, with every indicator an OWID `variableId` fetched from the Data API. `ConfigEditor` takes a grapher config, edits it against the live preview, and hands it back through `onSave`/`onChange`. An `IndicatorStore` makes "where indicator data and metadata come from" pluggable: OWID's Data API is one implementation, a host's own table or CSV another. Where the config lives becomes the host's business. `GrapherEditor` is the component a host mounts, and `/admin/editor-playground` lets you try both stores and shows the exact props being passed, next to Plotly's `react-chart-editor` for comparison.

**The admin's chart editor page runs on it (step 2).** Everything a chart has because it is a database row (revision logs, references, redirects, pageviews, tags, publishing, inheritance from its indicator) stays in the admin and plugs in through two props. `ChartEditor` is deleted.

**The narrative-chart pages run on it (step 3).** `NarrativeChartEditor` goes too, and with it every remaining `instanceof` branch in the tabs. The editor slice is now one editor class with two hosts inside the admin and one outside.

## The two design decisions worth your attention

**Inheritance collapses to one prop, `baseConfig`:** "the config this one is a patch against". The editor shows the base's values as inherited, saves the diff, and re-applies a swapped base underneath the user's edits. What the base is made of stays in the admin: `ChartEditorPage` merges the indicator's config (when the toggle is on) with the chart's ETL layer, refetches it when the first y indicator changes, and owns the on/off toggle. The editor never hears the words indicator, ETL layer or inheritance toggle. A narrative chart uses the same prop with its parent chart's full config.

**What only the host knows arrives as callbacks.** Admin-only UI goes through `extraTabs` and `renderSaveButtons`. Revisions, Refs and a new **Publishing** tab (URL/slug, inheritance toggle, tags, force-datapage) for charts; a "Narrative chart" tab and Refs for narrative charts. An earlier revision injected these controls into the bottom of the generic tabs through an `extensions` prop with named hooks; that is gone, so the editor makes no promises about the inside of its own tabs.

Two smaller ones carry facts the editor cannot work out from the config it was handed: `renderNote("map.colorScale")` puts the "last edited by …" line back under the map colours (the answer is in the diff between two revisions, which only a host with a revision log has), and `originUrlSuggestions()` fills the Origin url dropdown, for which the admin offers the posts that already show this chart and then every published topic page. Both are called while the section renders, so a note or a list the host is still loading appears on its own — a plain data prop would be swallowed by mobx-react's shallow prop compare.

Rejected: keeping `ChartEditor`/`NarrativeChartEditor` as subclasses. It would have kept the `instanceof` branches in the tabs and left the package with three editor classes to explain.

## Pitfalls

**Two small features are dropped:**

- the preview link's `?forceDatapage=true` query, so previewing an *unsaved* datapage toggle is lost; the preview route falls back to the chart's saved setting. Deliberate — every chart page is headed for a data page anyway
- the narrative Debug tab's bespoke "diff to parent", replaced by the generic "diff to base config" that admin charts with inheritance now get too

Two others were dropped and are back, through the callbacks above: the "last edited by …" note on the Map tab's colour scale, and the referenced-post suggestions in the Origin url autocomplete. The suggestions changed shape on the way: the Text tab used to fetch `/api/gdocs/publishedTopicSlugs` itself, and now takes the whole list from the host, so the editor no longer calls the admin API for it and narrative charts get the referenced posts too.

**The narrative patch is not the editor's generic diff.** `makeNarrativeChartPatchConfig` drops `id`/`slug`/`isPublished`/`version` and always persists chart type, tab and entity selection even when they match the parent. Rather than teach the editor a second diff strategy, the page computes that patch in `onSave`. Consequence: the narrative Debug tab and the inherited markers show the generic diff, not the saved one. The saved shape is unchanged.

**The slug field left the Text tab** for the top of the admin's Publishing tab. It was hidden for narrative charts by an `instanceof` check; with that gone, a chart's URL reads as a publishing concern. Narrative charts and CSV hosts don't see it, which is right — their configs never carry a slug.

## How to test it

Open [the playground on staging](http://staging-site-narrative-on-grapher-editor/admin/editor-playground) (or `/admin/editor-playground` on a dev stack). Switch **Data store** to "Pasted CSV" to see a chart built from a CSV with no OWID indicators; edit it, hit **Output config** to watch the live config and **Save config** to see what a host would receive. **Load config** takes your own JSON, and in CSV mode your own CSV and column definitions.

Then use the admin's own chart editor as you normally would and see whether anything feels different. The Publishing tab is the one new place to look.

Two spots worth a click, because they broke and were repaired: the **Map** tab of [chart 9052](http://staging-site-narrative-on-grapher-editor/admin/charts/9052/edit), where "Color scale" should read *"Last edited … by Bastian Herre"* — pointing at the 6 Sep 09:34 revision, not the two later saves that left the colours alone — and the **Origin url** field on the Text tab of a chart with references, which should offer those posts first, marked "(referenced by this chart)", ahead of the topic pages.

## How it works

The editor slice reached the admin API in two ways: the pages fetched chart-record data (patch, parents, logs, refs, tags) and passed it in through the existing `Manager` interface, and `ChartEditorView` itself called four endpoints directly for the indicator catalog and details-on-demand. The first already worked as a seam; the second became optional **providers** on the view manager (`indicators`, `details`), plus an optional `environment` for URLs. A host without providers gets an editor with no "Add indicator" button and no DoD validation, which is the right degradation.

**`IndicatorStore`.** Only the editor was welded to OWID indicators: every dimension had to be a `variableId` fetched from the Data API. A store makes that pluggable — it hands the editor a table holding the columns its dimensions name, with column metadata attached, and tells the picker which columns exist. `dataApiIndicatorStore` wraps today's caching fetcher and is the default; `tableIndicatorStore`/`csvIndicatorStore` serve a table the host already has, under the host's own column slugs. Dimensions name those columns by `slug` (#7248), so there is one config form and no translation layer: what the host passes in is what the editor edits and what `onSave` hands back, `display` overrides included. The result renders with `GrapherLoader.fromCsv` as is. Column metadata (name, unit, description, sources) is read-only under every store: it belongs to the source, not the chart.

**The flat form is read, not written.** A config that names its columns the short way (`ySlugs`, `xSlug`, `colorSlug`, `sizeSlug`) is turned into dimensions on the way in and comes back naming dimensions — valid, renders identically, but not the shape it arrived in. Nothing infers columns: a config naming none opens an empty chart and the picker, rather than plotting every numeric column in the table.

**Saving is the host's:** `onSave(config, editor)` does the PUT/POST and returns the server's `savedPatch`, which the editor then treats as the saved state, keeping today's behaviour where the server normalises the patch.

**`GrapherEditor` creates its editor once per mount** rather than as a `computed`. The computed version was rebuilt whenever the host re-rendered with new props (e.g. after a save pushed a new revision log), which reset the chart being edited. Hosts that change the store or the config remount via `key`.

**Props:** `config`, `store`, `onSave` required; `onChange`, `indicators`, `details`, `tabs`, `baseConfig`, `initialQueryParams`, `extraTabs`, `renderSaveButtons`, `renderNote`, `originUrlSuggestions` optional. `indicators` defaults to `store.catalog`; `null` disables the picker. The population/GDP shortcut ids moved onto the store (`variableIdsByCatalogPath`), since which indicator is "population" is a store question.

**`EditorTab` and `tab`:** `editor.tab` is a `string` so host tabs can use their own keys; `?tab=` from the URL is accepted if it is a built-in tab or one of the host's. Host tabs slot in before Export.

**Deleted:** `ChartEditor.ts` and `NarrativeChartEditor.ts`. Their non-class exports (`Log`, `NarrativeChartMinimalInformation`, `References`, `getFullReferencesCount`, `deleteChart`, `fetchChartConfigByIndicatorId`) moved to `adminChartApi.ts`; `findLastMapColorScaleEdit` moved there too, with tests, and feeds the note through `renderNote`. `AbstractChartEditor.references` is gone; the Refs tab label with its count is the page's.

**Tabs touched:** `EditorBasicTab` (tags/narrative sections and `updateParentConfig` gone), `EditorTextTab` (chart-only branches and the slug field gone), `EditorMapTab`/`EditorColorScaleSection` (the note now comes from `renderNote`), `EditorDebugTab` (config-only, with the generic base-config diff), `EditorHistoryTab` (takes `logs`), `EditorReferencesTab` (chart and narrative variants take props), `SaveButtons` (config-only; `ChartSaveButtons` and `NarrativeChartSaveButtons` are the pages').

**Create-page navigation:** after a successful POST the page marks the editor saved and sets `createdId`, and a `<Redirect>` navigates after React has re-rendered. A direct `history.push` ran before that render and fired the unsaved-changes prompt on our own redirect.

**Why the reaction reorder in `ChartEditorView.componentDidMount`:** without an indicator catalog there is nothing to `await` in `fetchData`, so `_isDbSet` flips synchronously inside `refresh()`. The reaction on `this.editor` was registered after `refresh()` and never saw the transition, leaving the preview at "No table loaded yet". Registering reactions first fixes both cases.

## Verified

Headless Chrome on the worktree dev stack against the local DB, plus two agents doing exploratory QA on the staging server (every issue they found is fixed in the branch).

- **Playground, Data API store:** loads the example config from the public Data API, all tabs render, editing the title updates the preview and the live config; "Save config" hands the edited config to `onSave` and the button disables until the next edit.
- **Playground, CSV store:** a made-up three-city CSV renders with the columnDefs' names, units and source in the footer; the picker lists the CSV's numeric columns; adding the second puts both series on the chart; the saved config comes back as `dimensions: [{ property: "y", slug: "rent_index" }, …]`, with no `variableId` anywhere.
- **Indicator picker off:** the chart still loads, existing dimensions stay editable, "Add indicator" is gone. The "lite" preset shows Basic/Data/Text/Customize/Map only.
- **`/admin/charts/64/edit`:** all tabs including Revisions and Refs (38), the five save buttons, the tags section, the preview link. Edited the title, hit **Update chart**: one `PUT /admin/api/charts/64?inheritance=disable&forceDatapage=false`, `success: true`, the editor kept the edit, a new revision at the top of Revisions. Reverted. Publishing tab shows URL, Inheritance (toggle + indicator link), Tags and Data page; switching inheritance on makes a "Base config" section appear on the Debug tab and off removes it. `/admin/charts/create` renders with **Create draft** only and the "can't tag until saved" note.
- **Narrative chart 1:** tabs Basic, Data, Text, Customize, Scatter, Narrative chart, Refs (0), Export, Debug; buttons Save narrative chart, Go to parent chart, Create DI. Edited the title and saved: one PUT with `success: true`, the patch carries the new title with `chartTypes`/`tab`/`selectedEntityNames` persisted and `id`/`slug`/`isPublished`/`version` absent. Reverted.
- **Narrative create page** with a multi-dim view config id and `grapherQueryStr=tab=table`: opens on the Table tab, creating navigates to the new chart's editor with no unsaved-changes prompt, the record exists via the API (deleted afterwards). With a plain chart's config id the server's 404 surfaces through the admin's error modal and the page stays put.
- **Unit tests:** `ConfigEditor.test.ts` (tab derivation, `onSave` receiving the config and clearing `isModified`, `onError` on a failed save, `onChange` firing), `ConfigEditorInheritance.test.ts` (base/patch merge, inherited-property detection, saving only the diff, a swapped base re-applied under the user's edits, the base going away), `indicatorStores.test.ts` (catalog from a CSV, the table served under the host's own slugs with its metadata). All admin client tests pass.

**Side-by-side against master.** A second staging server, `staging-site-parity-baseline`, runs the exact master commit this branch sits on, restored from the same data dump, so chart ids and configs match. Two agents drove both admins through the same six flows and I diffed the saved configs afterwards through the API:

| flow | chart | saved config identical across servers |
| --- | --- | --- |
| edit the title | 9052 | yes |
| add an indicator ("Population (historical)") | 64 | yes |
| switch chart type (stacked area → line) | 383 | yes |
| change the map colour scheme | 9052 | yes |
| toggle inheritance off | 7118 | yes |
| edit a narrative chart's subtitle | narrative 1 | yes |
| create a chart from scratch | new | yes (ids aside) |
| unpublish, then republish | 9052 | yes |

Every save produced a byte-identical patch config on both servers, and identical rendered configs. Creating a chart POSTs and lands on `/admin/charts/<id>/edit` with no unsaved-changes prompt on either server, and the two created charts differ only in id. Unpublish and republish fire the same requests and the same confirm dialogs, word for word including the references warning, and 9052 ends published on both. With the Origin url field cleared, chart 64's dropdown offers the same six entries in the same order on both servers, the five referencing posts each marked "(referenced by this chart)" ahead of the topic page — so the restored suggestions match master's byte for byte. Chart 383's save also normalised ~30 default-valued keys out of its stored patch — identically on both servers, so that is pre-existing `diffGrapherConfigs` behaviour, not something this PR introduces. The "Last edited … by …" note under the map colour scale reads word for word the same on both. The differences the agents found in the UI were the relocations this PR makes (Publishing tab; the narrative-chart section promoted to its own tab) plus one: master's inheritance copy rewrites itself when you flip the toggle, the Publishing tab's does not — the toggle state is the signal there instead.

One bug candidate was raised and disproved: chart 383's Y indicator briefly shows as its numeric id before the table loads. It does that on master too, for the same sub-second window.

Not verified: save-as-new, delete and save-as-narrative-chart beyond rendering (they open new tabs or destroy the row; the code moved verbatim); "Create DI" beyond rendering; the narrative Refs tab with non-zero references.

## Follow-ups this sets up

A public indicator list on R2 so the Data API store has a catalog outside the admin, and the extraction into `packages/@ourworldindata/grapher-editor`, which is #7235.

</details>

### DIFF
--- a/adminSiteClient/EditorBasicTab.tsx
+++ b/adminSiteClient/EditorBasicTab.tsx
@@ -41,6 +41,7 @@ import {
 } from "./EntityPresets.js"
 import {
     DimensionProperty,
+    ColumnSlug,
     OwidVariableId,
     OwidChartDimensionInterface,
     areSetsEqual,
@@ -129,10 +130,10 @@ export class DimensionSlotView<
         this.updateParentConfig()
     }
 
-    @action.bound private onRemoveDimension(variableId: OwidVariableId) {
+    @action.bound private onRemoveDimension(columnSlug: ColumnSlug) {
         void this.updateDimensionsAndRebuildTable(
             this.props.slot.dimensions.filter(
-                (d) => d.variableId !== variableId
+                (d) => d.columnSlug !== columnSlug
             )
         )
         this.updateParentConfig()
@@ -340,7 +341,8 @@ export class DimensionSlotView<
         const { isSelectingVariables } = this
         const { slot, editor, canSwapXAndY, onSwapXAndY } = this.props
         const dimensions = slot.dimensions.map((dim, index) => ({
-            id: dim.variableId,
+            // Every slot has a column slug; only indicator-backed ones have an id.
+            id: dim.columnSlug,
             dim,
             index,
         }))
@@ -386,7 +388,7 @@ export class DimensionSlotView<
                                     slot.isOptional
                                         ? () =>
                                               this.onRemoveDimension(
-                                                  d.dim.variableId
+                                                  d.dim.columnSlug
                                               )
                                         : undefined
                                 }
--- a/adminSiteClient/VariableSelector.tsx
+++ b/adminSiteClient/VariableSelector.tsx
@@ -528,15 +528,19 @@ export class VariableSelector<
         const { variableUsageCounts } = this.database
         const { dimensions } = this.props.slot
 
-        this.chosenVariables = dimensions.map((d) => {
+        this.chosenVariables = dimensions.flatMap((d) => {
+            // A slot naming a host-supplied column has no variable to pick.
+            const variableId = d.variableId
+            if (variableId === undefined) return []
+
             const { datasetName, datasetId } = d.column
             const dataset =
                 datasetId !== undefined ? datasetsById[datasetId] : undefined
 
             return {
                 name: d.column.name,
-                id: d.variableId,
-                usageCount: variableUsageCounts.get(d.variableId) ?? 0,
+                id: variableId,
+                usageCount: variableUsageCounts.get(variableId) ?? 0,
                 datasetId: datasetId ?? 0,
                 datasetName: datasetName || "",
                 catalogPath: undefined,
--- a/baker/updateChartEntities.ts
+++ b/baker/updateChartEntities.ts
@@ -18,6 +18,7 @@ import {
     OwidVariableDataMetadataDimensions,
     DbRawChartConfig,
 } from "@ourworldindata/types"
+import { excludeUndefined } from "@ourworldindata/utils"
 import * as db from "../db/db.js"
 import pMap from "p-map"
 import { mapEntityNamesToEntityIds } from "../db/model/Entity.js"
@@ -94,7 +95,9 @@ const obtainAvailableEntitiesForGrapherConfig = async (
     })
 
     // Manually fetch data for grapher, so we can employ caching
-    const variableIds = _.uniq(grapher.dimensions.map((d) => d.variableId))
+    const variableIds = _.uniq(
+        excludeUndefined(grapher.dimensions.map((d) => d.variableId))
+    )
     const variableData: MultipleOwidVariableDataDimensionsMap = new Map(
         await pMap(variableIds, async (variableId) => [
             variableId,
--- a/db/model/archival/archivalDb.ts
+++ b/db/model/archival/archivalDb.ts
@@ -412,6 +412,7 @@ export const getExplorerChecksumsFromDb = async (
         if (config.dimensions) {
             for (const dimension of config.dimensions) {
                 const variableId = dimension.variableId
+                if (variableId === undefined) continue
                 variablesByExplorerSlug.get(slug)?.add(variableId)
                 allVariableIds.add(variableId)
             }
--- a/devTools/svgTester/utils.ts
+++ b/devTools/svgTester/utils.ts
@@ -19,6 +19,7 @@ import {
     OwidVariableMixedData,
     OwidVariableWithSourceAndDimension,
     TESTING_ONLY_disable_guid,
+    excludeUndefined,
 } from "@ourworldindata/utils"
 import fs, { stat } from "fs-extra"
 import path from "path"
@@ -473,7 +474,9 @@ export async function saveGrapherSchemaAndData(
     )
 
     const grapher = initGrapherForSvgExport(config)
-    const variableIds = grapher.grapherState.dimensions.map((d) => d.variableId)
+    const variableIds = excludeUndefined(
+        grapher.grapherState.dimensions.map((d) => d.variableId)
+    )
 
     await Promise.allSettled([
         promise1,
--- a/packageDocs/docs/chart-config/index.md
+++ b/packageDocs/docs/chart-config/index.md
@@ -22,4 +22,6 @@ and also shipped with the package as `@ourworldindata/grapher/grapher-schema.jso
 
 A browsable, field-by-field rendering of that schema is available in the [schema reference](../schema-reference/index.md).
 
+A dimension names its column either by `variableId`, an OWID indicator, or by `slug`, a column of the table you supply; see [overriding metadata for one chart](../loading-data.md#overriding-metadata-for-one-chart).
+
 One note on requiredness: the schema describes **persisted** configs, which must carry `$schema` and `dimensions`. Configs passed to `GrapherLoader` are looser — `$schema` is never needed, and `dimensions` is only required for [`fromApi`](../api/index.md#grapherloaderfromapi-config-dataapiurl), where it says which indicators to fetch.
--- a/packageDocs/docs/loading-data.md
+++ b/packageDocs/docs/loading-data.md
@@ -118,3 +118,37 @@ GrapherLoader.fromCsv({
 ```
 
 The footer's attribution line is assembled from `sourceName` and the origins' producers (or set it directly via the config's `sourceDesc`). `additionalInfo`, `descriptionFromProducer`, and `presentation` (e.g. `titlePublic`, `attributionShort`) are also supported — see the `OwidColumnDef` type for the full surface.
+
+## Overriding metadata for one chart
+
+A column definition describes the column itself, so every chart drawn from that
+table sees it. When one chart needs to say something different, put a
+`dimensions` entry in its config that names the column by `slug` and carries a
+`display` block. It is laid over the column's own definition, and only for that
+chart:
+
+```js
+GrapherLoader.fromCsv({
+    config: {
+        title: "Cows per capita",
+        dimensions: [
+            {
+                property: "y",
+                slug: "cows", // a column of your table, not an OWID indicator
+                display: { name: "Cattle", numDecimalPlaces: 1 },
+            },
+        ],
+    },
+    csvUrl: "./cows.csv",
+    columnDefs: [{ slug: "cows", type: "Numeric", name: "Cows per capita" }],
+}).mount(container)
+```
+
+A dimension is the same object OWID's own charts use to name an indicator, with
+`slug` in place of `variableId`, so anything `display` supports works here too.
+`ySlugs` and its siblings remain the short way to say the same thing when no
+override is needed.
+
+If you build the table yourself and hand it to `fromTable`, this happens for
+you. If you assign `grapherState.inputTable` directly, call
+`applyDimensionDisplayOverrides(table, config.dimensions)` first.
--- a/packages/@ourworldindata/explorer/src/Explorer.tsx
+++ b/packages/@ourworldindata/explorer/src/Explorer.tsx
@@ -666,7 +666,9 @@ export class Explorer
             (dim) => dim.type === "slug"
         )
 
-        const yVariableIds = variableIdDimensions.map((dim) => dim.variableId)
+        const yVariableIds = excludeUndefined(
+            variableIdDimensions.map((dim) => dim.variableId)
+        )
         const ySlugs = slugDimensions.map((dim) => dim.slug)
 
         const partialGrapherConfig =
@@ -781,7 +783,7 @@ export class Explorer
         // Sort dimensions to match the order defined in config.ySlugs
         config.dimensions = _.sortBy(
             dimensions,
-            (dim) => indexMap[dim.slug ?? dim.variableId]
+            (dim) => indexMap[dim.slug ?? dim.variableId ?? ""]
         )
 
         this.inputTableTransformer = (table: OwidTable) => {
--- a/packages/@ourworldindata/grapher/src/chart/ChartDimension.test.ts
+++ b/packages/@ourworldindata/grapher/src/chart/ChartDimension.test.ts
@@ -1,4 +1,4 @@
-import { expect, it } from "vitest"
+import { describe, expect, it } from "vitest"
 
 // todo: remove this when we remove chartDimension
 
@@ -13,4 +13,39 @@ it("can serialize for saving", () => {
             { table: BlankOwidTable() }
         ).toObject()
     ).toEqual({ property: "x", variableId: 1 })
+})
+
+describe("a slot naming a host-supplied column", () => {
+    const manager = { table: BlankOwidTable() }
+
+    it("uses the authored slug as its column, with no variable id", () => {
+        const dimension = new ChartDimension(
+            { property: DimensionProperty.y, slug: "rent_index" },
+            manager
+        )
+        expect(dimension.columnSlug).toBe("rent_index")
+        expect(dimension.variableId).toBeUndefined()
+    })
+
+    it("writes the authored slug back out", () => {
+        const config = {
+            property: DimensionProperty.y,
+            slug: "rent_index",
+            display: { name: "Rents" },
+        }
+        const dimension = new ChartDimension(config, manager)
+        expect(dimension.toObject()).toEqual(config)
+    })
+
+    it("does not write out a slug derived from a variable id", () => {
+        const dimension = new ChartDimension(
+            { property: DimensionProperty.y, variableId: 815383 },
+            manager
+        )
+        expect(dimension.columnSlug).toBe("815383")
+        expect(dimension.toObject()).toEqual({
+            property: DimensionProperty.y,
+            variableId: 815383,
+        })
+    })
 })
--- a/packages/@ourworldindata/grapher/src/chart/ChartDimension.ts
+++ b/packages/@ourworldindata/grapher/src/chart/ChartDimension.ts
@@ -19,7 +19,8 @@ import { OwidTable, CoreColumn } from "@ourworldindata/core-table"
 // and a particular variable that it requests as data
 class ChartDimensionDefaults implements OwidChartDimensionInterface {
     property!: DimensionProperty
-    variableId!: OwidVariableId
+    // Undefined when the slot names a host-supplied column by slug instead.
+    variableId?: OwidVariableId
 
     // check on: malaria-deaths-comparisons and computing-efficiency
 
@@ -97,14 +98,24 @@ export class ChartDimension
 
         deleteRuntimeAndUnchangedProps(obj, new ChartDimensionDefaults())
 
+        // An authored slug is part of the config and has to survive the round
+        // trip; a slug derived from the variable id is not.
+        if (this._slug !== undefined) obj.slug = this._slug
+
         return trimObject(obj)
     }
 
-    // Do not persist yet, until we migrate off VariableIds
+    /**
+     * The slug as authored, when the config named a column of the host's own
+     * table. Indicator-backed dimensions leave this unset and derive their
+     * slug from the variable id instead, which is why it is only written back
+     * out in `toObject` when it was authored.
+     */
     _slug: ColumnSlug | undefined = undefined
 
     @computed get slug(): ColumnSlug {
         if (this._slug) return this._slug
+        if (this.variableId === undefined) return ""
         return getDimensionColumnSlug(this.variableId, this.targetYear)
     }
 
@@ -117,6 +128,6 @@ export class ChartDimension
     }
 
     @computed get columnSlug(): string {
-        return this.slug ?? this.variableId.toString()
+        return this.slug
     }
 }
--- a/packages/@ourworldindata/grapher/src/core/LegacyToOwidTable.ts
+++ b/packages/@ourworldindata/grapher/src/core/LegacyToOwidTable.ts
@@ -52,12 +52,20 @@ export const legacyToOwidTableAndDimensionsWithMandatorySlug = (
         | { [entityName: string]: string | undefined }
         | undefined
 ): OwidTable => {
-    const dimensionsWithSlug = dimensions?.map((dimension) => ({
-        ...dimension,
-        slug:
+    // A slot names its column either by slug, which the host's table already
+    // has, or by variable id, whose column is named on assembly. A slot with
+    // neither names nothing and is dropped.
+    const dimensionsWithSlug = dimensions?.flatMap((dimension) => {
+        const slug =
             dimension.slug ??
-            getDimensionColumnSlug(dimension.variableId, dimension.targetYear),
-    }))
+            (dimension.variableId !== undefined
+                ? getDimensionColumnSlug(
+                      dimension.variableId,
+                      dimension.targetYear
+                  )
+                : undefined)
+        return slug !== undefined ? [{ ...dimension, slug }] : []
+    })
     return legacyToOwidTableAndDimensions(
         json,
         dimensionsWithSlug,
@@ -96,6 +104,8 @@ export const legacyToOwidTableAndDimensions = (
     const variableTablesToJoinByDay: OwidTable[] = []
     const variableTablesWithYearToJoinByEntityOnly: OwidTable[] = []
     for (const dimension of dimensionColumns) {
+        // Slots naming a host-supplied column have no indicator to convert.
+        if (dimension.variableId === undefined) continue
         const variable = json.get(dimension.variableId)
 
         // TODO: this shouldn't happen but it does sometimes
--- a/packages/@ourworldindata/grapher/src/core/applyDimensionDisplay.test.ts
+++ b/packages/@ourworldindata/grapher/src/core/applyDimensionDisplay.test.ts
@@ -1,0 +1,90 @@
+import { expect, it, describe } from "vitest"
+import { OwidTable } from "@ourworldindata/core-table"
+import {
+    ColumnTypeNames,
+    DimensionProperty,
+    type OwidColumnDef,
+} from "@ourworldindata/types"
+import { applyDimensionDisplayOverrides } from "./applyDimensionDisplay.js"
+
+const csv = `entityName,year,rent_index,vacancy_rate
+Berlin,2015,100,3.1
+Berlin,2020,131.4,1.2
+Vienna,2015,100,4
+Vienna,2020,112.7,3.8`
+
+const columnDefs: OwidColumnDef[] = [
+    {
+        slug: "rent_index",
+        type: ColumnTypeNames.Numeric,
+        name: "Rent index",
+        unit: "index (2015 = 100)",
+        display: { numDecimalPlaces: 1 },
+    },
+    {
+        slug: "vacancy_rate",
+        type: ColumnTypeNames.Numeric,
+        name: "Vacancy rate",
+    },
+]
+
+const makeTable = (): OwidTable => new OwidTable(csv, columnDefs)
+
+describe(applyDimensionDisplayOverrides, () => {
+    it("lays a slot's display over the column's own definition", () => {
+        const table = applyDimensionDisplayOverrides(makeTable(), [
+            {
+                property: DimensionProperty.y,
+                slug: "rent_index",
+                display: { name: "Rents", numDecimalPlaces: 0 },
+            },
+        ])
+
+        const column = table.get("rent_index")
+        // The chart's name and rounding win...
+        expect(column.displayName).toBe("Rents")
+        expect(column.numDecimalPlaces).toBe(0)
+        // ...and what the chart says nothing about is left alone.
+        expect(column.unit).toBe("index (2015 = 100)")
+    })
+
+    it("leaves columns no slot points at untouched", () => {
+        const table = applyDimensionDisplayOverrides(makeTable(), [
+            {
+                property: DimensionProperty.y,
+                slug: "rent_index",
+                display: { name: "Rents" },
+            },
+        ])
+
+        expect(table.get("vacancy_rate").displayName).toBe("Vacancy rate")
+    })
+
+    it("returns the same table when no slot overrides anything", () => {
+        const table = makeTable()
+        expect(applyDimensionDisplayOverrides(table, undefined)).toBe(table)
+        expect(
+            applyDimensionDisplayOverrides(table, [
+                { property: DimensionProperty.y, slug: "rent_index" },
+                { property: DimensionProperty.y, variableId: 42 },
+                {
+                    property: DimensionProperty.y,
+                    slug: "not_a_column",
+                    display: { name: "Nope" },
+                },
+            ])
+        ).toBe(table)
+    })
+
+    it("ignores indicator-backed slots, whose display is folded in on assembly", () => {
+        const table = applyDimensionDisplayOverrides(makeTable(), [
+            {
+                property: DimensionProperty.y,
+                variableId: 815383,
+                display: { name: "Should not apply" },
+            },
+        ])
+
+        expect(table.get("rent_index").displayName).toBe("Rent index")
+    })
+})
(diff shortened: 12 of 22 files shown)


Marigold:
@marcel @sophiamersmann ready for review whenever you have time. No rush!

I've added back "Last changed by ..." in the palette and listing Origins in the dropdown as they seemed really useful. I dropped the preview link with `?forceDataPage=1`. And as I said, some things have moved to `Publish` tab. Other than that it should be identical.

I ran an extensive parity comparison between this PR and staging, and everything matched ✅. I skimmed the code but invested most of my time in testing via agents.

I'll merge this after https://github.com/owid/owid-grapher/pull/7248 that adds `dimensions[].slug`.

adminSiteClient/AbstractChartEditor.ts
```diff
@@ -163,16 +175,31 @@ export abstract class AbstractChartEditor<
...
 
+        // The saved baseline: set once the config (and its data, if it has
+        // any) is in, so a freshly opened chart doesn't count as modified.
         when(
-            () => this.grapherState.hasData && this.grapherState.isReady,
+            () =>
+                this.grapherState.isReady &&
+                (this.grapherState.hasData ||
+                    this.grapherState.dimensions.length === 0),
             () => (this.savedPatchConfig = this.patchConfig)
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Confirmed — the predicate was already true in the constructor, so the baseline was the empty config and every chart opened modified. Fixed in a834d782b5: the view calls `markAsSaved()` once it has applied the config and loaded its data, with a test that pins the constructor not to take a baseline.

adminSiteClient/ConfigEditor.ts
```diff
@@ -0,0 +1,155 @@
...
+                },
+                { equals: comparer.structural }
+            )
+        )
+    }
+
+    /** The host's config, translated into the dimension-based form the
+     *  editor works on (identity for the Data API store). */
+    override get originalGrapherConfig(): GrapherInterface {
+        return this.store.toEditorConfig(super.originalGrapherConfig)
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Real, and fixed in a834d782b5 — though not by translating the layers separately, which has the same problem from the other side: `dimensions` replace one another wholesale in a merge while `ySlugs`/`colorSlug` merge field by field, so a base's y column would still be dropped by a patch that only names a colour. The layers are now merged in the host's own form and the result translated once (`fromHostConfig`), with tests for both directions.

adminSiteClient/ConfigEditor.ts
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            saved = await this.manager.onSave(hostConfig, this)
+        } catch {
+            onError?.()
+            return
+        }
+        runInAction(() => {
+            this.savedPatchConfig = saved
+                ? this.store.toEditorConfig(saved)
+                : patchConfig
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Confirmed, and it bit the admin: a chart whose title is derived from the data stayed modified after a successful save. Fixed in a834d782b5 — the editor adopts what the host says it stored, but only when it differs in something the user authored (`version` and `id` excluded), so an ordinary save re-applies nothing.

adminSiteClient/ConfigEditor.ts
```diff
@@ -0,0 +1,155 @@
...
+                    const { patchConfig } = this
+                    runInAction(() => {
+                        this.parentConfig = base
+                    })
+                    this.updateLiveGrapher(
+                        mergeGrapherConfigs(
+                            this.activeParentConfig ?? {},
+                            patchConfig
+                        )
+                    )
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Fixed in a834d782b5: the base-swap reaction reloads the table. No host swaps a base that names columns today, but the store caches, so the reload is close to free when nothing changed.

adminSiteClient/NarrativeChartSaveButtons.tsx
```diff
@@ -0,0 +1,160 @@
...
+        const { editor } = this.props
+        // Save the narrative chart first if there are unsaved changes
+        if (editor.isModified) {
+            const shouldSave = window.confirm(
+                "You have unsaved changes to this narrative chart. The Data Insight will use the saved version. Do you want to save your changes now before creating the DI?"
+            )
+            if (!shouldSave) return
+            await editor.saveGrapher()
+        }
+        this.isCreateDataInsightModalOpen = true
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Fair, though it predates this PR — master's `SaveButtons` did the same. Fixed anyway in a834d782b5: the save reports failure through `onError` and the modal doesn't open.

adminSiteClient/EditorDebugTab.tsx
```diff
@@ -269,18 +72,29 @@ class EditorDebugTabForNarrativeChart extends Component<{
...
-        const { patchConfig, parentConfig, fullConfig, parentUrl } =
-            this.props.editor
+    @action.bound copyYamlToClipboard() {
+        // Avoid modifying the original JSON object
+        // Due to mobx memoizing computed values, the JSON can be mutated.
+        const patchConfig = {
+            ...this.props.editor.patchConfig,
+        }
+        delete patchConfig.id
+        delete patchConfig.dimensions
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Fixed in a834d782b5: the button serialises `hostConfig`, so a slug-based config keeps its bindings and an OWID one still drops `dimensions` for the ETL.

adminSiteClient/ChartEditorView.tsx
```diff
@@ -382,39 +372,50 @@ export class ChartEditorView<
...
+        const activeTab = availableTabs.includes(editor.tab)
+            ? editor.tab
+            : availableTabs[0]
+
+        // Hosts of the config-only editor plug their own tabs, save buttons
+        // and preview link in; the admin's chart-record editors get the
+        // built-in ones.
+        const configEditor = isConfigEditorInstance(editor) ? editor : undefined
+        const extraTabs = configEditor?.manager.extraTabs ?? []
+        const activeExtraTab = extraTabs.find((tab) => tab.key === editor.tab)
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Fixed in a834d782b5: `activeExtraTab` resolves from `activeTab`.

adminSiteClient/ChartEditorView.tsx
```diff
@@ -382,39 +372,50 @@ export class ChartEditorView<
...
+        const tabLabel = (tab: string): React.ReactNode =>
+            extraTabs.find((t) => t.key === tab)?.label ?? _.capitalize(tab)
+        // A config that carries a chart id came from the admin database, so
+        // the admin's preview page can show it.
+        const previewUrl = grapherState.id
+            ? `/admin/charts/${grapherState.id}/preview`
+            : undefined
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Agreed on the boundary, and this is the last admin URL the editor built for itself (master inferred the same one). Fixed in a834d782b5: `previewUrl` is a host prop. The admin passes its chart's preview page and the two narrative pages pass the parent chart's, which is what they linked to before; a config-only host gets no link.

adminSiteClient/ChartEditorPage.tsx
```diff
@@ -49,134 +87,187 @@ export class ChartEditorPage
...
+                // ignore a late answer for an indicator that was swapped again
+                if (this.indicatorId === newId) this.indicatorConfig = config
+            })
+        )
     }
 
-    async fetchRefs(): Promise<void> {
+    private async fetchChartJson<T>(
+        suffix: string,
+        pick: (json: any) => T,
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Fixed in a834d782b5: the helper takes `Json`, the type `admin.getJSON` already resolves to.

adminSiteClient/ChartEditorPage.tsx
```diff
@@ -49,134 +87,187 @@ export class ChartEditorPage
...
+                : undefined
+            runInAction(() => {
+                this.patchConfig = grapherConfig
+                this.indicatorConfig = parentConfig
+                this.indicatorId = parentIndicatorId
+            })
         }
+        // The store is built with these, so they have to be in before mount.
+        await this.fetchVariableIdsByCatalogPath()
+        runInAction(() => (this.isLoaded = true))
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Fixed in a834d782b5: the lookup can fail without keeping the editor behind its loading blocker — the population/GDP shortcuts are simply absent then.

adminSiteClient/ConfigEditor.ts
```diff
@@ -0,0 +1,177 @@
...
+        let saved: GrapherInterface | void
+        try {
+            saved = await this.manager.onSave(hostConfig, this)
+        } catch {
+            onError?.()
+            return
+        }
+        runInAction(() => {
+            const savedPatch = saved
+                ? this.store.toEditorConfig(saved)
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

Right, thanks — fixed in f927f77d0a, with the inference conditional rather than off: a returned patch is a patch when a base names the columns, but it is the whole chart when nothing sits underneath it, and turning inference off unconditionally would leave that second case with no dimensions at all.

adminSiteClient/ConfigEditor.ts
```diff
@@ -0,0 +1,177 @@
...
+            ),
+            // The host swapped the base config (e.g. the admin fetched the
+            // defaults of a newly picked indicator). Re-apply it underneath
+            // the user's edits: capture the patch against the *old* base
+            // first, otherwise values the old base supplied would be folded
+            // into the patch as if the user had authored them.
+            reaction(
+                () => this.manager.parentConfig,
+                (base) => {
+                    const { patchConfig } = this
```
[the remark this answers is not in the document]
Marigold: > _Written by Anthropic Claude Opus 5 — @Marigold at the wheel._

The mechanism is real but it isn't about the host's form, so computing the patch in slug form doesn't fix it: `diffGrapherConfigs` treats `dimensions` as one value, so as soon as any slot differs from the base the patch carries the whole array, inherited slots included. In slug form that same patch comes out as `ySlugs: "a", colorSlug: "b"` and overrides a new base just the same.

Making a swapped base reach individual slots means per-slot dimension inheritance, which needs #7248's `dimensions[].slug` first. Leaving it: the only host that swaps a base today is the admin, whose bases are indicator configs that carry no dimensions, and the narrative pages never swap theirs.

adminSiteClient/indicatorStores.ts
```diff
@@ -0,0 +1,145 @@
...
+    ColumnSlug,
+    OwidChartDimensionInterface,
+    OwidColumnDef,
+    OwidTableSlugs,
+} from "@ourworldindata/types"
+import { getCachingInputTableFetcher } from "@ourworldindata/grapher"
+import { IndicatorCatalog } from "./editorProviders.js"
+import { Dataset, IndicatorCatalogData } from "./EditorDatabase.js"
+
+type SelectedEntityColors =
```
sophiamersmann, line 24: Maybe move this into the type packages and use it elsewhere as well?

adminSiteClient/indicatorStores.ts
```diff
@@ -0,0 +1,145 @@
...
+     */
+    loadTable(
+        dimensions: OwidChartDimensionInterface[],
+        selectedEntityColors: SelectedEntityColors
+    ): Promise<OwidTable | undefined>
+    /**
+     * Well-known indicators the Basic tab offers as shortcuts (population,
+     * GDP per capita), keyed by catalog path. Absent → no shortcuts.
+     */
+    variableIdsByCatalogPath?: Record<string, number | null>
```
sophiamersmann, line 45: I don't think this belongs here. Can `variableIdsByCatalogPath` be dropped here and instead we add something like `scatterDefaults: OwidChartDimensionInterface[]` to the host?

adminSiteClient/indicatorStores.ts
```diff
@@ -0,0 +1,145 @@
...
+    const name = options.name ?? "Table"
+    // Categorical columns included: a scatter's colour dimension may well be
+    // a string column such as "continent".
+    const slugs = table.columnSlugs.filter(
+        (slug) => !STRUCTURAL_SLUGS.has(slug)
+    )
+
+    // Columns without a display name would show up as their slug anyway, but
+    // the picker reads `displayName`, so fill it in.
+    const named = table.updateDefs((def: OwidColumnDef) =>
```
sophiamersmann: If you switch the picker to `nonEmptyDisplayName`, you could get rid of this (assuming the comment is correct and that's the only reason it's done)

adminSiteClient/GrapherEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Receives the edited config. May return the config as actually stored,
+     * which then counts as the saved state.
+     */
+    onSave: ConfigEditorManager["onSave"]
+    /** Fires on every change of the edited config. */
+    onChange?: ConfigEditorManager["onChange"]
+    /** What "Add indicator" can offer. Defaults to `store.catalog`; pass
+     *  `null` for no picker. */
+    indicators?: IndicatorCatalog | null
```
sophiamersmann, line 44: Are there really legitimate use cases where what 'Add indicator' offers should differ from `store.catalog`? If the main reason for having this prop is to disable the picker, wouldn't the simplest approach be to not pass a `store.catalog`?

adminSiteClient/AbstractChartEditor.ts
sophiamersmann, line 48: If I understand correctly, 'revisions' and 'refs' doesn't belong here anymore, right?

adminSiteClient/GrapherEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+    /** Details on demand for validating text fields. Absent → none. */
+    details?: DetailsProvider
+    /** Restrict the tabs shown. */
+    tabs?: EditorTab[]
```
sophiamersmann, line 48: Could you rename the type to `EditorTabName`? I got confused because `tabs` and `extraTabs` hold very different things, and I think this would help a bit

adminSiteClient/ConfigEditor.ts
```diff
@@ -0,0 +1,155 @@
...
+                (config) => this.manager.onChange?.(config, this),
+                { equals: comparer.structural }
+            ),
+            // The host swapped the base config (e.g. the admin fetched the
+            // defaults of a newly picked indicator). Re-apply it underneath
+            // the user's edits: capture the patch against the *old* base
+            // first, otherwise values the old base supplied would be folded
+            // into the patch as if the user had authored them.
+            reaction(
+                () => this.manager.parentConfig,
```
sophiamersmann, line 75: Maybe it makes sense to make `parentConfig` a `@computed`, so this reaction doesn't unnecessarily fire on unrelated changes?

adminSiteClient/EditorPlaygroundPage.tsx
```diff
@@ -0,0 +1,805 @@
+/**
```
sophiamersmann, line 1: Do you intend to ship the playground?

adminSiteClient/AbstractChartEditor.ts
```diff
@@ -119,9 +189,21 @@ export abstract class AbstractChartEditor<
     // if inheritance is enabled, the parent config is applied to grapherState
     isInheritanceEnabled: boolean | undefined = undefined
```
sophiamersmann: All of these are now concerns of the host, so we should be able to drop them here, no?

adminSiteClient/AbstractChartEditor.ts
sophiamersmann, line 166: Now that `ConfigEditor` is the only subclass, I wonder if folding `AbstractChartEditor` into `ConfigEditor` is worth it. It would probably simplify some code and improve readability?

adminSiteClient/AbstractChartEditor.ts
```diff
@@ -47,12 +52,32 @@ const EDITOR_TABS = [
...
+    url: string
+    /** Why this URL is being offered, shown greyed after it. */
+    hint?: string
 }
 
 export interface AbstractChartEditorManager {
-    admin: Admin
+    // Only editors that talk to the admin API need this (charts, narrative
+    // charts). A config-only editor runs without it.
+    admin?: Admin
```
sophiamersmann, line 75: I don't quite understand the comment. Isn't the point of the refactor that the core editor no longer needs admin access? In other words, should this be droppable?

adminSiteClient/AbstractChartEditor.ts
sophiamersmann, line 190: All of these are now concerns of the host, so we should be able to drop them here, no?

adminSiteClient/AbstractChartEditor.ts
```diff
@@ -69,6 +94,57 @@ export interface AbstractChartEditorManager {
...
+]
+
+/**
+ * `ySlugs: "a b"` and friends as dimensions, which is what the editor edits.
+ * Both forms render the same and a dimension can say more about its column
+ * (`display`, a colour, a conversion factor), so the editor keeps only the
+ * one form and hands it back that way — a host that saved the flat form gets
+ * dimensions back. A config already naming dimensions is left alone.
+ */
+export function withDimensionsFromColumnSlugs(
```
sophiamersmann, line 131: I think we should remove support for `ySlugs` etc once explorers are removed from the codebase, so in my opinion, it’s fine if this PR doesn’t add support for them (in which case this function isn't needed, and maybe a few more things can be deleted)

adminSiteClient/ChartEditorView.tsx
```diff
@@ -382,39 +383,48 @@ export class ChartEditorView<
...
+        // before the config has loaded. Show the first available one instead
+        // without touching `editor.tab`, so the URL's intent survives.
+        const activeTab = availableTabs.includes(editor.tab)
+            ? editor.tab
+            : availableTabs[0]
+
+        // Hosts of the config-only editor plug their own tabs, save buttons
+        // and preview link in; the admin's chart-record editors get the
+        // built-in ones.
+        const configEditor = isConfigEditorInstance(editor) ? editor : undefined
```
sophiamersmann, line 411: Every editor is a ConfigEditor at the moment, right? If we don't intend to add other subclasses, that's probably another reason to fold AbstractChartEditor into ConfigEditor so this distinction goes away.

adminSiteClient/ChartEditorPage.tsx
```diff
@@ -188,42 +282,370 @@ export class ChartEditorPage
...
+            }
+        }
+    }
+
+    /**
+     * Everything about the chart as a row in our database that isn't its
+     * config: inheritance from the indicator, tags, the data-page override.
+     * Used to sit at the bottom of the Basic and Text tabs.
+     */
+    private renderPublishingTab(editor: ConfigEditor): React.ReactNode {
```
sophiamersmann, line 530: Can you move the publishing tab to its own component file for consistency?

sophiamersmann (commented):
Makes sense to me overall!

I did a little testing, but not much. I'm relying on your agents for that. I focused instead on the new interfaces and responsibilities, and on my own understanding. I also didn't review the parts related to the [new map config field](https://github.com/owid/owid-grapher/pull/7248), since I assume those will be reviewed on that branch.

Other than a few smaller things, two things stood out to me:

- I think we _probably_ want to get rid of the plain string fields `ySlugs`, `xSlug`, etc. in the future, so in my opinion there's no need to support them here.
- Unless there's a blocker or we expect more subclasses in the future, I think it makes sense to fold `AbstractChartEditor` into `ConfigEditor` to flatten the hierarchy and make the distinction between the two go away in code.

I have a bit of a dislike of the agent-heavy commenting style. I think most comments could either be removed or shortened.

PR #7230, opened 2026-09-09
Tag translated pageviews in Sentry

marcelgerber:
> _Written by Anthropic Claude Opus 5 — @marcelgerber at the wheel._

We detect browser translation but only report it to GA4, so translation isn't queryable in Sentry at all — confirming that an error came from a translated page currently means opening breadcrumbs one issue at a time.

This sets `page_translated` on the Sentry scope, plus `page_translated_to` for the target language, making translation filterable in the issue stream and usable in Discover and alerts. The tag is deliberately sticky for the rest of the pageview: readers toggle translation back off, but a page that was translated once keeps hitting errors afterwards, so "was translated at some point" is the signal worth keeping.

Nine lines in `site/SiteAnalytics.ts`, no behaviour change for readers.

### Why this is useful

Translation is the cause of two distinct error populations in the `website` project, and neither is currently filterable:

1. `RangeError: Maximum call stack size exceeded.` on WebKit — fragmented across 20+ issue groups because the minified symbol names shift underneath it. Confirmed translation-induced by sampling breadcrumbs in 9 of 9 groups, where a `Browser translation detected` console entry fires 6 ms–1 s before every error.
2. A much larger flood — ~250k events / ~2,700 users per 30 days — of `removeChild:`/`insertBefore: Cannot ...` messages that `site/hacks.ts` reports itself, on every engine, concentrated on `/grapher/*` in desktop Chrome.

With this tag, both become one query instead of a manual breadcrumb crawl.

### Implementation notes

- Tags are set in `sendTranslationAnalyticsEvent`, alongside the existing `console.info` and GA4 call, so there's a single place where translation is observed.
- `page_translated_to` is only set when the new language differs from the document's initial `lang`, so reverting to English doesn't overwrite it with `"en"`.
- `Sentry.setTag` is safe before/without `Sentry.init` — it writes to the current scope and is a no-op if nothing is sent — so no guard is needed for `LOAD_SENTRY=false`.

### Follow-up this enables

Once the tag is live, the RangeError can be fingerprinted on it in `beforeSend` so translated-page overflows group together while genuine stack overflows stay in their own group. Fingerprinting only affects new events; the existing groups would need Sentry's merge action.

### History

This PR started as a diagnostic branch to unminify the bundle and read the crash's stack traces. That work is preserved on `translate-crash-probe` and is worth keeping around — the on-device probe there is what established that the recursion is **not in our code**: the frames claim lines 196–232 of `/poverty` and `/search`, but those pages are 81 and 38 lines respectively, `owid.mjs` appears nowhere in the stack, and the recursion runs 250+ pairs deep against a DOM only 19 levels deep. It's the webview's injected translate script re-entering itself, which is why no source map will ever resolve those frames. See the comments below for the full trail.

Only the Sentry tag remains here, since it stands on its own.

</details>

### DIFF
--- a/site/SiteAnalytics.ts
+++ b/site/SiteAnalytics.ts
@@ -1,4 +1,5 @@
 import * as _ from "lodash-es"
+import * as Sentry from "@sentry/react"
 import { GrapherAnalytics, splitPathForGA4 } from "@ourworldindata/grapher"
 import {
     EventCategory,
@@ -415,6 +416,14 @@ export class SiteAnalytics extends GrapherAnalytics {
 
             // eslint-disable-next-line no-console
             console.info("Browser translation detected", ctx)
+
+            // Deliberately sticky for the rest of the pageview: readers toggle
+            // translation back off again, but a page that was translated once
+            // keeps hitting reconciliation errors afterwards.
+            Sentry.setTag("page_translated", "true")
+            if (newLang && newLang !== initialLang) {
+                Sentry.setTag("page_translated_to", newLang)
+            }
 
             this.logBrowserTranslationEvent(ctx)
         }


marcelgerber:
> _Written by Anthropic Claude Opus 5 — @marcelgerber at the wheel._

Added `385ad8e0e8`, because the debugger route is a dead end: the crash only reproduces on WebKit devices, BrowserStack Live gives a console at best, and desktop survives the same recursion since V8's stack is deeper. So the page reports its own stack instead.

`site/translationCrashProbe.ts` raises `Error.stackTraceLimit` from the default 10 to 500, then on the **first** RangeError collapses the stack into the frames that repeat and renders it as a `translate="no"` overlay you can read or screenshot on the device. It also mirrors the report to Sentry so it can be read from a desk. One-shot, so the burst of repeat crashes can't flood Sentry.

This supersedes the "read the stack in the console" note in the description — no console needed now.

Verified against a synthetic mutual recursion in local Chrome: 500 frames captured, and the overlay names the pair with counts (`250x insertOrAppendPlacementNode`, `250x getHostSibling`). Those names are the ones the real crash needs to confirm or kill — if the overlay shows that pair on a translated page, the `hacks.ts` no-op is the amplifier and the next step is the one-line experiment of disabling it.

marcelgerber:
> _Written by Anthropic Claude Opus 5 — @marcelgerber at the wheel._

**The premise of commit `1d88ff18c1` is void, and the hypothesis in the description is disproven.** The probe worked and produced the data that killed it.

On a translated page the overlay reported `248x Rk@/poverty:232:63` / `247x Tk@/poverty:232:408` — the *identical* symbols and line:col as the production events, despite the deployed bundle being unminified with React's development build. Three facts explain why:

1. **The frames aren't in our HTML.** They claim lines 196–232 of `/poverty` and `/search`. Production `/poverty` is **81 lines**; `/search` is **38**. Those lines do not exist.
2. **`owid.mjs` appears nowhere in the stack.** Not one frame.
3. **It isn't a DOM traversal.** The recursion runs 250+ pairs deep against a DOM that is **19** levels deep with 3,412 elements on `/poverty`.

So `Rk` and `Tk` are not `insertOrAppendPlacementNode` and `getHostSibling`. They're minified functions in the **translate script the iOS webview injects into the page** — WebKit attributes injected scripts to the page URL with their own line numbering, which is why no source map could ever resolve them and why unminifying our bundle changed nothing. And at 250+ levels against a 19-deep DOM, it's a re-entrant cycle, not a tree walk.

**Consequences**

- Drop commit `1d88ff18c1`. Unminifying our code cannot help when the code isn't ours.
- Keep `385ad8e0e8` (the probe) — it's what produced this.
- Keep `776fbe7734` (the Sentry tag) — still independently useful, still mergeable.
- The `hacks.ts` monkey patch is **not** where the overflow happens. It may still feed the loop by letting React keep reverting the translator's mutations, but the one-line disable experiment is no longer the priority.

**New direction:** the loop looks like React and the translator fighting — the translator rewrites a node, React reverts it, the translator re-translates and re-enters. The fix is to remove the fuel rather than out-race it: mark the subtrees React re-renders dynamically as `translate="no"`, starting with search results since `/search` is the top transaction. We already do this at `site/search/Autocomplete.tsx`.

Separately, the ~250k/30d `removeChild:`/`insertBefore: Cannot ...` flood from `hacks.ts` **is** our code, fires on every engine, and concentrates on `/grapher/*` in desktop Chrome. That's a real and larger problem, and it's not the same one as this RangeError.

PR #7243, opened 2026-09-09
🐝🤖 make the pre-commit hook executable so git actually runs it

mlbrgl:
> _Written by Anthropic Claude Fable 5.1 — @mlbrgl at the wheel._

The pre-commit hook has never run for anyone: `.husky/pre-commit` is tracked as a non-executable file, and with `core.hooksPath` pointing straight at `.husky` (no husky shim in this repo), git skips it with a warning on every commit. This makes it executable, so `lint-staged` (oxfmt on staged files) actually runs.

One thing to know: after this merges, commits on existing branches still won't run the hook until those branches rebase onto master — the file mode comes from the tree being committed on.

- The only change is the git file mode: `100644 → 100755` on `.husky/pre-commit`. Content unchanged (`npx lint-staged`).
- Root cause: git prints `hint: The '.husky/pre-commit' hook was ignored because it's not set as executable` on each commit. `core.hooksPath` is `.husky` here, so git executes the hook file directly rather than through husky v9's `.husky/_/h` shim, which would have run it via `sh` regardless of mode. There is no `prepare` script installing that shim.
- `lint-staged` is configured in `package.json` (`"*": "yarn oxfmt --no-error-on-unmatched-pattern"`) and installed, so enabling the hook doesn't introduce a failing step.
- Verified by committing with the fixed mode: lint-staged ran and the commit went through.

</details>


mlbrgl:
* **#7243** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7243?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

mlbrgl:
wrong resolution, needs `yarn husky`

PR #7249, opened 2026-09-09
🧪 Experiment: multi-indicator data pages (indicator switcher)

bnjmacdonald:
## What this does

Multi-indicator charts (several Y-indicators) currently render as plain grapher pages — the data-page metadata treatment only existed for single-indicator charts. This branch extends the redesigned metadata box (#6530) to multi-indicator charts:

- **`baker`**: enrolled charts are forced to bake as data pages (`shouldBakeAsDatapage` — the single seam that would later flip to `true` for an all-charts migration). Per-indicator metadata is fetched for each Y-indicator (S3 metadata, FAQs, dataset owners, primary topic), with per-pane failures degrading gracefully (a broken indicator drops its pane; a broken *primary* fetch falls back to the plain grapher page rather than failing the bake).
- **`site`**: the metadata box gains an **indicator switcher** — one metadata pane per Y-indicator, all panes baked into the HTML (crawlers and no-JS readers see every indicator's sources/citations/FAQs on a plain GET), only the active pane visible. Default control is horizontal pills with a "More" overflow; on mobile the pills swap (CSS-only) for a dropdown select on its own row. Three alternative switcher variants (`dropdown`, `h-tabs`, `v-tabs`) remain behind `?switcher=` for design comparison and will be deleted before any wider rollout.
- **`experiments`**: the `data-page-metadata-v1` path list now enrolls 24 multi-indicator pages (plus the existing 10 single-indicator pages, with two stale slugs replaced). Enrollment is the treatment arm of a page-level randomized design; the measurement plan is documented internally.

## Page-weight notes

All panes are baked into the HTML deliberately (agent/crawler readability). Measured on the heaviest page: raw HTML ~470 KB, **22 KB gzipped** (panes are highly redundant), hydration props ~51 KB after stripping an unused per-pane `chartConfig` (was 204 KB). Bake cost: ~1.5–2 s per heavy enrolled page, ~15–20 s total across all enrolled pages.

## Review

The branch was adversarially reviewed; confirmed findings fixed in the last six commits (mobile hydration mismatch from `useMediaQuery`, switch-tracking counting menu opens, title override leaking onto single-indicator pages, missing per-pane `primaryTopic` in citations, unguarded primary metadata fetch in the bake, config comment hygiene). Known deferred items: variant cull, per-pane `origins` dedupe in the hydration blob, batching the per-pane owners queries.

## Testing

- typecheck / lint / 2,511 unit tests green on the merged branch
- all enrolled pages render-checked locally (incl. 19–25-pane charts); non-enrolled pages verified to bake identically to master (only delta: a `"useNewDatapageDesign":false` key in the props blob)
- h-pills verified at 1440/950/375 px incl. the overflow-active case; hydration console clean of switcher-related errors


bnjmacdonald:
⚠️ Note for reviewers: the branch tip includes a TEMP commit (d3bab537fb) enrolling 26 extra multi-indicator pages so they bake as data pages on staging for QA. **It must be reverted before the experiment launches** — those pages are not part of the launch set. The `?forceDatapage=true` preview param (`/admin/grapher/<slug>?forceDatapage=true`) covers QA of any other chart.

PR #7251, opened 2026-09-10
Data page metadata experiment v2: randomise the redesign across 330 pages

bnjmacdonald:
> _Written by Claude Fable 5 — @bnjmacdonald at the wheel._

## What this is

A randomised test of the data page metadata redesign (#6579). The first run (`data-page-metadata-v1`) put 10 pages on the new design with no control group. This one puts **165 data pages on the redesign and keeps 165 comparable pages on the current design**, so we can measure the effect properly.

The assignment was drawn in advance with a fixed random seed and is recorded in owid/analytics#1038. This PR ships the treatment half of it.

**Merging this starts the experiment** on the next bake.

## What changes

1. **`config.ts`** — a new experiment entry, `data-page-metadata-v2`, set up exactly like v1: 100% treatment on its listed paths. The paths are the 165 treatment pages, copied from the analytics record. Control pages are not listed anywhere in this repo on purpose; the analysis identifies them from the analytics JSON by slug.
2. **`isDataPageMetadataRedesignActive(url)`** — one function meaning "this page is in v1 or v2", used by both the baker and `DataPageV2Content` so they cannot disagree.
3. **Control-arm click tracking** — the old design's sources, citation, "what you should know" and data-processing sections now emit the same GA events as the metadata box, so both arms can be compared on engagement. Previously the old design had no click tracking at all. Links inside rendered markdown are tracked the site-standard way: `SimpleMarkdownText` takes an optional `dataTrackNote` and puts `data-track-note` on every link it renders, `IndicatorProcessing` threads a `trackNote` to its pipeline link and processing notes, `IndicatorSources` threads a `descriptionTrackNote` to each source's description and "Data published by" text, and the existing document-level click listener does the logging (`owid.site_click`, `event_action` = note, `event_target` = link text).
4. **Search bar removed** from the redesign. It generated almost no organic use.
5. **Fix: list indentation in the box's "Documentation from data sources" section.** That markdown rendered outside the expander's prose wrapper, so the site-wide list reset left its bullets flush with the text (visible on e.g. `/grapher/co2-emissions-transport`). A box-scoped class restores the indent.
6. **Fix: `SimpleMarkdownText`'s custom link renderer no longer spreads react-markdown's `node` prop onto the `<a>`**, which serialised as `node="[object Object]"` in the HTML. Pre-existing on slideshow links via `openLinksInNewTab`; would otherwise have appeared on every tracked data page link.
7. **Details-on-demand hover location.** The same DoD term can appear in several places on a data page, and `owid.detail_on_demand` alone can't tell them apart. `SimpleMarkdownText` takes an optional `dodTrackNote` and stamps `data-dod-track-note` on each DoD span it renders (mirroring `dataTrackNote`); `IndicatorSources` and `IndicatorProcessing` pass it through. The shared DoD handler now hands the trigger span to its callback, and the site sends the span's location as `eventContext` (an existing param GTM already forwards; values are short fixed labels). Both designs use the same labels: `grapher` (an attribute on the existing chart containers), `wysk` (treatment: `wysk/preview`, `wysk/expanded`), `key_data`, `producer_documentation`, `processing`, `data_sources`, plus control-only `additional_information`. No wrapper elements are added. Hovers in FAQs and anywhere off data pages carry no location, as before.
8. **Tests** — rendering tests for `dataTrackNote` / `openLinksInNewTab` (attribute present only when asked, detail-on-demand terms untouched, no `node` attribute) and for `dodTrackNote`.

## What to check as a reviewer

- The `paths` list is not hand-edited: it must equal the `treatment` array in the analytics JSON. A script in the analytics PR verifies this against the pushed branches, and it passes.
- The control-arm event names match the metadata box's (`expand_expandable_toggle`, `copy_citation`, `retrieved_from`, `source_link`, `producer_link`, `wysk_link`, `processing_link`). FAQ answer links are deliberately not tracked.
- Nothing outside these files changes behaviour: the experiment framework itself is untouched, and the sources modal and bespoke metadata pages pass none of the new props so they render as before.
- DoD attribution in a browser: with `window.dataLayer = []` set in the console, hovering a dotted term in a metadata text prints an `owid.detail_on_demand` push whose `eventContext` names the section; in FAQs or on a non-data page it has no `eventContext`.

## Things to know

- **Cookie side effect, accepted.** Like v1, visiting a treatment page sets an `exp-data-page-metadata-v2=treatment` cookie that GA then sees on that visitor's later page views. So the GA experiment param is not a reliable arm indicator; the analysis uses page slugs instead.
- **`event_target` on link clicks is display text**, which browser translation can change; count on `event_action`. External link destinations are available from GA4's own `link_click` event.
- **Ending the experiment**: edit the config and rebake in one step.
- **Please avoid re-slugging or migrating enrolled pages** (treatment or control) during the run. The control list is in the analytics JSON.
- Staging: `staging-site-metadata-experiment-v2`.


packages/@ourworldindata/components/src/IndicatorKeyData/IndicatorKeyData.tsx
```diff
@@ -91,7 +91,16 @@ const splitSourceTextIntoFragments = (text: string | undefined): string[] => {
...
 
-export const makeLinks = ({ link }: { link?: string }): React.ReactNode => {
+export const makeLinks = ({
+    link,
+    trackNote,
+}: {
+    link?: string
+    /** When set, clicks on the rendered links are tracked under this note.
+     * Opt-in per call site so the same publisher URL is attributable to where it
+     * was clicked (a data page's sources section vs the grapher sources modal). */
```
ikesau: We can drop this comment, I think

packages/@ourworldindata/components/src/IndicatorSources/IndicatorSources.tsx
```diff
@@ -20,6 +20,11 @@ export interface IndicatorSourcesProps {
         index: number,
         isOpen: boolean
     ) => void
+    /** When set, clicks on the "Retrieved from" publisher links are tracked
+     * under this note. The data page metadata sections pass it; the grapher
+     * sources modal leaves it unset so the experiment's cross-arm publisher
+     * click-through metric isn't polluted by modal clicks. */
```
ikesau: We can remove this too

packages/@ourworldindata/utils/src/experiments/config.ts
```diff
@@ -118,6 +120,233 @@ export const experiments: Experiment[] = [
...
+     * Conditions:
+     * - (a) control: the current data page (AboutThisData + Sources/Reuse sections)
+     * - (b) treatment: the new metadata box in place of those sections
+     *
+     * Ending the experiment: edit this config and rebake in the same step.
+     * Letting the expiry date flip the gate would leave baked treatment markup
+     * hydrating against control on 165 pages until each one is rebaked. Remove
+     * or update the fixtures in experiments.test.ts in the same PR: they look
+     * this experiment up by id at collection time and fail CI if it is gone.
+     */
```
ikesau, line 137: Can this be pared down? General theme of my reviews is that I don't like overly long obviously agent-written comments 😅

packages/@ourworldindata/utils/src/experiments/config.ts
```diff
@@ -132,3 +361,20 @@ export function isUrlInActiveExperiment(rawId: string, url: string): boolean {
...
+/**
+ * True if the given data page url should render the redesigned metadata layout.
+ *
+ * Two experiments put a page on the new design, both at 100% treatment on their
+ * listed paths: v1 (10 pages) and v2 (the 165 treatment pages of the 330-page
+ * randomised design; its controls are not listed here). This is the single
+ * source of truth — the baker uses it to decide which pages get the extra
+ * per-indicator metadata loaded, and the data page component uses it to pick
+ * the markup, so the two can never disagree.
+ */
```
ikesau: Okay I'll stop pointing this out, but can you go through this whole PR and rewrite/remove comments to only their essential information? 🙏

packages/@ourworldindata/utils/src/experiments/experiments.test.ts
```diff
@@ -0,0 +1,104 @@
+import { expect, it, describe } from "vitest"
```
ikesau: I don't think we need these tests.

site/TrackedProseLinks.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * `data-track-note` of their own — and putting one on the wrapper would make the
+ * global click handler fire on every click in the block, text included. So this
+ * listens on the wrapper but only reports when the click actually landed on a
+ * link, sending the href as the label.
+ *
+ * Used on both data page designs' "What you should know about this indicator"
+ * text and their "How we process data" section, so click-through out of those
+ * is measured the same way in each arm.
+ */
+export default function TrackedProseLinks({
```
ikesau: What if we added a `data-track-note` prop to `SimpleMarkdownText` and just used the generic `startClickTracking` listener? Is it essential that we get the `href` and not just the `innerText`?

site/IndicatorMetadataBox.scss
```diff
@@ -74,3 +74,23 @@
...
 }
+
+// Producer documentation is free-form markdown rendered inside an
+// ExpandableToggle, outside the expander's `__prose` wrapper. The global reset
+// zeroes list padding, so restore the same list indentation the control
+// design's .key-info__content gives this text (AboutThisData.scss).
+.indicator-metadata-box__producer-docs {
+    ul,
+    ol {
+        padding-left: 1em;
```
ikesau: If anything, I feel like it should be indented more (also can you add a comment to this slug (`co2-emissions-transport`) so it's easier to check in the future?)

Image

ikesau (approved):
Apart from all the excessive commenting, looks pretty good to me! If you go for the markdown refactor, lmk if you want me to double-check it.

site/TrackedProseLinks.tsx
bnjmacdonald: ya good point – done ✅ . href is nice to have but not essential b/c the innerText is stored as the translated text when a user is viewing the translated page (minority of sessions, but means dropping data if using it in analysis). But we also get the href in the `click_url` on the `link_click` event, so it's available through that route if we want it (but then don't know exactly what component was clicked). I've gone with the data-track-note. 

If we want a more general solution for knowing what component a link was in when it was clicked, that's for another PR.

PR #7266, opened 2026-09-14
Improve backend tracing and reduce staging profiling

rakyi:
> _Written by OpenAI GPT-6 — @rakyi at the wheel._

We are exhausting Sentry's continuous profiling allowance, and recent profiles were dominated by standalone SQL transaction-control operations with little context about the work being done.

This change gives large bakes manageable traces for individual pages and shared preparation/cleanup phases, while smaller scripts use whole-job traces. Standalone database traces are excluded; database activity inside sampled work remains visible. Finishing telemetry delivery before scripts exit makes that coverage more reliable.

We keep Sentry's built-in profiling lifecycle and reduce staging's profiling session rate to 25%, retaining 100% in production. This reduces expected staging consumption while preserving some diagnostic coverage. Session sampling is decided once per process startup, so coverage is distributed across processes rather than requests. Trace sampling remains at 10% by default.

The tracing improvements do not themselves guarantee profiling savings: the installed SDK can profile non-recording spans too. We accept that behavior rather than maintain a custom profiler lifecycle. Actual profile-hour savings and the missing production profiles still need verification after deployment.

<details>
<summary>Details</summary>

Validation: typecheck, lint, formatting, and 10 focused tracing/shutdown tests pass. Baker tests also passed during implementation.

After deployment, compare continuous profile hours and confirm useful profiles arrive from production and sampled staging processes. Reassess session sampling if consumption remains excessive.
</details>


serverUtils/sentryTracing.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function sampleServerTrace({
+    attributes,
+    inheritOrSampleWith,
+}: Parameters<NonNullable<Sentry.NodeOptions["tracesSampler"]>>[0]): number {
+    // The sampler runs for trace roots, not their local child spans. Unparented
+    // SQL must not start a trace (and the profiler) for every BEGIN/COMMIT.
+    const op = attributes?.[Sentry.SEMANTIC_ATTRIBUTE_SENTRY_OP]
+    if (typeof op === "string" && (op === "db" || op.startsWith("db.")))
+        return 0
```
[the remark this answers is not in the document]
rakyi: Good to know.

rakyi:
@codex review

serverUtils/sentryTracing.ts
```diff
@@ -0,0 +1,47 @@
...
+export function sampleServerTrace({
+    attributes,
+    inheritOrSampleWith,
+}: Parameters<NonNullable<Sentry.NodeOptions["tracesSampler"]>>[0]): number {
+    // The sampler runs for trace roots, not their local child spans.
+    // Reject standalone BEGIN/COMMIT traces to reduce tracing noise. This
+    // does not suppress profiling lifecycle events in SDK 10.70.
+    const op = attributes?.[Sentry.SEMANTIC_ATTRIBUTE_SENTRY_OP]
+    if (typeof op === "string" && (op === "db" || op.startsWith("db.")))
+        return 0
```
marcelgerber: Just a question out of curiosity: Is it possible to only sample long-running DB operations over some limit, like >2s or something? Or do we always need to specify whether we sample something at the beginning of the span, not at the end?

Long-running DB spans could be insightful, but no worries if this is not possible.

baker/startDeployQueueServer.ts
```diff
@@ -29,6 +29,7 @@ const main = async () => {
     // Poll for changes every 5 seconds
     while (true) {
         try {
+            // This can run a full bake. Individual pages own their traces.
```
marcelgerber: this comment is unnecessary here

marcelgerber (approved):
thanks, looks good!

(I haven't paid line-by-line attention to every change in the baker, but think most changes are just wrapping it in a tracing session and that seems sensible)

serverUtils/sentryTracing.ts
rakyi: Sampling is decided before, not after.

I decided to remove this logic after all. I lowered the sample rate on staging, which might be enough to stay in quota. It's probably better to keep all the database calls even if most of them are uninteresting.

PR #7271, opened 2026-09-14
🎉🤖 (stacked-area-negatives) put the negative category at the bottom of the stack

sophiamersmann:

Drawing a negative category below the zero line left two artifacts, and they are the same problem seen from two sides. Where the bottom category crosses zero between two years, the band above it moves its lower edge to or from the zero line across that one interval, and a wedge opens and closes again. Where the negative category is not at the bottom, its own lower edge travels from the top of the stack beneath it down to zero, so the band sweeps through the categories underneath: 5 percent of the plot height filled twice on `electricity-mix-uk`.

The first has a geometry fix. The second does not, because it is the ordering that is wrong and the ordering is the author's to set, so the y dimension slot in the chart editor now says so.

- **Fix.** `toPlacedStackedAreaSeries` densifies before placing. A vertex goes into every series wherever the bottom series crosses zero, and the copy is re-stacked so the new points pick up offsets from the stacker that already exists. The placement code is untouched. On a 600px plot the worst run painted by neither of two adjacent bands drops from 0.81px to 0 on `co2-emissions-fossil-land` Germany and from 4.34px to 0 on United States.
- **Behaviour change.** Where the bottom category crosses zero, the top edge of the positive stack now shows a kink it did not before. The kink is real: at a crossing the category leaving the positive stack contributes nothing, so the sum genuinely dips, and the straight line drawn between the two years was hiding it.
- The fix is guarded to a lone negative series at the bottom of what actually gets drawn. `ozone-depleting-substance-consumption`, the chart that prompted this work, has five negative categories and is held out. Closing that case needs the up-stack total at every crossing of every series, which is the general mixed-sign stacked area chart.
- **Known limitation.** The warning has no magnitude threshold: a column counts as negative when its minimum is below zero. That fires on `share-adults-bank-account-financial-institution-mobile-money`, whose only negative value is −1.16e−08 in 264 points. Excluding it means deciding how negative a value has to be before it matters for layout, which is the same question the rendering path would have to answer, and neither answers it today.
- The warning is deliberately not routed through `errorMessagesForDimensions`, which feeds `editingErrors` and disables saving. A chart whose ordering is merely suboptimal has to stay saveable.
- `stackSeriesInBothDirections` writes `valueOffset` into the points it is handed rather than returning fresh ones, which is why the densified series are copied before being re-stacked. Making it pure would touch its other callers.

### The crossing fix

Reaches the eight published charts whose single negative indicator is already the last one, on an entity and time range where it crosses zero. Only `co2-emissions-fossil-land` does so in its default view; `foreign-aid-given-to-donor-countries-share` passes the guard but `minTime: latest` leaves it one time point and no intervals to fix.

[`co2-emissions-fossil-land`](http://staging-site-sa-neg-order/grapher/co2-emissions-fossil-land?country=~DEU), `foreign-aid-given-for-emergencies-and-longer-term-development`, `foreign-aid-given-for-emergencies-and-longer-term-development-per-capita`, `foreign-aid-given-to-donor-countries-share`, `foreign-aid-received-by-income-group`, `ghg-emissions-by-gas`, `ghg-emissions-by-world-region`, `global-co2-emissions-by-source-inc-land`

### The editor warning

19 of the 388 published stacked area charts. Eight have one negative indicator that is not last, and are told to drag it to the bottom (position in the y slot in brackets):

`cereal-land-spared` [1 of 2], `deaths-due-to-illicit-drug-use-by-age` [4 of 5], `electricity-mix-uk` [6 of 9], `electricity-source-wb-stacked` [1 of 4], `hivaids-deaths-and-averted-due-to-art` [1 of 2], `long-term-energy-transitions` [1 of 8], `number-of-new-hiv-child-infections-vs-number-of-infections-averted-due-to-pmtct` [1 of 2], `share-adults-bank-account-financial-institution-mobile-money` [1 of 3]

Eleven have more than one, where no single move helps, so they get the other message instead:

`contribution-to-temp-rise-by-gas`, `fertilizer-per-capita`, `foreign-aid-given-by-sector`, `foreign-aid-given-for-social-infrastructure-and-services`, `gender-gap-in-life-expectancy-by-age-group`, `global-warming-by-gas-and-source`, `government-spending-by-function`, `government-spending-by-function-as-share-of-gdp-across-countries`, `official-development-assistance-donor-bilateral-multilateral`, `ozone-depleting-substance-consumption`, `warming-fossil-fuels-land-use`

Measured over each indicator's full data, which is a stand-in for the admin's `column.minValue`. A chart whose author excluded entities could come out differently.


sophiamersmann:

Superseded by #7273, which combines both halves of this stack into one branch. The two could no longer be separated: the geometry change is now gated on the negative indicator being at the bottom, and that gate needs the predicate this branch introduced.

PR #7276, opened 2026-09-15
🔨🤖 (rounding) round font sizes consistently

sophiamersmann:
> _Written by Anthropic Claude Opus 5 — @sophiamersmann at the wheel._

Font sizes in the grapher package were rounded four different ways, or not at all: `Math.floor` at about thirteen sites, `Math.round` at six, `roundFontSize` at four, and nothing at the rest. They now all go through one function that scales a design size from the base and rounds to the nearest 0.5. The `GRAPHER_FONT_SCALE_*` constants are deleted with it, so `GRAPHER_FONT_SCALE_12 * fontSize` is no longer something you can write. Oxlint has no `no-restricted-syntax`, so taking the ingredient away was the only enforcement available.

At the default base size of 16 almost nothing moves. What does:

- **Behaviour change.** Scatter plots draw a different set of entity labels. Label sizes come off a continuous scale and feed the collision detection that picks which labels fit, so quantising them changes the outcome: two more labels at base 16 and one fewer at base 20 on [this chart](http://staging-site-round-font-sizes/grapher/access-to-clean-fuels-for-cooking-vs-gdp-per-capita). It is the only place this changes what a chart shows rather than how big its text is, and the part worth a look.
- **Behaviour change.** Categorical legend labels go 12.8px to 13px, comparison line labels 10px to 10.5px. Away from base 16 many more sites move, and footer text at base 20 goes 16px to 16.5px.
- **Fix.** The scatter size legend derives its label size from the circle radius and reached the DOM unrounded. It never used a scale constant, so deleting those did not surface it.
- Nothing replaces the deleted constants. `scaleFontSize(12, this.fontSize)` in a getter named `tickFontSize` already says what the 12 is, and the constants were named after their own values, so a value-named substitute would keep the ceremony and drop the only thing they did.
- `roundFontSize` stays alongside `scaleFontSize`, for the ten callers that round a size derived from another size, a clamp result, or an interpolated value.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7279, opened 2026-09-15
🔨🤖 (rounding) emit every SVG number at one precision

sophiamersmann:

Numbers written into SVG attributes carried six different precisions, sometimes two of them inside a single element: `ScatterPoints` used two decimals for coordinates and one for radii, the bar charts passed raw floats straight through. They now go through `roundForSvg`, applied where a number becomes an attribute rather than where layout is computed. Two decimals was picked by measurement: one decimal saves 1.65% of emitted bytes and three costs 1.33%, so size does not decide it, and two sits far below a device pixel at any ratio. The helper returns a number rather than a `.toFixed` string, so integers emit as `358` instead of `358.00`.

- **Behaviour change.** `Axis.snapToSubpixel` is gone. It quantised every placed coordinate to a tenth of a pixel before layout saw it, and label placement is decided from those coordinates, so two labels move: a [marimekko leader line](http://staging-site-round-svg-pixels/grapher/academic-freedom-index) changes stagger row, and a connected scatter draws two more labels. Keeping the quantisation but at two decimals produces exactly the same two changes, so this was a choice about quantising at all, not about the precision.
- **Fix.** `Triangle` was spreading `cx`, `cy` and `r` onto its `<polygon>`. They are consumed to build `points` and are not valid polygon attributes, so they are now omitted rather than rounded.
- **Fix.** Thirteen line endpoints in the numeric colour legend and the dumbbell row were emitting three and four decimals. They were invisible to the rendering check for most of this work, because its attribute pattern excluded digits and so never examined `x1`, `y1`, `x2` or `y2`.
- The helper lives in `@ourworldindata/utils`, not grapher, so `TextWrap` and `MarkdownTextWrap` can use it. Grapher depends on `components`, so a helper in grapher could never have reached the code that renders every `<text>` and `<tspan>`.
- Font-size attributes lost their `.toFixed(2)` rather than gaining the helper, since they are already multiples of 0.5 after the branch below. That let three `fontSize: string` props in `SeriesLabel` go back to `number`.
- Rounding sits at the SVG element, never at a caller passing coordinates into a component that renders one, so the same coordinate is not put through the rule twice.
- Three exceptions remain, all in the map and all named in code. Stroke widths and the projected-data pattern scale are divided down by the viewport scale, so two decimals would flatten them to zero when zoomed in. The choropleth transform matrix keeps three decimals because its scale components need more precision than a coordinate does.


sophiamersmann:
* **#7279** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7279?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7276** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7276?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sophiamersmann:
@codex review

sophiamersmann:
### Merge activity

* **Sep 15, 4:51 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/7279).
* **Sep 15, 4:53 PM UTC**: [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/7279) rebased this pull request as part of a merge.
* **Sep 15, 4:53 PM UTC**: @sophiamersmann merged this pull request with [Graphite](https://app.graphite.com/github/pr/owid/owid-grapher/7279).

PR #7280, opened 2026-09-15
🚧 [do not merge] staging check for owid/ops#671

Marigold:

**Throwaway PR — do not merge, do not review.** It exists only so a staging server picks up the matching ops branch (`staging-script`: an ops branch of the same name wins over `main`), to test owid/ops#671 before it is merged.

Same commits as owid/owid-grapher#7274. Labelled `staging-https` to exercise the mixed-content case, and `staging-bespoke` so the Cloudflare preview gets bundles too.

Will be closed once the check is done.


Marigold:
Check done — owid/ops#671 verified on staging-site-bespoke-relative-base-url over both HTTP and HTTPS. Closing; the real PR is #7274.

PR #7286, opened 2026-09-16
🔨🤖 resolve bespoke data URLs from the registry

sophiamersmann:
Follow-up to https://github.com/owid/owid-grapher/pull/7274

Metadata JSON files are used by both the Bespoke project and the Featured Viz Gdocs page. Previously, the URLs to these files were hard-coded in both places. This refactor centralises the URLs for both metadata and data files.

Each bespoke registry entry now defines:

- `dataUrl`: where the bundle's data lives, either relative to `BESPOKE_DATA_URL` or as an absolute URL.
- `metadataFilename`: the name of the metadata file inside the bundle.

`resolveBespokeComponentUrls` converts these values into environment-specific URLs, and `mount()` passes the resolved URLs to the bundle.

### What changed

- The registry replaces the module-global feed root in `bespoke/helpers/feedUrl.ts`
- The registry now lives in `bespoke/shared/`, so the dev server's demo page and the site use the same registry.
- All registered bundles now resolve to a metadata file. As a result, featured viz pages fetch metadata for every bundle at bake time instead of skipping the methods block for bundles without a configured metadata file.
- Metadata without `origins` or `descriptionKey` is no longer sent to Sentry. Metadata json that doesn't adhere to the schema required by the metadata box is still valid, so there is no need to warn about it.

### Special cases

`demography` and `migrant-demographics` publish their data manually rather than through the ETL. Their registry entries therefore use absolute `owid-public.owid.io` URLs, which resolve to the same file in every environment. The file served is unchanged. Only the URL's location has moved from the bundle into the registry. All five bundles now receive their URLs through `mount()`.

`example` has been removed from the registry because it has no data files. The other projects probably serve as examples well enough.


sophiamersmann:
* **#7286** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7286?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

bespoke/projects/causes-of-death/src/components/CausesOfDeathChart.tsx
```diff
@@ -34,53 +35,57 @@ const LATEST_YEAR = -1
...
+                    />
                 </QueryClientProvider>
             </NuqsAdapter>
         </EmbedConfigProvider>
     )
 }
 
 function CausesOfDeathChart(props: {
-    config?: CausesOfDeathConfig
+    config: CausesOfDeathConfig
```
sophiamersmann, line 56: Making config non-optional is an unrelated cleanup

bespoke/projects/migrant-demographics/src/variants/PyramidVariant.tsx
```diff
@@ -43,25 +44,13 @@ import { PyramidControls } from "../components/PyramidControls.js"
 
 export function PyramidVariant({
     config,
+    urls,
 }: VariantProps<PyramidVariantConfig>): React.ReactElement {
-    const { width, ref } = useContainerWidth()
-    const isNarrow = width > 0 && width < NARROW_BREAKPOINT
```
sophiamersmann, line 53: I pushed this down - also an unrelated cleanup

bespoke/shared/bespokeComponentRegistry.ts
```diff
@@ -0,0 +1,32 @@
+import type { BespokeComponentDefinition } from "./bespokeComponentTypes.ts"
+
+export const BESPOKE_COMPONENT_REGISTRY: Record<
+    string,
+    BespokeComponentDefinition
+> = {
+    "causes-of-death": {
+        scriptUrl: "/causes-of-death/index.js",
+        dataUrl: "ihme_gbd/latest/gbd_treemap_json",
```
sophiamersmann, line 9: Re-reading this code, I realise we could also specify this as `viz://bespoke/ihme_gbd/latest/gbd_treemap_json` to be a bit more explicit. The resolve function would then need to strip the `viz://bespoke/` part again...

db/model/Gdoc/GdocFeaturedViz.ts
```diff
@@ -75,27 +81,7 @@ export class GdocFeaturedViz
...
-            if (Object.keys(parsed.data).length > 0) {
-                await logErrorAndMaybeCaptureInSentry(
-                    new Error(
-                        `Metadata for bespoke component "${bundle}" at ${metadataUrl} carries some of the schema's fields but neither origins nor a descriptionKey, so "${this.slug}" will render without a methods block`
-                    )
-                )
-            }
-            return
-        }
+        if (!parsed.success || !shouldRenderBespokeMetadata(parsed.data)) return
```
sophiamersmann, line 84: Bespoke metadata json files combine shared fields that power the metadata box on featured-viz pages with whatever other fields are needed for rendering. It's fine for a metadata json file not to include any of the shared fields. In that case, we simply don't render a metadata box.

bespoke/shared/bespokeComponentRegistry.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+    migration: {
+        scriptUrl: "/migration/index.js",
+        dataUrl: "un_migration/latest/migration_stock_flows_json",
+        metadataFilename: "migration-stock-flows.metadata.json",
+    },
+    "migrant-demographics": {
+        scriptUrl: "/migrant-demographics/index.js",
+        dataUrl: "https://owid-public.owid.io/bespoke/migrant-demographics",
+        metadataFilename: "migrant-demographics.json",
```
sophiamersmann: `migrant-demographics` is an odd case because it currently doesn't have a separate metadata json file - I'm refactoring that now

Marigold (approved):
Using the registry is much nicer!

> demography and migrant-demographics publish their data manually rather than through the ETL. Their registry entries therefore use absolute owid-public.owid.io URLs, which resolve to the same file in every environment.

Should we publish `demographics` and `migrant-demographics` from ETL to have everything in one place? We could just commit the data files for now.

PR #7295, opened 2026-09-17
🐛🤖 Clarify search fallback notice

### LINKED ISSUE #6904: "No exact matches" notice shown even with an exact match
## Description

In search, it's possible to encounter the "No exact matches" notice, despite an exact match being visible.

## Expected behaviour

The notice is only shown when there are no exact matches visible.

## Steps to reproduce

1. Go to https://ourworldindata.org/search?q=deforestation&countries=Slovakia&resultType=all
2. Observe the "No exact matches" notice and the DI below it with "deforestation" in the title

## Screenshots

Image

rakyi:
Fixes #6904


site/search/SearchClosestMatchesNotice.tsx
```diff
@@ -1,17 +1,14 @@
...
  * queries.ts. Tells the user honestly that we don't have an exact answer,
  * while still offering the nearest content instead of a blank page.
  */
 export function SearchClosestMatchesNotice() {
-    const { state } = useSearchContext()
     return (
         <p className="search-closest-matches-notice">
-            No exact matches for <strong>“{state.query}”</strong> — showing the
-            closest matches instead.
+            No results match all your search terms — showing the closest matches
```
rakyi, line 10: I thought this was good enough instead of trying to correctly handle all kinds of search terms (topic/country/text).

PR #7303, opened 2026-09-18
Document the shared search layer

mlbrgl:

The site's search page and `/api/search` already share their relevance logic in `packages/@ourworldindata/utils/src/search/`, but the rule that they *must* agree has only ever existed as a `// Shared between…` comment repeated at the top of each shared module. Adding a fourth module in #7177 meant restating it rather than referencing it. This writes it down once, in a README beside the modules it governs, and collapses the three preambles to a one-line pointer that keeps the part that earned them — editing these files moves two surfaces at once.

**One claim is worth a second opinion before this merges:** that relevance is *not* part of `/api/search`'s stability contract. It's stated because it's what makes behavioural identity survivable — otherwise every relevance change on the search page is gated on the public API, and identity gets quietly abandoned instead — and because `docs/search-api.openapi.yaml` is `0.1.0` with no stability, versioning or deprecation language, so nothing has been promised yet. If that's not the intent, the contract section needs rewriting rather than editing.

Stacked on #7177 because it documents `searchTopicPages.ts`, which that PR adds.

### What the README covers

Four sections, deliberately short and free of anything `ls` already tells you:

- **The contract** — the same query returns the same results whichever surface asks; "the same" covers relevance (which records match, in which order, what happens when nothing does) but not delivery (response shape, pagination, sectioning, caching). Plus the stability claim flagged above.
- **What belongs here** — code that decides what to ask Algolia and how to read the answer, with no runtime dependency. Explicitly excludes React Query hooks, `SearchState`/URL sync, HTTP parsing and untrusted-input validation, `Env`, Sentry, and the API's response envelope.
- **What both sides inherit** — `baker/algolia/configureAlgolia.ts` is upstream of everything here; editing it moves both surfaces without either being touched. Uses the `path`-vs-`slug` topic page lookup as the worked example.
- **Consumers** — split into the two that answer queries and so owe each other the contract, and two that depend on the layer without being query surfaces.

### The line between validation and meaning

The "what belongs here" rule distinguishes *well-formedness* from *meaning*: whether a value is acceptable depends on how it arrived and who sent it, so it stays with the API; what a value denotes doesn't, so it's shared. Concretely — rejecting an unknown `pageTypes` value belongs to `/api/search`; deciding that a topic page request covers both of its gdoc subtypes would belong in this directory.

### The preamble collapse

Each preamble had two clauses. "Shared between `site/search/queries.ts` and `functions/api/search/searchApi.ts`" is now maintained once in the README's consumer list instead of three times. The second clause was a per-module restatement of the same claim ("…so identical filters produce identical requests", "…so the fallback behaves identically", "…so both recommend the same topic pages"), each near-tautological given the filename and the README's definition of identity.

The replacement keeps the warning that made the preambles useful:

```ts
// Shared by the site's search page and /api/search — see ./README.md.
```

In `searchFacetFilters.ts` the old preamble had drifted to the middle of the file, wedged between `MAX_FACET_VALUES` and `SearchFacetAttribute`, where it read as though it applied only to the type below. The marker now sits after the imports in all three.

### Why co-located rather than `docs/`

`docs/` here is setup guides and conventions with no ADR practice to slot into; a decision filed there is one nobody opens while editing the code. The rule being documented is "what belongs in this directory and why it exists", so it lives next to the modules it governs. Cross-links to `site/search/README.md` and `docs/search-api.openapi.yaml` are in the footer.

### Consumers were verified, not assumed

All seven modules in the directory are imported by `functions/`. Two consumers beyond the obvious pair: `functions/api/search/cached-queries.ts` proxies the site's empty-query "browse" requests and recognises them via `emptyQuerySearchPayload.ts` (the clearest case for the directory existing — the site produces the payload, the edge validates it), and `functions/_common/search/` builds a chart hit card's rich data, served off the grapher route and laid out by the same `Medium`/`Large` placement helpers the search UI uses.

### Checks

`yarn typecheck` clean (5 pre-existing `toHaveTextContent` errors in grapher test files are unrelated and present on a clean tree), `yarn testLintChanged` and `yarn testFormatChanged` clean, 83/83 tests in `packages/@ourworldindata/utils/src/search/` and `functions/api/search/index.test.ts`. Documentation only — no behaviour change.

</details>


mlbrgl:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7303?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7303** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7303?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7177** <a href="https://app.graphite.com/github/pr/owid/owid-grapher/7177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

PR #7319, opened 2026-09-22
Let the export tab remove the automatic tolerance notice

edomt:
> _Written by Claude Opus 5 — @edomt at the wheel._

Charts where tolerance is applied get an automatic notice appended to the note ("Where data for 2024 is unavailable, the value from the closest year between 2019 and 2023 is shown instead."). The export tab's "Displayed elements" list only showed a **Note** toggle when the chart had an *authored* note, so on a chart whose note is nothing but that notice there was no way to remove it before downloading a PNG/SVG.

The notice now gets its own toggle, and the two are independent: unticking **Note** removes only the authored note, unticking **Note: automatic tolerance notice** removes only the notice.

- `GrapherState`: new runtime display flag `hideToleranceNotice` (observable, not serialized — same shape as `hideTitle`/`hideNote`), also added to `GrapherProgrammaticInterface`.
- `GrapherState.effectiveNote` now resolves both flags itself: authored note dropped when `hideNote`, notice dropped when `hideToleranceNotice`, and the two are still run together as sentences when both are present.
- Consequently `Footer.showNote` no longer checks `manager.hideNote` (it would have hidden the notice too), and `hideNote` is gone from `FooterManager`. Nothing else read it — the admin export tab was its only caller.
- `EditorExportTab`: `hideToleranceNotice` threaded through `ExportSettings`/defaults/`saveOriginalSettings`/`currentSettings` (so it is applied to the preview, reset on unmount and remembered in session storage), plus the new toggle. Its visibility reads `grapherState.toleranceNotice` live rather than the `originalGrapher` snapshot taken in the constructor, because `toleranceNotice` is `undefined` until the grapher is ready and the flag itself never changes it.
- Tests: three cases in `GrapherState.test.ts` covering each flag on its own and both together. Full unit suite passes apart from the pre-existing timezone-dependent `archivalDate` failure; `yarn typecheck`, lint and format are clean.
- Not covered: the toggle only appears while the notice is actually on screen (i.e. on a chart or map tab, with tolerance in play) — same condition as the notice itself.

</details>


edomt:
> _Written by Claude Opus 5 — @edomt at the wheel._

Superseded by a one-line version that reuses the existing `hideNote` flag instead of adding a second toggle.

PR #7320, opened 2026-09-22
Offer the automatic tolerance notice as a removable export element

edomt:
> _Written by Claude Opus 5 — @edomt at the wheel._

Charts where tolerance is applied get an automatic notice appended to their note ("Where data for 2024 is unavailable, the value from the closest year between 2019 and 2023 is shown instead."). The export tab's "Displayed elements" list only showed a **Note** toggle when the chart had an *authored* note, so on a chart whose note is nothing but that notice there was no way to remove it before downloading a PNG or SVG.

The toggle is now gated on the effective note (authored note + notice) rather than the authored note alone. Unticking **Note** removes the whole note line, which is what `hideNote` already did — only the checkbox was missing.

- `adminSiteClient/EditorExportTab.tsx`: the toggle's condition becomes `grapherState.effectiveNote`, and `note` drops out of the `originalGrapher` snapshot since nothing reads it any more.
- The condition reads live grapher state rather than the constructor-time snapshot on purpose: `toleranceNotice` (and therefore `effectiveNote`) is `undefined` until the grapher is ready, so a snapshot could permanently hide the toggle. Reading live is safe because `hideNote` does not feed into `effectiveNote` — ticking the box off doesn't make the row disappear.
- No grapher-side changes: `Footer.showNote` already gates on `!hideNote && !!noteText`, and `noteText` already comes from `effectiveNote`.
- Considered and dropped: a second toggle to remove only the notice while keeping the authored note. It needs a new `hideToleranceNotice` flag plumbed through `GrapherState`/`Footer`, and stripping one footnote but not the other seems like a corner case for a static export. Easy to add later if it comes up.
- Checks: `yarn typecheck`, `yarn testLintChanged`, `yarn testFormatChanged` clean. No test added — the change is a JSX visibility condition in an admin component, and `effectiveNote` itself is already covered in `GrapherState.test.ts`.

</details>


edomt:
@codex review

edomt:
> _Written by Claude Opus 5 — @edomt at the wheel._

@sophiamersmann tagging you since this touches the automatic tolerance notice: the export tab's **Note** checkbox only appeared when a chart had an authored note, so on a chart whose note is nothing but the notice there was no way to untick it before downloading a PNG/SVG. The toggle now keys off `effectiveNote` instead. Verified visually on staging; CI is green.

sophiamersmann (approved):
Makes sense! I just removed the comment :)

